PR #1, opened 2018-04-18
Bump the rocket version to compile with current nightly

mprasil:
Hi this is just bumping some versions to compile with current nightly.
It generates a bunch of warnings about deprecated functionality in diesel, but that doesn't affect functionality for now, so I might look into it in a separate PR.

Thanks for your work on this, this is truly amazing bitwarden API implementation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -34,7 +34,7 @@ name = "base64"
 version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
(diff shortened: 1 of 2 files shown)


dani-garcia:
Thanks for this! I've been using and old nightly and didn't notice the issue.

I'm pretty sure that the deprecated issues with diesel are only because the large-tables feature, which is the default in diesel 1.2 and up, so I'm going to try to update the dependencies and remove that feature and see if that solves the issue.

PR #2, opened 2018-04-18
Fetch Vault from upstream when building Docker image

mprasil:
Hi, this extends the Dockerfile and builds the Vault files straight from upstream, so you can switch it easily. Should make it easier to follow the upstream.


dani-garcia:
Thanks again mprasil!

PR #3, opened 2018-04-19
Add bulk move and bulk delete

mprasil:
This adds support for `Move Selected` and `Delete Selected` in the Vault. It works, but there's a bit of code repetition in there which I'm not happy about, not sure how to do it nicer. It might require refactoring the current functions, I'll better leave that for another PR.

### DIFF
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -388,8 +388,80 @@ fn delete_cipher(uuid: String, headers: Headers, conn: DbConn) -> EmptyResult {
 fn delete_cipher_selected(data: Json<Value>, headers: Headers, conn: DbConn) -> EmptyResult {
     let data: Value = data.into_inner();
 
-    println!("{:#?}", data);
-    unimplemented!()
+    let uuids = match data.get("ids") {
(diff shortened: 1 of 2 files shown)


dani-garcia:
Cool! Thanks for this!

I personally think the repetition is mostly fine as it is now, as the functions are not particularly long or difficult to understand. If you can think of a better way to do it, don't hesitate to open another PR!

mprasil:
Thanks for the quick merge!

I was thinking about moving the delete cipher functionality out of the delete cipher handler and then just call if for each cipher in the bulk handler, but I'm still trying to wrap my head around the ownership handling in Rust, so not sure how to do it in some reasonable way.

The functions aren't that long right now, but my worry is that adding things like organizations might require some extra checks and we will have to add them in multiple spots, which will make to code harder to maintain and more error prone.

dani-garcia:
Yeah, you are right about the delete functionality. I've pushed some changes, basically made what you suggested. It didn't require a big change, just moving the functionality to a new function and making sure we are passing references to that function, to avoid taking ownership of the parameters.

PR #11, opened 2018-04-27
Save the name when updating the cipher data

mprasil:
Just a quick fix to save cipher name on save

### DIFF
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -188,6 +188,7 @@ fn update_cipher_from_data(cipher: &mut Cipher, data: CipherData, headers: &Head
 
     // Copy the type data and change the names to the correct case
     copy_values(&type_data, &mut values);
+    cipher.name = data.name;
 
     cipher.data = values.to_string();
 


mprasil:
Thank for that. While I was investigating this I've noticed you can't really remove Notes from cipher. When you remove the note text Vault will `POST` the `notes` field as `null` which we interpret as "no data sent" and keep the old text. Even a trick with space instead of text won't work as Vault does trim the data and again, will send `null`. Not sure if that's bug on our side or Vault?

mprasil:
Maybe we could set the data to empty strings before we call the `update_cipher_from_data` in the individual API handlers. That way the function can still update the cipher even with partial data without resetting missing fields.

dani-garcia:
I don't think that would be necessary, when updating a cipher, the client sends the entire updated cipher, with all the fields, so if a field is empty, it's because it was already empty or the user specifically deleted it.

So if we have a cipher with fields and notes and we update it to remove the notes, the server is going to receive the full fields correctly and the notes as empty. In that case we don't need to check if the data we receive is empty and we can assign it directly to the cipher.

PR #13, opened 2018-05-09
Implement Collection-Cipher mapping

mprasil:
This adds new table that maps ciphers into collections. Also implemented some parts of the API to assign cipher to collections.

### DIFF
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -322,7 +322,7 @@ fn post_collections_admin(uuid: String, data: Json<CollectionsAdminData>, header
 
     for collection in posted_collections.symmetric_difference(&current_collections) {
         match Collection::find_by_uuid(&collection, &conn) {
-            None => (), // Does not exist, what now?
+            None => err!("Invalid collection ID provided"),
             Some(collection) => {
                 if collection.is_writable_by_user(&headers.user.uuid, &conn) {
                     if posted_collections.contains(&collection.uuid) { // Add to collection
@@ -330,6 +330,8 @@ fn post_collections_admin(uuid: String, data: Json<CollectionsAdminData>, header
                     } else { // Remove from collection
                         CollectionCipher::delete(&cipher.uuid, &collection.uuid, &conn);
                     }
+                } else {
+                    err!("No rights to modify the collection")
                 }
             }
         }
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -320,7 +320,6 @@ fn post_collections_admin(uuid: String, data: Json<CollectionsAdminData>, header
     let posted_collections: HashSet<String> = data.collectionIds.iter().cloned().collect();
     let current_collections: HashSet<String> = cipher.get_collections(&conn).iter().cloned().collect();
 
-    //TODO: update cipher collection mapping
     for collection in posted_collections.symmetric_difference(&current_collections) {
         match Collection::find_by_uuid(&collection, &conn) {
             None => (), // Does not exist, what now?
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -318,7 +318,7 @@ fn post_collections_admin(uuid: String, data: Json<CollectionsAdminData>, header
     }
 
     let posted_collections: HashSet<String> = data.collectionIds.iter().cloned().collect();
-    let current_collections: HashSet<String> = cipher.get_collections(&conn).iter().cloned().collect();
+    let current_collections: HashSet<String> = cipher.get_collections(&headers.user.uuid ,&conn).iter().cloned().collect();
 
     for collection in posted_collections.symmetric_difference(&current_collections) {
         match Collection::find_by_uuid(&collection, &conn) {
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -1,4 +1,5 @@
 use std::path::Path;
+use std::collections::HashSet;
 
 use rocket::Data;
 use rocket::http::ContentType;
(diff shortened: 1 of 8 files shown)


src/db/models/cipher.rs
```diff
@@ -241,4 +241,11 @@ impl Cipher {
             .select(ciphers::all_columns)
             .load::<Self>(&**conn).expect("Error loading ciphers")
     }
+
+    pub fn get_collections(&self, conn: &DbConn) -> Vec<String> {
+        ciphers_collections::table
```
dani-garcia: We should probably filter these by user_uuid, and only show collections with the cipher and that the user has access to.

src/api/core/ciphers.rs
```diff
@@ -297,6 +298,46 @@ fn put_cipher(uuid: String, data: Json<CipherData>, headers: Headers, conn: DbCo
...
+        err!("Cipher is not write accessible")
+    }
+
+    let posted_collections: HashSet<String> = data.collectionIds.iter().cloned().collect();
+    let current_collections: HashSet<String> = cipher.get_collections(&conn).iter().cloned().collect();
+
+    //TODO: update cipher collection mapping
+    for collection in posted_collections.symmetric_difference(&current_collections) {
+        match Collection::find_by_uuid(&collection, &conn) {
+            None => (), // Does not exist, what now?
```
dani-garcia: I think in this case we should err! with something like "Invalid collection ids provided", even though it's something that shouldn't happen. Same thing with the else case of the is_writable_by_user a couple of lines down.

src/api/core/ciphers.rs
```diff
@@ -297,6 +298,46 @@ fn put_cipher(uuid: String, data: Json<CipherData>, headers: Headers, conn: DbCo
...
+    };
+
+    if !cipher.is_write_accessible_to_user(&headers.user.uuid, &conn) {
+        err!("Cipher is not write accessible")
+    }
+
+    let posted_collections: HashSet<String> = data.collectionIds.iter().cloned().collect();
+    let current_collections: HashSet<String> = cipher.get_collections(&conn).iter().cloned().collect();
+
+    //TODO: update cipher collection mapping
```
dani-garcia: Is there something else that needs to be done, or is this comment simply outdated?

src/db/models/cipher.rs
mprasil: I have no idea how does that work in upstream API. Do the users only see collections they have access to?

src/api/core/ciphers.rs
mprasil: My problem with err!-ing here is that this might happen after couple loops, so the changes for other collections could have been already applied. In other parts of the API (like the mass move) we ignore errors on individual operations and only error on common problems. (like cipher ownership in this case)

I  think it might make more sense to apply as much as we can and just skip whatever fails. I'm wondering how upstream deals with that, if there's some way to report partial success.

src/api/core/ciphers.rs
mprasil: :+1:  Yeah, had it there as a placeholder, forgot to remove it.

src/api/core/ciphers.rs
mprasil: I take that back, I went to check the code in the other cases (`/ciphers/delete`, `/ciphers/move`) and we indeed fail mid-loop. I'll add some `err!` here as well. Apologies for confusion.

src/db/models/cipher.rs
dani-garcia: I just tested this now in the upstream server to make sure, and when calling the `/api/ciphers` API, only the collections that the user has access appear in the collectionIds field.
But at the same time,  if the user is an owner or an admin, they can use the organization page to view all collections and ciphers, regardless of the setting.

So it seems that if the user calls the `/api/ciphers` or `/api/collections`, we should check if the user has access to the collection.
On the other hand, if the user is an admin or owner and they are using the organizations API, then we should show all collections. And by organizations API I mean:
`api/organizations/<uuid>/collections` and `api/ciphers/organization-details?organizationId=<uuid>`.

It's a bit of a weird system, but I think we can solve it simply by adding a boolean parameter `filter_collections` to the `to_json` method.

mprasil:
I've implemented the changes you suggested. At this point I'm a bit lost in the API, but I think it does what it should so far. The get collections function should now handle all three ways of getting access to collection - direct access, access_all and admin/owner of the organization. It does require some heavy query though, so second pair of eyes is definitely needed here. :nerd_face:

dani-garcia:
About the SQL query, I think it would be simpler to call Collection::find_by_organization_and_user_uuid() to get all the collections and then checking if the cipher is in any of them, something like this:

```
let collections = Collection::find_by_organization_and_user_uuid(&self.organization_uuid, user_uuid, conn);
let coll_uuids = collections.iter().map(|c| c.uuid);

// Find which collections contain the cipher
ciphers_collections::table.inner_join(collections::table)
      .filter(ciphers_collections::cipher_uuid.eq(&self.uuid))
      .filter(ciphers_collections::collection_uuid.eq_any(coll_uuids))
      .select(collections::all_columns)
      .load::<Self>(&**conn).expect("Error loading collections")
```
This way we don't have multiple queries doing nearly the same thing.

PD. I haven't actually tested the eq_any method in Diesel, but it should work
(http://docs.diesel.rs/diesel/expression_methods/trait.ExpressionMethods.html#method.eq_any).

mprasil:
I'm on the fence here. On one hand we would share some code between functions on other we do make 3 requests instead of one and with a bit of bad timing we might get inconsistent results as opposed to the atomicity of single query. I'd like to make the batch move/delete API calls atomic at some stage and this is kinda step in the other direction. I feel like the benefit of saving 4 lines of code might not be worth it. I dunno, what are your thoughts about this?

BTW, I've just rebased on top of your latest commit.

dani-garcia:
Fair enough, I don't think the savings in this case are worth that much, so I'm okay merging this as it is. Again, thanks for this!

mprasil:
:100: I think we can approach the code reuse on the diesel level, I'm not even sure how to achieve that properly, need to read a bit more on it, but I think we could extend the schema with some extra `joinable!` statements to reduce the repeated code.

PR #14, opened 2018-05-11
Also list shared ciphers in find_by_user

mprasil:
This splits the `find_by_user` into two separate functions: `find_owned_by_user` which retains previous functionality and lists only ciphers directly owned by user and `find_by_user` which now also lists all ciphers shared with user.

As a result shared ciphers are now visible in user's vault.

There's also small change in the `/api/ciphers/purge` handler to use the `find_owned_by_user` instead to only purge user's content and not the shared stuff.

### DIFF
--- a/src/api/core/accounts.rs
+++ b/src/api/core/accounts.rs
@@ -168,7 +168,7 @@ fn delete_account(data: Json<PasswordData>, headers: Headers, conn: DbConn) -> E
     }
 
     // Delete ciphers and their attachments
-    for cipher in Cipher::find_by_user(&user.uuid, &conn) {
+    for cipher in Cipher::find_owned_by_user(&user.uuid, &conn) {
         for a in Attachment::find_by_cipher(&cipher.uuid, &conn) { a.delete(&conn); }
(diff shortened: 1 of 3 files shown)


dani-garcia:
With some light testing I've noticed a bug: If a cipher is added to more than one collection, the method returns multiple copies of the cipher, which seems to break the web vault listing, you need to make sure that the method doesn't return duplicated ciphers.

You should also change the `/accounts/delete` handler to also use the new owned function as well.

mprasil:
Well spotted! I'll have a look at that.

mprasil:
I believe the bug should be fixed with `distinct()`. I've also updated `/accounts/delete` to use proper function.

BTW `/accounts/delete` and `/ciphers/purge` have a lot in common, I might look into moving the common functionality under `cipher.rs` in some later PR if you agree.

dani-garcia:
Okay, I've tested it again and it works correctly now, so I'll merge this. If you want to reduce the duplication in those functions in another PR, i'm OK with it.

PR #25, opened 2018-05-19
Implement deleting Organization

mprasil:
This implements API needed to delete Organization. As part of these changes I moved deleting CollectionUser directly into UserOrganization::delete() where it's used wen user is removed but also when organization is deleted.


mprasil:
Thanks for pulling bunch of PRs so fast! Much appreciated. :+1:

dani-garcia:
They were all pretty straight-forward, so it was no problem!

Issue #6, opened 2018-04-26
The models save() and delete() method should probably return Result instead of bool


mprasil:
I find that the bool actually masks the underlying issue, returning some sort of Result would probably be much better. Also the Result should actually be checked and API should respond appropriately. As a real world example, it took me quite a while to figure out that program couldn't write to DB due to permissions.

Here's the list of functions that are left to be updated:

- attachment.rs:
  - [x] Attachment::save()
- cipher.rs:
  - [x] Cipher::save()
- collection.rs:
  - [x] Collection::save()
  - [x] CollectionUser::save()
  - [x] CollectionUser::delete()
  - [ ] CollectionCipher::save()
  - [ ] CollectionCipher::delete()
- device.rs:
  - [ ] Device:save()
  - [ ] Device::delete()
- folder.rs:
  - [ ] Folder::save()
- organization.rs:
  - [ ] Organization::save()
  - [ ] UserOrganization::save()
- user.rs:
  - [ ] User::save()
  - [ ] User::delete()

dani-garcia:
Yeah, I figured that as we are using an integrated SQLite instance that we control, it was a safe-ish asumption that there wouldn't be any problems, but it would be a good idea to handle those kinds of errors in a better way.

mprasil:
These are generally a good first issues if someone feels like taking one function at a time.

Baelyk:
@mprasil I created a pr (#161) for the Attachment model.

In the future (and I guess present) what is the etiquette regarding good first issues? Now having completed a first issue (and part of this one), should I leave them in case other people are interested in contributing for the first time? Would it have been okay to do several of the models here?

And then regarding the bug issue previously, if I see something like that, is it okay for me to immediately work on that? Or should I ask permission in the issue/matrix?

Sorry thats a lot of questions but I am interested in contributing more but have never made it beyond the stage of "good first issue" of a project before and I guess I'm just not sure on how it usually works :/

mprasil:
That's great!

[...]

Issue #5, opened 2018-04-20
Implement organizations [wip]


mprasil:
There's some ground work laid out already, let's track the progress here. 
Endpoints to implement:
- [x] /api/organizations
- [x] /api/organizations/{orgId}/collections
- [x] /api/organizations/{orgId}/users
- [x] /organizations/{org_id}/collections/{coll_id}/users
- [x] /api/organizations/{orgId}/users/{orgUserId}/accept
- [x] /api/organizations/{orgId}/users/{orgUserId}/confirm
- [x] /api/organizations/{orgId}/users/invite
- [x] /api/organizations/users
- [x] /api/users/{userId}/public-key
- [x] /organizations/{org_id}/delete
- [x] /api/ciphers/{cipher_id}/collections-admin (#13)

Other changes:
- [x] /accounts/profile needs to list organizations for user
- [x] separate folder id from cipher to allow per-user folders (#10)
- [x] Respect the read_only bit when a user tries to modify a Cipher from a Collection
- [x] List all user Ciphers, including those from Collections the User has access to (#14)

dani-garcia:
Yeah, the work on organizations is barely started. I don't have a lot of use for that functionality myself, and it's a pretty big change, so I don't have time at the moment to do it myself. That said, if you or anyone else would want to implement it, I'd gladly accept the PRs!

dani-garcia:
With your PR and my latest fix (7e7c368) we can finally share ciphers between users with an organization.
For now, the Ciphers from the Organization are only available from the Organization page.

mprasil:
Uhm.. so do we miss anything for this?  If we're ready to close it, can I suggest bumping the version to 0.2.0 or something? ✌️

dani-garcia:
I forgot that we have a version 🤔 . We probably would be at 0.6.0 at least!

Yeeah, I think it's done, as far as I know. With this, the only thing left to be feature-equivalent with upstream would be: 
- implementing other 2Factor methods,  like YubiKey or U2F (I don't know how these work and I don't have the hardware to test, so I won't be personally implementing this)
- Enterprise features, like directory sync (I don't have a way to test this either, and I don't see much use for it, so I won't implement it)

I'll bump the version to 0.9.0, as in my view it's basically feature complete, but there might be bugs still. Once some time has passed and if there aren't any bugs left to fix, we could promote it to 1.0.

PR #32, opened 2018-05-29
Support listing and deleting users from collection

mprasil:
This implements ` /organizations/{org_id}/collections/{coll_id}/users` for #5, it also adds `/organizations/<org_id>/collections/<col_id>/delete-user/<org_user_id>` that is used for deleting individual users from collection.

This also improves `CollectionUser::delete()` to work on `self` rather than requesting IDs for lookup. It also returns `QueryResult` that is now checked. (#6)


mprasil:
Makes sense. I created this PR before #34, I can probably remove half of those match statements just by changing header type.

PR #34, opened 2018-05-30
Implement request guards for organization

### LINKED ISSUE #33: Protect Organization related GET requests
It seems that we don't really check user's access rights when doing `/organization/*` API calls. While this is mostly harmless in theory, we should probably protect all of these under some common check.

It think implementing some form of request guard would probably be the best way to handle that.

mprasil:
This implements `OrgHeaders`, `AdminHeaders` and `OwnerHeaders`, which are drop-in replacements for `Headers`, that also check appropriate `Organization` access level. 

I've updated all appropriate organization `GET` handlers with the above to show the usage. The other handlers should be updated as well, but  these will require removing some unnecessary check code and I wanted to keep this PR as minimal as possible, so I'd rather leave that for another, separate PR.

This should resolve #33

### DIFF
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -6,7 +6,7 @@ use db::DbConn;
 use db::models::*;
 
 use api::{PasswordData, JsonResult, EmptyResult, NumberOrString};
-use auth::Headers;
+use auth::{Headers, AdminHeaders, OwnerHeaders};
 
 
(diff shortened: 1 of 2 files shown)


dani-garcia:
This looks great, thanks!

mprasil:
Thanks for fast merge! Glad you like it.

PR #36, opened 2018-05-31
Better check for cipher access

mprasil:
This checks out two TODOs: 
1. Extends `Cipher::is_write_accessible_to_user` with a check for any collection that might give R/W access to cipher.
2. Implements `Cipher::is_accessible_to_user` with a proper check for any RO access. 

The current solution works quite well, but I don't like the code repetition. Any input would be greatly appreciated how to solve this in some reasonable manner. I was thinking about using `into_boxed()` at some stage to promote some code reuse, but I just don't know enough about diesel to pull this off successfully.

Other approach would be to split it into multiple sub-checks and aggregate those but that would lead to 5x the amount of queries, which I'd like to avoid.

Any ideas?

### DIFF
--- a/src/db/models/cipher.rs
+++ b/src/db/models/cipher.rs
@@ -194,29 +194,57 @@ impl Cipher {
     }
 
     pub fn is_write_accessible_to_user(&self, user_uuid: &str, conn: &DbConn) -> bool {
-        match self.user_uuid {
-            Some(ref self_user_uuid) => self_user_uuid == user_uuid, // cipher directly owned by user
-            None =>{
-                match self.organization_uuid {
-                    Some(ref org_uuid) => {
-                        match users_organizations::table
-                        .filter(users_organizations::org_uuid.eq(org_uuid))
-                        .filter(users_organizations::user_uuid.eq(user_uuid))
-                        .filter(users_organizations::access_all.eq(true))
-                        .first::<UserOrganization>(&**conn).ok() {
-                            Some(_) => true,
-                            None => false //TODO R/W access on collection
-                        }
-                    },
-                    None => false // cipher not in organization and not owned by user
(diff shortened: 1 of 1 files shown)


dani-garcia:
Yeah I don't really know diesel that well either. We could try doing what you say about boxed, or maybe the `impl Trait` change that arrived in Rust 1.26 can help?

I can think of a hack to solve the curernt problem, though, the problem is that one query has
```rust
users_collections::user_uuid.eq(user_uuid).and(users_collections::read_only.eq(false))
```

while the other has the first part.

We could change the second part of the query so it uses eq_any and we pass an array of booleans, that way in the function with the extra query we have:
```rust
users_collections::user_uuid.eq(user_uuid).and(users_collections::read_only.eq_any(&[false]))
```
While on the other we would have:
```rust
users_collections::user_uuid.eq(user_uuid).and(users_collections::read_only.eq_any(&[true, false]))
```
And in this case the second part would be doing nothing.

Then the entire query would be equal except for the array, so we could extract it to a separate function.

mprasil:
`impl Trait` sounds like better option to me. `eq_any(&[true, false])` is just one nasty hack. :smile:

I'll see if I can get the Trait implementation working somehow. I'd be willing to have some repetition with the `.filter()` calls if I could at least avoid the `left_join()` part there. I believe that is used in identical form in some other function down the file.

mprasil:
Okay I give up on this one. I just couldn't find nicer way to write it. `.into_boxed()` sounds like nice idea and we could probably use it to clean up some code, but once there's any `*_join()` call present, things get complicated beyond my comprehension.

I'd suggest to merge this PR to fix the functionality and if you feel like create an issue to rewrite it in some nicer way.

PR #41, opened 2018-06-12
Use proper toolchain in Dockerfile

mprasil:
This copies the rust-toolchain file into the build container to make sure we use the proper toolchain from the beginning. I've also switched the build image to official `rust` as we're going to pull requested toolchain anyways.


dani-garcia:
Thanks, I thought the official image was stable rust only, it's great that it uses rustup!

mprasil:
I think the light version might be stable only.

PR #61, opened 2018-07-06
updated readme to include information about creating backups

### LINKED ISSUE #59: Additional documentation for performing a backup
I'm trying to do a backup of my vault. Reading the official documentation, there are some differences for the rust docker implementation (https://help.bitwarden.com/article/backup-on-premise/). In the docker implementation, by default all 3 folders mentioned in that link are combined into the data folder. I assume that backing up that one /data/ folder is sufficient for a full backup. Assuming that is correct, I can put in pull request into the readme to add a backup section to clarify this if you want.

Also the official docs talk about an automatic nightly backup of the database. Is that also the case with this sqlite implementation?

laxmanpradhan:
fixes #59

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,3 +109,13 @@ diesel migration redo
 # This step should be done automatically when using diesel-cli > 1.3.0
 # diesel print-schema > src/db/schema.rs
 ```
+
+## Backing up your Vault
+The official bitwarden backup documentation can be found here:
+https://help.bitwarden.com/article/backup-on-premise/
+
+There are 3 folders that need to be backed up in the official install. In the bitwarden_rs install, the default docker setting will have all 3 of these folders in same location: /data/.
+Backing up this one folder will backup your sqlite database, key files and attachments. 
+
+Note that the official bitwarden install uses a MySQL database which is automatically backed up nightly. This is not the case with the bitwarden_rs. Instead you should use a backup program to create incremental backups.
+


mprasil:
Hey I know the review is up for @dani-garcia, so take my comment with grain of salt, but I'd like to raise some concerns with this PR. 

First thing is, that I find the mention of upstream documentation confusing because bitwarden_rs data is nothing like the upstream one. It doesn't help much people moving from official API as it's very different anyways and it doesn't help at all people that never used the official server. I believe there's also an misinformation in regards of the official database used. (I believe it's not MySQL, but MS SQL unless I'm terribly wrong) Not that it matters much as the entire part is not really relevant to the project.

Now to the actual backing up instructions: While the incremental backups of data folder should work (given the usual DB workload) it is certainly not the safest way to do it in terms of backing up the DB as external software would have no way of ensuring the backup consistency. This is why sqlite provides `.backup` command. So the recommended way should probably be regularly running something like:

```bash
sqlite3 data/db.sqlite3 ".backup 'backup/path/backup.sq3'"
```
And then perhaps moving that sq3 file to some safe location or backing it up incrementally with `attachments` folder. (and perhaps also `icon_cache`)

Otherwise you run some (probably pretty small) risk that you introduce DB inconsistency into your backups.

Just my 2c.

dani-garcia:
I think I agree with @mprasil, linking the official documentation here is going to be confusing for the users.

I think we should also mention the only files that really have to be backed up are the attachments folder and the sqlite database. The icon_cache would be regenerated with use, and the keys are only for the login tokens, so removing them would only log out the users.

As for backing up the database, if the server is turned off it's as simple as copying the `.sqlite3` file. In the case of an online backup, though, you should use the command @mprasil wrote.

laxmanpradhan:
thank you for your suggestions. I have updated my own backup protocol as well given the information about the database. I will make a new pull request with the updated info.

Issue #62, opened 2018-07-07
Offical Docker Hub image


jkaberg:
Would you please setup an automated build for bitwarden_rs @ [Docker Hub](https://hub.docker.com)? 

Documentation found [here](https://docs.docker.com/docker-hub/builds/)

Thanks for the hard work you put into this 😄

mprasil:
Hi @jkaberg, you can use [my image](https://hub.docker.com/r/mprasil/bitwarden/) if you'd like. It's automated build based on top of this source code, the only difference is docker hub focused README. (Just a commit I rebase on top of latest changes in this repo)

@dani-garcia the linked hub image accumulated quite a lot of users (to my surprise really) not sure what to do with it at this stage. I don't want to just discontinue it (I'm using it as well) and I don't mind maintaining it at all, but if you like I can give you access there or something to make it "official" and to give you some control over it if you're interested.

Whatever you prefer really as long as it's not confusing for current and future users.

dani-garcia:
@mprasil personally I'm indifferent about it. If you want, you can PR the changes to the README to avoid having to rebase it every time and we can call your version the official.

I'm not sure how Docker Hub works, but if we do that maybe you can point your image to this repository directly and avoid having to use a fork to pull the changes? That way it would be completely automated and save you time.

You are right that now that people are using your image we simply can't take it down.

mprasil:
So to sum up, the official image is available [here](https://hub.docker.com/r/mprasil/bitwarden/). This is now following the upstream source code. I think we can close this now.

PR #64, opened 2018-07-08
Update readme for docker hub

mprasil:
Hi, as per our discussion at #62 this is a PR to merge in the README changes. I think most of the information should be preserved, some of it is moved into separate BUILD.md file.

As for the docker hub side of things, if you're interested, create an account there and let me know, I can add you as collaborator, this should hopefully make things easier in the future. If you're not interested having yet another account in your bitwarden Valut, that's fine, but we need to figure out how can we get in touch to set up build trigger so that the whole thing is automated. (email me at miroslav@prasil.info perhaps?)

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -1,0 +1,69 @@
+## How to compile bitwarden_rs
+Install `rust nightly`, in Windows the recommended way is through `rustup`.
+
+Install the `openssl` library, in Windows the best option is Microsoft's `vcpkg`,
+on other systems use their respective package managers.
+
+Then run:
+```sh
+cargo run
+# or
+cargo build
+```
+
+## How to install the web-vault locally
+If you're using docker image, you can just update `VAULT_VERSION` variable in Dockerfile and rebuild the image.
+
+Install `node.js` and either `yarn` or `npm` (usually included with node)
+
+Clone the web-vault outside the project:
+```
+git clone https://github.com/bitwarden/web.git web-vault
+```
+
+Modify `web-vault/settings.Production.json` to look like this:
+```json
+{
+  "appSettings": {
+    "apiUri": "/api",
+    "identityUri": "/identity",
+    "iconsUri": "/icons",
+    "stripeKey": "",
+    "braintreeKey": ""
+  }
+}
+```
+
+Then, run the following from the `web-vault` dir:
+```sh
+# With yarn (recommended)
+yarn
+yarn gulp dist:selfHosted
+
+# With npm
+npm install
+npx gulp dist:selfHosted
+```
+
+Finally copy the contents of the `web-vault/dist` folder into the `bitwarden_rs/web-vault` folder.
+
+## How to recreate database schemas
+Install diesel-cli with cargo:
+```sh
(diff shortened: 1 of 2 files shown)


dani-garcia:
Alright, I've created a Docker Hub account, at https://hub.docker.com/u/danigarcia/.

I'll merge this now, thanks.

mprasil:
Added you as a collaborator on the repository. Hopefully you can see the [Build Setting](https://hub.docker.com/r/mprasil/bitwarden/~/settings/automated-builds/) now. If you scroll down, you should see build trigger, just add that as a webhook in the github repo settings and see if you can trigger a build this way. Once that's done, this should trigger a build when there's a git push.

mprasil:
I'm still trying to move the source repository path, so maybe wait a bit until that's done. I'll let you know.

mprasil:
Actually I'm afraid that I can't change the source to your repository as I'm not owner or collaborator here. Can you follow the [steps here](https://github.com/docker/hub-feedback/issues/313#issuecomment-265843090) to try and setup the trigger?

dani-garcia:
I've tried that, but that means creating a new image under my name. I've also tried accessing the build settings for your image but apparently collaborators don't have access to that.

The way I see it is we have three options:
- I'll mark you as collaborator here (I'm fine with it if you are too) and hope that it's enough to change the source repo. Then I'll probably have to set-up the webhook to build on git push and then the image should be automated.
- You continue as before, pulling the changes here to your repo, which should be easier now. This would mean no automation, but this proyect doesn't receive important changes that often, so it woudn't be that problematic.
- I create an automated image under my name, and you modify the README of yours to indicate the users to migrate to the new one. This would require action by the users, but as both images are practically the same, the migration shouldn't be complicated.

mprasil:
Man I'd be honoured to be contributor on this project. If you want to go ahead with this, I'm definitely interested. Let's hope that will be enough to change the source repo. (I think it should be)

If that won't work, then I'd rather do pulls manually.

From my experience changing the Readme doesn't really work to redirect people as most will only see it once and then they just use the image. Case in point is my other bitwarden-ruby image which I created as an test image from one of the forks of the ruby API implementation. It raked in 10k+ downloads in days which definitely surprised me (being an non-automatic build from non-main branch code and all that..) but I had no intentions keeping it alive long-term really. I tried to redirect people in Readme and it just doesn't work. That ruby thing is still being downloaded despite the warning and no updates for long time.. (I plan to eventually remove it which might make some of the users look at the Readme, but at that stage you're essentially breaking people's setup)

dani-garcia:
Alright, I've invited you. Hopefully that's enough to solve the problem. 
You are now officially a collaborator of the proyect! 🎉

mprasil:
:confetti_ball: :fireworks: Yay, I think I should celebrate. :wine_glass: 

But the bad news is I still can't setup the repository to do the automated build. I guess as long as I don't have access to repository settings, it won't give me that option. (It for example adds deploy read-only ssh key, which I definitely can't do here) But no worries I can sync the repo manually for now, now that the readme is merged it's not much work and I can eventually automate the sync.

PR #65, opened 2018-07-09
updated README to include backup information

laxmanpradhan:
New backup information for users


dani-garcia:
Great! This is looking much better! Thanks!

Issue #69, opened 2018-07-11
Dockerfile improvements


dustyrip:
Hello, first of all thanks you @dani-garcia and @mprasil for this this project.

Is there a reason for [exposing port 80](https://github.com/dani-garcia/bitwarden_rs/blob/master/Dockerfile#L79) by default from Dockerfile? I think this should by up to user to decide with `docker run`.

Also can we get `ROCKET_ENV` set to `prod` or make it user-customizable with  `docker run ... -e ROCKET_ENV=prod...` it should give us [better performance](https://github.com/SergioBenitez/Rocket/issues/315#issuecomment-308723800) so its worth to try.

**EDIT**: I have created [PR](https://github.com/dani-garcia/bitwarden_rs/pull/70).

mprasil:
Hi, it looks like your problem with `EXPOSE` comes from misunderstanding the [EXPOSE directive](https://docs.docker.com/engine/reference/builder/#expose). This just defines the port, that will be exposed when you run container with the `-P` option. It does not prevent you in any way to [specify your own desired port mapping](https://docs.docker.com/engine/reference/run/#expose-incoming-ports). So it's merely a default and that default is set based on the fact that server indeed starts on port 80 if you don't specify otherwise via `ROCKET_PORT`

dustyrip:
Yes, but I still think EXPOSE should not be used in Dockerfiles.  `docker container ls` shows these ports even if not mapped into host so its confusing.

mprasil:
Ah I see what you mean there. Unfortunately AFAIK there's no way to tell docker daemon during startup on which port is the application running internally. I'd still leave it in as it is the correct port as long as you don't deliberately change it in which case you're aware that it's different.

Having the `EXPOSE` there makes it a bit easier to use with some systems that can use the port automatically, so it is useful in some cases. Also for example this is handy when running the container with [rkt](https://coreos.com/rkt/) when you try to [expose the port](https://coreos.com/rkt/docs/latest/networking/overview.html#exposing-container-ports-on-the-host).

mprasil:
I'm for leaving it in if you don't mind @dustyrip. The `EXPOSE` directive is definitely useful in it's default configuration. Systems like Nomad, Traefik or Kubernetes can route to the proper container port based on that information without specifying any port mapping.

PR #70, opened 2018-07-11
Enable user to change ROCKET_ENV for container

dustyrip:
And allow users to specify ROCKET_ENV when starting container, remove default exposing of port 80.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ FROM node:9-alpine as vault
 
 ENV VAULT_VERSION "1.27.0"
 ENV URL "https://github.com/bitwarden/web/archive/v${VAULT_VERSION}.tar.gz"
+ENV ROCKET_ENV "staging"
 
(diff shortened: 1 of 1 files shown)


mprasil:
Hi, I've commented on the `EXPOSE` part in #69 

As for the ROCKET_ENV part, I agree it can be configurable and we shouldn't override it, however for this we just need to add the `ENV` directive and then change the cmd to:

```Dockerfile
CMD ./bitwarden_rs
```
There's no need to re-define the variable again as it is already present from the added `ENV` directive or specified by user via `-e ROCKET_ENV=production`.

Also I'd rather leave the default to staging as that has been the default previously and the change could break people's setup. (AFAIK there is no performance difference between staging and production, just different parts of `Rocket.toml` file are applied)

dustyrip:
Updated.

mprasil:
Thanks for that, I think this is ready to be merged @dani-garcia unless you have something to add?

PR #72, opened 2018-07-12
Add a link to Archlinux AUR packages

mqus:
Hi!
I made two packages for Archlinux, one for the purists (without web interface) and one with the web interface and wanted to link to them here.
I'm not sure if that is the best position to advertise this, let me know if you think the links are better put elsewhere.

Apart from that: If you have any other suggestions/remarks for repackaging, please tell me. 

Currently I'm linking to the latest master commit, but I would really welcome it if you could make a release every now and then.
Thanks for your great work!


mprasil:
This is very cool! Thanks for that. Can we move this somewhere more appropriate though? Perhaps somewhere with the buid instructions?

The problem is, that the readme file is directly used as an description/help for the Docker image by Docker Hub and I don't want to have unrelated information (in the scope of Docker Hub) directly on top.

mprasil:
Thanks, that's perfect. 

Going to merge this if you don't mind @dani-garcia

PR #84, opened 2018-07-14
Reflect changes in Archlinux packaging

mqus:
I changed the way bitwarden_rs is packaged (the web interface is now an addon-package instead of bundled) and added a 'stable' package which follows recent releases.
 I assume that following releases instead of the master branch is encouraged so I removed the link to the (still existing) bitwarden_rs-git package which does the latter.

OT-Comment on the diff: huh? I edited the file in the github-editor but somehow that last line got a change I didn't intent to make and I can't seem to find a difference... strange

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,8 +254,7 @@ For building binary outside the Docker environment and running it locally withou
 
 ### Arch Linux
 
-Bitwarden_rs is already packaged for Archlinux thanks to @mqus. There is an AUR package [with](https://aur.archlinux.org/packages/bitwarden_rs-vault-git/) and 
-[without](https://aur.archlinux.org/packages/bitwarden_rs-git/) the vault web interface available.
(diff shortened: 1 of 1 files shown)


mprasil:
Did you make sure at least include empty index.html file so that the server works without Vault? [There's a check](https://github.com/dani-garcia/bitwarden_rs/blob/55b7a3e4d1e49eb68b22b15dfaa02379755c7d5a/src/main.rs#L138) when server starts.

mprasil:
Btw that last line change is just github adding newline after the last line.

mqus:
vault is disabled per the .env file (which is patched when building) 
It is re-enabled when installing bitwarden_rs-vault. 
If vault is disabled, it doesn't check for the index.html.

mqus:
(see https://aur.archlinux.org/cgit/aur.git/tree/0001-Disable-Vault.patch?h=bitwarden_rs and https://aur.archlinux.org/cgit/aur.git/tree/bitwarden_rs-vault.install?h=bitwarden_rs-vault )

dani-garcia:
Thanks!

PR #92, opened 2018-07-18
Return 404 in case the path doesn't match instead of 500

mprasil:
Currently we return 404 if `WEB_VAULT_ENABLED` is set to `false` and 500 if Vault is enabled. Upstream API responds with 404 for non-existent endpoints, so we should do the same.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bitwarden_rs"
-version = "0.10.0"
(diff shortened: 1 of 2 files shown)


mprasil:
I've also bumped the version for this change ans it somewhat changes the behavior, can you please do a release after it's merged? (and maybe merge the workers PR first so we have both in the new version)

dani-garcia:
Done 👍

PR #104, opened 2018-07-26
Update matrix.to link in the README

jcgruenhage:
Using the room ID instead of an alias isn't supposed to be working for joining rooms, and doesn't work when joining over federation. It only works when your server is already participating in the room.


mprasil:
Thanks for that, we didn't have the proper shortlink set when I updated the readme

PR #107, opened 2018-07-31
Bug/attachments for orgs

shauder:
Fixes attachments when sharing a cipher that contains attachments with collections/orgs


mprasil:
@dani-garcia, I think we should bump the version number after merging this one.

shauder:
I synced the branches/repos and bumped the version up.

dani-garcia:
Alright, thanks for your work!

Issue #99, opened 2018-07-21
PUT /api/ciphers/move not implemented


mprasil:
New vault (#93) calls `PUT` on `/api/ciphers/move` to "Move selected" ciphers. We have `POST` handler [already implemented](https://github.com/dani-garcia/bitwarden_rs/blob/48e69cebab29585684be4bf22282ff1dd4697d24/src/api/core/ciphers.rs#L486) for this, so this should be easy to fix.

krankur:
Hi, I am new to Rust and wanted to start contributing to open source projects. This bug looked like a good beginner task, so I decided to give it a shot. I forked and cloned the project, but on "cargo run" I am getting the following error message:
```
JWT keys don't exist, checking if OpenSSL is available...
OpenSSL detected, creating keys...
Error creating keys, exiting...
```
Can someone please guide me on what can be the issue here?

mprasil:
Hi, I think the program fails to create the keys. Normally they are in the `data` directory, so make sure that whatever user you're running it under can create that folder and/or write to it.

krankur:
So, finally I found out the issue. It was the old openssl version being used. Even though I had `1.0.2l` version installed through brew, the old version at `/usr/local/bin/openssl` was being picked up. I had to create a symlink to the newer version to make it work. https://apple.stackexchange.com/questions/126830/how-to-upgrade-openssl-in-os-x

krankur:
Another question. While doing the setup I was following the instructions in `Build.md` file. Some of the instructions are out of date now, as in the latest vault release, project structure has been changed and the file `settings.Production.json` no longer exists. I had to use the previous release to build the project. Is the build process with the latest vault version documented anywhere?

Also, as per my understanding, the aim of this task is to accommodate the changes introduced as part of the latest vault release. So, to work on this bug I will need to setup the new vault. Is my understanding correct?

dani-garcia:
It’s not documented at the moment, but you can look at the lines 16-25 of the dockerfile to get an idea of the setup.

Basically, clone the 2.0 repo, apply the patch and run the npm commands.

mqus:
I packaged vault for archlinux and as of my knowledge, you don't have to change anything anymore and the build process is as following (taken from build.sh in the vault repo):

	$ npm install
	$ npm run sub:update
	$ npm run dist:selfhost
and the output directory is then `build/`

I have tested it and it worked fairly well but I can still be wrong.

mprasil:
[...]

Issue #98, opened 2018-07-21
DELETE /api/ciphers not implemented


mprasil:
New vault (#93) calls `DELETE` on `/api/ciphers` to "Delete selected" ciphers. We have `POST` handler [already implemented](https://github.com/dani-garcia/bitwarden_rs/blob/48e69cebab29585684be4bf22282ff1dd4697d24/src/api/core/ciphers.rs#L465) for this, so this should be easy to fix.

mannp:
Is the stable release of v2 tagged for the docker image at all?

Currently using the beta build to get v2, but unclear if there is another build I should be using?

Thanks

mannp:
Ok, so i updated my beta to the latest and it still doesn't delete any entries.

mprasil:
This is why this issue was open. 😄 It hasn't been fixed yet, I think I've confused you somewhere into thinking it's been resolved.

mprasil:
Resolved in #115

Issue #97, opened 2018-07-21
Vault 2 is using PUT instead of POST when sharing


mprasil:
As part of testing for #93, I've found, that the share functionality is broken in new Vault. The API endpoint used is `/api/ciphers/{{cipher_id}}/share`. We have the functionality already [implemented](https://github.com/dani-garcia/bitwarden_rs/blob/48e69cebab29585684be4bf22282ff1dd4697d24/src/api/core/ciphers.rs#L331) for `POST` so this should be easy backwards compatible fix.

dani-garcia:
This was fixed by #111

PR #108, opened 2018-07-31
Implementing PUT for 'api/ciphers/move' (#99)

krankur:
Have made changes to support `PUT` call for `/api/ciphers/move` instead of `POST` as  new vault version (2.0) makes a `PUT` call.

### DIFF
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -483,8 +483,9 @@ fn delete_cipher_selected(data: JsonUpcase<Value>, headers: Headers, conn: DbCon
     Ok(())
 }
 
-#[post("/ciphers/move", data = "<data>")]
+#[put("/ciphers/move", data = "<data>")]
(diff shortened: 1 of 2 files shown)


mprasil:
Hi @krankur, you don't want to replace POST with PUT as we still need POST for backwards compatibility. You need to create new route for PUT and call the POST function handler from within.

mprasil:
Almost there, but there's no need to write the function from scratch. You can simply call the other function from inside.

mprasil:
That's perfect.

@dani-garcia I guess this one is ready to merge unless you want the commits squashed to avoid confusion?

dani-garcia:
No, It’s good as it is. Thanks krankur!

krankur:
Thanks everyone for being patient with me. This was my first ever contribution to an open source project. Yay!

mprasil:
@krankur #97 and #98 seem to be still unclaimed, feel free to grab those if you want to contribute some more. The fixes should be pretty similar to what you did just now.

krankur:
Thanks @mprasil . I will look into those.

Issue #100, opened 2018-07-21
PUT /api/ciphers/share not implemented


mprasil:
New vault (#93) calls `PUT` on `/api/ciphers/share` to "Share selected" ciphers. This functionality wasn't present in the previous vault so implementation will be a bit more involved, but should be pretty similar to "move selected".
 
(also woohoo! 100th issue :smile:)

tgturner:
I would love to take a look at this one!

mprasil:
Go for it and don't hesitate to ask for help.

krankur:
I guess we need to add an organization in order to test the "Share" functionality, but I am not able to figure out how to add an organization? I tried uploading a random JSON file as the license file but I am getting a 401 error. Can someone please guide me on this?

mprasil:
Hi, there's no licence required - just create free organization (the first option) and it should have all features after that.

dani-garcia:
I think that happens in the self hosted version (when you build with npm run dist:selfhost)

It should work if you build the normal version (with npm run dist)

mprasil:
@dani-garcia I submitted #109 to merge recent changes from master. We should do this before @krankur touches the sharing API as there were some major changes and will be pain to merge later.

mprasil:
Thanks for merging it. @krankur make sure you're working with latest beta code.

krankur:
@mprasil Will do. Thanks for the heads up.

mprasil:
I think this was resolved in #116

PR #116, opened 2018-08-05
Implemented bulk cipher share (share selected) #100

krankur:
Implemented bulk cipher share "api/ciphers/share", as the new Vault version (2.0) supports bulk share.

### DIFF
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -87,6 +87,8 @@ fn get_cipher_details(uuid: String, headers: Headers, conn: DbConn) -> JsonResul
 #[derive(Deserialize, Debug)]
 #[allow(non_snake_case)]
 struct CipherData {
+    // Id is optional as it is included only in bulk share
+    Id: Option<String>,
     // Folder id is not included in import
     FolderId: Option<String>,
     // TODO: Some of these might appear all the time, no need for Option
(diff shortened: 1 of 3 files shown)


src/api/core/ciphers.rs
```diff
@@ -332,6 +334,65 @@ struct ShareCipherData {
...
+fn put_cipher_share_seleted(data: JsonUpcase<ShareSelectedCipherData>, headers: Headers, conn: DbConn) -> EmptyResult {
+    let mut data: ShareSelectedCipherData = data.into_inner().data;
+    let mut cipher_ids: Vec<String> = Vec::new();
+
+    if data.Ciphers.len() == 0 {
+        err!("You must select at least one cipher.")
+    }
+    
+    if data.CollectionIds.len() == 0 {
+        err!("You must select at least one collection.")
```
mprasil: I'm not sure this is true. I *think* you should be able to share even without collection selected and in this case the cipher should be in Organization but not in specific collection, so only users with access to all Organization ciphers will be able to see it.

This is definitely the case when you share individual cipher, I'd expect it's the same when you "share selected".

krankur: Initially, I didn't add this validation. But, while checking out the "Share Selected" functionality on the official Bitwarden web app, I stumbled across this behavior. If no collection is selected, the server throws an error saying "You must select at least one collection." Same is true for individual cipher share. Maybe, the behavior has been changed in the new version.

mprasil: This is strange, I tested the current beta and client does not complain that the share doesn't have collection selected. (just like before) Could that be some limitation of the Org account you have on the official server? I find it weird, that client wouldn't check for this and it would only be checked on the server side.

krankur: Not sure. I have attached a screenshot of what I see when I try to share without selecting any collection. Please have a look.
screen shot 2018-08-06 at 3 18 13 pm

krankur: Following is the response that I am getting.
screen shot 2018-08-06 at 3 24 29 pm

mprasil: I stand corrected, looks like this [changed in the upstream couple days ago](https://github.com/bitwarden/core/commit/99df132cc7afc47aeaf9dd19109d40413c0397dc#diff-84b6270bfcdf5955cc371f94f8a65b03L160).

mprasil: I've submitted an [issue in the upstream API repo](https://github.com/bitwarden/core/issues/334). To me it still looks like bug in the upstream code. Let's see what Kyle has to say about this.

dani-garcia:
I’ve merged this for now, if the upstream code changes to admit empty collections list, well change it later. Thanks everyone!

Issue #85, opened 2018-07-15
Password hints not available


Gigi1237:
The api endpoint for getting password hints appears to be unimplemented. When either through a client or through the webapp, a user requests a password hint an `Errors have occurred` message pops up.

The server log reports:
```
POST /api/accounts/password-hint application/json; charset=UTF-8:
    => Error: No matching routes for POST /api/accounts/password-hint application/json; charset=UTF-8.
    => Warning: Responding with 404 Not Found catcher.
    => Response succeeded.
```

mprasil:
We do save the hint on account creation and we also have `find_by_mail` implemented for `User`, so this one should be fairly easy to implement.

dani-garcia:
That could be possible, it depends on if the web-vault actually shows the error message sent, because in some cases it just shows a generic error message.

stammw:
Hello,

I'd like to try to solve this one, I'm really excited to try contributing to a rocket-rs based project!

Should I implement it returning the hint into the error response, or could I try to implement a minimal SMTP support ? Following these [implementation](https://github.com/bitwarden/core/blob/b2e6b49b14a98cc2b315a1017cf3473eddba0f33/src/Core/Services/Implementations/HandlebarsMailService.cs#L108) and [configuration](https://github.com/bitwarden/core/blob/master/util/Setup/EnvironmentFileBuilder.cs#L57) ?

mprasil:
Hey, why not both? 😄

Returning the hint in error message is unverified concept. We're not sure the client will display it, but if it's same as the other endpoints, it should. It should be fairly easy to test and implement though, so I'd advise to start with this.

The smtp implementation is going to be quite a lot of code. For example we need to add email verification first otherwise the api could be abused for spam. Also we need to make sure all of this will stay optional.

Gigi1237:
If it is possible to have the client display a custom error message, how about having the server optionally display the hint there? I know it's quite insecure but it could still be useful for small/single user installations where you are mindful of the hint being public, so it's something that meaningful only to you. This would allow forgetful people, like me, to still get hints without the trouble of setting up SMTP.

mprasil:
Yeah, that's the idea behind the first approach. Server can return error json data with the message as one of the fields. Normally the handler on the client side shows the message as an error notification.

stammw:
On what form should the option be handled? There can be a boolean option like "show password hint", or the controller could return the hint only in the case of absence of a valid SMTP configuration?

[...]

PR #125, opened 2018-08-10
Make password hints available in the error message #85

stammw:


stammw:
Sorry, I just clicked by mistake. 

Hope this PR is good OK! :)

.env
```diff
@@ -5,7 +5,7 @@
 # DATA_FOLDER=data
 
 ## Individual folders, these override %DATA_FOLDER%
-# DATABASE_URL=data/db.sqlite3
```
mprasil: You probably didn't want to commit this change?

stammw: Oops, no.

stammw: I just amended the commit, it should be fine now ?

stammw:
I just amended the commit, it should be fine now?

src/api/core/accounts.rs
```diff
@@ -247,3 +247,23 @@ fn revision_date(headers: Headers) -> String {
...
+#[post("/accounts/password-hint", data = "<data>")]
+fn password_hint(data: JsonUpcase<PasswordHintData>, conn: DbConn) -> EmptyResult {
+    let data: PasswordHintData = data.into_inner().data;
+
+    if !CONFIG.show_password_hint {
+        return Ok(())
+    }
+
+    match User::find_by_mail(&data.Email, &conn) {
+        Some(user) => err!(user.password_hint.to_owned().unwrap_or("".to_string())),
```
dani-garcia: You can probably use unwrap_or_default() here, which I think is cleaner.
It would be better also to return a message instead of just the hint, something like:
Your password hint is: HINT

Also, did you test this with the old vault or the new version in beta? We need to make sure if it works in the beta branch, as that would probably be merged soon.

Thanks for your work, stammw!

stammw: I tested it with the old vault I guess. I didn't find a way to build web-2.X, so I used web-1.X.

I'll update the PR with your recommendations.

A pleasure to start working on a real project !

mprasil: Yeah agree with @dani-garcia here, there should be some message explaining the error is not really an error. Something like: 

```rust
        Some(user) => err!(
            format!(
                "Can't send email, password hint: {}",
                user.password_hint.to_owned().unwrap_or_default()
            )
        )
```

dani-garcia:
I've looked at the code of the 2.0 vault and there isn't a way to modify the error message. The only thing I think we can do is to patch the web vault like we do with the URLS (it would probably be a one-line change).

Otherwise, as it is now, there is no way to fix it in the new vault.

@stammw, make the change to the message (something like what @mprasil said) and I'll merge this. Then we can think later how to make it work in the beta version.

Issue #110, opened 2018-08-01
Change email of an account


blacs30:
I am running the latest tag of mprasil's bitwarden docker image.

When I want to change my email address I get asked to enter a code that is supposedly sent to me via mail.
I don't  a receive a mail and on the missing features email is listed so I don't expect mails.

Question:
Is there a way to change the accounts email addres?

mprasil:
There is a [handler](https://github.com/dani-garcia/bitwarden_rs/blob/56b3afa77ca12aba6c3d11edf30b8b16378cbfe5/src/api/core/accounts.rs#L163) `post_email_token` that handles this, it seems to verify your password and some other stuff, but it doesn't really change anything. We need to implement that maybe. We don't have email sending functionality so I guess we could just change it straight away?

blacs30:
From my perspective (user and admin of that bitwarden instance) a direct change would be acceptable. 
If I would host an instance for a lot of people and it's more publicly accessible then an email verification code would make sense for security reasons but I understand that this would be a complete different dimension of time and code.

mprasil:
Well we don't use that email for anything at all really. It's more like an username and bitwarden_rs wouldn't even mind if it wasn't valid email address. So I guess there's little harm in not validating it other than one user can "steal" the account with his email if such email wasn't already registered.

If we decide to just change the email here, it should be pretty simple change.

krankur:
Not sure if I understand the above conversation fully.  As per my understanding, if the user wants to change their email, they need to:

- Go to the "My account" screen
- In the "Change Email" section, enter "Master Password" and "New Email"
- Click ''Continue"

On clicking "Continue", the server API that is called (`api/accounts/email-token`), should update the email. But this doesn't seem right, as the user will be asked to enter the code on clicking "Continue", which will cause confusion. So, am I missing something here?

mprasil:
@krankur you're right it's going to be confusing, but the suggested way to go about it (without implementing and requiring SMTP) would be to just change the email straight away as it is sent and ignore the token part completely.Sure, people will be confused, but at least the email is changed.

mprasil:
Okay, so it turns out this was already [implemented](https://github.com/dani-garcia/bitwarden_rs/commit/1c45c2ec3a716392997b5f60dbbdde77385a7fb2). (without the email verification)

When you want to change your email, you just need to enter any random token when prompted. We should probably document this functionality.

mprasil:
[...]

PR #130, opened 2018-08-12
Changes to change email without email verification code (#110)

krankur:
Changes to change email address on click of "Continue" on "My Accounts" screen under "Change Email", without the need of email verification code.


mprasil:
As per discussion on chat, I'm closing this as it was already implemented before.

PR #134, opened 2018-08-13
Add aliases for PUTs and DELETEs on collections, organizations and org users

mprasil:
The Vault 2 uses PUT instead of POST, this should fix it.


mprasil:
I've added a few more aliases while I was at it.

PR #135, opened 2018-08-14
Deserialize "null" to empty Vec for Collections

mprasil:
When adding/editing user and giving access to all collections, the collection list is now set to _null_ and causes deserialization failure. This change deserializes this back to empty Vec.

### DIFF
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -8,6 +8,8 @@ use db::models::*;
 use api::{PasswordData, JsonResult, EmptyResult, NumberOrString, JsonUpcase};
 use auth::{Headers, AdminHeaders, OwnerHeaders};
 
+use serde::{Deserialize, Deserializer};
+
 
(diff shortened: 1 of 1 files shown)


dani-garcia:
Is there a reason to create a custom deserializer instead of using an optional value?

Or using a serde attribute to set the default?
(https://serde.rs/attr-default.html)

mprasil:
I just wanted to preserve the previous functionality, in which the handler just got vector. I think it captures the difference between getting null and not getting the value at all while keeping the handler code the same.

I don't really care either way. Would any of those preferred by you? (Would the default be used if we get null? Wouldn't it throw error instead?)

dani-garcia:
Well I just checked and you are right, the default is only if it’s missing, not when it’s null.

In that case, I think it’s fine as it is, we’d have to check in the case of using an optional value anyway, so it wouldn’t be much better.

PR #136, opened 2018-08-14
Remove analytics from Vault

### LINKED ISSUE #121: One-time opt-out of Google Analytics
By default, the Bitwarden apps come with Google Analytics activated. To opt-out, you need to visit each app/plugin instance and tick the box. This can really add up: With my three machines, two user accounts on each and two browsers, I need to opt-out twelve times. Surely I've forgotten it on one or two instances. This is problematic from a data protection standpoint, especially in Europe.

Some ideas for solutions:
* Allow to change the Google Analytics setting server-wide, allowing the administrator to set it to opt-in.
* Acknowledge the user's decision and store it as a preference server-side.
* Have the Docker container's entrypoint remove the Google Analytics code from the downloaded Vault's JavaScript code (configurable via environment variable).

mprasil:
This is to resolve #121

### DIFF
--- a/docker/set-vault-baseurl.patch
+++ b/docker/set-vault-baseurl.patch
@@ -1,6 +1,14 @@
 --- a/src/app/services/services.module.ts
 +++ b/src/app/services/services.module.ts
-@@ -119,10 +119,8 @@
+@@ -116,17 +116,15 @@ const exportService = new ExportService(folderService, cipherService, apiService
+ const importService = new ImportService(cipherService, folderService, apiService, i18nService, collectionService);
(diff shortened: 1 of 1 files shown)


dani-garcia:
Does this change allow to create organizations? I remember when it was still in beta it required uploading some kind of license file.

mprasil:
Oh I wasn't aware of that. Need to test this.

mprasil:
Hi, I've pushed new branch, this disables analytics via simple patch and should keep rest of the functionality intact. (I've verified I can still create Organization as usual)

PR #137, opened 2018-08-15
SMTP implementation, along with password HINT email 

stammw:
relates to #85

### DIFF
--- a/src/api/core/accounts.rs
+++ b/src/api/core/accounts.rs
@@ -264,7 +264,7 @@ fn password_hint(data: JsonUpcase<PasswordHintData>, conn: DbConn) -> EmptyResul
     let data: PasswordHintData = data.into_inner().data;
 
     if !is_valid_email(&data.Email) {
-        return Ok(());
+        return err!("This email address is not valid...");
     }
 
     let user = User::find_by_mail(&data.Email, &conn);
--- a/src/api/core/accounts.rs
+++ b/src/api/core/accounts.rs
@@ -5,6 +5,7 @@ use db::models::*;
 
 use api::{PasswordData, JsonResult, EmptyResult, JsonUpcase, NumberOrString};
 use auth::Headers;
+use mail;
 
 use CONFIG;
 
@@ -258,15 +259,23 @@ struct PasswordHintData {
 fn password_hint(data: JsonUpcase<PasswordHintData>, conn: DbConn) -> EmptyResult {
     let data: PasswordHintData = data.into_inner().data;
 
-    if !CONFIG.show_password_hint {
-        return Ok(())
+    let user = User::find_by_mail(&data.Email, &conn);
+    if user.is_none() {
+        return Ok(());
     }
 
-    match User::find_by_mail(&data.Email, &conn) {
-        Some(user) => {
-            let hint = user.password_hint.to_owned().unwrap_or_default();
-            err!(format!("Your password hint is: {}", hint))
-        },
-        None => Ok(()),
+    let user = user.unwrap();
+    let hint = user.password_hint.to_owned().unwrap_or("You don't have any...".to_string());
+
+    if let Some(ref mail_config) = CONFIG.mail {
+        if let Err(e) = mail::send_password_hint(&user.email, &hint, mail_config) {
+            err!(format!("There have been a problem sending the email: {}", e));
+        }
     }
+
+    if !CONFIG.show_password_hint {
+        err!(format!("Your password hint is: {}", &hint));
+    }
+
+    Ok(())
 }
--- a/src/api/core/accounts.rs
+++ b/src/api/core/accounts.rs
@@ -265,15 +265,16 @@ fn password_hint(data: JsonUpcase<PasswordHintData>, conn: DbConn) -> EmptyResul
     }
 
     let user = user.unwrap();
-    let hint = user.password_hint.to_owned().unwrap_or("You don't have any...".to_string());
+    let hint = match user.password_hint {
+        Some(hint) => hint,
+        None => return Ok(()),
+    };
 
     if let Some(ref mail_config) = CONFIG.mail {
         if let Err(e) = mail::send_password_hint(&user.email, &hint, mail_config) {
             err!(format!("There have been a problem sending the email: {}", e));
         }
-    }
-
-    if !CONFIG.show_password_hint {
+    } else if CONFIG.show_password_hint {
         err!(format!("Your password hint is: {}", &hint));
     }
 
--- a/src/api/core/accounts.rs
+++ b/src/api/core/accounts.rs
@@ -82,7 +82,10 @@ fn post_profile(data: JsonUpcase<ProfileData>, headers: Headers, conn: DbConn) -
     let mut user = headers.user;
 
     user.name = data.Name;
-    user.password_hint = data.MasterPasswordHint;
+    user.password_hint = match data.MasterPasswordHint {
+        Some(ref h) if h.is_empty() => None,
+        _ => data.MasterPasswordHint,
+    };
     user.save(&conn);
 
     Ok(Json(user.to_json(&conn)))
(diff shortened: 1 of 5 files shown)


stammw:
Hello, 

Do you think I should implement mail body templating to match bitwarden's email Look ? Or having the same message in text content would be sufficient ? Can I use rocket's template feature to do that?

src/api/core/accounts.rs
```diff
@@ -258,15 +259,23 @@ struct PasswordHintData {
...
+        }
     }
 
-    match User::find_by_mail(&data.Email, &conn) {
-        Some(user) => {
-            let hint = user.password_hint.to_owned().unwrap_or_default();
-            err!(format!("Your password hint is: {}", hint))
-        },
-        None => Ok(()),
+    if !CONFIG.show_password_hint {
```
mprasil: Isn't this if actually other way around?

src/api/core/accounts.rs
```diff
@@ -258,15 +259,23 @@ struct PasswordHintData {
...
 
-    if !CONFIG.show_password_hint {
-        return Ok(())
+    let user = User::find_by_mail(&data.Email, &conn);
+    if user.is_none() {
+        return Ok(());
+    }
+
+    let user = user.unwrap();
+    let hint = user.password_hint.to_owned().unwrap_or("You don't have any...".to_string());
```
mprasil: I think we should stop straight away here if there's no hint and just return `Ok(())`. (AFAIK the client will say something like "If there was hint set for this email, you will receive it", so that would be consistent with the client side)

Alternatively at least be more specific. Something like "No hint was set for this account" would be more appropriate. If the user receives message saying: 

> Your hint is:  "You don't have any..."

They might think this is an actual hint and try to figure out what they don't have that could be password. :smile:

mprasil:
Hi, thanks for that, that was faster than expected. :smiley:  I've added my comments to the code. As for templating the email, I think you can leave that for another PR. 

I have two worries here:
1) Why is the username and password mandatory? what about setups where SMTP authorization isn't required?
2) We need to verify or at least validate the email first. As it is now, it might not even be valid address format as we didn't validate it before. Not sure if that's verified anywhere.

stammw:
1. I admit I haven't been very interested into implementing this with a non credential based setup in mind, as setting up the test environment it is more complex. But I'll try to do something better on this side.

2. Do you mean that verifying the SMTP setup on startup would be the best solution ? I think the email is verified on the frontend, and the best place to do it is [in the registration](https://github.com/dani-garcia/bitwarden_rs/blob/19e0605d30384e04b5bba4a3aa34d40619f61add/src/api/core/accounts.rs#L34)

src/api/core/accounts.rs
mprasil: Your code is:

```
    if !CONFIG.show_password_hint {
        err!(format!("Your password hint is: {}", &hint));
    }
```
This would show hint if `show_password_hint` is false. (remember that `err!` shows hint in old vault) It should be other way around shouldn't it?

mprasil: I think if the email fails, it should follow the `show_password_hint` setting and only show the hint if it's set to true.

mprasil:
1) As for testing with local smtp and no authorization, maybe try [this one](https://hub.docker.com/r/gessnerfl/fake-smtp-server/)? (if you use docker) I think it's pretty common scenario where you have locally running smtp relay, that requires no authorization and that forwards mail somewhere appropriate, so it would be good to have that as an option if you'd be willing to implement that. (or leave it for another PR if you feel like it's too much work)

2) We can't rely on client side verification, you can hit the API directly and register with any non-email "email" right now. We should probably verify it during the registration also. (we just didn't care so far, because we didn't really use the address as address) But that wouldn't solve the problem for accounts that are already registered. What I'm wondering is, what will happen when you try to send an hint for an account that has malformed address? If it just fails, maybe it's okay I'm more worried about possible attack vectors like inserting custom mail headers that way and such. Looking at lettre it might already handle that? If it does, ignore my comment entirely. :wink:

dani-garcia:
Thanks for your work, stammw!

After a quick revision, it seems to be going the right way.

The SmtpTransport created by the mailer method can probably be reused, to avoid creating multiple connections each time.

I think the best way to do that would be to let Rocket manage it, like we are doing with the database connection pool already.

https://rocket.rs/guide/state/

mprasil:
@dani-garcia Not sure if this isn't over-engineering things a bit. Remember the number of calls generating email is going to be fairly low. (unlike DB where virtually any call results in DB connection)

I mean if @stammw is up for the challenge of wrapping this via r2d2 and request handler, by all means do this, it will be nice exercise, but realistically your typical server will probably see one email a day on average with hundreds of users.

dani-garcia:
Creating a connection pool would probably be overkill, yes.

But keeping only one SmtpTransport and setting it to reuse a connection instead of opening and closing each time seems like a reasonably small change.

Anyway, if the code works fine (I can’t really test it at the moment), I’m fine with leaving it as it is, it’s true that there won’t be a ton of calls sending mails.

About the emails contents, I’m ok with just a plain text message, maybe we could use some HTML to make links clickable, but other than that, I don’t really see the point.

mprasil:
I don't think that lettre actually does keepalive connections. Most smtp servers would close it after few seconds of inactivity. However if @stammw is up for doing this, it would have one benefit - the code would be much nicer and we could easily identify handlers that can use SMTP just by searching for appropriate request guard. So efficiency aside, the code would be nicer for sure.

stammw:
@dani-garcia Great! Happy to read that! 

About the connection pool, lettre-rs has an [open issue for an r2d2 implementation](https://github.com/lettre/lettre/issues/193). I'll maybe give a ping there.

Yes do it through a request_guard would be much more rocket-rs style compliant.

Thank you!

src/main.rs
```diff
@@ -153,6 +158,57 @@ lazy_static! {
...
+        
+        // When SMTP_HOST is absent, we assume the user does not want to enable it.
+        if smtp_host.is_none() {
+            return None
+        }
+
+        let smtp_ssl = util::parse_option_string(env::var("SMTP_SSL").ok()).unwrap_or(true);
+        let smtp_port = util::parse_option_string(env::var("SMTP_PORT").ok())
+            .unwrap_or_else(|| {
+                if smtp_ssl {
```
dasjoe: Please always submit mails on port 587, port 25 is for SMTP relaying only.
See https://www.mailgun.com/blog/25-465-587-what-port-should-i-use

stammw:
Hi @dani-garcia, if you want I can help setting a testing environment for validation.

dani-garcia:
Sorry for the wait, I completely forgot that we had this pending PR!

I've tested it with my GMail account and it works perfectly!
One small note, though: I think we should do something other than fail silently when there is no hint. Either error and show it on the webpage or send an email saying there is no hint available. Otherwise it may lead someone to believe that the email system isn't working. 

We could also print an error when the email isn't valid, as it wouldn't pose any kind of security threat and may notify the user of a typo on their part.


As a note for anyone else, the GMail config is as follows:
```
SMTP_HOST=smtp.gmail.com
SMTP_PORT=25
SMTP_SSL=true
SMTP_FROM=<user>@gmail.com
SMTP_USERNAME=<user>@gmail.com
SMTP_PASSWORD=<App-Specific-Password>
```

And the password is obtained from https://myaccount.google.com/apppasswords

mprasil:
I'd be careful with the invalid email error, it might lead to email/username enumeration weakness if there isn't some sort of rate limit.

dani-garcia:
Yeah, that's why I said only when invalid and not when the email is not asocciated to a user. In that second case we would have an enumeration exploit, but I don't think erroring when an invalid email is introduced would cause any problems.

mprasil:
Ah you mean invalid as in not in valid format? That makes sense.

stammw:
Here are the changes you requested. I took the liberty to change `/accounts/profile` behavior, saving `None` when the password hint is empty. I hope this makes sense.

dani-garcia:
Alright, I think this is ready to be merged. We'd probably need some documentation in the README before the next release, but we can worry about that later.

Thanks a lot for your work, stammw!

PR #140, opened 2018-08-15
Update the error format to show message in new Vault

mprasil:
This follows the new error object format as observed in upstream implementation.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -3,19 +3,20 @@
 ///
 #[macro_export]
 macro_rules! err {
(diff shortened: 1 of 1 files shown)


dani-garcia:
Can you test another one of the apps to see if they also follow this new style? (Mobile, desktop, browser).

I think the easiest case to test would be invalid password.

If they use the old style maybe we should merge both for backwards compatibility.

mprasil:
From what I understood, the Vault was the only old style client before 2.0. I can test with Desktop for sure.

PR #146, opened 2018-08-21
Update affected users revision on cipher and folder change

mprasil:
This should update revision for all affected users on cipher and folder changes. The `find_by_cipher_and_org` definitely needs another pair of eyes to double-check. (from my limited testing it does what it should)


dani-garcia:
In some light testing it seems to work fine. As it isn’t doing anything problematic (only updates the date) I merged it. If there are any rough edges, well fix them as they appear

PR #148, opened 2018-08-21
Merge Beta to master

mprasil:
I believe all of the bugs we spotted with vault 2.0 are now resolved. 

This also bumps up version to 1.0.0 to mark new epoch :champagne:


dani-garcia:
Alright, it's done. Good work, everyone! 🎉

PR #150, opened 2018-08-23
Update the build instruction for new Vault

mprasil:


mprasil:
I've changed the build instructions to clone repository instead of downloading archive. Thanks to @mqus for raising the issue in #151

PR #151, opened 2018-08-23
Update BUILD.md for web-vault 2.x.x

mqus:
Update BUILD.md to reflect changes to the web-vault build process

I switched to `git clone` instead of unpacking release archives because there is a git submodule (jslib) which is not packed into the archives. One could also download the specific commit archive of [bitwarden/jslib](https://github.com/bitwarden/jslib) for the release archive and unpack it into the `jslib` directory, but this would be a much more complex process, because we would also have to explain how to find the matching commit hash.

I also removed the instructions to change upstream files because `npm run dist:selfhost` already applies everything changed by the docker patch, if I'm not mistaken(See [bitwarden/web/package.json](https://github.com/bitwarden/web/blob/master/package.json#L15) for setting the env variables and [services.modules.ts](https://github.com/bitwarden/web/blob/master/src/app/services/services.module.ts#L127) which uses it).

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -17,28 +17,22 @@ cargo build --release
 When run, the server is accessible in [http://localhost:80](http://localhost:80).
 
 ### Install the web-vault
-Download the latest official release from the [releases page](https://github.com/bitwarden/web/releases) and extract it.
-
-Modify `web-vault/settings.Production.json` to look like this:
-```json
-{
-  "appSettings": {
-    "apiUri": "/api",
-    "identityUri": "/identity",
-    "iconsUri": "/icons",
-    "stripeKey": "",
-    "braintreeKey": ""
-  }
-}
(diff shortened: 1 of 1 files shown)


mqus:
whoops, missed #150 . But my first point still stands

mprasil:
Hey, good catch with the clone vs download, however you also want to avoid `npm run dist:selfhost` as that will basically disable Organizations part of functionality. Instead you need to apply the patch. (see my PR)

I can change my PR to use `git clone` instead of release download or you can update your build steps, I don't mind either way.

mqus:
Then I'll leave it up to you! I struggled with the clone vs download yesterday when packaging for archlinux ;)

mprasil:
👍  Okay, thanks for raising the issue, I'll use your approach there.

PR #152, opened 2018-08-25
Add ip and username to failed login attempts

### LINKED ISSUE #119: Feature Request: include IP and username in logging for failed attempts
# Feature Request
### Issue
Currently, failed login attempts are logged like so: 
```json
{"log":"ERROR: Username or password is incorrect. Try again.\n","stream":"stdout","time":"2018-08-08T19:33:20.892869034Z"}
```

Ideally, these log messages would include the IP address of the host attempting login, and username that was attempted.  Including this information would allow for better monitoring and alerting, as well as blocking of bad actors.

### Suggestion
Include IP address and username in log message.
Example:
```json
{"log":"ERROR: Username or password is incorrect. Try again.\n","IP":"192.0.2.23","username":"foo@example.com","stream":"stdout","time":"2018-08-08T19:33:20.892869034Z"}
```

Baelyk:
Resolves #119

I added the IP and Username of the failed login attempt to the error message describing the attempt, as request in issue #119.

### DIFF
--- a/src/api/identity.rs
+++ b/src/api/identity.rs
@@ -1,4 +1,5 @@
 use std::collections::HashMap;
+use std::net::{IpAddr, Ipv4Addr, SocketAddr};
 
 use rocket::request::{self, Form, FormItems, FromForm, FromRequest, Request};
 use rocket::{Outcome, Route};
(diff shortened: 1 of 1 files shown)


dani-garcia:
Thanks for this, Baelyk!
I just tested this and it seems to work fine when running standalone so I'll merge it now.
For those behind proxy, you might need to configure the proxy to create a `X-Real-IP` header with the external IP of the request.

PR #159, opened 2018-08-28
Allow non-Admin user to share to collection (fixes #157)

### LINKED ISSUE #157: User type 'User' cannot share password with collection
A user who has access to the collection (and not 'Readonly') cannot share a password with that collection -- when attempting to do so the following error is shown:
image

The only work-around is to set the user type to 'Admin' (or 'Owner'), however this allows access to all collections regardless of the explicit collection permissions

mprasil:
This should allow non-admin user to share directly to collection, if they have write access at least there.

### DIFF
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -132,8 +132,8 @@ fn post_ciphers(data: JsonUpcase<CipherData>, headers: Headers, conn: DbConn) ->
     Ok(Json(cipher.to_json(&headers.host, &headers.user.uuid, &conn)))
 }
 
-fn update_cipher_from_data(cipher: &mut Cipher, data: CipherData, headers: &Headers, is_new_or_shared: bool, conn: &DbConn) -> EmptyResult {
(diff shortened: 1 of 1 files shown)


src/api/core/ciphers.rs
```diff
@@ -147,6 +147,9 @@ fn update_cipher_from_data(cipher: &mut Cipher, data: CipherData, headers: &Head
         } else {
             cipher.user_uuid = Some(headers.user.uuid.clone());
         }
+    } else if let Some(org_uuid) = data.OrganizationId { // Shared to collection, but still belongs to Organization
+        cipher.organization_uuid = Some(org_uuid);
```
dani-garcia: We aren’t validating that the user is part of the org here, this could allow a user to update a  local cipher (so the is_new is false) and make it part of an organization, no?

mprasil: You're right, I'll have to split the Boolean into two and handle accordingly. It's gonna be more readable anyways.

PR #167, opened 2018-09-01
Add support for cipher update notifications

shauder:
This I think covers all the update cipher options (not create/delete)

EDIT: Now includes create and delete updates for ciphers

### DIFF
--- a/src/db/models/cipher.rs
+++ b/src/db/models/cipher.rs
@@ -130,19 +130,25 @@ impl Cipher {
         json_object
     }
 
-    pub fn update_users_revision(&self, conn: &DbConn) {
+    pub fn update_users_revision(&self, conn: &DbConn) -> Vec<String> {
+        let mut user_uuids = Vec::new();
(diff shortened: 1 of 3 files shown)


src/db/models/cipher.rs
```diff
@@ -358,4 +358,21 @@ impl Cipher {
         .select(ciphers_collections::collection_uuid)
         .load::<String>(&**conn).unwrap_or(vec![])
     }
+
+    pub fn get_users(&self, conn: &DbConn) -> Vec<String> {
```
mprasil: This function is nearly identical with `update_users_revision`. Can you maybe rewrite the update function to just use results from this function? The idea is that we avoid duplicating the code.

shauder: I had originally thought of doing that but wasn't sure the name fit for what it did but I have made this change in the latest commit

dani-garcia:
Thanks, shauder!

PR #170, opened 2018-09-04
Fix editing users in Organization

### LINKED ISSUE #163: As an org admin, assigning myself (or other admin) to a collection results in an error
Reproduction steps:
- Using the latest docker container.
- Log in to the web UI as user who is an organization admin.
- Select organization, click on the Manage tab, select People.
- Click on yourself or another organization admin.
- Under "Access Control" assign the user to any collection or select "This user can access and modify all items".
- Click "Save".

Expected result:
The request succeeds and the edited user gains access to the selected collections.

Actual result:
Red error message pops up saying "Only Owners can grant Admin or Owner type".

Background:
Failing request is:
PUT https://<HOSTNAME>/api/organizations/<ORG_ID>/users/<USER_ID>

mprasil:
This fixes multiple issues:

1. [Make sure](https://github.com/mprasil/bitwarden_rs/blob/049aa33f179076a6b1ba3b71351b5d7b95ef4fdc/src/api/core/organizations.rs#L481) the user is actually member of organization we're managing.
2. Fix #163 by [allowing](https://github.com/mprasil/bitwarden_rs/blob/049aa33f179076a6b1ba3b71351b5d7b95ef4fdc/src/api/core/organizations.rs#L486-L497) admin users to edit other admin users.

    The implemented change allows admin users to change collection access for other admin users and themselves. Admin user is however not allowed to promote other users to Admin type or drop Admin privileges for other admin users.

3. Change `user_id` dynamic parameters to `org_user_id` as this better reflects the actual information that's provided.

### DIFF
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -217,7 +217,7 @@ fn delete_organization_collection_user(org_id: String, col_id: String, org_user_
         }
     };
 
-    match UserOrganization::find_by_uuid(&org_user_id, &conn) {
+    match UserOrganization::find_by_uuid_and_org(&org_user_id, &org_id, &conn) {
         None => err!("User not found in organization"),
         Some(user_org) => {
             match CollectionUser::find_by_collection_and_user(&collection.uuid, &user_org.user_uuid, &conn) {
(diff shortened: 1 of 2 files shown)


mprasil:
Added another commit and changed few other checks that could be theoretically abused for privilege escalation.

src/api/core/organizations.rs
```diff
@@ -464,33 +456,36 @@ struct EditUserData {
...
+    let mut user_to_edit = match UserOrganization::find_by_uuid_and_org(&org_user_id, &org_id, &conn) {
         Some(user) => user,
         None => err!("The specified user isn't member of the organization")
     };
 
-    if new_type != UserOrgType::User as i32 &&
+    if new_type != user_to_edit.type_ as i32 && (
+            user_to_edit.type_ >= UserOrgType::Admin as i32 ||
+            new_type >= UserOrgType::Admin as i32
+        ) &&
```
dani-garcia: Isn't the comparison backwards here? The >= in 479 and 480 should be <=, because the struct values are sorted the other way (Owner = 0, Admin = 1, User = 2). At least if I understand the logic correctly.

mprasil: You're right. I got confused there.

mprasil:
Updated the logic there.

Issue #40, opened 2018-06-12
Implement poor man's invitation-only registration


mprasil:
In current state one can either enable or disable user registration. I think it should be possible to implement some exception to the rule:

Admin and Owner of the Organization should be able to invite users not currently registered. For these special record would be made (possibly in a separate table?) that would allow registration of an user with specified email even if the registration is otherwise turned off. That way admin could turn off  the self-registration, but he could still invite users if needed without having to restart Bitwarden to enable/disable it as it is implemented right now.

I'm not sure if I understand the invitation process on the client side properly, but I think it shouldn't clash with how it works right now.

Any thoughts?

dani-garcia:
I thought about a similar solution to yours some time ago. The problem is that we don't validate emails at all, so anyone that knows the email of the invited user could register an account first.
I guess the danger of that would be pretty low, knowing the user needs to be confirmed by an admin o owner to have access to the organization, so I'm okay with implementing this as long as there is a way to disable it.

Your idea seems alright, we would need to create a table for the invites (with email, organization id, user type, and collections the user can access) and basically delay the creation of the userorganization until the user is created. The problem with this is the invites would not appear on the organization user list, because that only shows userorganization and those don't exist for the invited users yet. We could change it so it shows both the userorganization and the invites but that seems like a hack.

Another option would be to pre-create the user and mark it as disabled until the user registers and the admin confirms the invite, which to me seems like the best option, but we don't have a disabled flag in the users table and I don't feel like doing an entire migration for that (maybe we can assume that if the masterPasswordHash field is empty that means the user is disabled or something).

NickBusey:
Quick question, how does one disable user registration? I don't see that option in the Readme.

mprasil:
Just set

mprasil:
Apologies, missclicked close.

dani-garcia:
What @mprasil said is correct. You can also modify the `.env` file if you prefer. I've added a short explanation of configuration options to the README in the last commit.

jakob42:
I would love that feature as well, but two things would mitigate the problem in a different fashion for me and might be easier:

- Either have an option to send a notification mail to a defined admin address every time a user registers
- Or have an option with an array of email domains allowed (my couple users would come from two domains only.)

Thanks for your work guys!

NickBusey:
I would say have an array of emails OR email domains.

So I could specify `gmail.com` or just `user@gmail.com`

ViViDboarder:
[...]

PR #173, opened 2018-09-10
Implement poor man's invitation via Organization invitation

mprasil:
This should allow invitation-only registration via Organization invitations. The way it works follows the suggested approach [outlined](https://github.com/dani-garcia/bitwarden_rs/issues/40#issue-331759674) in #40. When Admin/Owner invites user with an email that is not found, a new record in `invitations` table is created. At the same time an `User` is created with random password and key generated just as a "placeholder".

When user tries to register, they will be allowed to do so even if SIGNUPS_ALLOWED is set to false if their email is in the invitations table.

This should also allow user to register without the placeholder `User` already being present, which would accommodate some external integration like the one [suggested](https://github.com/dani-garcia/bitwarden_rs/issues/40#issuecomment-405399443) by @ViViDboarder. But we'd have to document that kind of usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ _*Note, that this project is not associated with the [Bitwarden](https://bitward
   - [Updating the bitwarden image](#updating-the-bitwarden-image)
 - [Configuring bitwarden service](#configuring-bitwarden-service)
   - [Disable registration of new users](#disable-registration-of-new-users)
+  - [Disable invitations](#disable-invitations)
   - [Enabling HTTPS](#enabling-https)
   - [Enabling U2F authentication](#enabling-u2f-authentication)
   - [Changing persistent data location](#changing-persistent-data-location)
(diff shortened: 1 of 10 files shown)


dani-garcia:
Is there a reason for generating a random password and hashing it instead of just setting the password_hash to an empty vec directly? That way there is no need for a new dependency.

Also, all the Invitation::find_by_mail are followed by a delete, so maybe we can create a method take_by_mail or something that returns the invitation if there is one and also deletes it. It would remove some matches and make the code easier to follow. (Maybe we can just use the diesel::delete function in that hypothetical take_by_mail, which I think returns the number of rows deleted, as a check that the invitation existed).

mprasil:
The reason mostly is, that User::new() will take that empty vec and will hash it to save as password hash, so this would in theory add valid user with blank password, that could actually log in. I'm not sure if that's bad thing. I can definitely remove that part if you think that makes sense.

There is one exception to that delete rule: When the invitation is present, but the `INVITATIONS_ALLOWED` are set to false. I didn't want to delete the invitation in case the invitations are allowed later on.

dani-garcia:
Well, if you use the User::new method as it is, then yeah anyone sending a blank masterPasswordHash could log in, because that would produce a valid encrypted hash. But if you directly set the password_hash vec to empty, bypassing that method, then the crypto::verify_password_hash method would fail (we could even add a check in there to make sure, but no input should produce and empty vec hash).

About the invitations thing, that could be solved by checking the INVITATIONS_ALLOWED flag first, that way we also do one less DB call.

mprasil:
Thanks a lot for your input. I've rewrote a lot of the implementation to hopefully make it nicer. I've also added documentation and I've added code to make invited, but unregistered users handled as "Invited" and not "Accepted" which makes it work a bit better in the Org admin interface.

I've squashed all the commits into one as it was already a mess.

dani-garcia:
I think it does look nicer now.

The user::new_invited won't work correctly like that, that empty string would get hashed correctly here:
https://github.com/dani-garcia/bitwarden_rs/blob/master/src/db/models/user.rs#L48.

As it is anyone sending an empty masterPasswordHash could login in that account.

I think the best solution would be to remove the password parameter from the user::new function and set the password_hash to Vec::new(), then in the cases where we do set the password, we would explicitly call the already existing set_password function (Which I think it's only done in the register method).

mprasil:
Oh, my bad! I forgot to handle the hash manually as you suggested while working on the other changes.

mprasil:
OKay I've did the changes. We now need to set the key and password in extra steps, the `User::new()` will set those to blank.

dani-garcia:
Okay, looking great now, thanks!

PR #175, opened 2018-09-11
Documentation for SMTP and password hint configuration

stammw:


README.md
```diff
@@ -248,6 +250,36 @@ docker run -d --name bitwarden \
   mprasil/bitwarden:latest
 ```
 
+### SMTP configuration
+
+You can configure bitwarden-rs to send emails via a SMTP agent:
```
fbartels: The project usually refers to itself as bitwarden_rs, not bitwarden-rs

README.md
```diff
@@ -248,6 +250,36 @@ docker run -d --name bitwarden \
...
+  -v /bw-data/:/data/ \
+  -p 80:80 \
+  mprasil/bitwarden:latest
+```
+
+When `SMTP_SSL` is set to `true`(this is the default), only TLSv1.1 and TLSv1.2 protocols will be accepted and `SMTP_PORT` will default to `587`. If set to `false`, `SMTP_PORT` will default to `25` and the connection won't be encrypted. This can be very insecure, use this setting only if you know what you're doing.
+
+### Password hint display
+
+Usually, password hints are sent by email. But as bitwarden-rs is made with small or personal deployement in mind, hints are also available from the password hint page, so you don't have to configure an email service. If you want to disable this feature, you can use the `SHOW_PASSWORD_HINT` variable:
```
fbartels: For instead of or

mprasil: "or" makes more sense IMO.

stammw: I'll leave it as it is then.

fbartels: ok, was maybe a bit too early for me then. indeed both make sense. Then there is just the deployment typo in this line.

mprasil:
@stammw there are some merge conflicts, would you mind rebasing your branch on top of current master to make it a bit easier to merge?

dani-garcia:
I solved the merge conflicts, apparently Github has a nice editor to solve them, hopefully that won't break anything.

Anyway, I'll merge this now and create a new release, as it's been some time since the last one.

Thanks for your work, everyone!

PR #182, opened 2018-09-14
Update README.md

dobunzli:
Added infos about enabling https when softwares getting certs are using symlinks


dani-garcia:
Thanks!

PR #183, opened 2018-09-14
traefik proxy example

jkaberg:
Added an example on how to proxy Bitwarden properly with [Traefik](https://traefik.io/)


dani-garcia:
Thanks, jkaberg!

PR #191, opened 2018-09-18
Update Vault to v2.3.0

mprasil:
The patch seems to apply correctly and it seems to build fine, but I haven't yet tested this extensively, so maybe try to test it first before merging.


dani-garcia:
With some light testing, it seems to work fine. I checked the commits after the 2.2 release and it seems the biggest change is that now the kdf parameters are user configurable from the settings page, but that doesn't break anything and we can implement it later. Thanks!

PR #209, opened 2018-10-05
Return proper error message for org reinvite

### LINKED ISSUE #188: Resend invitation functionality not implemented
This popped up in #187, when inviting users that aren't registered yet, they will be set to `Invited` state. In this state there's an option to resend invitation in the Organization admin interface:

image

We should handle this API call more graciously as it's failing with `404` at the moment. We should return proper error message explaining the functionality is not implemented.

janost:
Fixes #188.

Displays an error message in the log when the user clicks the `Resend invitation` option for a user on the Organization/People admin panel.

### DIFF
--- a/src/api/core/mod.rs
+++ b/src/api/core/mod.rs
@@ -112,6 +112,7 @@ pub fn routes() -> Vec<Route> {
         put_organization_user,
         delete_user,
         post_delete_user,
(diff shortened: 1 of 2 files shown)


mprasil:
Thanks a lot @janost

PR #213, opened 2018-10-06
CollectionCipher::save() and delete() should return QueryResult instead of bool

janost:
Related to #6.


mprasil:
I think the error messages could be a little more descriptive. Something like "failed to add cipher to collection" or "remove from" for the opposite action.

The idea is that this error will be also displayed in the client application and users have no idea what CollectionCipher is as this is just mapping done "behind the scenes".

janost:
@mprasil thank you for the review. I changed the error messages.

dani-garcia:
Thanks!

Issue #212, opened 2018-10-06
Admin Panel


joe217:
Hi, 

Any thoughts on getting the admin panel working?

Usually accessible at my.bitwardendomain.com/admin?

fbartels:
I think that has been previously discussed at https://github.com/dani-garcia/bitwarden_rs/issues/202

mprasil:
Yeah, there actually isn't much we could use from upstream in this case and generally is too much work for little benefit. If there's something specific you'd like to do in it, it's usually easy enough to do directly with sqlite client on the db.

Feel free to ask if you want to do something specific.

I'm closing this as duplicate of #202

joe217:
Thank you for the prompt response...and my apologies for missing the recent similar post.

Two quick follow-up quesitons:
1. Isn't the database encrypted?  How would I list the user names?
2. If the user names are unencrypted, could you list a sample CLI command that I could use to list them?  I tried doing some research on sqlite, but may be a bit above my head.  A few simple commands may point me in the right direction for making the actual bash scripts...

Thanks!

fbartels:
@joe217 the linked ticket has this example: `sqlite3 data/db.sqlite3 "select * from users;"`

joe217:
Ok. Thank you. Figured that out now. That produces mostly gibberish,
though. The username emails are in plaintext. How would I go about deleted
them?

Also, would working on the bw-data file be equivalent to working in the
actual container?

Thanks.

fbartels:
You can find a comprehensive tutorial on sqlite at https://www.tutorialspoint.com/sqlite/

The reason for the "gibberish" is that it outputs all the data in the table. Looking at it a bit closer you the following probably give a more meaningful output `sqlite3 data/db.sqlite3 "select uuid,email from users;"`

In the end you probably want something like: `sqlite3 data/db.sqlite3 "delete from user where email='bla@company';"` (untested)

mprasil:
Careful with the deletes though, the user has a lot of attached objects in the DB. The best thing to do would be to change his password to some string like "deactivated" or something like that. This will never match the hashed passphrase and will effectively disable the account.

joe217:
https://github.com/dani-garcia/bitwarden_rs/issues/212#issuecomment-428093496

`sqlite3 data/db.sqlite3 "select uuid,email from users;"

[...]

Issue #202, opened 2018-09-29
Support for admin panel


plasmite:
Use this every day! Thank you for your excellent work!

Are there any plans to support the admin panel for user management [as described here](https://help.bitwarden.com/article/admin-portal/)? Or is there a way to enable it in the current implementation? The instructions on the linked page do not appear to apply and I do not see any env variables that would allow enabling this feature.

Thank you!

dani-garcia:
The admin panel source is not included with the web vault, and as far as I know it's not publicly available, so we'd have to recreate it from scratch.

At the same time, I don't think implementing this is worth it, to be honest. It would probably be easier to view or edit the sqlite database directly. And if someone needs to view or delete users and organizations very often, they could just create a bash script. Something like:

```sh
sqlite3 data/db.sqlite3 "select * from users;"
```

plasmite:
Thanks I figured this was the case--was poking around in the db already but didn't want to go mucking anything up if there was an alternative. Thanks again!

PR #216, opened 2018-10-12
Implement poor man's admin panel

mprasil:
This enables you to configure one email account to be server administrator via the `SERVER_ADMIN_EMAIL` environment variable. This will give the user extra functionality and privileges to manage users on the server. In the Vault, the user will see a special (virtual) organization called `bitwarden_rs`. Deleting users from this organization will actually completely delete the user from the server. Inviting users into this organization will just invite the user so they are able to register, but will not grant any organization membership.

This should hopefully cover use cases like the one in #212 and #202 without actually creating the admin panel.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ _*Note, that this project is not associated with the [Bitwarden](https://bitward
 - [Configuring bitwarden service](#configuring-bitwarden-service)
   - [Disable registration of new users](#disable-registration-of-new-users)
   - [Disable invitations](#disable-invitations)
+  - [Configure server administrator](#configure-server-administrator)
   - [Enabling HTTPS](#enabling-https)
   - [Enabling WebSocket notifications](#enabling-websocket-notifications)
   - [Enabling U2F authentication](#enabling-u2f-authentication)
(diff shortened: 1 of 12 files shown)


Biepa:
Great! Thank You :tada: :grinning:

dani-garcia:
Alright, I'll create a new release now that this is merged.

jfreak53:
Quick question on this, will this list orgs that other users have created allowing me to remove them? Also when removing a user, will it erase all their passwords, and other info from them from the db?

mprasil:
No, this will not show any organizations created, just users. And yes it will delete user with all attached data just as if the user deleted their own account from Vault.

This is the reason it might fail to delete users sometimes - when they are the only owner of some organization. (It doesn't want to leave lingering organization)

jfreak53:
Ok, thanks. Is there anyway in a future release to get it to list user orgs, so admins can remove? I am using in a corporate environment and we need control of the server. I can do SQL but I'd prefer if the panel did it for quickness.

My1:
sorry for necro but I think here seems the best place to ask, might it be an interesting Idea to create a "meta orgtanization" of admins on first setup which the first user belongs to and can add any other users and users of that orga can go into the admin panel?
how does that sound

dani-garcia:
By meta organization do you mean an organization inside the bitwarden instance? (Instead of the separate admin interface with a separate admin token).

If so that's what this PR did initially, but eventually it was replaced with the current admin panel because we were limited by what we could do with the clients and also because the clients didn't like having this fake meta organization, which would cause subtle bugs.

To do something similar to what you want, you could start the server with SIGNUPS_ALLOWED=false and INVITATIONS_ALLOWED=true, and then invite all the users from the admin panel.

My1:
almost.

basically the idea would be keeping the admin panel but having this organization (which would be pretty normal for all other purposes except for regulating the admin access) but having a normal login over there and users in that organization can log in onto it, basically just swap the admin token with a login and boom over

no idea what the specifics of the meta organization are and why clients didnt like it but if the organization would behave normal for all other intents and purposes it would be weird for bugs to appear

My1:
also it's not about inviting new users, but rather having the admin panel access not hang on a single funny token, but rather on the user accounts making it easy to give and take away access

dani-garcia:
Ah I can see the usecase for creating a real organization and "blessing it" as the admin organization. The problem is that it would require implementing at least some of the crypto that is currently used in the clients in the admin page itself, while this far we've tried to keep away from doing just that.

The problems I mentioned about the old organization approach were because the organization didn't actually exist, it was just virtual, which allowed us to not implement the client side crypto.

My1:
I see, I guess that is because there is no direct way to check passwords? because at least currently the admin panel seems to be able to see who is on which org (at least currently where I only have 1 user)

btw talking about passwords, how about a way of giving out the admin token not in plaintext but as a hash of some sort?

PR #217, opened 2018-10-14
Device::save() should return QueryResult instead of bool

janost:
Related to #6.

### DIFF
--- a/src/api/identity.rs
+++ b/src/api/identity.rs
@@ -46,16 +46,17 @@ fn _refresh_login(data: &ConnectData, _device_type: DeviceType, conn: DbConn) ->
     let orgs = UserOrganization::find_by_user(&user.uuid, &conn);
 
     let (access_token, expires_in) = device.refresh_tokens(&user, orgs);
-    device.save(&conn);
-
-    Ok(Json(json!({
-        "access_token": access_token,
-        "expires_in": expires_in,
-        "token_type": "Bearer",
-        "refresh_token": device.refresh_token,
-        "Key": user.key,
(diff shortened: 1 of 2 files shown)


src/api/identity.rs
```diff
@@ -128,7 +131,10 @@ fn _password_login(data: &ConnectData, device_type: DeviceType, conn: DbConn, re
     let orgs = UserOrganization::find_by_user(&user.uuid, &conn);
 
     let (access_token, expires_in) = device.refresh_tokens(&user, orgs);
-    device.save(&conn);
+    match device.save(&conn) {
```
mprasil: Same as your other PR, this would be better handled by `is_err()`

janost: Thank you. In the meantime I realized that, working on changing this.

src/api/identity.rs
```diff
@@ -46,7 +46,10 @@ fn _refresh_login(data: &ConnectData, _device_type: DeviceType, conn: DbConn) ->
...
```
mprasil: Instead of doing `Ok(()) => (),` here, you can move entire `Ok(Json(json!({` block that's following the match into the match statement. So `Ok(())` will return the json and Err will error as it does now.

mprasil:
Thanks for this, I've suggested two small changes there in the code comments.

src/api/identity.rs
janost: Thank you, fixed!

janost:
Thank you for the review @mprasil I implemented changes based on your suggestions.

dani-garcia:
Thanks!

PR #218, opened 2018-10-14
Folder::save() should return QueryResult instead of bool

janost:
Related to #6.

### DIFF
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -229,11 +229,15 @@ fn post_ciphers_import(data: JsonUpcase<ImportData>, headers: Headers, conn: DbC
     let data: ImportData = data.into_inner().data;
 
     // Read and create the folders
-    let folders: Vec<_> = data.Folders.into_iter().map(|folder| {
-        let mut folder = Folder::new(headers.user.uuid.clone(), folder.Name);
-        folder.save(&conn);
-        folder
-    }).collect();
+    let mut folders: Vec<_> = Vec::new();
+    for folder in data.Folders.into_iter() {
+        let mut new_folder = Folder::new(headers.user.uuid.clone(), folder.Name);
+        if new_folder.save(&conn).is_err() {
+            err!("Failed importing folders")
+        } else {
+            folders.push(new_folder);
+        }
+    }
 
     // Read the relations between folders and ciphers
     use std::collections::HashMap;
--- a/src/api/core/folders.rs
+++ b/src/api/core/folders.rs
@@ -47,7 +47,9 @@ fn post_folders(data: JsonUpcase<FolderData>, headers: Headers, conn: DbConn, ws
 
     let mut folder = Folder::new(headers.user.uuid.clone(), data.Name);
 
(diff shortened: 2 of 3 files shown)


src/api/core/ciphers.rs
```diff
@@ -231,7 +231,10 @@ fn post_ciphers_import(data: JsonUpcase<ImportData>, headers: Headers, conn: DbC
     // Read and create the folders
     let folders: Vec<_> = data.Folders.into_iter().map(|folder| {
         let mut folder = Folder::new(headers.user.uuid.clone(), folder.Name);
-        folder.save(&conn);
+        match folder.save(&conn) {
```
mprasil: This still kinda masks the issue as `true`/`false` values aren't used anywhere. Something like this would be better in my opinion:

```rust
        match folder.save(&conn) {
            Ok(()) => folder,
            Err(_) => err!("Failed importing folders")
        }
```

src/api/core/folders.rs
```diff
@@ -47,7 +47,10 @@ fn post_folders(data: JsonUpcase<FolderData>, headers: Headers, conn: DbConn, ws
 
     let mut folder = Folder::new(headers.user.uuid.clone(), data.Name);
 
-    folder.save(&conn);
+    match folder.save(&conn) {
```
mprasil: These would be probably better handled by [`is_err` method](https://doc.rust-lang.org/nightly/std/result/enum.Result.html#method.is_err). Something like:

```rust
if folder.save(&conn).is_err() {
    err!("Failed to save folder")
}
```

src/api/core/folders.rs
```diff
@@ -73,7 +76,10 @@ fn put_folder(uuid: String, data: JsonUpcase<FolderData>, headers: Headers, conn
 
     folder.name = data.Name;
 
-    folder.save(&conn);
+    match folder.save(&conn) {
```
mprasil: Same as the above.

mprasil:
Hi @janost, thanks for the PR, much appreciated. I've added couple comments and suggested some small improvements.

src/api/core/ciphers.rs
dani-garcia: This is not going to be that easy of a change, the err! is returning the closure from the map call, not the external post_ciphers_function.

We could change it to something like:
```
match folder.save(&conn) {
    Ok(()) => Ok(folder),
    Err(_) => err!("Failed importing folders")
}
```

but then we'd also have to change some other code because currently the `folders` vec only contains `Folder` and with this change it would contain `Result<Folder>`

janost:
@mprasil thank you for the review, I implemented changes based on your feedback. Please let me know if what I did in `post_ciphers_import` is correct, I needed to refactor that function a bit so I could prevent the compiler to complain about type issues in the `folders` vector, since it believed that the result of `err!` could end up in `folders`.

mprasil:
Nice! Thanks for refactoring that. @dani-garcia, I think we can merge it now?

dani-garcia:
Thanks, @janost!

Issue #203, opened 2018-10-01
Second organization collection is working incorrectly in Vault until log out/in is done by the owner


shauder:
Starting with a fresh database to make sure I didn't have any lingering issues from previous testing.

Observations:
Create new account
Create new org
* This seems to work as intended - I get a default collection and they show up

Create second org

* This creates a default collection but it is without a name and does not show up in your collections (unable to decrypt) on applications outside web vault

Create new collection in first org

* Seems to work and shows up in my collections

Create new collection in second org

* Not seeing it show up in my collections and seems to also have trouble decrypting

Adding a credential to the second org and then adding it to one of the collections seems to work initially. After reloading the web-vault the credential is no longer able to be decrypted and is lost.

mprasil:
This issue is weird, but it looks like there's something wrong in the client data when you create the second organization that causes the subsequent issues. If you create the second Organization and **then reload the browser page** or log off/log back in, everything seems to be working fine. The default collection shows up normally.

shauder:
You are correct the second default collection from the second org seems to work fine after a web-vault refresh in the browser. This seems to break the second collection I created in that org if it was made before a full refresh of the web-vault.

mprasil:
Okay, so after some in depth testing it looks like this only shows up if websockets aren't configured properly - the Vault app will get confused on unexpected error trying to sync and this is how this whole issue seems to start.

I've submitted PR #221 to hide the websockets negotiation behind configuration option and have that off by default as it will break Vault unless users also make sure proxy is configured correctly.

mprasil:
#221 is now merged, I think we can close this one. Images are [just building](https://hub.docker.com/r/mprasil/bitwarden/builds/btmucyxbckjewpmrw3rqttn/).

PR #221, opened 2018-10-14
Disable WebSockets negotiation by default

mprasil:
This let's users enable WebSockets by setting the `WEBSOCKET_ENABLED` variable to `true`. The reason why I suggest to have this off by default is that for users that won't do that extra proxy configuration this causes issues like the one observed in #203 - essentially breaking the organization functionality in unpredictable way. 

Users using WebSockets need to do some extra configuration anyways and while this change will break their configuration, it will not break the core functionality, just the WebSockets won't be used until they turn it back on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,11 +212,22 @@ docker run -d --name bitwarden \
 *Important: This does not apply to the mobile clients, which use push notifications.*
 
 To enable WebSockets notifications, an external reverse proxy is necessary, and it must be configured to do the following:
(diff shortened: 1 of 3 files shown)


dani-garcia:
Maybe we should also avoid starting the websockets server if we aren't going to use it in any way?

Just putting the  `thread::spawn` call from `start_notification_server` inside an if should do the trick.

mprasil:
Good idea! I've added that there.

PR #225, opened 2018-10-17
Don't send Domains if excludeDomains=true on /sync

### LINKED ISSUE #222: /sync should send null for Domains if excludeDomains=true
Vault will call `/api/sync?excludeDomains=true` when it wants to do sync without Equivalent Domains information. While returning it anyway doesn't break the client, it makes the response unnecessarily big. (about 5x the response size with the default domain list)

janost:
Fixes #222.

### DIFF
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -22,8 +22,14 @@ use auth::Headers;
 
 use CONFIG;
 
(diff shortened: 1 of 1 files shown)


mprasil:
You're a star @janost!

dani-garcia:
Thanks!

PR #226, opened 2018-10-18
Fix /sync without query string

janost:
Due to Rocket being a bit weird with query strings currently (see [this Rocket issue](https://github.com/SergioBenitez/Rocket/issues/376)), my last change (#225) broke clients which call `/sync` without the `?excludeDomains=` query string. At least the Firefox addon is affected.
Until Rocket releases their new query string parsing logic [implemented here](https://github.com/SergioBenitez/Rocket/issues/608) in `0.4.0`, we're probably stuck with this rather ugly workaround.

### DIFF
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -53,6 +53,13 @@ fn sync(data: SyncData, headers: Headers, conn: DbConn) -> JsonResult {
     })))
 }
 
+#[get("/sync")]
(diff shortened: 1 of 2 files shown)


dani-garcia:
Thanks, I didn’t think to test it with anything other than the web client!

We already have to make similar workarounds for endpoints that accept both PUT and POST, so I don’t think this is any worse.

PR #231, opened 2018-10-22
Accept PUT and POST on /settings/domains, returns JsonResult, fixes saving Custom Equivalent Domains

### LINKED ISSUE #230: Custom Equivalent Domains API results in 404
When trying to set a new "Custom Equivalent Domains" (under Settings > Domain Rules) I receive an error popup: "An error has occurred. An unexpected error has occurred.". Looking at the developer tools I see that the API call `PUT https://[domain]/api/settings/domains` gets a 404 response.

Using Version 2.4.0 in a docker container.

janost:
Fixes #230.

It seems the Web Vault actually expects a JSON instead of any empty response too, so I changed that as well. Seems to work as expected.

### DIFF
--- a/src/api/core/mod.rs
+++ b/src/api/core/mod.rs
@@ -121,6 +121,7 @@ pub fn routes() -> Vec<Route> {
 
         get_eq_domains,
         post_eq_domains,
+        put_eq_domains,
 
     ]
 }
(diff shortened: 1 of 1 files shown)


dani-garcia:
Thanks!

PR #233, opened 2018-10-25
Update PROXY.md

Toucan-Sam:
Removed: `include conf.d/proxy-confs/proxy.conf;` lines because they're specific to user (shauder) and will break nginx if copy-pasted/don't exist.
Changed: Moved listen value and server_name to top as is standard for nginx configs
Changed: Commented out SSL config as it's specific to user (shauder) and will break if copy-pasted/don't exist. But is still useful and a good idea for simplifying nginx config.
Changed: Rearranged location blocks because OCD. First /, then /notifications/hub, then /notifications/hub/negotiate because it looks nicer in a tree where each location grows.

### DIFF
--- a/PROXY.md
+++ b/PROXY.md
@@ -31,25 +31,24 @@ localhost:443 {
 ## Nginx (by shauder)
 ```nginx
 server {
-  include conf.d/ssl/ssl.conf;
-
   listen 443 ssl http2;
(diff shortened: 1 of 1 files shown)


dani-garcia:
Thanks, @Toucan-Sam!

PR #234, opened 2018-10-26
Add Kubernetes deployment reference

icicimov:
As per our discussion, adding the Kubernetes deployment reference as subsection.

Cheers,
Igor

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ _*Note, that this project is not associated with the [Bitwarden](https://bitward
 - [Building binary](#building-binary)
 - [Available packages](#available-packages)
   - [Arch Linux](#arch-linux)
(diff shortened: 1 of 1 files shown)


mprasil:
For reference, we're talking about discussion [here](https://github.com/mprasil/bitwarden_rs/pull/3).

@icicimov sorry for not being clear there. I think it would be better to actually add some small description and link to the repository in the document rather than just linking to it from the TOC as people generally don't expect TOC to link to different document. (and also is autogenerated) Hope that makes sense.

icicimov:
I'm sorry, it was getting late here and I was too tired. A properly done PR will follow shortly.

mprasil:
That's perfect, just one typo: `bitworden_rs` should be `bitwarden_rs`.

dani-garcia:
Thanks icicimov!

PR #236, opened 2018-10-28
Grammar fixes to README.md

xMateMCPE:
I think you meant below instead of bellow.

Definition of bellow (intransitive verb):
1 : to make the loud deep hollow sound characteristic of a bull
2 : to shout in a deep voice


mprasil:
:sweat_smile: thanks for that

Issue #242, opened 2018-11-06
Adding item to organization fails


ottidmes:
When I try to add an item to an organization I get a similar error as mentioned in issues #37 and #178. The error occurs under the latest commit at this time: 8f5bfe7938b9037fb802c0811b28f2575f3b0e89.

The logs:
```
GET /api/organizations/8a9b9e44-bae1-4ebe-a6d5-648dbf234368/collections application/json:
    => Matched: GET /api/organizations/<org_id>/collections
    => Outcome: Success
    => Response succeeded.
GET /api/ciphers/organization-details?organizationId=8a9b9e44-bae1-4ebe-a6d5-648dbf234368 application/json:
    => Matched: GET /api/ciphers/organization-details?<data>
    => Outcome: Success
    => Response succeeded.
POST /api/ciphers/create application/json; charset=utf-8:
    => Matched: POST /api/ciphers/<uuid>
    => Error: Couldn't parse JSON body: Error("missing field `Type`", line: 1, column: 366)
    => Outcome: Failure
    => Warning: Responding with 400 Bad Request catcher.
    => Response succeeded.
GET /images/totp-countdown.png:
    => Matched: GET /<p..>
    => Outcome: Success
    => Response succeeded.
POST /api/ciphers/admin application/json; charset=utf-8:
    => Matched: POST /api/ciphers/admin
    => Error: Couldn't parse JSON body: Error("missing field `Type`", line: 1, column: 530)
    => Outcome: Failure
    => Warning: Responding with 400 Bad Request catcher.
    => Response succeeded.
```

Request payload `POST /api/ciphers/admin`:
```
[...]

PR #243, opened 2018-11-07
Bump vault version to 2.5.0 [wip]

mprasil:
I've bumped Vault to 2.5.0 but didn't have a time to extensively test. There were some new features added, so it's possible some things might be broken.

### DIFF
--- a/src/db/models/organization.rs
+++ b/src/db/models/organization.rs
@@ -86,7 +86,7 @@ impl PartialOrd<i32> for UserOrgType {
 
     fn gt(&self, other: &i32) -> bool {
         match self.partial_cmp(other) {
-            Some(Ordering::Less) => false,
+            Some(Ordering::Less) | Some(Ordering::Equal) => false,
             _ => true,
         }
     }
--- a/src/db/models/organization.rs
+++ b/src/db/models/organization.rs
@@ -1,3 +1,4 @@
+use std::cmp::Ordering;
 use serde_json::Value as JsonValue;
 
 use uuid::Uuid;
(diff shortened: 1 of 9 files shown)


dani-garcia:
According to [OrganizationUserType.cs](https://github.com/bitwarden/core/blob/master/src/Core/Enums/OrganizationUserType.cs) the new manager role has a value of three, so we need to make sure that it doesn't break anything, as we do a lot of `thisrole > thatrole`.

Also, #242 should be fixed before we merge this, otherwise adding ciphers to organizations probably won't work.

mprasil:
Yeah there's multiple things to fix before this can be merged to master. [I triggered docker image build from the branch](https://hub.docker.com/r/mprasil/bitwarden/builds/b3ctmsekiigttrjnqrqsmxd/), so that people can test what works and what is broken.

mprasil:
Yeah, I'm afraid we will start checking for manager once we implement some manager-related functionality. But as you say, I don't see how we could make that better with diesel.

mprasil:
Okay, I've fixed #242 and also improved the `PartialOrd` handling of unknown `i32` type. I think it should be ready to merge if you didn't spot any other issues.

src/db/models/organization.rs
```diff
@@ -83,6 +83,21 @@ impl PartialOrd<i32> for UserOrgType {
         }
         return None
     }
+
+    fn gt(&self, other: &i32) -> bool {
+        match self.partial_cmp(other) {
+            Some(Ordering::Less) => false,
+            _ => true,
```
dani-garcia: This would return true when equal

mprasil: I.. ..totally.. uhm.. left that there to test if you read my code. Definitely not a mistake. :sweat_smile:

mprasil:
Fixed the `gt()` bug. Well spotted!

dani-garcia:
Alright, I made a new release with the vault 2.4.0, and now this is ready to merge.

mqus:
@dani-garcia Could you please re-release 1.4.0 (or 1.4.1)? the changes of this PR were merged after the tagged commit...

mprasil:
@mqus, that was the plan. 😄 We expect some bugs with Vault 2.5.0, so people wanting more stable experience will use 1.4.0 (and the appropriate docker image version)

If you want this code, use master and the latest docker image that should be ready in couple hours. (Currently in the pipe)

mqus:
... I had somehow read that vault 2.5.0 support is in 1.4.0... sry, my fault (It's 2.4.0 (which was also in 1.3.0 already))

Issue #235, opened 2018-10-26
Cannot register U2F key


odensc:
When I go to register a U2F key in the web vault, it errors on `POST /api/two-factor/get-u2f-challenge`.

mprasil:
Can you post the whole error log message?

dani-garcia:
It seems the web vault has changed the APIs used for u2f, probably in version 2.4.0 when support for multiple keys was implemented.

For now, if you need to use u2f keys, you can use bitwarden_rs 1.2.0, which uses an older web vault version that should hopefully work.

Hopefully next week I'll have some time to investigate further and fix this.

pwlgrzs:
I presume this is related, I wanted to activate Duo on my instance but getting this:
```
POST /api/two-factor/get-duo application/json; charset=utf-8:
    => Error: No matching routes for POST /api/two-factor/get-duo application/json; charset=utf-8.                                          
    => Warning: Responding with 404 Not Found catcher.
    => Response succeeded.
```

dani-garcia:
Duo support is not implemented at the moment

RomanHargrave:
The "/api/two-factor/got-u2f" endpoint appeared in jslib at [4b7962d - "add support for u2f setup apis"](/bitwarden/jslib/commit/4b7962dc8fba73003be1ae651013f5f817496551)

RomanHargrave:
I have tested and confirmed that _at minimum_, adding the new endpoint is enough to make u2f enrollment and authentication function. 

Of course, this all is still single-key.

dani-garcia:
The PR has been merged now, so I'll close this. If anyone finds another problem, please open another issue.

PR #244, opened 2018-11-09
Make U2F work with vault 2.4.0 changes

RomanHargrave:
The Bitwarden API for u2f enrollment changed. Specifically, challenge generate was removed from `/api/two-factor/get-u2f` in to `/api/two-factor/get-u2f-challenge`. 

Presently, this PR restores the minimal functionality, but the desire is to implement multi-enrollment.

RE: #235

### DIFF
--- a/src/api/core/mod.rs
+++ b/src/api/core/mod.rs
@@ -80,6 +80,7 @@ pub fn routes() -> Vec<Route> {
         activate_authenticator,
         activate_authenticator_put,
         generate_u2f,
+        generate_u2f_challenge,
         activate_u2f,
         activate_u2f_put,
 
--- a/src/api/core/two_factor.rs
+++ b/src/api/core/two_factor.rs
@@ -273,6 +273,34 @@ fn generate_u2f(data: JsonUpcase<PasswordData>, headers: Headers, conn: DbConn) 
     })))
 }
 
+#[post("/two-factor/get-u2f-challenge", data = "<data>")]
+fn generate_u2f_challenge(data: JsonUpcase<PasswordData>, headers: Headers, conn: DbConn) -> JsonResult {
+    let data: PasswordData = data.into_inner().data;
+
+    if !headers.user.check_valid_password(&data.MasterPasswordHash) {
+        err!("Invalid password");
+    }
+
+    let user_uuid = &headers.user.uuid;
+
+    let u2f_type = TwoFactorType::U2f as i32;
+    let register_type = TwoFactorType::U2fRegisterChallenge;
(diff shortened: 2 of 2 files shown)


dani-garcia:
Thank you, @RomanHargrave!

With this working, adding multiple devices should hopefully be easy. Currently there is a `TODO` in the call to `U2F.register_response` to save multiple registrations, where we could do something like this (untested):

```rust
match U2F.register_response(challenge.clone(), response) {
            Ok(registration) => {
                // Get the current saved u2f if there is one
                let user_u2f = TwoFactor::find_by_user_and_type(
                    &headers.user.uuid,
                    TwoFactorType::U2f as i32,
                    &conn,
                );

                let tf_registration = match user_u2f {
                    Some(user_u2f) => {
                        // If there is one already, add the current registration to the list
                        let mut registrations = _parse_registrations(user_u2f.data);
                        registrations.push(registration);

                        user_u2f.data = serde_json::to_string(&registrations).unwrap();
                        user_u2f
                    }
                    None => {
                        // Otherwise, create it, and add the current registration
                        let mut registrations = Vec::new();
                        registrations.push(registration);

                        TwoFactor::new(
                            headers.user.uuid.clone(),
                            TwoFactorType::U2f,
                            serde_json::to_string(&registrations).unwrap(),
                        )
                    }
                };

                tf_registration
                    .save(&conn)
                    .expect("Error saving U2F registration");
// The rest of it...
```

We may also need to remove the `break` from the `validate_u2f_login` function, as otherwise only the first registration would be checked.

And as far as I know, the rest of the process doesn't care if there is one or multiple registrations, unless the web vault does something strange about it.

RomanHargrave:
@dani-garcia It looks like implementing multiple U2F in such a way that it can be managed by the vault is going to require a little bit more than that.

Looking at the models that the vault expects from `/api/twofactor/get-u2f`, it is apparent that at the very least an additonal `name` field will be needed in the `twofactor` table.

```ts
export class TwoFactorU2fResponse {
    enabled: boolean;
    keys: KeyResponse[];

    constructor(response: any) {
        this.enabled = response.Enabled;
        this.keys = response.Keys == null ? null : response.Keys.map((k: any) => new KeyResponse(k));
    }
}

export class KeyResponse {
    name: string;
    id: number;
    compromised: boolean;

    constructor(response: any) {
        this.name = response.Name;
        this.id = response.Id;
        this.compromised = response.Compromised;
    }
}
```

There is also the question of whether it is safe or not to provide the vault with the internal `twofactor` ID, since a distinguishing ID will be needed to differentiate keys at the API level.

Finally, an endpoint `DELETE /two-factor/u2f` will need to be added which validates the master password and deletes the key identified by the uuid in `id`.

Issue #247, opened 2018-11-13
Add cron job for sql backup for docker image 


ghost:
Hi all,

Is it possible to add cron job for sql backup in docker image? Everytime I want update container I need to do database backup manually, it would be better to have cron job for this inside container running every hour or day or possibility to configure this variable during container configuration.

Strayer:
You essentially have two options, both require a container to share the volume that bitwarden_rs uses.

A) Run a sqlite container periodically via the Docker hosts crontab
B) Build a container with a cron daemon (e.g. dcron) and run the sqlite backup command in it

Option A is trivial, option B is a little more complicated since its horribly annoying getting a cron daemon to work correctly in a Docker container. I'm technically doing something similar than option  A… I'm rebuilding and restarting all containers on my hosts daily with a cronjob anyway to cover potential security updates and simply run the sqlite backup before, directly on the Docker host.

This obviously doesn't work on a Kubernetes/Docker Swarm setup, but then again… if you are running a cluster you should be able to get option B running easily.

Haxy89:
I am asking about B). There is a docker image but without cron job https://hub.docker.com/r/mprasil/bitwarden/

Bruceforce:
> Having said that, if someone feels like writing a documentation about how to achieve cron backup with external container and shared volume, the PR would be very appreciated.

@mprasil Would something like this be useful for your README.md?
https://gitlab.com/1O/bitwarden_rs-backup/tree/master

I outlined an automatic and a manual solution there.

mprasil:
I think it would! You can create a PR to update the readme and link to your repo or something like that.

PR #258, opened 2018-11-18
Update backup section of README.md

Bruceforce:
As discussed in #247 it would be nice if the documentation describes a way to do a backup of the database without installing sqlite to the host system. I also added the mkdir in Line 450 because it doesn't exist in the first place.


dani-garcia:
Thanks!

PR #263, opened 2018-11-20
Update dockerfile for alpine, armv7 and aarch64

### LINKED ISSUE #259: docker alpine builds are failing
The recently merged pr #257 has caused the alpine docker builds (`Dockerfile.alpine`) to fail:

```
Step 11/33 : FROM clux/muslrust:nightly-2018-08-24 as build
nightly-2018-08-24: Pulling from clux/muslrust
Digest: sha256:e113c43250b90138797363777ae3c603c19c4e85248470bbaf9a1ce013bdf1eb

[...]

Step 19/33 : RUN cargo build --release
 ---> Running in d35d6089b138
   Compiling bitwarden_rs v1.0.0 (file:///volume)
error[E0658]: access to extern crates through prelude is experimental (see issue #44660)
   --> src/api/identity.rs:278:69
    |
278 |                 let yubikey_metadata: two_factor::YubikeyMetadata = serde_json::from_str(&twofactor.data).expect("Can't parse Yubikey Metadata");
    |                                                                     ^^^^^^^^^^
    |
    = help: add #![feature(extern_prelude)] to the crate attributes to enable

error: aborting due to previous error

For more information about this error, try `rustc --explain E0658`.
error: Could not compile `bitwarden_rs`.
```

The debian-based docker version builds without problems.

mprasil:
This updates the alpine Dockerfile to fix #259, it also simplifies the Dockerfiles for alpine, raspberry and aarch64 to build a little bit faster on docker hub.

### DIFF
--- a/Dockerfile.aarch64
+++ b/Dockerfile.aarch64
@@ -39,14 +39,7 @@ RUN apt-get update \
 ENV CARGO_HOME "/root/.cargo"
 ENV USER "root"
 
-# Creates a dummy project used to grab dependencies
-RUN USER=root cargo new --bin app
 WORKDIR /app
-
-# Copies over *only* your manifests and vendored dependencies
-COPY ./Cargo.* ./
-COPY ./libs ./libs
-COPY ./rust-toolchain ./rust-toolchain
 
 # Prepare openssl arm64 libs
 RUN sed 's/^deb/deb-src/' /etc/apt/sources.list > \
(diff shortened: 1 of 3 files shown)


dani-garcia:
The point of doing the build in multiple steps like that was to make sure we had a cached copy of the compiled dependencies, which would make the compilation faster. Does docker hub not use the cache?

mprasil:
No, docker hub doesn't share anything between builds, so it's built from scratch every time. So it's just making it take a bit longer really. (I think it might even rebuild some dependencies twice this way as they might be compiled differently based on how dummy vs real application uses them)

I've left the x86 build in the old format as that's the one people would use for development and the benefit is there.

Hope that makes sense.

dani-garcia:
Oh that does makes sense. I agree that leaving an image for development is a good idea.

PR #284, opened 2018-12-11
Close #264. Username and IP address is logged on successful authentication

### LINKED ISSUE #264: Log messages for successful login
We have a login message for failed login attempts, but no counterpart for successful login attempts.

It should include username, and IP address/host, as with the failed attempts.

Algebro7:
Closes #264 by using the new rocket logging features to write IP addresses and usernames to logs on a successful login.


dani-garcia:
Thanks!

PR #289, opened 2018-12-13
Fix failing alpine build

mprasil:
Currently docker hub fails to build alpine image, this should fix it.


dani-garcia:
Was the target removed from the docker image? That seems strange.

mprasil:
Maybe it has something to do with move to Rust 2018? It seems strange, but it definitely makes the Alpine image compile. (also bumping the image version to be on par with rust-toolchain)

PR #291, opened 2018-12-14
Add memory usage note to the vault build documentation

mqus:
As there were multiple people having problems with the build I thought that maybe adding a note would help.


dani-garcia:
Thanks!

PR #293, opened 2018-12-15
Update web.rs

dheimerl:
Add frame-ancestors to allow U2F to work in Chrome (and possibly Firefox) extension

### DIFF
--- a/src/api/web.rs
+++ b/src/api/web.rs
@@ -56,8 +56,8 @@ impl<'r, R: Responder<'r>> Responder<'r> for WebHeaders<R> {
                 res.set_raw_header("X-Frame-Options", "SAMEORIGIN");
                 res.set_raw_header("X-Content-Type-Options", "nosniff");
                 res.set_raw_header("X-XSS-Protection", "1; mode=block");
-                res.set_raw_header("Content-Security-Policy", "frame-ancestors chrome-extension://*");
-                res.set_raw_header("Content-Security-Policy", "frame-ancestors moz-extension://*");
+                let csp = "frame-ancestors chrome-extension://nngceckbapebfimnlniiiahkandclblb moz-extension://* ".to_owned() + &CONFIG.domain + ";";
+                res.set_raw_header("Content-Security-Policy", csp);
 
                 Ok(res)
             },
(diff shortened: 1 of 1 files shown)


dani-garcia:
Setting the second header like that would make Rocket replace the first one, I think. It's better to do both at the same time:
```rust
res.set_raw_header("Content-Security-Policy", "frame-ancestors chrome-extension://* moz-extension://*;"); 
```

Also, can you try removing the wildcard and setting the [actual ID](https://chrome.google.com/webstore/detail/bitwarden-free-password-m/nngceckbapebfimnlniiiahkandclblb)? I think firefox uses a random ID generated at install time, so that one would have to stay like that, at least for now.

```rust
let csp = "frame-ancestors chrome-extension://nngceckbapebfimnlniiiahkandclblb moz-extension://*;"
res.set_raw_header("Content-Security-Policy", csp); 
```

I'd appreciate if you could test it, it seems just a little bit safer that way.

Thanks for your help!

dheimerl:
I tested using your suggestion, that caused the web vault to not work for U2F. I added the configured domain to the list and it worked in the web vault and extensions for Chrome and Firefox.

dheimerl:
I tested with 'self' and that works.

dani-garcia:
Thanks, @dheimerl!

tycho:
Hmm, I just looked up the `moz-extension://*` wildcard thing, because that felt bad. I think Firefox uses the extension GUID for that URL. And those aren't randomly assigned GUIDs per install, they're part of the .xpi manifest.json file. Like this (@dheimerl can you give this a try in your `frame-ancestors` specification on Firefox?):

```
moz-extension://446900e4-71c2-419f-a6a7-df9c091e268b
```

dheimerl:
That doesn't work, just tried it. There is the extension ID (446900e4-71c2-419f-a6a7-df9c091e268b) but when the extension is used, it uses the internal UUID, which is generated each install. I just installed Firefox on a second PC to verify and it does generate a new UUID each time the extension is installed.

tycho:
That's a real bummer. I wonder why Mozilla decided to do this. It makes securing this kind of thing properly that much harder. :disappointed:

PR #294, opened 2018-12-15
Add systemd setup instructions to README.md

i-ghost:


mqus:
Could you maybe clarify in the section title that those are the instructions for the docker image and not bitwarden itself? I'm considering adding a similar section for a binary install with the already existing [archlinux systemd file](https://aur.archlinux.org/cgit/aur.git/tree/bitwarden_rs.service?h=bitwarden_rs).

README.md
```diff
@@ -490,6 +495,115 @@ If it works correctly and your IP is banned, you can unban the ip by running:
...
+After=docker.service
+Requires=docker.service
+
+[Service]
+EnvironmentFile=/etc/bitwarden.service.conf
+TimeoutStartSec=0
+-snip-
+```
+
+### Running
```
mprasil: I believe this should be one level deeper to be under the `systemd-docker` chapter.

README.md
```diff
@@ -490,6 +495,115 @@ If it works correctly and your IP is banned, you can unban the ip by running:
 sudo fail2ban-client unban XX.XX.XX.XX bitwarden
 ```
 
+### Running with systemd
```
mprasil: This should say "Running with systemd-docker" IMHO, as plain systemd kinda implies binary directly or docker directly and this is actually using one more middleware in between.

README.md
```diff
@@ -490,6 +495,115 @@ If it works correctly and your IP is banned, you can unban the ip by running:
...
+
+It's possible to directly specify environment variables in the unit file using the `-e` option of `docker`.
+In this case, you can omit the `--env` option shown in the example above.
+
+If you want to maintain environment settings separately see the subsections below.
+
+To verify that your environment variables are set correctly, check the output of `systemctl show bitwarden.service`
+for an `Environment` line.
+
+##### Using a service subdirectory
```
mprasil: I'd omit this section completely. It's a bit confusing, because the main systemd configuration isn't package-provided file, so people can just modify it directly. For those that really wand service.d override, they can look it up in systemd configuration.

Instead I'd just show here simple use of `Environment` and `EnvironmentFile` bellow. Possible slimming it down a bit. Make it short and to the point.

i-ghost:
Thanks for the feedback both. I've pushed a commit which should address your feedback.

mprasil:
Nice, this is much better. I guess we can merge it @dani-garcia?

dani-garcia:
Thanks!

PR #295, opened 2018-12-16
Add Email Invite Functionality

njfox:
This PR adds support for emailing invite tokens to users who are invited to organizations. `EMAIL_INVITATIONS` has been added as a config option that, when enabled, will cause the /invite endpoint to generate a JWT which is valid for 5 days and email the user a link. When the user clicks the link, their UserOrganization status is set to "Accepted", which is what happened automatically before. The previous behavior is unchanged when `EMAIL_INVITATIONS` is not explicitly enabled.

Note the TODO in src/api/core/organizations.rs:511--currently, when the user clicks the invite link, they are just presented with a blank page. I couldn't figure out how to get redirection working in rocket but ideally they would be automatically redirected to the registration page with their email address pre-filled (ie something like `rocket::response::Redirect::to(format!("/#/register?email={}", invite_claims.email))`.

I think a good next step after getting the redirect working would be to email the user who sent the invitation to let them know it has been accepted, and it would be trivial to add to the new `#[get("/organizations/<org_id>/users/<org_user_id>/accept?<token>")]` endpoint added in this PR.

### DIFF
--- a/src/api/core/accounts.rs
+++ b/src/api/core/accounts.rs
@@ -4,7 +4,7 @@ use crate::db::models::*;
 use crate::db::DbConn;
 
 use crate::api::{EmptyResult, JsonResult, JsonUpcase, NumberOrString, PasswordData, UpdateType, WebSocketUsers};
-use crate::auth::Headers;
+use crate::auth::{Headers, decode_invite_jwt, InviteJWTClaims};
 use crate::mail;
 
 use crate::CONFIG;
@@ -44,6 +44,8 @@ struct RegisterData {
     MasterPasswordHash: String,
     MasterPasswordHint: Option<String>,
     Name: Option<String>,
+    Token: Option<String>,
+    OrganizationUserId: Option<String>,
 }
 
 #[derive(Deserialize, Debug)]
@@ -59,22 +61,37 @@ fn register(data: JsonUpcase<RegisterData>, conn: DbConn) -> EmptyResult {
 
     let mut user = match User::find_by_mail(&data.Email, &conn) {
         Some(user) => {
-            if CONFIG.mail.is_none() {
-                if Invitation::take(&data.Email, &conn) {
+            if Invitation::find_by_mail(&data.Email, &conn).is_some() {
+                if CONFIG.mail.is_none() {
                     for mut user_org in UserOrganization::find_invited_by_user(&user.uuid, &conn).iter_mut() {
                         user_org.status = UserOrgStatus::Accepted as i32;
                         if user_org.save(&conn).is_err() {
                             err!("Failed to accept user to organization")
                         }
                     }
+                    if !Invitation::take(&data.Email, &conn) {
+                        err!("Error accepting invitation")
+                    }
                     user
-                } else if CONFIG.signups_allowed {
-                    err!("Account with this email already exists")
                 } else {
-                    err!("Registration not allowed")
+                    let token = match &data.Token {
+                        Some(token) => token,
+                        None => err!("No valid invite token")
+                    };
+                    let claims: InviteJWTClaims = match decode_invite_jwt(&token) {
+                        Ok(claims) => claims,
+                        Err(msg) => err!("Invalid claim: {:#?}", msg),
+                    };
+                    if &claims.email == &data.Email {
+                        user
+                    } else {
+                        err!("Registration email does not match invite email")
+                    }
                 }
+            } else if CONFIG.signups_allowed {
+                    err!("Account with this email already exists")
             } else {
-                user
+                err!("Registration not allowed")
             }
         }
         None => {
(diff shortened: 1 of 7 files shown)


src/api/core/accounts.rs
```diff
@@ -59,22 +59,27 @@ fn register(data: JsonUpcase<RegisterData>, conn: DbConn) -> EmptyResult {
...
+                } else {
+                    err!("Registration not allowed")
                 }
-                user
-            } else if CONFIG.signups_allowed {
-                err!("Account with this email already exists")
             } else {
-                err!("Registration not allowed")
+                // User clicked email invite link, so they are already "accepted" in UserOrgs
+                user
```
njfox: Pretty sure we need to check for an outstanding invitation here, otherwise users could theoretically register as existing users without an invitation. Will update soon.

dani-garcia:
Some quick comments, as I don't think I'll have time to actually test this thoroughly for a couple of days.

I just sent an invite using the official server to check what the official workflow is, and the link they use is different, it's the following format:
```
https://vault.bitwarden.com/#/accept-organization?organizationId=<>&organizationUserId=<>&email=<>&organizationName=<>&token=<>
```

From that link, the user can choose to login or register:
- If they register, the normal register endpoint is called, but there is an extra `token` parameter, which we can decode to get the email and org id necessary to add the user (Note: You'll need to add the orgid to the jwt)
- If they login, after the login the accept endpoint is called automatically.

Doing this, you might not need to access the accept endpoint directly, and thus won't need a redirect.

Another thing, if an email is sent to the user, then the user has to be marked as invited. At the moment, all the invited users are marked as accepted.

Also, there is a resent invitation endpoint tha would also need to be implemented. The url is:
```
/api/organizations/<org_uuid>/users/<user_uuid>/reinvite
```

Last for now, and this is more subjective, is the email_invitations option really needed? I think it would be far to assume that anyone that has configured smtp would use that option.

njfox:
Thanks for the feedback--I'll work on integrating these changes in new commits when I have time.

> I just sent an invite using the official server to check what the official workflow is, and the link they use is different

Yeah, I saw the web-vault link and I agree that would be the ideal way to go. I believe we would have to change the current way we do org invites, as the web vault page will only let you accept a single userorg, but currently we are accepting all userorgs when the user accepts the invitation. I don't think it would be a big deal, but you'd have to send multiple invites to invite a user to multiple organizations.

> Another thing, if an email is sent to the user, then the user has to be marked as invited. At the moment, all the invited users are marked as accepted.

I'm a little confused by https://github.com/njfox/bitwarden_rs/blob/invite_emails/src/api/core/organizations.rs#L427--newly invited users are temporarily marked as accepted, but their status is changed to invited if the user doesn't exist, which is expected. However, if the user does exist, but doesn't belong to that organization, the status is left as accepted. Is that correct?

> Last for now, and this is more subjective, is the email_invitations option really needed? I think it would be fair to assume that anyone that has configured smtp would use that option.

I agree.

dani-garcia:
I think having to send and accept an invite per organization makes sense. I don't expect most users to be in more than one, maybe two organizations, so it's not like they would be saving a lot of time anyway.

Yes, that's correct. Currently, because we don't have emails it works like this:
- Invite is sent to user
  - If the user exists, it's automatically marked as accepted (because we can't make the user accept without email)
  - If it doesn't, and invitations are enabled, we mark it as invited until the user is registered, when it changes to accepted
- At this point the status is accepted, waiting for confirmation from the admins

The way I imagine invite status when email is enabled is:
- Invite is sent to user (marked as invited, regardless if the user exists or not)
- User opens the invitation
  - If the user doesn't have an account, they can register, which automatically accepts the invite
  - If they have an account, they can login and accept the invite
- At this point the status is accepted, waiting for confirmation from the admins

Does that make sense?

njfox:
Yep, makes sense, thanks!

src/api/core/accounts.rs
dani-garcia: We can check the token here (it's sent as an extra parameter when registering after an invite). With that we can validate that the user has actually been invited, and isn't trying to overwrite another user.

src/api/core/organizations.rs
```diff
@@ -476,12 +481,80 @@ fn send_invite(org_id: String, data: JsonUpcase<InviteData>, headers: AdminHeade
...
 }
 
+#[derive(Deserialize)]
+#[allow(non_snake_case)]
+struct AcceptData {
+    Token: String,
+}
+
+#[post("/organizations/<org_id>/users/<org_user_id>/accept", data = "<data>")]
+fn accept_invite(org_id: String, org_user_id: String, data: JsonUpcase<AcceptData>, conn: DbConn) -> EmptyResult {
```
dani-garcia: The first two params of this function are unused. That's fine, really, because we are already checking the token, but in that case I'd prefer to append an underscore to their names to avoid the warning, and a short comment explaining why they aren't needed.

dani-garcia:
Great, thanks @njfox!

quthla:
Little off topic but I feel like the recent changes should be documented in README.md because else it's not easy to find out what's possible

njfox:
agreed, I’m planning to update the README within the next couple of days.

quthla:
Was not referring to your pr in particular as there's also the icon cache pr which is yet to be documented

Issue #255, opened 2018-11-16
Setup with Fail2Ban


itr6:
Hello all! First off, thank you for this! It is amazing!
I was talking to u/me-ro on the homelab subreddit and they suggested I raise an issue to see if I can get some help with setting up bitwarden and Fail2Ban. He mentioned it logged all the password attempts but I can't find any of the logs.

Can someone please help?

Thanks!

mprasil:
Yeah, this was added in PR #152 but keep in mind, that bitwarden_rs logs to stdout, so, you need to configure docker (or however you run the service) to log to file instead.

As for setting up Fail2Ban, it seems like the only missing bit might be setting up the filter. Maybe @Baelyk submitted the PR for this purpose and already has some configuration?

itr6:
I run it in a standard docker container with all persistent data in a 'data' folder. How would I set it up to log the attempts into a file that Fail2Ban can read?

mprasil:
You can configure [logging driver](https://docs.docker.com/config/containers/logging/configure/) for container or the daemon. Maybe something like `syslog` would work to get the logs somewhere where fail2ban can read them?

itr6:
Sorry if this is really basic, I'm just starting out with docker.
But I need to add:
```
services:
  bitwarden:
    logging:
      driver: "syslog"
```
to my docker-compose file?

mprasil:
I think that should work.

itr6:
I added that and started the container, now when I run `docker inspect --format='{{.LogPath}}' Bitwarden` it doesn't return anything. Before I added the log driver it showed a log. I dont know if I have taken a step backward or what. I guess I don't know enough about this yet.

mprasil:
I guess that's expected as it now logs to syslog. Now you need to figure out where does syslog store the messages.

itr6:
OK so I've made a little progress. I found where the syslog is and it works. Mostly. It logs attempts made if the username looks like an email address, like asdf@asdf.com. But if I try to log in with a random username, like "asdfl" then no logging is made. I'm assuming this is expected behavior?

itr6:
Gotcha, Now I have the syslog working and made it a little more human friendly with the 'tag' option. Now I just need to create the fail2ban filter.

itr6:
GOT IT! OK, here we go.
I added the following to my docker-compose file:

```
    logging:
      driver: "syslog"
      options:
        tag: "Bitwarden"
```
[...]

PR #296, opened 2018-12-16
updated fail2ban setup to match changes in logformat for 1.5.0

jeslinmx:
as mentioned in issue #255 from [this comment](https://github.com/dani-garcia/bitwarden_rs/issues/255#issuecomment-447383930) onwards.


dani-garcia:
Thanks!

PR #297, opened 2018-12-16
Update vault to 2.7.1

mprasil:
Haven't really fully tested yet, but it looks like the core functionality should be OK.


mprasil:
To me 2.7.1 sounds like 2.6 + some useful password checks. I'd like to see it in the 1.5.0 release.

Issue #299, opened 2018-12-17
negative icon caching


tycho:
I notice that bitwarden_rs doesn't remember failing to download icons which don't exist. Every time I refresh my vault in a web browser, I notice it logs 404 errors on the server-side, trying to fetch icons via https://icons.bitwarden.com. I would expect these requests to fail until the domains in question add icons. It seems bad to retry downloading these *every* time a user refreshes their vault page, and we could reasonably do negative caching for a long time. I'd probably cache the negative responses for 3 days or longer.

Also it's possible for icons to change on the upstream pages, so it'd probably be nice to have a time-to-live on the cached icons in `ICON_CACHE_FOLDER`.

To avoid DoS-ing icons.bitwarden,com, the TTL should probably have a bit of variance to it as well.

tycho:
How does something like this look?

https://github.com/tycho/bitwarden_rs/commit/1b1f8b400398c26f3078a5a6bd439e7716316bb7 - add `ICON_CACHE_TTL` and `ICON_CACHE_NEGTTL` config variables
https://github.com/tycho/bitwarden_rs/commit/603dce21723130af96507de7c88cf328ce241d40 - implement icon cache TTLs

Be forewarned, this is my first time touching Rust code since before the first version of Rust was released and I'm a little bit concerned with whether my error handling patterns and coding style are sane or not.

I'm also not sure if people want to run bitwarden_rs on non-Unixy systems where `std::os::unix::fs::symlink` is unavailable, or `std::fs::Metadata::modified` could fail. If we do want to support such systems, the icon cache expiration timestamps should probably be stored in a database table instead of based on symlinks and filesystem timestamps like I use in these patches.

mprasil:
Yeah something like that could work. I kinda like the symlink approach, it leaves no temporary data in the DB. There should be some way to turn off negative caching completely and then we can provide that as fallback on systems that might have issues with symlinks? Do a PR so we can reason about it a bit better.

dani-garcia:
I'd prefer to keep compatibility with as many systems as possible unless completely necessary. While I like the idea behind the symlinks, and storing that data in the database is a bit more clunky, I don't think the benefits are worth dropping support for non-unix.

mprasil:
[...]

PR #302, opened 2018-12-17
implement TTLs for icon cache

tycho:
For previous discussion, see also #299.

### DIFF
--- a/src/api/icons.rs
+++ b/src/api/icons.rs
@@ -1,5 +1,7 @@
 use std::io::prelude::*;
-use std::fs::{create_dir_all, File};
+use std::fs::{symlink_metadata, create_dir_all, remove_file, File};
+use std::time::SystemTime;
+use std::error::Error;
 
 use rocket::Route;
 use rocket::response::Content;
@@ -44,12 +46,23 @@ fn get_icon (domain: &str) -> Vec<u8> {
         },
         Err(e) => {
             error!("Error downloading icon: {:?}", e);
+            mark_negcache(&path);
             get_fallback_icon()
         }
     }
 }
 
 fn get_cached_icon(path: &str) -> Option<Vec<u8>> {
+    // Check for expiration of negatively cached copy
+    if icon_is_negcached(path) {
+        return Some(get_fallback_icon());
+    }
+
+    // Check for expiration of successfully cached copy
+    if icon_is_expired(path) {
+        return None
+    }
+
     // Try to read the cached icon, and return it if it exists
     if let Ok(mut f) = File::open(path) {
         let mut buffer = Vec::new();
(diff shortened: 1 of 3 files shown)


tycho:
Oh, one thing to note about this PR -- it *should* build and run on Windows now, since I added the "symlink" crate. But I haven't tested it on there  at all yet. I need to get a Rust development environment set up on Windows first.

tycho:
I tested this on Windows and it seems to work okay for positive cache TTLs, but the negative cache symlinking is denied unless the user is running bitwarden_rs as an administrator:

> [2018-12-18][01:46:14][bitwarden_rs::api::icons][ERROR] Error creating symbolic link from "c:/dev/bitwarden_rs/target/debug/data/icon_cache/elbereth.uplinklabs.net.png" to fallback icon: Os { code: 1314, kind: Other, message: "A required privilege is not held by the client." }

So yeah, that's not good.

One option is to just create an empty file in place of where the icon would be, or drop some other kind of marker on the filesystem indicating it couldn't be obtained. The symlink was a convenient approach on non-Windows, but it's not the only option. I'll think about it a bit.

src/api/icons.rs
```diff
@@ -62,6 +75,49 @@ fn get_cached_icon(path: &str) -> Option<Vec<u8>> {
...
+}
+
+fn mark_negcache(path: &str) {
+    let miss_indicator = path.to_owned() + ".miss";
+    File::create(&miss_indicator).expect("Error creating negative cache marker");
+}
+
+fn icon_is_expired(path: &str) -> bool {
+    let expired = file_is_expired(path, CONFIG.icon_cache_ttl);
+    !expired.is_ok() || expired.unwrap()
```
dani-garcia: This can be changed to:
```rust
expired.is_err() || expired.unwrap()
```

Or even shorter to:
```rust
expired.unwrap_or(true)
```

src/api/icons.rs
```diff
@@ -62,6 +75,49 @@ fn get_cached_icon(path: &str) -> Option<Vec<u8>> {
...
+    let modified = meta.modified()?;
+    let age = SystemTime::now().duration_since(modified)?;
+
+    Ok(ttl > 0 && ttl <= age.as_secs())
+}
+
+fn icon_is_negcached(path: &str) -> bool {
+    let miss_indicator = path.to_owned() + ".miss";
+    let expired = file_is_expired(&miss_indicator, CONFIG.icon_cache_negttl);
+    if expired.is_ok() {
```
dani-garcia: I think a match here would make it better to understand:

```rust
match expired {
   Ok(true) => {
        <remove_file>
        false
   },
   Ok(false) => true,
   Err(_) => false,
}
```

tycho: That's *waay* more readable! Another case of not being used to Rust. I keep forgetting that the match Ok/Err thing exists.

dani-garcia:
Great, with those changes, this is ready.
Thanks, @tycho!

tycho:
Thanks @dani-garcia!

PR #308, opened 2018-12-23
Fix email invitations for existing users and update README

njfox:
This PR updates the invitation /accept logic to rely on JWT validation for inviting existing users to new organizations. Previously, the invitations SQLite table was used, which resulted in a bug when an existing user was invited to join an organization because no invitation was created. 

I've also updated the README to document the new functionality surrounding email invitations and how it's necessary to set the `DOMAIN` configuration option so that links are generated correctly.


dani-garcia:
Thanks!

PR #315, opened 2018-12-28
Restrict join on users_collections to current user (fixes #313)

### LINKED ISSUE #313: Mobile app fails to sync when user is in organizations with multiple collections
#### Description
I managed to setup a user that can no longer sync. The user is in an organization where he is only associated to one of two collections but has admin rights. Therefore he is still allowed to see all collections. The end result is, that the mobile app crashes immediately (right after sync and then right upon startup).
In a debug session on Android I found that the exception is apparently a duplicate key while creating the dictionary for the Collections the user can see. Indeed, one collection UUID is in there twice.

#### Root cause
From what I can tell, the root of that evil is in `collections.rs`: `find_by_user_uuid`. The result of that function is a vector containing the *sum* of two queries, therefore having the same UUID (possibly) twice in the resultset.

#### Solution
Either cleanup beforehand or use a `HashSet` instead of a `Vec`  (I didn't want to simply change that, since my Rust knowledge is rudimentary and you may have a good reason to use a `Vec` instead of a 
`HashSet`).

#### Remarks
I'm not sure why the .NET dictionary mapping is so fragile regarding duplicate keys, but I think the proper solution here is to not send duplicate keys in the first place.

aksdb:
It fixes #313 at least partially ... the other part being fixed by @mprasil's PR #314.

The problem was, that when multiple users in the same collection are Admin or Owner, the previous JOIN would include multiple rows for the same cipher. The PR fixes this by restricting the joined users_collections in the first place.
(It also makes another JOIN a little more explicit.)

### DIFF
--- a/src/db/models/cipher.rs
+++ b/src/db/models/cipher.rs
@@ -293,7 +293,7 @@ impl Cipher {
             .first::<Self>(&**conn).ok()
     }
 
-    // Find all ciphers accesible to user
+    // Find all ciphers accessible to user
     pub fn find_by_user(user_uuid: &str, conn: &DbConn) -> Vec<Self> {
         ciphers::table
         .left_join(users_organizations::table.on(
@@ -303,7 +303,9 @@ impl Cipher {
                 )
             )
         ))
-        .left_join(ciphers_collections::table)
+        .left_join(ciphers_collections::table.on(
+            ciphers::uuid.eq(ciphers_collections::cipher_uuid)
+        ))
         .left_join(users_collections::table.on(
             ciphers_collections::collection_uuid.eq(users_collections::collection_uuid)
         ))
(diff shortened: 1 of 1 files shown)


src/db/models/cipher.rs
```diff
@@ -352,7 +354,9 @@ impl Cipher {
             )
         ))
         .left_join(users_collections::table.on(
-            users_collections::collection_uuid.eq(ciphers_collections::collection_uuid)
+            users_collections::collection_uuid.eq(ciphers_collections::collection_uuid).and(
+                users_collections::user_uuid.eq(user_id)
+            )
         ))
         .filter(ciphers_collections::cipher_uuid.eq(&self.uuid))
         .filter(users_collections::user_uuid.eq(user_id).or( // User has access to collection
```
dani-garcia: We are checking that same thing here right?, maybe we should remove this one then?

Also, currently we are checking that either:
- The user has an entry in users_collections, which means they can access that collection
- The user has access all
- The user is an admin or owner

If we move that check there, do the last two cases still work correctly?

dani-garcia:
Alright, thanks @aksdb!

PR #321, opened 2019-01-01
Remove redundant information from README after moving it to the wiki

njfox:
Now that the current README, PROXY, and BUILD instructions have been added to the wiki, I've removed all of that redundant information from the README and replaced it with some basic installation/usage sections and point users to the wiki for more information.


dani-garcia:
Nice, I just had a look at the wiki and it's looking a lot better!

Now that it's all moved there, shouldn't we also remove the BUILD and PROXY files?

Also, small note: some links in the Wiki don't work anymore, like for example:
- https://github.com/dani-garcia/bitwarden_rs/wiki/Differences-from-the-upstream-API-implementation
- https://github.com/dani-garcia/bitwarden_rs/wiki/Enabling-WebSocket-notifications (This works, but should link to the Wiki entry)

njfox:
Sounds good, I've also updated all the links in the wiki that were previously pointing to other sections in the README or PROXY/BUILD.

PR #323, opened 2019-01-03
Send email notifications when invitations are accepted/confirmed

njfox:
This PR sends email notifications to users when their invitee accepts an invitation so they know when to confirm them. It also sends an email to the invitee when they are confirmed, which is the behavior indicated by the web vault when they call the /accept endpoint.

Additionally, I've replaced the specific `send_invite` function with a generic email sending function. This requires the subject and body to be formatted by the caller instead of the callee, which I think offers some flexibility for sending many different types of email, but I am open to discussion on whether we want to keep this approach.

### DIFF
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -4,7 +4,6 @@ use serde_json::Value;
 use crate::api::{JsonResult, JsonUpcase};
 use crate::CONFIG;
 
-use crate::auth::{encode_jwt, generate_invite_claims};
 use crate::mail;
 use crate::db::models::*;
 use crate::db::DbConn;
@@ -48,17 +47,8 @@ fn invite_user(data: JsonUpcase<InviteData>, _token: AdminToken, conn: DbConn) -
     if let Some(ref mail_config) = CONFIG.mail {
         let mut user = User::new(email);
         user.save(&conn)?;
-        let org_id = String::from("00000000-0000-0000-0000-000000000000");
-        let claims = generate_invite_claims(
-            user.uuid.to_string(),
-            user.email.clone(),
-            org_id.clone(),
-            None,
-            None,
-        );
         let org_name = "bitwarden_rs";
-        let invite_token = encode_jwt(&claims);
-        mail::send_invite(&user.email, &org_id, &user.uuid, &invite_token, &org_name, mail_config)?;
+        mail::send_invite(&user.email, &user.uuid, None, None, &org_name, None, mail_config)?;
     }
 
     Ok(Json(json!({})))
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -4,6 +4,8 @@ use serde_json::Value;
 use crate::api::{JsonResult, JsonUpcase};
 use crate::CONFIG;
 
+use crate::auth::{encode_jwt, generate_invite_claims};
+use crate::mail;
 use crate::db::models::*;
 use crate::db::DbConn;
 
@@ -31,7 +33,7 @@ fn get_users(_token: AdminToken, conn: DbConn) -> JsonResult {
 #[post("/invite", data = "<data>")]
 fn invite_user(data: JsonUpcase<InviteData>, _token: AdminToken, conn: DbConn) -> JsonResult {
     let data: InviteData = data.into_inner().data;
-
+    let email = data.Email.clone();
     if User::find_by_mail(&data.Email, &conn).is_some() {
         err!("User already exists")
     }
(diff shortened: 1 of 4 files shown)


src/api/admin.rs
```diff
@@ -43,7 +45,21 @@ fn invite_user(data: JsonUpcase<InviteData>, _token: AdminToken, conn: DbConn) -
...
+        let claims = generate_invite_claims(
+            user.uuid.to_string(),
+            user.email.clone(),
+            org_id.clone(),
+            None,
+            None,
+        );
+        let org_name = "bitwarden_rs";
+        let invite_token = encode_jwt(&claims);
+        mail::send_invite(&user.email, &org_id, &user.uuid, &invite_token, &org_name, mail_config)?;
```
dani-garcia: The generate_claims method receives pretty much the same parameters as the send_invite, and it's value is only used there, maybe we should just move the generate_claims call inside the send_invite?

Also, I don't really like the idea of using magic values for the org_id like that, maybe we can change it to optional like the user_org_id after it? We'll need to provide a default value for the vault URL with `unwrap_or`, if you set it to `_`, does the web vault complain? The link is already big enough, and that fake id is both a url parameter and part of the token, so we could save some space.

src/api/core/organizations.rs
```diff
@@ -617,6 +604,20 @@ fn accept_invite(_org_id: String, _org_user_id: String, data: JsonUpcase<AcceptD
         None => err!("Invited user not found"),
     }
 
+    if let Some(ref mail_config) = CONFIG.mail {
+        let org_name = match Organization::find_by_uuid(&claims.org_id, &conn) {
+                Some(org) => org.name,
+                None => String::from("bitwarden_rs"),
+        };
+        if claims.invited_by_email.is_some() {
```
dani-garcia: This is more clear as `if let Some(invited_by_email) = claims.invited_by_email { ...`

src/mail.rs
```diff
@@ -87,17 +76,54 @@ pub fn send_invite(
...
+    new_user_email: &str,
+    address: &str,
+    org_name: &str,
+    config: &MailConfig,
+) -> EmptyResult {
+    let (subject, body) = {
+        ("Invitation accepted",
+        format!(
+            "<html>
+             <p>Your invitation to <b>{}</b> to join <b>{}</b> was accepted. Please log in to the bitwarden_rs server and confirm them from the organization management page.</p>
```
dani-garcia: Wouldn't it make more sense for it to say:
Your invitation **for** `email` to join `org` was accepted

src/mail.rs
```diff
@@ -87,17 +76,54 @@ pub fn send_invite(
...
+    let email = EmailBuilder::new()
+    .to(address)
+    .from((config.smtp_from.clone(), "Bitwarden-rs"))
+    .subject(subject)
+    .header(("Content-Type", "text/html"))
+    .body(body)
+    .build()
+    .map_err(|e| Error::new("Error building email", e.to_string()))?;
+
+mailer(config)
```
dani-garcia: This is formatted strangely, it looks like it's out of the function

njfox: Oops, good catch. Fixed in 2f5bdc2.

src/mail.rs
njfox: Fixed in 2f5bdc2.

src/mail.rs
```diff
@@ -87,17 +76,54 @@ pub fn send_invite(
...
+pub fn send_invite_confirmed(
+    address: &str,
+    org_name: &str,
+    config: &MailConfig,
+) -> EmptyResult {
+    let (subject, body) = {
+        (format!("Invitation to {} confirmed", org_name),
+        format!(
+            "<html>
+             <p>Your invitation to join <b>{}</b> was accepted. It will now appear under the Organizations the next time you log into the web vault.</p>
```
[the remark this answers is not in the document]
njfox: Agreed, fixed in 2f5bdc2.

src/api/core/organizations.rs
njfox: Yeah, I was originally going to do that but I often get a little confused about the `if let Some()` syntax. Thanks for the example--I added this in 7db66f7.

src/api/admin.rs
njfox: > Also, I don't really like the idea of using magic values for the org_id like that, maybe we can change it to optional like the user_org_id after it? We'll need to provide a default value for the vault URL with unwrap_or, if you set it to _, does the web vault complain?

I agree, I don't like using magic values either. Just to make sure I understand, are you suggesting we replace the `org_id` URL parameter with `_` in the invite links?

dani-garcia: Yeah, in the struct we can have an optional value, but in the URL, we'll need something for the case when org_id is none. With `_`, the URL would end up like `/#/accept-organization/?organizationId=_&organizationUserId...`.

We could try removing the query parameter from the URL completely, but that seems like too much work for something that may not even work. Using an empty string may work though, if the web vault doesn't choke on it.

njfox: Good idea. I've implemented this in 0a74e79 and gone through the normal org invite flow as well as inviting from the /admin page and both seem to be working fine.

dani-garcia:
Great, thanks @njfox!

PR #326, opened 2019-01-05
Update Vault to 2.8.0

mprasil:
Update the Dockerfiles once Vault builds are ready


dani-garcia:
I tested it a bit, and the  org users don't show anything to indicate that 2fa is enabled, maybe we need to send `twofactorenabled` in the user json. We can look into that later.

The new json export at least exports fine, I imagine there is no problem with the import either.  Same with the new importers.

Thanks!

PR #347, opened 2019-01-17
Add Feature-Policy header

TBK:
I have omitted _encrypted-media, sync-xhr and usb_ since I am not certain if vault uses it or not.

https://scotthelme.co.uk/a-new-security-header-feature-policy/
https://github.com/w3c/webappsec-feature-policy/blob/master/features.md


dani-garcia:
Thanks, I didn't know about this header!

About the encrypted-media, that seems exclusively about DRM for videos (like Netflix) so we can safely disable that one too.

The vault needs XHR to `self` to work correctly, but there are also requests to `https://haveibeenpwned.com` and `https://twofactorauth.org` for the reports. I'm not sure if they are synchronous though, but if we add those three it should be fine.

About usb I'm not sure, U2F tecnically uses a USB device, but it does so through the U2F API, instead of the WebUSB API, so maybe that can be enabled too?

TBK:
I did not either until today when I did a routine check of my websites with https://securityheaders.io/.

I have added all 3. WebUSB can be used for phishing of U2F (https://www.yubico.com/support/security-advisories/ysa-2018-02/ & https://www.cbronline.com/news/yubico-chrome-vulnerability) and U2F is its own browser API (https://www.w3cschool.cn/doc_browser_support_tables/browser_support_tables-u2f.html) so now I am 99% sure it is alright to disable usb.

dani-garcia:
Great, thanks!

PR #365, opened 2019-01-25
Fix the list of users with access to a collection to display correctly.

srwhite:
This addresses the display issues noted on https://github.com/dani-garcia/bitwarden_rs/issues/364


dani-garcia:
Uh, they must have changed this at some point and we didn't notice. Thanks, @srwhite!

About the managers mentioned in the issue, they aren't implemented in bitwarden_rs, they are considered the same as normal users in practice.

PR #372, opened 2019-01-29
Changed the way to fix the href

BlackDex:
- Using url from reqwest to fix href, this fixes:
   + "//domain.com/icon.png"
   + "relative/path/to/icon.png"
   + "/absolute/path/to/icon.png"
 - Removed fix_href function
 - Some variable changes


dani-garcia:
Nice! The clone makes sense, otherwise we couldn't keep the url reference past the call to `Soup::from_reader(content)`, which consumes the variable.

PR #377, opened 2019-01-31
Added cookies to the icon download request.

BlackDex:
Some sites use XSRF Tokens, or other Tokens to verify a subseqense
response. The cookies which are sent during the page request are now
used when downloading the favicon.

A site which uses this is mijn.ing.nl.

### DIFF
--- a/src/api/icons.rs
+++ b/src/api/icons.rs
@@ -8,6 +8,8 @@ use rocket::Route;
 
 use reqwest::{header::HeaderMap, Client, Response};
 
+use rocket::http::{Cookie};
+
 use regex::Regex;
 use soup::prelude::*;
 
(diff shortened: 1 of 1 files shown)


src/api/icons.rs
```diff
@@ -147,18 +149,28 @@ struct IconList {
...
+
     let resp = get_page(&ssldomain).or_else(|_| get_page(&httpdomain));
     if let Ok(content) = resp {
         // Extract the URL from the respose in case redirects occured (like @ gitlab.com)
         let url = content.url().clone();
+        let raw_cookies = content.headers().get_all("set-cookie");
+        raw_cookies.iter().for_each(|raw_cookie| {
+            let cookie = Cookie::parse(raw_cookie.to_str().unwrap_or_default()).unwrap();
+            cookie_str.push_str(&format!("{}={}; ", cookie.name(), cookie.value()));
+        });
```
dani-garcia: I think this could look better like this:

```rust
    let cookie_str;
    // ...
    cookie_str = raw_cookies.iter().map(|raw_cookie| {
        let cookie = Cookie::parse(raw_cookie.to_str().unwrap_or_default()).unwrap();
        format!("{}={}; ", cookie.name(), cookie.value())
    }).collect::<String>();
```

You'll have to test it to make sure it works, but I'm pretty sure that you can collect Strings to a String.

BlackDex:
Ok, i have changed it. But i have to initialize the cookie_str, else it could return an uninitialized variable.

dani-garcia:
That makes sense, thanks!

PR #378, opened 2019-02-01
only use dotenv in debug builds, not in release builds

gdamjan:
dotenv will look for .env files in the current and all parent
directories. this is potentionaly dangerous in production use.

make it dependent on --debug (the default) builds

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -125,7 +125,9 @@ impl Config {
 
     fn load() -> Self {
         use crate::util::{get_env, get_env_or};
-        dotenv::dotenv().ok();
+        #[cfg(debug_assertions)] {
+            dotenv::dotenv().ok();
+        }
 
         let df = get_env_or("DATA_FOLDER", "data".to_string());
         let key = get_env_or("RSA_KEY_FILENAME", format!("{}/{}", &df, "rsa_key"));


mprasil:
Can we discus a way to still keep this as an option somehow? I'm just concerned, that this could break a lot of people's setup. I agree that it can be quite dangerous considering it tries to find the file all the way up to the root dir.

dani-garcia:
I think we can use `dotenv::from_path(".env").ok();`, to allow loading the `.env` from the current directoy without the whole parent traversal thing, I'll have to check the source code to make sure.

gdamjan:
> I'm just concerned, that this could break a lot of people's setup

You think? how many users do that? the dotenv docs do have this:

> This library is meant to be used on development or testing environments in which setting environment variables is not practical.

mprasil:
Well many users don't even know about the `dotenv` crate, so they'd never see the docs there. I personally don't use it, but I know that many people asked how to use that file inside the container. I'm afraid it's quite a lot of people. I wouldn't really care about the break of functionality as such as it can be configured via variables directly. But this can potentially lead to leaving the server with registrations open while people might think they disabled it. This is my main concern.

dani-garcia:
Sorry I forgot about this, I actually checked the function I mentioned above and it seems to have the expected result, if the `.env` file is in the current directory it's loaded, but if it isn't dotenv won't look anywhere else.

I think that's an acceptable solution to this problem, and I don't expect any user to depend on the parent traversal functionality.

dani-garcia:
I fixed this in the new config panel branch that I just merged, using the function I mentioned to keep compatibility, so I'll close this.

Thank you for reporting this issue, @gdamjan!

PR #382, opened 2019-02-04
Loop through the iconlist until an icon is found

BlackDex:
Loop for a maximum of 5 times through the iconlist or until a
successful download of an icon.

### DIFF
--- a/src/api/icons.rs
+++ b/src/api/icons.rs
@@ -137,19 +137,19 @@ struct IconList {
     href: String,
 }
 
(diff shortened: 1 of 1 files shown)


BlackDex:
Wait, i forgot that the list can of course be smaller.
Going to fix that!

dani-garcia:
I was going to say that too. You also might want to use truncate to limit the Vec length: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.truncate.

PR #393, opened 2019-02-12
Update ring dependency

archfan:


dani-garcia:
This can't be done for now, until all the dependencies that use ring update to the new version. Those dependencies are `rocket` (which uses `cookie`), `u2f` and `jsonwebtoken` as far as I know.

I'm watching those dependencies for updates, but for now this has to be put on pause. Thanks, anyway!

donmor:
Now that `ring 1.17.0` and `jsonwebtoken 9.0.0` are present, we can reopen this.

PR #401, opened 2019-02-16
Images in Email Templates

TheMardy:
Added static routes to Rocket and added Image links to the Email templates


BlackDex:
Is it possible with lettre to embed the images?
Else you always need a public faceing server, and this is not something which could always be the case.


The emails will get larger, but that shouldn't matter that much.

dani-garcia:
As far as I know, lettre only supports normal attachments, not embedding.

What we could do is embed the images in base64 ourselves:
```html
<img alt="My Image" src="data:image/png;base64,<image_bytes_in_base64>" >
```
I'm not sure how well supported that is in the email clients. And for what is worth, the images are small and not very important to the structure of the email, so I don't think it's a big problem if they don't load.


@TheMardy, the images we are using are PNG, the browsers are going to detect the image correctly anyway, but it's better to set the correct MIME Type:
```rust
ContentType::new("image", "png");
```

ViViDboarder:
There are definitely some email clients that do not accept base64 images.

I've found the most consistent way to get images to render is by a public URL or attaching them and using a cid.

dani-garcia:
Thanks! Sorry for the wait!

Issue #399, opened 2019-02-15
Bitwarden_rs Docker on LCOW: Database errors


djpbessems:
I'm trying to run the bitwarden_rs dockerimage on Windows Server 2019 with Linux Containers On Windows (LCOW).

I'm using the command  
`docker run -d --name bitwarden -v C:/path/to/bitwarden/conf:/data -p 80:80 -e "RUST_BACKTRACE=1" mprasil/bitwarden:latest`  
which results in the error  
`thread 'main' panicked at 'Failed to turn on WAL: DatabaseError(__Unknown, "disk I/O error")'` (it creates several files in C:\path\to\bitwarden\conf)  
and subsequent runs change the error to  
`thread 'main' panicked at 'Can't run migrations: QueryError(DatabaseError(__Unknown, "disk I/O error"))'`

Permissions for the path C:\path\to\bitwarden\conf are set to FullControl for Users (temporarily for testing purposes); but the errors remain.

What possible troubleshooting steps could I still follow?

BlackDex:
What i can find quickly on the internet is that this is an LCOW issue.
I read stuff about disable the shared drive, reboot, re-enable the shared drive again. Or try the reset credentials link at that same menu. It seems to have something to do with the credentials not being set right for the LCOW usage.

This issue has nothing to do with bitwarden_rs it self, since it only uses normal filesystem operations and is restricted by what the docker environment allows.

Hope the credentials part in the LCOW settings will help you.
Please let us know if that solves your issue.

mprasil:
Yeah, we're [enabling `WAL` here](https://github.com/dani-garcia/bitwarden_rs/blob/6027b969f542b29365875fa78dbf77765b12d290/src/main.rs#L172-L174) for performance reasons and it sounds like your filesystem (or whatever windows emulates) doesn't support that in current configuration. Try to follow @BlackDex's advice to see if you can enable it.

Alternatively we might look into optionally not enabling [`WAL`](https://sqlite.org/wal.html). I think it would be a useful feature for cases like this, but in the past we had some `SQLITE_BUSY` related issues where some operations (especially "heavy ones") sometimes (rarely, but still) failed. We've since added some code handling these problems, but disabling `WAL` would have to go with the understanding, that it's somewhat non-standard configuration that might sometimes cause issues.

dani-garcia:
[...]

PR #404, opened 2019-02-18
Add an option to not enable WAL (should help in #399)

mprasil:


dani-garcia:
Two things, first I would especify that WAL doesn't work over network filesystems.

Second, enabling WAL is permanent, so we need to make sure that the notice mentions that the flag is required all the time, and that setting it to false after the first launch will do nothing.

mprasil:
Yeah I was thinking about adding the instructions to turn off the WAL into the wiki as it's a bit more complex topic. I'm going to add an information that turning it on will require restart and that it might prevent service from working on some filesystems.

mprasil:
Okay, I've added the setting into the .env.template file and I now refer users to the wiki for more details. As agreed on chat, this won't show in the admin interface as it requires server restart (and sometimes also manual changes to DB) to actually apply the setting completely.  I'm going to create wiki if you're okay merging this.

PR #437, opened 2019-03-21
Split up long line to stop SMTP from breaking

njfox:
Bitwarden invite emails are failing DKIM checks because one of the lines in the HTML template ends up being > 1000 characters, while the SMTP spec sets the max line length at 1000. Postfix ends up adding a newline at the 1,000th character, which causes the body to change and thus the DKIM signature to fail.

This PR fixes this problem by adding a line break to separate some of the HTML attributes from the invite token link, which shortens the line and allows it to pass DKIM checks.


njfox:
Just a note, @BlackDex is looking into a more robust way to address this problem using base64 and other types of encoding

PR #440, opened 2019-03-25
Fixed long e-mail message extending 1000 lines.

BlackDex:
- Added quoted_printable crate to encode the e-mail messages.
- Change the way the e-mail gets build to use custom part headers.

### DIFF
--- a/src/mail.rs
+++ b/src/mail.rs
@@ -1,8 +1,9 @@
 use lettre::smtp::authentication::Credentials;
 use lettre::smtp::ConnectionReuseParameters;
 use lettre::{ClientSecurity, ClientTlsParameters, SmtpClient, SmtpTransport, Transport};
-use lettre_email::EmailBuilder;
+use lettre_email::{EmailBuilder,PartBuilder,MimeMultipartType};
 use native_tls::{Protocol, TlsConnector};
+use quoted_printable::encode_to_str;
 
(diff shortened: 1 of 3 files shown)


src/mail.rs
```diff
@@ -135,11 +136,28 @@ pub fn send_invite_confirmed(address: &str, org_name: &str) -> EmptyResult {
...
+        .message_type(MimeMultipartType::Alternative)
+        .child(text)
+        .child(html);
+
     let email = EmailBuilder::new()
         .to(address)
         .from((CONFIG.smtp_from().as_str(), CONFIG.smtp_from_name().as_str()))
         .subject(subject)
-        .alternative(body_html, body_text)
+        .child(alternative.build())
```
dani-garcia: Looking at the source of lettre, this alternative call does pretty much the same you did, but it also sets the message type to mixed: https://docs.rs/lettre_email/0.9.0/src/lettre_email/lib.rs.html#352.

Maybe we should do the same? I don't really now if that affects anything.

BlackDex: It is already there. see line 152 :).
adding the .child(alternative.build()) will get it wrapped up within a "MimeMultipartType::Mixed".
That is exactly how the RFC is determining how it should be.

dani-garcia: Oh, right! I didn't check inside the `update_headers` call.

PR #455, opened 2019-04-11
Add new endpoint for retrieving all users

ViViDboarder:
Mostly grabbed from an earlier version. Adding this endpoint so that LDAP syncing can be improved as described in #396

### DIFF
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -6,7 +6,7 @@ use rocket::response::{content::Html, Flash, Redirect};
 use rocket::{Outcome, Route};
 use rocket_contrib::json::Json;
 
-use crate::api::{ApiResult, EmptyResult};
+use crate::api::{ApiResult, EmptyResult, JsonResult};
 use crate::auth::{decode_admin, encode_jwt, generate_admin_claims, ClientIp};
 use crate::config::ConfigBuilder;
 use crate::db::{models::*, DbConn};
@@ -21,6 +21,7 @@ pub fn routes() -> Vec<Route> {
 
     routes![
         admin_login,
+        get_users,
         post_admin_login,
         admin_page,
         invite_user,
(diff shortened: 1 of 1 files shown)


dani-garcia:
Thanks!

ViViDboarder:
I guess I didn't think this completely through. This patch is great to avoid overhead of trying to invite existing users, but it doesn't do anything to prevent invites to already invited users.

Gonna have to try and build a new endpoint for that too.

ViViDboarder:
Actually... Looks like there are two different types of invites.

If mail is enabled, a `User` is created and an invite email sent to them. Otherwise, no `User` is created and instead an `Invitation` is created.

This could go a few ways:

1. Create a new endpoint to retrieve all invites from the `Invitation` table
2. Update the existing endpoint to return all users and invites
3. Update the Invitation logic to be consistent and create a `User` regardless of email being enabled or not

Thoughts?

dani-garcia:
The problem I see is that as you say, we have separate invitation systems and they don't act the same way. So the current system is:
- If email is enabled, we create the user always, but never create an Invitation, this is by design, as Invitations are insecure (anyone can register with the email from the Invitation, not just the owner of that email). We send a secure invitation token instead.
- If email is disabled, we always create an Invitation (because there is no way to send the token) and we create the user only sometimes. I think the only time we don't create it is in the admin page invite endpoint, but we'll need to recheck.

To me the best solution would be to just create a new user always, to be consistent.

ViViDboarder:
That makes sense. I can try to work on that, but I pushed a get_invites endpoint for the time being. What do you think about merging that before implementing a refactored invite system?

dani-garcia:
I just checked the code, and I was right, it was only on the admin page where the user wasn't created, so I went ahead and changed it in the latest commit in master. I also made a small change that might help you out, users have a _Status field, with a value of 1 when invited and a value of 0 when they register and become full users, is that enough to solve your problem? Now you can get the list of registered users and invited users in one API call.

ViViDboarder:
Awesome! Thanks for pushing that so quick. I’ll update and test the sync tool with the new field.

ViViDboarder:
@mslocrian your AI agent is talking nonsense. Your comment has nothing to do with this 7 year old PR.

PR #456, opened 2019-04-11
Add get invites endpoint

ViViDboarder:
Update to #455 to allow retrieval of invites.


ViViDboarder:
Not needed anymore now that invited users are included in the Users table.

PR #460, opened 2019-04-14
Fixed purging organization vault

### LINKED ISSUE #458: Purging organization vault purges user vault instead
The `api/ciphers/purge` route does not check whether `organizationId` is set and therefore always purges the user vault.

janost:
Fixes #458.

The new logic checks for the `organizationId` query parameter.
If there is none, it assumes the users wants to purge their own vault.
If there is, it only purges the vault of the organization if the user is an owner of the organization.

### DIFF
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -870,8 +870,14 @@ fn move_cipher_selected_put(
     move_cipher_selected(data, headers, conn, nt)
 }
 
-#[post("/ciphers/purge", data = "<data>")]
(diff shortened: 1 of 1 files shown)


dani-garcia:
Thank you for this @janost! I really didn't want to leave such a dangerous bug unfixed, but I couldn't find time to actually fix it. Really appreciate your help here!

SebastianS90:
Thank you guys for the quick fix!
Btw, no harm has been done since I was only testing some features and not using my live wallet :relieved:

PR #475, opened 2019-05-03
Create Backup funcitonality

TheMardy:
Added create backup functionality to the admin panel

### DIFF
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -9,7 +9,7 @@ use rocket_contrib::json::Json;
 use crate::api::{ApiResult, EmptyResult, JsonResult};
 use crate::auth::{decode_admin, encode_jwt, generate_admin_claims, ClientIp};
 use crate::config::ConfigBuilder;
-use crate::db::{models::*, DbConn};
+use crate::db::{models::*, DbConn, backup_database};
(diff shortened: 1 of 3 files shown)


mprasil:
Nice, but makes me wonder if we could maybe run the SQL directly rather than invoking external command. That way it would work with our current Docker image that _doesn't_ have the sqlite3 binary bundled in. Something with `diesel::sql_query()` like we do [here](https://github.com/dani-garcia/bitwarden_rs/blob/a8a92f6c51f3c60cd5c4288395b5ad93221b4572/src/main.rs#L140) should work I guess?

TheMardy:
@mprasil That would indeed be a better way to do it, but as far as I know there is no way to do it in a sql query. That's why I used the .backup command.

Skeen:
One option to do it in SQL might be the `VACUUM INTO`: https://www.sqlite.org/lang_vacuum.html#vacuuminto

mprasil:
Yeah, 3.27.0 was released 2019-02-07, so only few months ago.

dani-garcia:
I've merged this as it is, if we find a way to not depend on the SQLite binary we can change it later.
I'll also make some changes after the MySQL support is merged to hide this too.

Issue #208, opened 2018-10-03
Use mariadb database


BobWs:
Hi is it possible to use mariadb instead of SQLite?

mprasil:
See #87, it's the same for MariaDB.

dani-garcia:
To keep the issue tracker more focused, I'm closing this issue in favor of the meta issue at #246.

Issue #87, opened 2018-07-15
Use postgresql database


rzig:
Hi! Would it be possible to add support for PostgreSQL? I'd like to use a database independent of the docker container.

mprasil:
It's definitely possible - diesel that's used ORM layer does support Postgres, howeven it's not an easy task to do that. Having DB backed swapable in place is not [goal of diesel](https://github.com/diesel-rs/diesel/issues/980#issuecomment-312849513) and hence you need some work around this.

Having said that I don't think we use SQL directly outside of the migrations code, so if someone is motivated enough to dive into the code to do that, patches would be welcome.

mprasil:
Note that we're using [replace_into](https://docs.rs/diesel/1.2.2/diesel/fn.replace_into.html) a lot. It appears, that this might not be supported with Postgres.

jgrip:
Upserts were added in 9.5, so this is supported on version 9.5 and up.
http://www.postgresqltutorial.com/postgresql-upsert/

Not looked at the bw_rs code to see how much work it would be yet.

odarriba:
It would be very helpful tonbe able to deploy bw_rs in Kubernetes and ease the backup of the files! Also running multiple nodes without direct file sharing

dani-garcia:
To keep the issue tracker more focused, I'm closing this issue in favor of the meta issue at #246.

PR #486, opened 2019-05-20
[WIP/RFC] MySQL / MariaDB database backend support

Skeen:
Hello,

This PR introduces support for MySQL as database backend (and in the current state, completely breaks support for SQLite in the process), thus resolving #208. MySQL was chosen due to [diesel::replace_into ](https://docs.rs/diesel/1.2.2/diesel/fn.replace_into.html) support, as noted as a requirement here: #87.

The changes are mainly:
* Replacing SQLite* with mysql*
* Reworking migrations to name `key` to `akey` and `type` to `atype` (`key` and `type` are reserved keywords in MySQL).
* Reworking migrations to use `VARCHAR(40)` instead of `TEXT` for UUIDs (MySQL cannot use `TEXT` as keys, as it only (as far as I know) indexes up to 255 characters).

The PR can be tested out, by starting a mysql docker container:
```
docker run --name mysql -e MYSQL_ROOT_PASSWORD=my-secret-pw -e MYSQL_DATABASE=bitwarden -p 3306:3306 -d mysql:5.7
```
or a mariadb docker container:
```
docker run --name mysql -e MYSQL_ROOT_PASSWORD=my-secret-pw -e MYSQL_DATABASE=bitwarden -p 3306:3306 -d mariadb:10.3
```
And building the project via `cargo run`, althrough we now need the `libmysql++-dev` debian package.

Now for the questions:
1. Are these changes welcome?
- I imagine changes to migration files could be a big issue for current installations.
- The naming `akey` and `atype` is up for discussion, I just wanted to avoid the conflicts.

2. How do we peserve sqlite support while supporting mysql?
- The connection import + type alias in `db/mod.rs` will have to be dynamic and configuration dependent.
- If we combine support for both in one image, it will be larger than the pure sqlite3 one.


dani-garcia:
Hey @Skeen, would love to have this in! We need a way to keep both backends available though.

The type changes to the migrations should be fine, as SQlite ignores the varchar length and uses the text type anyway. The name changes would be a problem though. Instead of modifying the existing migrations, we'd have to create a new one, where we create new tables with the new fields, move all the data there and then delete the old tables. This can create problems with foreign keys of tables pointing at the tables that have been recreated, so those might need to be recreated too (or at least the foreign keys would need to be removed and readded). I'll leave that for last for now.

I'm not completely sure that we can easily support both both databases with a runtime config option though, or at least not easily. I think the best alternative would be a compile time flag, and for that we would need to distribute a separate image (which is fine by me anyway I don't expect people to switch between dbs very often if at all).

About the renames, I can see multiple options:
- Change the names of the SQLite schema to use different names (akey and atype are as good as any other). Seeing as the main tables are modified, this would pretty much require the complete migration mentioned which can be error prone. We can mitigate a lot of the dangers if we plan carefully though. This should in theory allow us to keep the same set of migrations and schema for both dbs, but it might create problems later down the line if we try to add support for another db.

- Create a separate set of migrations and schemas per db type, we would have to switch between them with some sort of `#[cfg(feature="mysql")]` and we'd have to do the same with the structs fields to make sure they are renamed to the correct name in the Rust structs. I feel like this would make the code hard to understand and modify, so probably not the best idea.

- A bit of a middle ground, do the massive migration to make sure the names aren't using reserved keywords in any db we could support (either now or in the future, it would be a good idea that if we do this we make sure the names are also available in postgres for example). And keep the separate migrations and schemas which would allow to change the types as long as Diesel supports them.

I'm not sure which way would be most appropiate to follow at the moment to be honest. I'll give it some thought over the next couple of days, if anyone else has another idea i'd love to hear.

janost:
This is great, thank you!

I believe we don't have to rename fields which are also reserved keywords in mysql, [we could escape them in queries](https://stackoverflow.com/questions/2889871/how-do-i-escape-reserved-words-used-as-column-names-mysql-create-table).

Skeen:
@janost That looks great! - Do you know if diesel automatically escapes the names when used in generated queries? - Excuse me for asking if it's obvious, I'm not a rust programmer, and I'm not known around the frameworks.

janost:
@Skeen I'm very new to Rust as well, so don't worry about it. :)
I actually have no idea if diesel does this automatically. I'd say it definitely should, but it might not, we probably need to test.

migrations/2018-01-14-171611_create_tables/up.sql
```diff
@@ -1,14 +1,14 @@
 CREATE TABLE users (
-  uuid                TEXT     NOT NULL PRIMARY KEY,
+  uuid                VARCHAR(40) NOT NULL PRIMARY KEY,
```
janost: Setting this and other UUID fields to `CHAR(40)` instead of `VARCHAR(40)` might be beneficial, since UUIDs are fixed in length and looking up fixed-length fields are usually faster.

Skeen: `CHAR(36)`, but I agree, using `CHAR` directly will be preferable when we know it is indeed fixed-width.

Alternatively there might be UUID specific types, like in [Postgres](https://www.postgresql.org/docs/11/datatype-uuid.html), or [binary + functions like in MySQL](https://mysqlserverteam.com/mysql-8-0-uuid-support/).

endyman: 👍 for the efforts! We were about to submit a merge request as we are currently trying to get mysql up an running, too and have almost the the same chnages - apart from backend specific schema directories and build args in order to support multiple backends. What I'm currently struggling with are Database Errors on Mysql due to Foreign Key Constraints

`[CAUSE] DatabaseError(
    ForeignKeyViolation,
    "Cannot delete or update a parent row: a foreign key constraint fails (`bitwarden`.`devices`, CONSTRAINT `devices_ibfk_1` FOREIGN KEY (`user_uuid`) REFERENCES `users` (`uuid`))",
)`

This happens when trying to update the master password or email address. I think it would make sense to merge our efforts @dani-garcia what do you think? should we simply create a second MR or would it be better to just point to our fork/branch?

Skeen: @endyman

That sounds super lovely, if your implementation is more throughout than what I did, let's work with that. My PR was literally made in an afternoon after work, and if you put more work into yours, maybe we should focus on that. We could also try to merge the efforts, although if we work is similar, it might not be worth the hassle.

endyman: Short update: I identified the SQL issue in our code - we tried to not rename ```key``` to ```akey``` by using a different syntax in migrations an schema (``` `key` ```)but it looks like diesel does not work well with reserved column names. I'm currently evaluating various options on how to implement this in a non breaking way.

endyman:
@Skeen @dani-garcia I create a new PR (#493) including the code from @skeen and our code which supports multiple backends.

dani-garcia:
I think with @endyman's PR we should focus our efforts there, so I'll close this one.

Thanks for your work, @Skeen!

Skeen:
Yea I agree with going forwards with @endyman's PR.

PR #493, opened 2019-05-26
Initial support for mysql

endyman:
This is a merged PR based on the work of @skeen in #486 and our efforts to support additional backends.

Using build args the desired backend can be selected, the default is sqlite. You can start a mysql build using the following command: `docker build --build-arg DB=mysql .`

Here is a simple test setup:
```
docker run --name mysql --net <some-docker-network>\
 -e MYSQL_ROOT_PASSWORD=<my-secret-pw>\
 -e MYSQL_DATABASE=bitwarden\
 -e MYSQL_USER=<bitwarden_user>\
 -e MYSQL_PASSWORD=<bitwarden_pw> -d mysql:5.7

docker run -d --name bitwarden --net <some-docker-network>\
 -v $(pwd)/bw-data/:/data/ -v <Path to ssl certs>:/ssl/\
 -p 443:80 -e ROCKET_TLS='{certs="/ssl/<your ssl cert>",key="/ssl/<your ssl key>"}'\
 -e RUST_BACKTRACE=1 -e DATABASE_URL='mysql://<bitwarden_user>:<bitwarden_pw>@mysql/bitwarden'\
 -e ADMIN_TOKEN=<some_random_token_as_per_above_explanation>\
 -e ENABLE_DB_WAL='false' <you bitwarden image build above>
```
Current supported features
- [x] initial mysql and sqlite support
- [x] per backend schemas and migrations
- [x] Make backends (features) mutually exclusive 
- [x] backward compatibility through migrations and sqlite defaults (needs testing)
- [x] Adapt Docker files for other archs

What still needs to be done apart from review and backward compatibility and migration testing:
- once this is in, documentation on how to use the new backend in wiki


src/db/mod.rs
```diff
@@ -12,16 +15,25 @@ use rocket::{Outcome, Request, State};
 use crate::CONFIG;
 
 /// An alias to the database connection used
+#[cfg(feature = "sqlite")]
 type Connection = SqliteConnection;
+#[cfg(feature = "mysql")]
+type Connection = MysqlConnection;
```
dani-garcia: This is a bit nitpicky but i'd rather do:

```rust
#[cfg(feature = "sqlite")]
type Connection = diesel::sqlite::SqliteConnection;
#[cfg(feature = "mysql")]
type Connection = diesel::mysql::MysqlConnection;
```

and remove the imports and cfgs from lines 5-8.

Also the Pool comment in next line would probably be better without specifically mentioning either sqlite or mysql

src/main.rs
```diff
@@ -123,24 +123,8 @@ fn chain_syslog(logger: fern::Dispatch) -> fern::Dispatch {
...
-    }
-
-    // Turn on WAL in SQLite
-    if CONFIG.enable_db_wal() {
-        use diesel::RunQueryDsl;
-        let connection = db::get_connection().expect("Can't conect to DB");
-        diesel::sql_query("PRAGMA journal_mode=wal")
-            .execute(&connection)
-            .expect("Failed to turn on WAL");
-    }
```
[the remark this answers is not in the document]
Skeen: This is probably an oversight by @endyman, it looks like a reminent of my changes, from when I butchered the code to make my MySQL proof of concept.

Cargo.toml
```diff
@@ -47,10 +49,10 @@ log = "0.4.6"
 fern = { version = "0.5.8", features = ["syslog-4"] }
 
 # A safe, extensible ORM and Query builder
-diesel = { version = "1.4.2", features = ["sqlite", "chrono", "r2d2"] }
-diesel_migrations = { version = "1.4.0", features = ["sqlite"] }
+diesel = { version = "1.4.2", features = ["mysql", "sqlite", "chrono", "r2d2"] }
```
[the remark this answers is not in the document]
endyman: Can you provide me with a hint how to make `libsqlite3-sys` optional? I'll tried 
```
sqlite = ["diesel/sqlite", "diesel_migrations/sqlite", "libsqlite3-sys"]
...
libsqlite3-sys = { version = "0.12.0", features = ["bundled"], optional = true }
```
But that results in a linker error (sqlite3 lib is missing?!) when building the mysql feature.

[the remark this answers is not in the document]
dani-garcia: I think the problem is that diesel is still including both its "mysql" and "sqlite" features, and this second one is pulling "libsqlite3-sys" transitively, but without the bundled feature, which means it requires libsqlite to be installed in the system for it to compile.

Now that our own "mysql" and "sqlite" features activate the necessary diesel features, you should be able to remove them from the diesel dependency:

```toml
diesel = { version = "1.4.2", features = ["chrono", "r2d2"] }
diesel_migrations = { version = "1.4.0" }
```

I think that should work.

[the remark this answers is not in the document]
endyman: Good point - will give it a try!

Cargo.toml
```diff
@@ -13,6 +13,8 @@ build = "build.rs"
 [features]
 # Empty to keep compatibility, prefer to set USE_SYSLOG=true
 enable_syslog = []
+mysql = ["diesel/mysql", "diesel_migrations/mysql"]
+sqlite = ["diesel/sqlite", "diesel_migrations/sqlite", "libsqlite3-sys"]
```
tycho: How tough would it be to add PostgreSQL support as well? Diesel apparently wrote their [getting started guide](https://diesel.rs/guides/getting-started/) around using it.

endyman: According to the diesel docs, diesel currently does not support `replace_into` with PostgreSQL http://docs.diesel.rs/diesel/fn.replace_into.html (MySQL & SQLite only). `replace_into` is currently  used in models so it might be a bigger challenge.

src/main.rs
```diff
@@ -45,6 +45,9 @@ fn main() {
         init_logging().ok();
     }
 
+    #[cfg(all(feature = "sqlite", feature = "mysql"))]
+    compile_error!("Can't enable both backends");
```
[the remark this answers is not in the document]
endyman: I'll see your point. It's mostly about having per Backend Connections which is out of scope of Diesel https://github.com/diesel-rs/diesel/issues/1881 - On the other hand this would pull in the dependencies for each type of database which would lead to a bigger footprint. From my perspective it would be quite hard to get dynamic backends in.

Cargo.toml
endyman: See also comments in this issue: https://github.com/dani-garcia/bitwarden_rs/issues/87

tycho: Ah, bummer.

src/main.rs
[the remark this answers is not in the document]
tycho: Okay, no problem. It's a nice-to-have.

Cargo.toml
Skeen: Potentially [relevant ticket](https://github.com/diesel-rs/diesel/pull/1884) for diesel.
It will require us to use `upsert` instead of `replace`, but they are essentially functionally equivalent.

endyman: Maybe it makes sense to split this? Get multibackend + mysql in by this PR and open up a second as soon as diesel provides crossbackend upsert following the patterns used here?

endyman:
I'll do some more testing but currently don't have any pending commits.

Cargo.toml
Skeen: Yea I do believe it should be split.

Create multibackend support + mysql support here, and do postgresql support and thus changing `replace_into` with `upsert` once upsert is supported upstream in diesel.

Skeen:
Once this is merged, I'll update my Kubernetes helm chart to support MySQL as backend.

endyman:
@dani-garcia anything else I can do from your side like squash commits?

dani-garcia:
No it's good as it is! I don't have a MySQL db to test this, but the changes are reasonably easy to understand so no problems there.

A couple of things though, in the CI files there is no need to enable the `enable_syslog` feature as it does nothing anymore.

Also when compiling directly using `cargo build`, the error message is not obvious that there needs to be a backend enabled, we could do a couple of things, create another check like we do when both are enabled: 
```rust
#[cfg(not(any(feature = "sqlite", feature = "mysql")))]
    compile_error!("Need to enable one DB backend, do cargo build --features blah blah...");
```

Or we could set `sqlite` as a default feature, which would allow for this to work when using SQLite:
```
cargo build
```

But would require doing this to enable MySQL instead
```
cargo build --features mysql --no-default-features
```

Alternatively, we could just create a feature for mysql and only enable sqlite when there are no other backend features selected:
```rust
#[cfg(all(feature = "mysql", feature = "postgres")))]
    //  Multiple backend error
#[cfg(not(any(feature = "mysql", feature = "postgres")))]
    // SQLite backend
#[cfg(feature = "mysql")]
    // MySQL backend
#[cfg(feature = "postgres")]
    // Postgres backend
```

Any suggestions about this? I know most users prefer using the docker image where this isn't a problem but I would rather not break anything for those who build the image themselves.

tycho:
I think explicitly asking for a DB backend at build time is better than implicitly getting one and being forced to override it if another is desired.

I can imagine a future where defaults change and someone builds without explicitly asking for a DB backend and gets a build with features they didn't expect. I don't know that the default DB backend would ever be changed, but it's a design precedent for other mutually-exclusive features in the future too.

dani-garcia:
If we were building this from the start I'd agree, but we could have users already that depend on `cargo build --release` working, so requiring a flag now will break those builds. It shouldn't be a big deal, but it does require the users to make some changes. That's why I think keeping the current behavior as the default is the best option.

In any case, at a minimum we should improve the error message if no backends are enabled.

tycho:
I don't think it's too bad to break packagers until they fix their scripts. Your idea of including:

```rust
#[cfg(not(any(feature = "sqlite", feature = "mysql")))]
    compile_error!("Need to enable one DB backend, do cargo build --features blah blah...");
```
would really help this situation because packagers would have clear direction of what to do in order to get a working build again. We probably should mention in that message something like "the default used to be `--features sqlite`, and is probably what you want"

endyman:
I've added the compile_error and reference to previous defaults and removed the syslog feature from the azure ci pipeline.

dani-garcia:
Alright, thank you so much for your work, @Skeen and @endyman! Maybe after this we should create a new 2.0 release? 🎉

uchagani:
I'm trying to run `docker build --build-arg DB=mysql .` right now and in the logs I see this while building:

```
Step 10/34 : ARG DB=sqlite
 ---> Running in a45b8efcf005
```

is that expected or just a typo?

mprasil:
That's expected. The sqlite there is just default in case you didn't specify otherwise. See [Dockerfile ARG documentation](https://docs.docker.com/engine/reference/builder/#arg).

PR #513, opened 2019-06-24
Update Docker Base-Images

JacobiteDE:


mprasil:
What is the rationale behind this? Id prefer to have the base image version explicitly defined rather than using the `latest` version that can change anytime.

Dockerfile.alpine
```diff
@@ -20,17 +20,17 @@ RUN ls
 
 ########################## BUILD IMAGE  ##########################
 # Musl build image for statically compiled binary
-FROM clux/muslrust:nightly-2018-12-01 as build
+FROM clux/muslrust:stable as build
```
fbartels: the explicit version tag for the last stable is `1.35.0-stable`

PR #521, opened 2019-07-03
Update Alpine Image to 3.10

Jacobite89:


docker/amd64/mysql/Dockerfile.alpine
```diff
@@ -20,18 +20,17 @@ RUN ls
 
 ########################## BUILD IMAGE  ##########################
 # Musl build image for statically compiled binary
-FROM clux/muslrust:nightly-2018-12-01 as build
+FROM clux/muslrust:1.35.0-stable as build
```
dani-garcia: We need to use the nightly images as rocket will not work in stable rust just yet, but we should still update it to the latest nightly available. Same for the other dockerfile.

Jacobite89: Prefer the nightly, 1.37.0-nightly or nightly-2019-07-02 tag?

dani-garcia: I think it's better to specify the date in this instance, mostly because we've had cases of new nightlies breaking the build.

Also if you don't mind, I'd appreciate if you also update the `rust-toolchain` file at the root of the project to the same date, to keep parity between CI and the dockerfiles.

PR #525, opened 2019-07-04
use hadolint for linting Dockerfiles

fbartels:
- add hadolint to travisfile

fixes https://github.com/dani-garcia/bitwarden_rs/issues/522

### DIFF
--- a/docker/aarch64/mysql/Dockerfile
+++ b/docker/aarch64/mysql/Dockerfile
@@ -2,18 +2,20 @@
 # 	https://docs.docker.com/develop/develop-images/multistage-build/
 # 	https://whitfin.io/speeding-up-rust-docker-builds/
 ####################### VAULT BUILD IMAGE  #######################
-FROM alpine as vault
+FROM alpine:3.10 as vault
 
 ENV VAULT_VERSION "v2.10.1"
 
 ENV URL "https://github.com/dani-garcia/bw_web_builds/releases/download/$VAULT_VERSION/bw_web_$VAULT_VERSION.tar.gz"
 
-RUN apk add --update-cache --upgrade \
+RUN apk add --no-cache --upgrade \
     curl \
     tar
 
 RUN mkdir /web-vault
 WORKDIR /web-vault
+
+SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
 
 RUN curl -L $URL | tar xz
 RUN ls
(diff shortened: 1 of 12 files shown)


dani-garcia:
The command is making the build fail because it returns a non zero error code, we could just do something like `command || true` but then the warnings would be hidden unless someone checks travis log. Maybe there is a way to tell travis to just warn on error for a specific step?

fbartels:
Yes, i would even say the failure is exactly the idea. I will take care to resolve all errors and will remove the "WIP" once this is ready to be squashed and merged.

fbartels:
@dani-garcia with the last commit the build should now succeed. There is quite a lot of duplication in the Dockerfiles, I am wondering if this could be simplified a bit (for the different architectures for example) when using build arguments.

docker/aarch64/mysql/Dockerfile
```diff
@@ -2,32 +2,35 @@
...
 
-RUN apk add --update-cache --upgrade \
+RUN apk add --no-cache --upgrade \
     curl \
     tar
 
 RUN mkdir /web-vault
 WORKDIR /web-vault
 
+SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
```
dani-garcia: What does this line do exactly?

dani-garcia:
Yeah it would be great if the setup could be simplified a little bit, but I don't know enough about docker or what docker hub supports to do it myself.

docker/aarch64/mysql/Dockerfile
fbartels: The following line uses a pipe (curl | tar). This line makes the docker build fail if the pipe fails.

mprasil:
@fbartels we actually attempted to use the build arguments [and it didn't work](https://github.com/dani-garcia/bitwarden_rs/issues/512#issuecomment-504917374). I actually think we could benefit from multiple Dockerfiles and limit the dependencies being installed.

fbartels:
Yes, I've also done a bit of research meanwhile and when just relying on the Docker Hub and not travis in combination with some other scripting (make or a shell script) this does not really seem possible. I was not even referring to sqlite backend vs. mysql backen, but was just thinking about the builds for the different architectures.

But in general I would say that the pr is complete for the moment.

dani-garcia:
Great, thanks for your work, @fbartels!

PR #541, opened 2019-07-22
Add "New Device Logged In From" email

vverst:
- Added both templates for the New Device Logged In From email.
- Added `send_new_device_logged_in` email function.
- Moved logic of retrieving or creating device to `get_device` function (Is that ok?). When the device is new an email is sent.

### DIFF
--- a/src/api/identity.rs
+++ b/src/api/identity.rs
@@ -101,38 +101,7 @@ fn _password_login(data: ConnectData, conn: DbConn, ip: ClientIp) -> JsonResult 
         )
     }
 
-    // On iOS, device_type sends "iOS", on others it sends a number
-    let device_type = util::try_parse_string(data.device_type.as_ref()).unwrap_or(0);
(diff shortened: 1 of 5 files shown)


vverst:
I noticed that it was sending the email before the 2fa check so I moved the email to after the two factor authentication check.

dani-garcia:
Nice! Thanks for your work! Sorry I let this unanswered for some time, I just came back from vacation.

vverst:
Thanks! No problem, hope you had a nice vacation.

strikerlulu:
is there any way to disable this feature ? 
i'm using bitwarden as k8s external secrets provider https://external-secrets.io/main/examples/bitwarden/
and its ending this new device login mail too often, can i disable it for a instance with env variable or for single user ?

BlackDex:
No, there is no option for that, other then to disable SMTP.

Or, somehow let that feature use a fixed appId.
The `appId` is stored in `~/.config/Bitwarden CLI/data.json`

PR #555, opened 2019-08-04
Add Email 2FA login

vverst:
- Added email code templates for sending emails
- Moved twofactor options into different files/modules
- Tried using TOTP for email token codes but that using that didn't make sense. Currently generating a code, serializing and saving it into the `data` field. Is this a reasonable solution?
- Using a `TwoFactorType ::EmailVerificationChallenge` to show that an email has not been verified yet. After verification it is of `TwoFactorType::Email` type.

If merged:
- [ ] web vault builds change required so it no longer hides the 2FA Email button


dani-garcia:
I think generating random codes and saving them is pretty good, though instead of adding a new crate I’d use the random functions we use already, which depend on the ones provided by ring:

https://github.com/dani-garcia/bitwarden_rs/blob/8be2ed6255f6137267dde218ca8b53ef38321a1a/src/api/core/ciphers.rs#L674

vverst:
Thanks for the suggestion. That makes more sense.
I've now updated it to use `get_random`.

src/api/core/two_factor/email.rs
```diff
@@ -97,13 +99,12 @@ struct SendEmailData {
...
-    (0..TOKEN_LEN)
-        .map(|_| {
-            let num = rng.gen_range(0, 9);
-            char::from_digit(num, 10).unwrap()
+    crypto::get_random(vec![0; TOKEN_LEN])
+        .iter()
+        .map(|byte| { (byte % 10)})
+        .map(|num| {
+            dbg!(num);
+            char::from_digit(num as u32, 10).unwrap()
```
dani-garcia: This could be simplified as: 
```rust
let bytes = crypto::get_random(vec![0; 4]) // Four u8 to fill a u32, with a u32 we could use up to 9 digits, could use u64 for up to 19 digits
let number = u32::from_be_bytes(bytes) % 10u32.pow(TOKEN_LEN); // Make sure the number doesn't exceed the length
let number_str = format!("{:0size$}", number, size=TOKEN_LEN); // Pads the number with zero from the left if it's smaller
```

I think it might be a good idea to expose both the token size and the expiration times of those tokens as config options for the more security-conscious users.

I won't be able to use my main dev machine for at least a little over a week, so I probably won't be able to test the code a bit more in depth, or upload a new web vault build until then.

vverst: No problem. There is no hurry.

About the code snippet; I had to add a conversion from Vec to array (u8,4) since `from_be_bytes` only accepts a fixed u8 array.
```rust
let bytes_array = [bytes[0], bytes[1], bytes[2], bytes[3]];
let number = u32::from_be_bytes(bytes_array) % 10u32.pow(TOKEN_LEN);
```
is what I had. Is there a better way?

About adding configuration
- Could be added but not sure if it has any benefit. Official bitwarden also uses a fixed length of 6 numbers. If we add a config for this token size we should maybe also add a config for TOTP token size? Do we need to set a minimum length of 6 or allow any length?
- Right now the email time is `600` seconds.
We could add a configuration item. Same question; allow any expiration time or set a maximum amount of time?

Or maybe good defaults are enough? Not sure.

Also, we might need to prevent brute forcing the token. Benchmarking locally got around 20000 requests in 10 minutes. Maybe we should limit the amount of attempts (keeping track of amount of attempts)?

dani-garcia: Oh right yeah I forgot fixed size arrays can be weird to handle sometimes. The reason for the Vec parameters in the get_random method was because there isn't a way to generalize over array size (at least until const generics land).

You could do this instead:
```rust
    let bytes_array = [0u8; 4]; // This value is going to be replaced by next line, so just set to 0
    bytes_array.copy_from_slice(&bytes);
    u32::from_be_bytes(bytes_array);
```

About the options, personally i'm fine with just having reasonable defaults, but as adding a config option is basically two lines, it's not a big effort. As it is we don't do many checks in the config values, and expect the admins to read the documentation and know what they are doing, but we could validate them sure (I think we check that the admin token is not an empty space, but mostly to avoid a security issue).

And yeah, we should keep a count of failed attempts and delete the token after a couple of tries (3-5 sound good, but it could also be configurable), or at least just delete it after the first failed try, which would require less work for us at the cost of being more cumbersome for the users.

vverst: I've added the configuration options. Still using the old `generate_token` that uses a single byte per number.
Can't immediately get the new way to work for large numbers since `10u32.pow()` gives an overflow error.

vverst:
I'm not sure how to proceed with testing. I would like to at least test the logic that authorizes the user when using email codes (`validate_email_code_str`).

However the function depends on `DbConn` which would require a database and does not allow me to replace it with fixed testing data. For example a Repository pattern with IUserRepository which would be implemented for db access DatabaseUserRepository and TestUserRepository for a simple test user repository would allow me. Would this be a UserRepository trait in Rust?

 Instead of changing the `DbConn` dependency I could also start an in memory database which would be easier to setup in the short-term but this seems less clean.

src/api/core/two_factor/email.rs
```diff
@@ -0,0 +1,328 @@
...
+/// Validate the email code when used as TwoFactor token mechanism
+pub fn validate_email_code_str(user_uuid: &str, token: &str, data: &str, conn: &DbConn) -> EmptyResult {
+    let mut email_data = EmailTokenData::from_json(&data)?;
+    let mut twofactor = TwoFactor::find_by_user_and_type(&user_uuid, TwoFactorType::Email as i32, &conn)?;
+    let issued_token = match &email_data.last_token {
+        Some(t) => t,
+        _ => err!("No token available"),
+    };
+
+    if issued_token != &*token {
```
dani-garcia: You might be able to use `crypto::ct_eq` for a constant time comparison here

src/api/core/two_factor/email.rs
dani-garcia: The 10u32.pow(x) should work for x equal or less than 9. Otherwise it would overflow yeah, but for the default value of 6 it works. Changing the number to 10u64 would allow up to 19 digits which I think is more than enough.

vverst: I have changed it to the `u64::from_be_bytes` way and added some verification checks to the config.

src/api/core/two_factor/email.rs
vverst: Changed it to use `crypto::ct_eq`.

vverst:
> Yeah, the current state of testing tooling for Rust is not that great and diesel doesn't allow mocking the database connections as far as I know. The repository pattern with traits would work but I think it might be a lot of work to implement for not a huge benefit.
> 
> The in memory database is probably the easiest option as I'm pretty sure you can just set `DATABASE_URL` to `:memory:` while using SQLite.

I tried the `:memory:` way and it works but it would be nicer to work with tests if we could separate it.  If possible I would like to work on the repository pattern using Traits idea if that is a viable solution or look around for other solutions.

dani-garcia:
Okay I think this is pretty much ready for merging, I'm going to make a new release to include all the MySQL changes and merge this after. That way we've got some time to uncover any bugs before a stable release.

dani-garcia:
Done! If you find a decent solution to the testing issue, I'd love to know!

nicolaspernoud:
Hi, @vverst : is it possible to enable 2FA with email with sqllite yet ?
If so, how to ?
Best regards,

hank:
I'm confused as well - how do we enable this?

nicolaspernoud:
@hank : start the container with the following options (or set those in admin panel) : 
```
    -e _ENABLE_EMAIL_2FA=true \
    -e INPUT_EMAIL_TOKEN_SIZE=6 \
    -e INPUT_EMAIL_EXPIRATION_TIME=600 \
    -e INPUT_EMAIL_ATTEMPTS_LIMIT=3 \
```
And set email 2FA in your account.

hank:
Still not working, can't activate 2FA in my account settings. Sending emails works for password hints. How do I get to the admin panel?
```
docker run -d 
--name='Bitwarden' 
--net='bridge' 
--log-opt max-size='50m' 
--log-opt max-file='1' 
-e TZ="America/New_York" 
-e 'SIGNUPS_ALLOWED'='false' 
-e 'SMTP_HOST'='****' 
-e 'SMTP_FROM'='***@***' 
-e 'SMTP_PORT'='***' 
-e 'SMTP_SSL'='false' 
-e '_ENABLE_EMAIL_2FA'='true' 
-e 'INPUT_EMAIL_TOKEN_SIZE'='6' 
-e 'INPUT_EMAIL_EXPIRATION_TIME'='600' 
-e 'INPUT_EMAIL_ATTEMPTS_LIMIT'='3' 
-e 'DOMAIN'='https://****' 
-e 'SMTP_EXPLICIT_TLS'='true' 
-p '8343:80/tcp' 
-v '/bitwarden/':'/data/':'rw' 
'mprasil/bitwarden' 
```

hank:
I just realized I'm using a different docker image; trying yours now!

hank:
OK it appears to be working now, but the recovery code option doesn't work. I have a green checkmark next to the email option. When will this be available? Thanks!

dani-garcia:
Fixed in the latest commit https://github.com/dani-garcia/bitwarden_rs/commit/e449912f05d63a3499609ae00184796dd7390bf0. You should delete the email 2fa and then reenable it again for the changes to apply

Issue #570, opened 2019-08-12
icon not downloading over outgoing proxy


kTitan:
Currently It looks like that my bitwarden_rs installation is not able to fetch icons, because it runs inside a restrictive network. 
Outgoing connections are not allowed, only via an normal proxy.
For that I tried to set these environment variables for the docker container:
```
http_proxy: 'http://proxy:8888/'
https_proxy: 'http://proxy:8888/'
```

But after that I still get these errors:

```
[2019-08-11 08:31:24][rocket::rocket][INFO] GET /icons/accounts.autodesk.com/icon.png image/png:
[2019-08-11 08:31:24][_][INFO] Matched: GET /icons/<domain>/icon.png (icon)
[2019-08-11 08:31:39][bitwarden_rs::api::icons][INFO] Download failed for https://accounts.autodesk.com/favicon.ico
[2019-08-11 08:31:39][bitwarden_rs::api::icons][ERROR] Error downloading icon: Empty response. Empty response
thread '<unnamed>' panicked at 'Error creating negative cache marker: Os { code: 2, kind: NotFound, message: "No such file or directory" }', src/libcore/result.rs:1051:5
```

I have seen that the rust lib https://github.com/seanmonstar/reqwest/releases added the support for use of system proxy in the latest release, and that you are already using it.
But I do not know how much work it would be to activate the feature inside the code.

BlackDex:
@kTitan I have created a pull-request for this feature. It was a small update.
I tested it and works as it should.

kTitan:
@BlackDex thanks for this quick pull request. 
I just have seen that the `hibp_breach` function is also not working, because it also needs the proxy. 
Could you add it there also? If you want I can also open an new issue for that.

BlackDex:
@kTitan There are some more locations where this could be needed.
the DUO auth is something i can see quicly.

BlackDex:
Maybe it is better for this to wait for the updated reqwest module, since that will enable the use_sys_proxy by default.

kTitan:
But there is no release date scheduled for the 0.10 version if it, as far as I can see. So it will perhaps takes months/years :/

PR #571, opened 2019-08-12
Added reqwest proxy support

BlackDex:
As requested in issue #570 i added system proxy support.
I have tested this it it works via the proxy by setting the environment variable.
Maybe we can create something to have this set in the environment, but i think that is up to docker it self.

### DIFF
--- a/src/api/icons.rs
+++ b/src/api/icons.rs
@@ -27,6 +27,7 @@ const ALLOWED_CHARS: &str = "_-.";
 lazy_static! {
     // Reuse the client between requests
     static ref CLIENT: Client = Client::builder()
+        .use_sys_proxy()
         .gzip(true)
         .timeout(Duration::from_secs(CONFIG.icon_download_timeout()))
         .default_headers(_header_map())


dani-garcia:
Nice, this shouldn't break anything, unless someome has a misconfigured proxy, but I think we can ignore those cases anyway as something was bound to break sometime.

We should document that you might need to set both the `HTTP_PROXY` and `HTTPS_PROXY` variable in the wiki somewhere too.

oguh22:
Hello,

Is it possible to use HTTP_PROXY env variable for tthee admin/diagnostics page ?

BlackDex:
@oguh22, what isn't working now?
I just tested it and the HTTP_PROXY and HTTPS_PROXY env variables seem to work for me.

oguh22:
@BlackDex ,
uuuhm unfortunatly  not ... it is working fine to get websiite images, on my proxy I can see request going to all differents websites.
When I do to the admin/diagnostics page, I see no logs on my proxy and bitwarden_rs is not able to perform checks

PR #585, opened 2019-08-24
Allow explicitly defined smtp auth mechansim

ViViDboarder:
This allows using `Xoauth2` or forcing an auth mechanism over an unencrypted connection.

### DIFF
--- a/src/mail.rs
+++ b/src/mail.rs
@@ -42,7 +42,7 @@ fn mailer() -> SmtpTransport {
 
     let smtp_client = match &CONFIG.smtp_auth_mechanism() {
         Some(auth_mechanism_json) => {
-            let auth_mechanism = serde_json::from_str::<SmtpAuthMechanism>(&auth_mechanism_json.clone());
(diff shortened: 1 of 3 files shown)


ViViDboarder:
@dani-garcia this is based on what we discussed on Matrix.

src/mail.rs
```diff
@@ -39,6 +40,17 @@ fn mailer() -> SmtpTransport {
         _ => smtp_client,
     };
 
+    let smtp_client = match &CONFIG.smtp_auth_mechanism() {
+        Some(auth_mechanism_json) => {
+            let auth_mechanism = serde_json::from_str::<SmtpAuthMechanism>(&auth_mechanism_json.clone());
```
dani-garcia: As we aren't using the value anywhere else, there is no need to clone it here.
```rust
    let smtp_client = match CONFIG.smtp_auth_mechanism() {
        Some(auth_mechanism_json) => {
            let auth_mechanism = serde_json::from_str::<SmtpAuthMechanism>(&auth_mechanism_json);
```

PR #589, opened 2019-08-27
Adds Healthcheck for default docker container

H3npi:
As of https://github.com/dani-garcia/bitwarden_rs/issues/246#issuecomment-524299620
a basic healthcheck for the container is implemented with this commit.

When everything is okay the same healthcheck will be applied to the other dockerfiles.

### DIFF
--- a/docker/aarch64/mysql/Dockerfile
+++ b/docker/aarch64/mysql/Dockerfile
@@ -97,5 +97,9 @@ COPY Rocket.toml .
 COPY --from=vault /web-vault ./web-vault
 COPY --from=build /app/target/aarch64-unknown-linux-gnu/release/bitwarden_rs .
 
+COPY docker/healthcheck.sh ./healthcheck.sh
+
+HEALTHCHECK --interval=10s --timeout=1s CMD bash healthcheck.sh || exit 1
+
 # Configures the startup!
 CMD ["./bitwarden_rs"]
--- a/docker/aarch64/sqlite/Dockerfile
+++ b/docker/aarch64/sqlite/Dockerfile
@@ -97,5 +97,9 @@ COPY Rocket.toml .
 COPY --from=vault /web-vault ./web-vault
 COPY --from=build /app/target/aarch64-unknown-linux-gnu/release/bitwarden_rs .
 
+COPY docker/healthcheck.sh ./healthcheck.sh
+
+HEALTHCHECK --interval=10s --timeout=1s CMD bash healthcheck.sh || exit 1
+
 # Configures the startup!
 CMD ["./bitwarden_rs"]
--- a/docker/amd64/mysql/Dockerfile
+++ b/docker/amd64/mysql/Dockerfile
@@ -94,5 +94,9 @@ COPY Rocket.toml .
 COPY --from=vault /web-vault ./web-vault
 COPY --from=build app/target/release/bitwarden_rs .
 
+COPY docker/healthcheck.sh ./healthcheck.sh
+
+HEALTHCHECK --interval=10s --timeout=1s CMD bash healthcheck.sh || exit 1
+
 # Configures the startup!
 CMD ["./bitwarden_rs"]
--- a/docker/amd64/mysql/Dockerfile.alpine
+++ b/docker/amd64/mysql/Dockerfile.alpine
@@ -76,5 +76,9 @@ COPY Rocket.toml .
 COPY --from=vault /web-vault ./web-vault
 COPY --from=build /app/target/x86_64-unknown-linux-musl/release/bitwarden_rs .
 
+COPY docker/healthcheck.sh ./healthcheck.sh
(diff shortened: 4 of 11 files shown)


dani-garcia:
As I mentioned in the comment, we also need to handle the case where HTTPS is enabled inside the container, either by testing both HTTP and HTTPS and succeeding if one does, or more naively by checking if the ROCKET_TLS variable is set and if then testing HTTPS.

H3npi:
> or more naively by checking if the ROCKET_TLS variable is set and if then testing HTTPS.

For better readability i'd create a bash script for the health check and call the bash script via `HEALTCHECK`. Any recommendations where you want the file to be located?

dani-garcia:
The best place is probably in the docker folder right next to the multiple architecture folders.

dani-garcia:
Okay a couple of small details now, use `COPY` instead of `ADD` , so that the CI doesn't complain (I don't think there is a difference between those two commands in this case anyway).

I'm not sure if all the containers have `bash` available, can we use `sh` instead? We could test anyway so no big deal either way. Same with the shebang, might be better to do `#!/usr/bin/env sh` if we can.

The interval seems kinda low to me, I would have chosen something like 10-20 seconds.

Other than that you can apply it to all the dockerfiles and i'll merge it.

mprasil:
In the health check script you should also use the configured HTTP(S) port if it's defined (`ROCKET_PORT`) otherwise the health check will fail for users that force specific port.

I don't think any of these can be set via admin interface (and config file) @dani-garcia right? Otherwise we would also have to check these.

mprasil:
Something like this should do the trick:

```sh
curl --fail "http://localhost:${ROCKET_PORT:-"80"}/alive" || exit 1
```

mprasil:
I realized that these values can be in theory set via env file. We need to be able to handle that otherwise the health check will start to fail for users using the file which would very likely lead to service unavailability.

In theory we could maybe just `source` the file if it exists?

dani-garcia:
At the moment you can’t modify the rocket options from the config no, but you can use the env file you are right.

Sourcing the file before launching the server sounds like it could work yeah.

mprasil:
What about `Rocket.toml`?

dani-garcia:
Well that file is only added at build time and I don’t think we mention anywhere that it can be edited so it could be considered an unsupported option.

But now that I say that, the env file is not available in the docker image either is it? It’s read from the application root directory but the user can only really touch inside the data folder, so it might not be a concern after all.

mprasil:
Well you can mount the files inside the container easily. I believe some people do mount the env file this way, I remember some issues regarding that.

Having said that, I'd agree that it might be considered unsupported as there are generally better alternatives, but we should be mindful of the fact that this is potentially breaking change.

H3npi:
> I realized that these values can be in theory set via env file. We need to be able to handle that otherwise the health check will start to fail for users using the file which would very likely lead to service unavailability.

I am unsure about the service unavailability. AFAIK the only effect is that the container is marked as unhealthy. The service should still be available, but the status will be `(health: starting)` forever?

BlackDex:
And what about the websocket port?
That should be something which also need to be checked i think.

Ayitaka:
To add another twist, ROCKET_PORT can also be set as an environment variable within docker-compose.yml, which I don't think can be sourced?

(I suppose I should move all those things to an env file, but the wiki example I used when starting had settings defined in docker-compose.yml so I just built out from there.)

mprasil:
@Ayitaka the variables set in docker compose are eventually exposed as environment variables that the script already checks, so I think it should be OK there.

@H3npi well that depends, some tools like kubernetes or traefik won't route traffic to a backend that reports unhealthy status. In those cases the service would be down.

dani-garcia:
The websockets server can be queried like this:
```sh
curl -i -N -H "Sec-WebSocket-Key: test" localhost:3012
```

But this is likely to cause a panic in the websockets code because at the moment it isn't handling requests that send no data. It also would keep the connection open and I don't know of a way to close it other than sending `^C` to the curl command.

For now I think the basic version that just handles the port already in the environment is good enough:
```sh
curl --fail "http://localhost:${ROCKET_PORT:-"80"}/alive" || exit 1
```

The only other ways that port could be changed are the `Rocket.toml` file and the `.env` file, and both would require someone to build the image themselves so they could always also modify the script.

BlackDex:
For the websocket maybe adding --max-time with a low value (in seconds) is an option.

dani-garcia:
That would mean that all requests end in a time out error, both the successful ones and the errored ones.

We could always parse the output in stdout but that seems dirty.

H3npi:
@mprasil You are right, i just thought of plain docker.

@dani-garcia I just added the environment port.

dani-garcia:
Okay then, I think this is ready. Thanks!

PR #621, opened 2019-09-12
Adds support for PostgreSQL which adds #87 and is mentioned in #246.

swedishborgie:
The biggest change to existing code is that `replace_into()` isn't supported by Diesel for the PostgreSQL back end, instead requiring the use of `on_conflict()`. This unfortunately requires a branch for `save()` on most of the models and requires PostgreSQL >= 9.5.

I've included `Dockerfile`'s for amd64 for both the Debian and Alpine variants, it should be straight forward to create new files for the other architectures but I don't currently have access to the other architectures and I didn't want to submit anything untested.

I've only done a couple hours worth of testing but everything seems to work. The migration SQL and model schema are both very similar to the existing two back ends.

Please let me know if you'd like any changes or if I missed anything.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -116,6 +116,7 @@ dependencies = [
  "num-derive 0.2.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "num-traits 0.2.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "oath 0.10.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "openssl 0.10.24 (registry+https://github.com/rust-lang/crates.io-index)",
  "percent-encoding 2.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "quoted_printable 0.4.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "regex 1.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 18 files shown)


dani-garcia:
Very nice job @swedishborgie!

It all looks good to me, pity that replace_into isn't supported, but the workaround is simple enough in my view.

One small question I have is the addition of the openssl dependency, as you say for SSL support, but it isn't actually used anywhere, is diesel using it? And in that case shouldn't diesel require it or expose it as a feature flag? (Not that I'd have a problem adding the dependency, it just seems a bit strange)

swedishborgie:
I ran into that issue when trying to build the Alpine image. The Rust PostgreSQL bindings link against libpq, which links against OpenSSL to provide encrypted connection support. The libpq-dev package on Alpine requires that the OPENSSL_config struct be defined while linking and without it the project fails to link. Explicitly including the openssl crate seemed to pull in the right symbols. On Debian everything seemed OK without the import.

Here's the issues I followed to get to the solution I ended up with:
- https://github.com/emk/rust-musl-builder/issues/64
- https://github.com/sgrif/pq-sys/issues/25

dani-garcia:
Makes sense, thanks for your work!

ptman:
Is it possible to migrate from sqlite to postgresql?

bjo81:
Yes. Let it create the postgresql-database, then use pgloader to migrate with the following config:

```
load database
     from sqlite:///var/lib/bitwarden_rs/db.sqlite3 
     into postgresql://user:pass@localhost/database
     with data only, include no drop, reset sequences
     excluding table names like '__diesel_schema_migrations'
;
```

PR #624, opened 2019-09-16
Fix issue with downloading attachments and PostgreSQL backend.

swedishborgie:
Apologies for the rapd-fire PR's. While testing today I realized that downloading attachments with the PostgreSQL backend is currently not working.

The root cause is that since attachments don't use UUID identifiers like the majority of other tables the usage of the `CHAR(36)` type causes the identifiers to get padded out to 36 characters with spaces because of the [odd way PostgreSQL handles the CHAR type](https://www.postgresql.org/docs/current/datatype-character.html). Since the request URL's get trimmed this causes attachments to be not retrievable.

Since CHAR() in PostgreSQL is significantly different in behavior from MySQL and sqlite I added a migration that converts the `CHAR()` types to `VARCHAR()`. This appears to fix the attachments issue and should prevent any future padding issues caused by PostgreSQL's handling of the `CHAR()` type.

Please let me know if you'd like any changes.


dani-garcia:
Thanks! I agree that trying to match the two other backends with regards to varchar is probably the right move here.

If you are curious, the reason for the different id format was to match the upstream server, but I don't think that the clients care about the format, so we could have just used uuids like everywhere else anyway.

PR #631, opened 2019-09-23
Use Access-Control-Allow-Method

vverst:
- It's not `Access-Control-Allow-Methods` but `Access-Control-Allow-Method` where the value is the HTTP method that caused the preflight request.
- Changed the name to something shorter since it appears in the logging.


dani-garcia:
Thanks!

PR #638, opened 2019-09-30
Add sqlite binary into the docker images

mprasil:
This is done to enable backup functionality in the admin interface while
we're waiting for the libsqlite-sys 0.17 to bubble up in the upstream
dependencies. Then we can start using `VACUUM INTO`

This also extends the check for the sqlite binary to also try `sqlite3`
as this is the name of the binary in baseimage distributions we use.

### DIFF
--- a/docker/aarch64/sqlite/Dockerfile
+++ b/docker/aarch64/sqlite/Dockerfile
@@ -82,6 +82,7 @@ RUN apt-get update && apt-get install -y \
     openssl \
     ca-certificates \
     curl \
+    sqlite3 \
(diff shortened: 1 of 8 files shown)


dani-garcia:
If we need to check for both binaries here we’d also need to check when we run the `.backup` command, no? I think we do that in db/mod.rs but not sure.

mprasil:
Good point! The backup was actually using `sqlite3` command. I guess we can remove the check for `sqlite` as I can't find system where it would be called that way and not `sqlite3` and on those systems backup never worked anyways.

PR #639, opened 2019-10-01
Change CORS headers

### LINKED ISSUE #629: "Method PUT is not allowed by Access-Control-Allow-Methods"
Getting the following error when trying to save an edit in the Bitwarden Safari extension 1.42.0, after updating Safari to 13.0, running bitwarden_rs 1.10.0-4c07f05b:

"Method PUT is not allowed by Access-Control-Allow-Methods"

vverst:
All CORS header used to be added always even when that was not necessary.
Changed to only add Allow-Origin to all requests and move the others to preflight OPTIONS request.

Extension sends the request with `file://` Origin. This seems to cause problems.
If Origin is `file://` change it to the wildcard.

This seems to correspond with the responses from the official API.

Before merging
- [x] Confirmation of fix from other Safari 13 extension user

Fixes #629

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -42,6 +42,13 @@ impl CORS {
             _ => "".to_string(),
         }
     }
+
+    fn valid_url(url: String) -> String {
+        match url.as_ref() {
+            "file://" => "*".to_string(),
+            _ => url,
+        }
+    }
 }
 
 impl Fairing for CORS {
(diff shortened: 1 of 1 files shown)


seanfarley:
I can confirm that deleting and editing an item via Safari 13's extension now work. I got a weird "Cypher not found" error that disappeared too quickly and never showed up again ... 😬

vverst:
I think the `Cipher not found` error is because it was actually already deleted (I also had that error). During the deletion it wasn't returning `Access-Control-Allow-Origin: *` because it was set only on ~`Content-Type: json` responses and not on the response to the `DELETE` http request. This caused the extension to think it was not deleted but it actually was.

seanfarley:
Ah, fair enough then. Well, anyways, it works for me ^_^

vverst:
That’s good 🙂. Thanks for testing it.

PR #643, opened 2019-10-05
Updated icon blacklisting.

BlackDex:
- Blacklisting was not effective for redirects and rel href
- Able to blacklist non global IP's like RFC1918, multicast etc...

### DIFF
--- a/src/api/icons.rs
+++ b/src/api/icons.rs
@@ -1,12 +1,13 @@
 use std::fs::{create_dir_all, remove_file, symlink_metadata, File};
 use std::io::prelude::*;
+use std::net::ToSocketAddrs;
 use std::time::{Duration, SystemTime};
 
 use rocket::http::ContentType;
 use rocket::response::Content;
 use rocket::Route;
 
-use reqwest::{header::HeaderMap, Client, Response};
+use reqwest::{header::HeaderMap, Client, Response, Url};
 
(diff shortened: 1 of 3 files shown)


src/api/icons.rs
```diff
@@ -60,22 +61,47 @@ fn icon(domain: String) -> Content<Vec<u8>> {
...
+fn check_icon_domain_is_blacklisted(domain: &str) -> bool {
+    let mut is_blacklisted = false;
+    if CONFIG.icon_blacklist_non_global_ips() {
+        is_blacklisted = (domain, 0)
+            .to_socket_addrs()
+            .map(|x| {
+                for ip_port in x {
+                    if !ip_port.ip().is_global() {
+                        warn!("IP {} for domain '{}' is not a global IP!", ip_port.ip(), domain);
+                        return true;
```
dani-garcia: Instead of doing to_socket_addrs including a port, wouldn't it be easier to parse the string into an IpAddr directly?
https://doc.rust-lang.org/std/net/enum.IpAddr.html#examples

BlackDex: I Tried that.
But it failed on me with DNS records having more then one A record set.
It always generated a AddrParseError for some reason.

dani-garcia: Fair enough, it isn't a big deal, it's okay as it is. Thanks!

PR #657, opened 2019-10-10
Updated authenticator TOTP

BlackDex:
- Added security check for previouse used codes
- Allow TOTP codes with 1 step back and forward when there is a time
drift. This means in total 3 codes could be valid. But only newer codes
then the previouse used codes are excepted after that.


src/api/core/two_factor/authenticator.rs
```diff
@@ -94,27 +94,69 @@ fn activate_authenticator_put(data: JsonUpcase<EnableAuthenticatorData>, headers
...
+    if generated_now == totp_code && time_step_now > twofactor.last_used {
+        twofactor.last_used = time_step_now;
+        twofactor.save(&conn)?;
+        return Ok(());
+    } else if generated_now == totp_code && time_step_now <= twofactor.last_used {
+        warn!("This or a future TOTP code has already been used!");
+        err!("Invalid TOTP code!");
     }
 
-    Ok(())
```
dani-garcia: There's a lot of repetition here, you could probably do something like:
```rust
pub fn validate_totp_code(user_uuid: &str, totp_code: u64, secret: &str, conn: &DbConn) -> EmptyResult {
    use oath::{totp_raw_custom_time, HashType};
    use std::time::{UNIX_EPOCH, SystemTime};

    let decoded_secret = match BASE32.decode(secret.as_bytes()) {
        Ok(s) => s,
        Err(_) => err!("Invalid TOTP secret"),
    };

    let mut twofactor = TwoFactor::find_by_user_and_type(&user_uuid, TwoFactorType::Authenticator as i32, &conn)?;

    // Get the current system time in UNIX Epoch (UTC)
    let current_time: u64 = SystemTime::now().duration_since(UNIX_EPOCH)
        .expect("Earlier than 1970-01-01 00:00:00 UTC").as_secs();

    // Until here it's the same
    
    let steps = 1; // How many steps to either side of current to check
    for i in -steps..=steps {
        let time_step = (current_time / 30) as i32 + i;
        let generated = totp_raw_custom_time(&decoded_secret, 6, 0, 30, current_time + i * 30, &HashType::SHA1);
        if generated == totp_code && time_step > twofactor.last_used {

            if i != 0 {
                info!("TOTP Time drift detected. The time step offset is {}", i);
            }

            twofactor.last_used = time_step;
            twofactor.save(&conn)?;
            return Ok(());
        } else if generated_now == totp_code && time_step <= twofactor.last_used {
            warn!("This or a future TOTP code has already been used!");
            err!("Invalid TOTP code!");
        }
    }

    // Else no valide code received, deny access
    err!("Invalid TOTP code!");
}
```

BlackDex:
@dani-garcia, looking at those libraries. The only thing i am missing is that they do not return the time step it was valid for. Just a true.

Now i could fork/pr this of course.
What do you think?

I think having that extra check is an useful addition to the security.

dani-garcia:
Well the easiest solution would be to save the latest step in the range we admit, though that could mean a user wouldn't be able to login in two devices if they try to do it nearly at the same time.

I think forking or making a PR would be too much work for something that we already have working here, so maybe we'll keep using the current library. It's not like the TOTP algorithm is very complex anyway.

BlackDex:
I will adapt the code to be more nice and update the pr.

PR #670, opened 2019-10-16
Remove unsupported WS transport

Jellyfrog:
This stops the spamming:
```
[2019-10-16 13:06:27][bitwarden_rs::api::notifications][INFO] Server got message 'Binary Data<length=3>'.                                           
[2019-10-16 13:06:42][bitwarden_rs::api::notifications][INFO] Server got message 'Binary Data<length=3>'.                                           
[2019-10-16 13:06:57][bitwarden_rs::api::notifications][INFO] Server got message 'Binary Data<length=3>'.                                           
[2019-10-16 13:07:12][bitwarden_rs::api::notifications][INFO] Server got message 'Binary Data<length=3>'.                                           
[2019-10-16 13:07:27][bitwarden_rs::api::notifications][INFO] Server got message 'Binary Data<length=3>'.                                           
[2019-10-16 13:07:42][bitwarden_rs::api::notifications][INFO] Server got message 'Binary Data<length=3>'.                                           
[2019-10-16 13:07:57][bitwarden_rs::api::notifications][INFO] Server got message 'Binary Data<length=3>'.   
```

I'm not sure if it is unsupported, nor if it breaks for someone else.
Please treat it as "Works for me".

### DIFF
--- a/src/api/notifications.rs
+++ b/src/api/notifications.rs
@@ -26,7 +26,7 @@ fn negotiate(_headers: Headers, _conn: DbConn) -> JsonResult {
     let mut available_transports: Vec<JsonValue> = Vec::new();
 
     if CONFIG.websocket_enabled() {
-        available_transports.push(json!({"transport":"WebSockets", "transferFormats":["Text","Binary"]}));
+        available_transports.push(json!({"transport":"WebSockets", "transferFormats":["Text"]}));
     }
 
     // TODO: Implement transports


dani-garcia:
Well technically we do use the binary format for the responses (it's MessagePack), so simply disabling it might lead to more problems down the line, particularly with the browser extensions which are the ones that started using the binary format if i recall correctly.

When I created the notifications backend, the clients only ever sent one message to connect for the first time, and then the server sent some pings from time to time to avoid connection timeouts, but at some point it seems the clients started to send frequent pings, probably to avoid dropped connections, we seem to be handling the response correctly (it just echoes back the message which we already do).

I think it would be safer to just remove the line that creates that log entry, which is not very useful in the first place anyway: https://github.com/dani-garcia/bitwarden_rs/blob/2ccbc1cd3817451b0087ba1eacdf13551bd55f75/src/api/notifications.rs#L160

PR #696, opened 2019-11-01
Compile sqlite version by default

mprasil:
This sets the default feature to sqlite, so that it will correctly
compile "out of the box". Users can still choose different backend by
running:

```
cargo build --no-default-features --features <backend>
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,11 @@ publish = false
 build = "build.rs"
 
 [features]
+# By default compile the sqlite version
+# different DB backend can be compiled by running:
+# cargo build --no-default-features --features <mysql|postgresql>
+default = ["sqlite"]
+
 # Empty to keep compatibility, prefer to set USE_SYSLOG=true
 enable_syslog = []
 mysql = ["diesel/mysql", "diesel_migrations/mysql"]
--- a/build.rs
+++ b/build.rs
@@ -2,14 +2,14 @@ use std::process::Command;
 
 fn main() {
     #[cfg(all(feature = "sqlite", feature = "mysql"))]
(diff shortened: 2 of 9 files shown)


dani-garcia:
We talked about this when the mysql PR was still open and it seems the preferred option was to explicitly require one option: https://github.com/dani-garcia/bitwarden_rs/pull/493#issuecomment-497058038.

Any reason you think it would be better this way? It would mean requiring anyone compiling it to change their scripts again. Personally I don't have a strong opinion either way, especially knowing that there's an error message https://github.com/dani-garcia/bitwarden_rs/blob/master/build.rs#L12.

mprasil:
Yeah, so my selfish motivation is that RLS in VS Code fails to compile it, because it can't use the features flag. I wonder if more such system exist? Could be beneficial there.

Other than that I just feel that things should do the right thing by default. We shouldn't rely on error message to build the most likely version users want.

Just my opinion, no strong feelings either way.

dani-garcia:
I use `"rust.features": ["sqlite"]` in the workspace config for VSCode. It would be good if there was a way to configure RLS from the Cargo.toml but I don't think that's possible.

You are right about doing the right thing, but I expect most users building it are reading the instructions through the wiki, and we do include the features there: https://github.com/dani-garcia/bitwarden_rs/wiki/Building-binary.

mprasil:
Those are fair points. I don't share your optimism about users reading documentation, but the error message does lead the right direction.

It just irks me that it does not build "out of the box", so I figured I reopen the discussion with PR. If you think current setup is okay, feel free to close the PR, I don't have strong opinions either way.

PR #703, opened 2019-11-05
Don't include excluded global equivalent domains during sync

### LINKED ISSUE #681: Excluded domain rules still in effect
In the web vault you can configure domain rules for equivalent domains. This works for extra custom domains. But if I want to exclude a predefined `Global Equivalent Domains`, the logins for excluded domains are still listed together.

For example:
1. Exclude item `mi.com, xiaomi.com` from the vault.
2. Create one item for `mi.com`, one item for `xiaomi.com`.
3. Force sync browser extension.
4. Visit `mi.com`.

Expected Outcome:
Only item for `mi.com` is listed in the extension.

Actual Outcome:
Both items are listed.

This is tested on the Chrome extension. Tested with official bitwarden vault and it does not have this problem. Browser extension honored the exclusion correctly.

patrickli:
Fixes #681

Please excuse my rusty rust.

### DIFF
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -88,7 +88,7 @@ fn sync(data: Form<SyncData>, headers: Headers, conn: DbConn) -> JsonResult {
     let domains_json = if data.exclude_domains {
         Value::Null
     } else {
-        api::core::get_eq_domains(headers).unwrap().into_inner()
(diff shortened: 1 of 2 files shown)


dani-garcia:
Strange decision by bitwarden to just remove the elements here instead of respecting the Excluded boolean, but we'll have to follow it.

Your code looks fine to me, thanks!

PR #716, opened 2019-11-11
Add vim modeline for GitHub Linguist

ThomDietrich:
Super minor change to get correct syntax highlighting

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -170,3 +170,5 @@
 # SMTP_PASSWORD=password
 # SMTP_AUTH_MECHANISM="Plain"
 # SMTP_TIMEOUT=15
+
+# vim: syntax=ini


dani-garcia:
Huh, neat! Didn't think Github would support that. Thanks!

ThomDietrich:
;) ran into the issue in another more annoying case

chris-morgan:
I request that this be changed to something other than a Vim modeline, because this actually *breaks* highlighting in Vim. Before, Vim decided on sh due to the .env file suffix, which highlighted things decently. By using the wonky `set syntax=ini` (the actual differences between `'syntax'` and `'filetype'` are fuzzy to me, but `'filetype'` is the one to use, and using `'syntax'` will broadly just break everything), this breaks that, and doesn’t load a new syntax file in its stead, as there is no bundled file syntax/ini.vim. There’s syntax/dosini.vim, which you could use—no idea whether Linguist would catch that.

ThomDietrich:
Hello @chris-morgan,
I am very sorry for introducing this new bug! Had to check twice but turns out my custom vim settings also include a ini syntax file. Thinking about it I don't believe `ini` is the wrong syntax and am not sure if it is a shell script - even though it does not violate the sh syntax.

Solutions:
1. Vim modeline `dosini`which is a valid Linguist alias for `INI`
2. Vim modeline `sh`which is a valid Linguist alias for `Shell`
3. Remove vim modeline and add [Linugist only gitattribute](https://github.com/github/linguist#using-gitattributes)

Wdyt?

Issue #718, opened 2019-11-12
podman error on running bitwardenrs/server:latest


gnu300:
Since Fedora 31 defaults to cgroupv2 [1] podman is one of the recommended alternatives. Unfortunately podman can't run bitwarden_rs:

```
[gnu300@fuso container]$ docker run -d --name bitwarden -v /bw-data/:/data/ -p 8888:80 bitwardenrs/server:latest
Trying to pull docker.io/bitwardenrs/server:latest...
Getting image source signatures
Copying blob de171338eba4 done
Copying blob 3332f6a74903 done
Copying blob 4f3b3bfff359 done
Copying blob 68c6b84d8d28 done
Copying blob 8d691f585fa8 done
Copying blob 35e7083657ef done
Copying blob fcb7d34ccc30 done
Copying config 97d8cc3df2 done
Writing manifest to image destination
Storing signatures
Error: executable file not found in $PATH: No such file or directory: OCI runtime command not found error
```

In order to get the container running with podman you'll have to change:
`CMD ["./bitwarden_rs"]`
to
`CMD ./bitwarden_rs`

Tested on Fedora 31 with podman version:
Version:            1.6.2
RemoteAPI Version:  1
Go Version:         go1.13.1
OS/Arch:            linux/amd64

[1] https://www.redhat.com/sysadmin/fedora-31-control-group-v2

mprasil:
Exec form of specifying `CMD` is actually preferred according to official documentation. But maybe a slightly cleaner approach would be to set `WORKDIR /` and `CMD ["/bitwarden_rs"]`. (note the removed dot to make the path absolute)

I can't reproduce your issue on my system, can you test above changes?

gnu300:
With the suggested changes i am able to build and run the container with podman. Here is the PR for it. #720 

Thanks!

mprasil:
I think we can close this with PR almost out of the door. Thanks for working on this and thanks for finally giving me good reason to install podman. :smile:

PR #720, opened 2019-11-13
cleaner startup exec in order to build the image and run the containe…

gnu300:
…r with podman/libpod

### DIFF
--- a/docker/amd64/sqlite/Dockerfile
+++ b/docker/amd64/sqlite/Dockerfile
@@ -94,4 +94,5 @@ COPY docker/healthcheck.sh ./healthcheck.sh
 HEALTHCHECK --interval=30s --timeout=3s CMD sh healthcheck.sh || exit 1
 
 # Configures the startup!
-CMD ["./bitwarden_rs"]
+WORKDIR /
+CMD ["/bitwarden_rs"]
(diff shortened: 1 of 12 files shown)


mprasil:
For reference this change was discussed under #718 to enable running on podman with cgroupv2, but the change should be also an improvement for regular docker by making the `WORKDIR` and binary location more explicit and less reliable on defaults.

@gnu300 can you also make the same change to the rest of the Dockerfiles?

docker/amd64/sqlite/Dockerfile
```diff
@@ -94,4 +94,5 @@ COPY docker/healthcheck.sh ./healthcheck.sh
 HEALTHCHECK --interval=30s --timeout=3s CMD sh healthcheck.sh || exit 1
 
 # Configures the startup!
-CMD ["./bitwarden_rs"]
+WORKDIR /
+CMD ["/bitwarden_rs"]
```
mprasil: Change the rest of the Dockerfiles please

gnu300: np -> done

mprasil: Almost there! I think you missed the alpine Dockerfile?

gnu300: omg, yes of course.

dani-garcia:
Looks good to me, thanks!

PR #728, opened 2019-11-16
Add the ability to disable signups, but allow signups from a whitelist

### LINKED ISSUE #727: Limit signups to specific domains?
Is there a way to specify a list of domains that can be used to sign up? I'd like to limit signups to email addresses I control, but I don't want to disable signups alltogether.

tomuta:
This feature can be enabled by setting SIGNUPS_ALLOWED=false and
providing a comma-separated list of whitelisted domains in
SIGNUPS_DOMAINS_WHITELIST.

Fixes #727

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -95,6 +95,14 @@
 ## Controls if new users can register
 # SIGNUPS_ALLOWED=true
 
+## Controls if new users from a list of comma-separated domains can register
+## even if SIGNUPS_ALLOWED is set to false
+##
+## WARNING: There is currently no validation that prevents anyone from
+##          signing up with any made-up email address from one of these
+##          whitelisted domains!
+# SIGNUPS_DOMAINS_WHITELIST=example.com,example.net,example.org
+
 ## Token for the admin interface, preferably use a long random string
 ## One option is to use 'openssl rand -base64 48'
 ## If not set, the admin panel is disabled
--- a/src/api/core/accounts.rs
+++ b/src/api/core/accounts.rs
@@ -90,7 +90,7 @@ fn register(data: JsonUpcase<RegisterData>, conn: DbConn) -> EmptyResult {
             }
         }
         None => {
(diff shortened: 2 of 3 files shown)


dani-garcia:
Couple of things, at the moment we don't validate that the user owns the email when they register, so anyone could potentially register with one of those domains, so at the moment this is more security by obscurity. That said I'm fine adding this but we should make the fact that we don't validate more obvious in the comments in the .env file at least.

The emailaddress crate that you add doesn't make any complicated attempt at parsing the email, beyond splitting it in two chunks by the @, so I'd prefer to remove the dependency and do it directly:
https://github.com/pwoolcoc/emailaddress-rs/blob/master/src/simple.rs#L8

Lastly, instead of allocating a Vec, it's better to work with the Iterator you get from split directly:
```rust
    let edom= "example.com";
    let doms= "example.com,other.org,dani.xxx";
    
    doms.split(",").any(|d| d == edom)
```

tomuta:
Thanks, will work on your suggestions. This is my very first time dabbling in rust, so I apologize for that.

tomuta:
How does this look? I think to add a validation step I'll work on a separate merge request.

dani-garcia:
Looks good to me!

For the email validation, we'd have to send an email to the address with a link that would basically set a boolean in the database to true, and if email validation is enabled we shouldn't allow the users to login until that happens. For that I'd recommend to follow whatever upstream bitwarden is doing, but a simple JWT in the email link should be enough to validate the user, this is similar to what we are doing with email organization invites.

It might be a bit of a complex task for a new user, but if you are up for it we might be able to help here and there.

jjlin:
> Couple of things, at the moment we don't validate that the user owns the email when they register, so anyone could potentially register with one of those domains, so at the moment this is more security by obscurity. That said I'm fine adding this but we should make the fact that we don't validate more obvious in the comments in the .env file at least.

For traceability, email verification was subsequently added in #730. There are still some obsolete references to this caveat in the wiki pages.

PR #730, opened 2019-11-17
Implement change-email, email-verification, account-recovery, and welcome notifications

tomuta:
- [x] Settings -> Change Email: Send out change verification emails and validate token (also validate against domain whitelist, if applicable)
- [x] In password hint email: Add a link to account recovery page where user can request a delete account verification email, and when they click the enclosed link then the account gets deleted (e.g. they cannot remember the password at all)
- [x] Main vault page: Show "Verify Email" tile and send verification email when button is clicked, and mark account as validated when link is clicked
- [x] Send welcome email upon registration (with verification link, if verification is required)
- [x] Reject login after signup if email verification is required and the user hasn't verified it yet. Send a few reminders if they attempt to login later.
- [x] Add schema changes for pgsql and mysql


tomuta:
I'm a little stuck now, I can't figure out how to get the "Verify Email" tile to show. The user's "EmailVerified" field shows as `false` as expected.

tomuta:
@dani-garcia I am getting really close to finishing up most of these features, but for the life of it I cannot figure out why the `app-verify-email-token` component doesn't show up. It's supposed to key off of the `EmailVerified` field (which shows `false` as expected), but I'm not sure how to work with Angular, so I'd really appreciate any help/hints as to why it's not showing up.

dani-garcia:
Well to be honest I don't know how the client works either, maybe you can ignore that for now, and when a non validated user tries to login, just return an error and send another email?

tomuta:
> Well to be honest I don't know how the client works either, maybe you can ignore that for now, and when a non validated user tries to login, just return an error and send another email?

Actually, I am not yet trying to solve this particular issue yet. But yes, that was the approach I planned on to validate signups.

What I'm stuck with right now is trying to get that "VERIFY EMAIL" box on the top-right to show:
Screenshot from 2019-11-19 10-16-52

As far as I can tell, this box should be visible if `EmailVerified` is set to `false`.

dani-garcia:
Right, that's what I mean, instead of allowing the user to login and displaying that box, simply return an error if the user is trying to login with an unverified email, then later we can add an option to let unverified users login, which should show that window.

tomuta:
Oh I see. Well, I wanted to get the "verify email" option to work first, because that's more in line with regular Bitwarden.  The "signup verification" feature I want to add later on is a little bit of a departure from the normal Bitwarden flow, and my that in my mind should be optional and disabled by default.

dani-garcia:
@tomuta I've looked into it and there are two places to return the email verified:

Here is the other one: https://github.com/dani-garcia/bitwarden_rs/blob/c5b7447dac3aa7620f039d2849d0fa149105adf4/src/db/models/device.rs#L90
Where it should probably be changed from true to user.verified_at.is_some()

tomuta:
> @tomuta I've looked into it and there are two places to return the email verified:
> 
> Here is the other one:
> 
> https://github.com/dani-garcia/bitwarden_rs/blob/c5b7447dac3aa7620f039d2849d0fa149105adf4/src/db/models/device.rs#L90
> 
> Where it should probably be changed from true to user.verified_at.is_some()

Wow, that did the trick! Thank you so much!

tomuta:
@dani-garcia I think I got all of this working, including account recovery/deletion (which is on the wishlist in #246). Does this look good to you?

dani-garcia:
Looking at the code it looks all very good to me! Hopefully I'll get some time to properly test it in the weekend, but overall very high quality PR, congrats @tomuta! 👍

src/crypto.rs
```diff
@@ -52,6 +52,15 @@ pub fn get_random(mut array: Vec<u8>) -> Vec<u8> {
     array
 }
 
+pub fn get_random_numeric_digits(n: usize) -> String {
+    let mut ret = String::new();
```
dani-garcia: This should probably use the same code in https://github.com/dani-garcia/bitwarden_rs/blob/eab145f343f7316c9c1c53af6c8d1a851e129fc8/src/api/core/two_factor/email.rs#L113-L126

We should extract that one which I feel is easier to understand to the crypto.rs file and call it from both places.

tomuta: Ah, I didn't realize there was such code already. I moved this function into crypto.rs as you suggested.

dani-garcia:
Thanks, merged!

PR #741, opened 2019-11-22
Add an option to fetch and parse href="data:image"

BlackDex:
Some sites are using base64 encoded inline images for favicons.
This will try to match those with some sane checks and return that.
These icons will have lower prio then the icons with a normal URL.


dani-garcia:
Looks good to me, but what’s the reason for the git dependency? Is the last version on crates.io not updated?

BlackDex:
Nope it isn't updated. Although not that much changed, but he removed some deprecated functions.
Seemed to me it's better to use the latest one, especially with the latest rust version we use here.

PR #755, opened 2019-12-02
Create an issue template

mqus:
I'm not sure if this is needed but I think it could be useful in lessening the workload.

As this is only something I threw together in a few minutes, I'm eager to hear any constructive criticism or changes (I already see some spelling errors :wink: ).
If you think this is too restrictive or not needed you can also close this request right away, I just felt that the discussion would be helpful at 462 closed (solved?) issues.

If this was discussed already, feel free to point it out as I'm not aware of it.

One point where I wasn't sure: All non-headlines (which should not be present in a finished issue) could also be wrapped in a `<!-- -->` to make them disappear when rendered.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,28 +1,33 @@
-Please fill out the following template to make solving your problem easier and faster for us. 
+<!--
+Please fill out the following template to make solving your problem easier and faster for us.
 This is only a guideline. If you think that parts are unneccessary for your issue, feel free to remove them.
 
 Remember to hide/obfuscate personal and confidential information, 
-such as Names, global IP/DNS adresses and especially passwords, if neccessary.
+such as names, global IP/DNS adresses and especially passwords, if neccessary.
+-->
 
 ### Subject of the issue
-Describe your issue here.
+<!-- Describe your issue here.-->
 
 ### Your environment
-* Version of bitwarden_rs, mysql/postgresql/sqlite, docker/package/built from source?
-* If applicable: client used and version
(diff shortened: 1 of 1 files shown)


dani-garcia:
Yeah this could be a good idea, indeed.

I'd do as you say and put most of the non-headlines in a comment block, to make sure they don't show in the final version, maybe with the exception of the environment block, where I think it makes more sense as a series of bullet points that you could answer directly next to each one.

I'd also separate it a bit more to make it easier to answer directly:
```markdown
<!-- The version number, obtained from the logs or the admin page -->
* Bitwarden_rs version: 
<!-- How the server was installed: Docker image / package / built from source -->
* Install method: 
* Database used: 
* Clients used: <if applicable>
* Reverse proxy and version: <if applicable>
* Version of mysql/postgresql: <if applicable>
* Other relevant information
```

What do you think?

dani-garcia:
Nice, thanks!

mqus:
Sure! I should also mention that:
* I haven't done this before, so I don't know if you have to set something in the repo settings to make this work
* Github [suggests](https://help.github.com/en/github/building-a-strong-community/about-issue-and-pull-request-templates#issue-templates) to use an `ISSUE_TEMPLATE` directory with separate templates for feature requests, bug reports, questions, etc., but I think that this simple file fits better and is a broader approach for everything. We can expand on this later if neccessary.

dani-garcia:
I just checked and it shows in new issue for me directly, so no problems there.

Having multiple templates seems a bit overkill to me at the moment, but if we discover it to be an issue we can change to the directory method.

PR #779, opened 2019-12-23
Add Github build Action

publicarray:
Sometimes It's nice to have a binary to download

### DIFF
--- a/.github/workflows/rust-linux.yml
+++ b/.github/workflows/rust-linux.yml
@@ -14,7 +14,7 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - name: Install dependencies
-      run: sudo apt update && sudo apt install --no-install-recommends openssl sqlite3 libmysql++-dev
+      run: sudo apt update && sudo apt install --no-install-recommends openssl sqlite3 libmysql++-dev libpq-dev
 
     - name: Cache cargo registry
       uses: actions/cache@v1.0.3
--- a/.github/workflows/rust-linux.yml
+++ b/.github/workflows/rust-linux.yml
@@ -1,0 +1,34 @@
+name: build-linux
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: install dependencies
+      run: sudo apt update && sudo apt install --no-install-recommends openssl ca-certificates curl sqlite3
+    - name: Cache
+      uses: actions/cache@v1.0.3
+      with:
+        path: target
+        key: ${{ runner.os }}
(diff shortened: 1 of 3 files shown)


dani-garcia:
Nice, I thought about integrating this in the past but never had the time to properly look into it.

A couple of comments:

- Travis for example also caches the .cargo directory, not just the target folder, any reason for choosing only that one? The cache example in actions_rs does a bit more https://github.com/actions/cache/blob/master/examples.md#rust---cargo.

- Currently the actions are using what I asume is the latest nightly, is there a way to make it respect the rust-toolchain file? This shouldn't be a major problem as lately we haven't had any problems with broken nightlies but would be good to make the building as reproducible as we can.

- Where do the results of the upload artifact action go? In the releases page or somewhere else?

- Also do you think it would be possible to also build the mysql and postgres versions in the same action? Maybe by running the cargo build and upload artifact actions again with a different name?

dani-garcia:
Okay that seems to solve my questions so far, thanks!

publicarray:
No worries, unfortunately the windows (mysql/postgresql) version does not build successfully yet: https://github.com/publicarray/bitwarden_rs/actions d

dani-garcia:
Yeah Windows is a pain with dependencies, but at least I'm happy with having linux binaries.

Don't know if you saw it already, but the diesel project has azure pipelines running on windows, and it seems to be the same format so maybe it can help:
https://github.com/diesel-rs/diesel/blob/master/azure-pipelines.yml#L132

dani-garcia:
Also not sure if also included in the images, but I sucessfully used [vcpkg](https://github.com/microsoft/vcpkg) in Windows in the past to install OpenSSL libraries, so that could be an option.

publicarray:
Cool thanks for the tips! I'll it try again later.

publicarray:
So I don't have a windows install to use so I was just guessing where things are installed. So far I was not able to build the windows version but I think I've made some improvements to the build process (It's just a single file). If you like them I can send you a new PR https://github.com/publicarray/bitwarden_rs/blob/master/.github/workflows/workspace.yml Someone else will need to implement the windows version sorry.

publicarray:
All someone needs to do is set the following on the windows build and it should work

```yml
env:
        OPENSSL_DIR: C:\Program Files\Path to OpenSSL\
        MYSQLCLIENT_LIB_DIR: C:\Program Files\Path to MySQL\
        PQ_LIB_DIR: C:\Program Files\Path to PostgreSQL\
```

dani-garcia:
Okay yeah, the one file version looks nicer indeed. I'll try later today to dust off my old windows laptop and see it I can get it to work, but no big deal if it doesn't, I assume most people only want the linux binaries. Thanks for your work!

PR #784, opened 2019-12-28
Use Debian base image for all steps of the build process

ypid:
No need to use two different base images. Debian buster is pulled later anyway so we can just use it for the vault stage as well.

My reason for this change is partly to avoid redundancy and partly to make it easier to build everything yourself. When all the build environment is based on Debian than you just have to figure out how to build a Debian Docker base image (ref: https://github.com/ypid/docker-makefile).

~~WIP: This PR is incomplete. I first wanted to check if this is acceptable two you because I put together a short script to replace this in all the Dockerfiles.~~

I assume the broken windows CI build is not my fault. Why support Windows anyway ;-)

### DIFF
--- a/docker/Dockerfile.j2
+++ b/docker/Dockerfile.j2
@@ -37,10 +37,14 @@ RUN apk add --no-cache --upgrade \
     curl \
     tar
 {% else %}
-ENV DEBIAN_FRONTEND=noninteractive \
-    LANG=C.UTF-8 \
-    TZ=UTC \
-    TERM=xterm-256color
+
+# Build time options to avoid dpkg warnings and help with reproducible builds.
+ARG DEBIAN_FRONTEND=noninteractive
+ARG LANG=C.UTF-8
+ARG TZ=UTC
+ARG TERM=xterm-256color
+
+RUN echo $TERM
 
 RUN apt update -y \
     && apt install -y \
@@ -79,6 +83,12 @@ ARG DB=mysql
 # set postgresql backend
 ARG DB=postgresql
 {% endif %}
+
+# Build time options to avoid dpkg warnings and help with reproducible builds.
+ARG DEBIAN_FRONTEND=noninteractive
+ARG LANG=C.UTF-8
+ARG TZ=UTC
+ARG TERM=xterm-256color
 
 # Don't download rust docs
 RUN rustup set profile minimal
--- a/docker/Dockerfile.j2
+++ b/docker/Dockerfile.j2
@@ -156,7 +156,7 @@ RUN apt-get update && apt-get install -y \
 
 {% endif %}
 # Creates a dummy project used to grab dependencies
-RUN USER=root cargo new --bin app
+RUN USER=root cargo new --bin /app
 WORKDIR /app
 
 # Copies over *only* your manifests and build files
--- a/docker/Dockerfile.j2
+++ b/docker/Dockerfile.j2
@@ -87,6 +87,18 @@ RUN rustup set profile minimal
 ENV USER "root"
 
 {% endif %}
+{% if "aarch64" in target_file or "armv" in target_file %}
+# Install required build libs for {{ package_arch_name }} architecture.
+RUN sed 's/^deb/deb-src/' /etc/apt/sources.list > \
+        /etc/apt/sources.list.d/deb-src.list \
+    && dpkg --add-architecture {{ package_arch_name }} \
+    && apt-get update \
+    && apt-get install -y \
+        --no-install-recommends \
+        libssl-dev{{ package_arch_prefix }} \
+        libc6-dev{{ package_arch_prefix }}
+
+{% endif -%}
 {% if "aarch64" in target_file %}
 RUN apt-get update \
     && apt-get install -y \
@@ -142,19 +154,6 @@ RUN apt-get update && apt-get install -y \
 # Creates a dummy project used to grab dependencies
 RUN USER=root cargo new --bin app
 WORKDIR /app
-
-{% if "aarch64" in target_file or "armv" in target_file %}
-# Install required build libs for {{ package_arch_name }} architecture.
-RUN sed 's/^deb/deb-src/' /etc/apt/sources.list > \
-        /etc/apt/sources.list.d/deb-src.list \
-    && dpkg --add-architecture {{ package_arch_name }} \
-    && apt-get update \
-    && apt-get install -y \
-        --no-install-recommends \
-        libssl-dev{{ package_arch_prefix }} \
-        libc6-dev{{ package_arch_prefix }}
-
-{% endif -%}
 
 # Copies over *only* your manifests and build files
 COPY ./Cargo.* ./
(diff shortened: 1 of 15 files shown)


dani-garcia:
Oh yeah this makes perfect sense, you are right. No need to download a separate image when not needed. We could also consider just moving the web vault building inside the rust build image and just save one step.

The windows failure is cause by dependency problems with the database libraries and not related to this, so shouldn't be an issue.

I'm not a huge shell expert, what do these changes do?
https://github.com/dani-garcia/bitwarden_rs/blob/f250c5481302280360a89c25a5fd02607b91c53f/docker/amd64/sqlite/Dockerfile#L11-L14

https://github.com/dani-garcia/bitwarden_rs/blob/f250c5481302280360a89c25a5fd02607b91c53f/docker/amd64/sqlite/Dockerfile#L24

ypid:
> bash -o nounset -o pipefail -o errexit

The options `errexit` (`-e`) and `pipefail` where already used before. Good work! The only one missing was `nounset` which is not really relevant here. My other change is to always use the long option names to help users understand what we do and make reviewing easier.

> DEBIAN_FRONTEND=noninteractive

Important for apt/dpkg to not try to ask questions.

The other environment variables help with reproducible builds which should be mostly irrelevant as this part of the multistaged build only unpacks a tar (what could go wrong there ;-) ).

ypid:
All CI builds seem to be passing. Note that I have only reviewed the change, not tested all of them :)

docker/aarch64/mysql/Dockerfile
```diff
@@ -1,21 +1,27 @@
-# Using multistage build: 
+# Using multistage build:
 # 	https://docs.docker.com/develop/develop-images/multistage-build/
 # 	https://whitfin.io/speeding-up-rust-docker-builds/
 ####################### VAULT BUILD IMAGE  #######################
-FROM alpine:3.11 as vault
+FROM balenalib/aarch64-debian:buster as vault
```
dani-garcia: The web vault download steps can't use the balenalib/* ones, those are just for running on ARM hardware and the images get built on x86 machines, so they'll give an error when running. In those cases we'll need to use the same image as the one used to build the rust code, or move the dowload code inside the build image.

docker/aarch64/mysql/Dockerfile
```diff
@@ -42,29 +48,50 @@ RUN apt-get update \
...
+    --no-install-recommends \
+    libmariadb-dev:arm64 \
+    && rm -rf /var/lib/apt/lists/*
+
+# Creates a dummy project used to grab dependencies
+RUN USER=root cargo new --bin app
 WORKDIR /app
 
-# Prepare openssl arm64 libs
+# Install required build libs for arm64 architecture.
```
dani-garcia: This step would need to be moved before the Install step above, line 52 in this file, otherwise the library above won't install because the architecture hasn't been added yet.

docker/amd64/mysql/Dockerfile.alpine
```diff
@@ -32,24 +32,38 @@ RUN rustup set profile minimal
 
 ENV USER "root"
 
-# Install needed libraries
+# Install MySQL package
 RUN apt-get update && apt-get install -y \
     --no-install-recommends \
-    libmysqlclient-dev \
+    libmariadb-dev \
```
dani-garcia: This library doesn't exist in the muslrust image, we should keep the original one.

docker/amd64/mysql/Dockerfile.alpine
```diff
@@ -32,24 +32,38 @@ RUN rustup set profile minimal
...
     && rm -rf /var/lib/apt/lists/*
 
+# Creates a dummy project used to grab dependencies
+RUN USER=root cargo new --bin app
```
dani-garcia: The muslrust images seem to have a workdir of /volume as oposed to / in the others so doing cargo new like this would create the folder in /volume/app. Let's use `RUN USER=root cargo new --bin /app` in all images to be more exact.

docker/amd64/mysql/Dockerfile.alpine
```diff
@@ -32,24 +32,38 @@ RUN rustup set profile minimal
...
 # To avoid copying unneeded files, use .dockerignore
 COPY . .
 
-RUN rustup target add x86_64-unknown-linux-musl
-
 # Make sure that we actually build the project
 RUN touch src/main.rs
 
-# Build
+RUN rustup target add x86_64-unknown-linux-musl
```
dani-garcia: I don't think this is doing anything right now because the target is probably installed already but it should be moved before the cargo build in line 53.

docker/armv6/mysql/Dockerfile
```diff
@@ -42,31 +48,50 @@ RUN apt-get update \
...
+RUN cargo build --features ${DB} --release
+RUN find . -not -path "./target*" -delete
 
 # Copies the complete project
 # To avoid copying unneeded files, use .dockerignore
 COPY . .
 
-# Build
-RUN rustup target add arm-unknown-linux-gnueabi
-RUN cargo build --features ${DB} --release --target=arm-unknown-linux-gnueabi
```
dani-garcia: This build step dissapeared

docker/armv7/mysql/Dockerfile
```diff
@@ -31,43 +37,46 @@ ARG DB=mysql
 # Don't download rust docs
 RUN rustup set profile minimal
 
-RUN apt-get update \
-    && apt-get install -y \
-        --no-install-recommends \
-        gcc-arm-linux-gnueabihf \
-    && mkdir -p ~/.cargo \
-    && echo '[target.armv7-unknown-linux-gnueabihf]' >> ~/.cargo/config \
-    && echo 'linker = "arm-linux-gnueabihf-gcc"' >> ~/.cargo/config
```
dani-garcia: This also dissapeared

dani-garcia:
Also note that the CI is only checking that the code compiles, it doesn't actually test the docker images.

ypid:
I addressed all of your comments, thanks for the thoughtful review and sorry that I missed quite a few things.

dani-garcia:
I gave it a look yesteday before new years and it all looked good to me, I'll see if I can find sometime today to build all the images and make sure it all works. Thanks for your work!

docker/Dockerfile.j2
```diff
@@ -0,0 +1,310 @@
...
+RUN touch src/main.rs
+
+# Builds again, this time it'll just be
+# your actual source files being built
+{% if "amd64" in target_file %}
+RUN cargo build --features ${DB} --release
+{% elif "aarch64" in target_file %}
+RUN cargo build --features ${DB} --release --target=aarch64-unknown-linux-gnu
+{% elif "armv6" in target_file %}
+RUN cargo build --features ${DB} --release --target=arm-unknown-linux-gnueabi
```
dani-garcia: The armv7 build is still missing, but other than that the rest seems okay and builds correctly for me.

ypid:
You are very welcome. Fixed :)

dani-garcia:
Okay I think that is all!

PR #800, opened 2020-01-04
Update README.md

andrzejwp:
Provide link to cdwv/bitwarden-k8s - helm chart for installing bitwarden_rs inside a Kubernetes cluster


dani-garcia:
We already have some kubernetes links in the wiki: https://github.com/dani-garcia/bitwarden_rs/wiki/Kubernetes-deployment.

What's the difference between this new one and those?

mprasil:
I think the Readme change is a little bit confusing as it's talking about K8S installation, but the actual code provided is Helm package - not everyone is using Helm with k8s.

Also I have to agree with @dani-garcia that the wiki page is better place for this. Feel free to add your chart under Helm section, we can have multiple charts there no prob.

Ornias1993:
Most users don't run this within kubernetes, the Readme is supposed to be somewhat if a short-and-dirty quickstart guide.

Kubernetes is not something that belongs in a quickstart guide on non-kubernetes-centric project imho.

BlackDex:
This can be closed i think?

PR #812, opened 2020-01-14
Fixes #635 - Unique constraint violation when using U2F tokens on PostgreSQL

### LINKED ISSUE #635: U2F not working when using Postgresql
An error when trying to login with U2F and using postgresql db:
[CAUSE] DatabaseError(
bitwarden bitwarden_rs[108]:     UniqueViolation,
bitwarden bitwarden_rs[108]:     "duplicate key value violates unique constraint \"twofactor_user_uuid_atyp
bitwarden bitwarden_rs[108]: )

swedishborgie:
Because of differences in how .on_conflict() works compared to .replace_into() the PostgreSQL backend wasn't correctly ensuring the unique constraint on user_uuid and atype wasn't getting violated.

This change simply issues a DELETE on the unique constraint prior to the insert to ensure uniqueness. PostgreSQL does not support multiple constraints in ON CONFLICT clauses.

Please let me know if you'd like to see any changes.

### DIFF
--- a/src/db/models/two_factor.rs
+++ b/src/db/models/two_factor.rs
@@ -73,6 +73,16 @@ impl TwoFactor {
 impl TwoFactor {
     #[cfg(feature = "postgresql")]
     pub fn save(&self, conn: &DbConn) -> EmptyResult {
+        // We need to make sure we're not going to violate the unique constraint on user_uuid and atype.
+        // This happens automatically on other DBMS backends due to replace_into(). PostgreSQL does
+        // not support multiple constraints on ON CONFLICT clauses.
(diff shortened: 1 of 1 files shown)


swedishborgie:
Pull updated to switch out for the ? operator, sorry about that, I'm not primarily a Rust developer. Learned something new!

The delete doesn't appear to return an error on 0 rows affected; I did test the code using a TOTP token. I couldn't find anything in [the documentation](https://docs.diesel.rs/diesel/query_builder/struct.DeleteStatement.html) explicitly stating that it doesn't throw an error but the examples seem to suggest that checking the affected row counts are the responsibility of the caller:

    let deleted_rows = diesel::delete(users)
        .filter(name.eq("Sean"))
        .execute(&connection);
    assert_eq!(Ok(1), deleted_rows);

dani-garcia:
Okay makes sense, thanks!

Issue #792, opened 2020-01-01
Unable to change email address


yegle:
# Subject of the issue

It looks like I'm unable to change the email address from the settings page.

The request payload looks like this:
```
{"newEmail":"foo@example.com","masterPasswordHash":"some-hash"}
```

The response payload looks like this:
```
{"ErrorModel":{"Message":"Email cannot be changed to this address","Object":"error"},"Message":"","Object":"error","ValidationErrors":{"":["Email cannot be changed to this address"]},"error":"","error_description":""}
```

Server side log looks like this:

```
[2020-01-01 03:32:43][error][ERROR] Email cannot be changed to this address
[2020-01-01 03:32:43][response][INFO] POST /api/accounts/email-token (post_email_token) => 400 Bad Request
```

### Your environment
* Bitwarden_rs version: 1.13.0-95dd1cd7
* Install method: docker
* Clients used: web UI
* Reverse proxy and version: probably not applicable
* Version of mysql/postgresql:  probably not applicable
* Other relevant information: 

### Steps to reproduce
Change email address from the web under settings

### Expected behaviour
The email address changed.

### Actual behaviour
An error appear in the web:

An error has occurred.
Email cannot be changed to this address

WinkelB:
Having the Same Problem.

aneisch:
Seeing this as well. It looks like this is due to the following line:
**[https://github.com/dani-garcia/bitwarden_rs/blob/8d1b72b9512b90775e671b7ba08cd552a0aabd13/src/api/core/accounts.rs#L374](https://github.com/dani-garcia/bitwarden_rs/blob/8d1b72b9512b90775e671b7ba08cd552a0aabd13/src/api/core/accounts.rs#L374)**

If I allow signups I'm then able to change an email address.

tomuta:
I think this might be a problem in `can_signup_user()`. I believe the function should return `true` if the whitelist is empty. Right now it only returns `true` if a match for the domain is found.

dani-garcia:
This should be fixed now in the latest master, thanks to @tomuta

PR #824, opened 2020-01-25
Fix change email when no whitelist is configured

tomuta:
Fixed issue #792

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -510,7 +510,12 @@ impl Config {
             warn!("Failed to parse email address '{}'", email);
             return false;
         }
-        self.signups_domains_whitelist().split(',').any(|d| d == e[0])
+        
+        // Allow signups if the whitelist is empty/not configured
+        // (it doesn't contain any domains), or if it matches at least
+        // one domain.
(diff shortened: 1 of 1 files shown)


dani-garcia:
The change seems correct, but I think it would be easier to read as something like:
```rust
 let whitelist_str = self.signups_domains_whitelist();
whitelist_str.is_empty() || whitelist_str.split(',').any(|d| d == e[0])
```

That way it's purpose is more clear, as it basically spells what your code comment says.

tomuta:
But that wouldn't catch a semi-empty list, e.g. ",,"
Is there a way to specify that split should not ever output empty tokens?  Then this would work.

dani-garcia:
You could do something like
```rust
whitelist_str.split(',').filter(|s| !s.is_empty()).any(|d| d == e[0])
```

That said, I'd consider that case to be an invalid configuration, so I don't think it's a big deal if it isn't handled perfectly

tomuta:
Does this look better?

dani-garcia:
Yep, thanks!

PR #864, opened 2020-02-16
Do not disable invitations via admin API

mprasil:
This was brought up today:

https://github.com/dani-garcia/bitwarden_rs/issues/752#issuecomment-586715073

I don't think it makes much sense in checking whether admin has the
right to send invitation as admin can change the setting anyway.

Removing the condition allows users to forbid regular users from
inviting new users to server while still preserving the option to do so
via the admin API.

### DIFF
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -153,10 +153,6 @@ fn invite_user(data: Json<InviteData>, _token: AdminToken, conn: DbConn) -> Empt
         err!("User already exists")
     }
 
-    if !CONFIG.invitations_allowed() {
-        err!("Invitations are not allowed")
-    }
-
     let mut user = User::new(email);
     user.save(&conn)?;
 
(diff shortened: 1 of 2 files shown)


dani-garcia:
Seems reasonable to me, but note that User::take() only works with invitations enabled, so we'd have to change that there too.

mprasil:
Ah true! Forgot there's another check there.

mprasil:
I've checked the spots when `Invitation::new()` and `Invitation::take()` are used and it seems like all spots are already correctly gated. So I've just removed the check in `Invitation::take()`.

There's only one issue with this change - any invitations that were already in place (for example from before disabling invitations) will now be active. I suppose this won't be a major concern?

dani-garcia:
I think keeping the invitations is reasonable, if it becomes a concern we can add a button in the admin page to delete any pending invitations.

Issue #865, opened 2020-02-16
Feature request: Disable attachments


Dubz:
Title says it all.

I want to be able to disable the attachments feature on my system. Attachments take up storage space. Storage costs me money. I do not want to be paying excessive costs to host user attachments.

Logins aren't as much space so I'm alright with that, but when you can multiply someone's storage usage with a few attachments, that's not really something I'd like to have to pay for. They may add a few, or many, files, which would grow the used space exponentially.

mprasil:
That sounds like good idea. I've created a PR to enable that.

dani-garcia:
I've implemented the size limit functionality, with two limits, one for users and another for organizations, so depending where the cipher is located it can be one or the other, and setting it to zero disables uploading at all. The options are `USER_ATTACHMENT_LIMIT` and `ORG_ATTACHMENT_LIMIT` and they use KB.

https://github.com/dani-garcia/bitwarden_rs/commit/325039c31695ac981da3b88dbbe6c6f40c6a180d

Issue #807, opened 2020-01-07
Question: where to set the 1GB attachment limit?


pdarcos:
Hi guys,

Maybe a dumb question , but where in the code is this logic handled?
Is it possible to set limits per account/organization or is it a global setting for everyone?

Cheers

dani-garcia:
At the moment there isn't any attachment size limit.

pdarcos:
Ah ok. Thanks for the answer @dani-garcia 

Are there plans to implement that? I don't want some of my users abusing it and uploading GBs of data.

Where in the code should I be looking to implement this? Either per user or globally

Thanks

dani-garcia:
At the moment there aren't any plans in progress but it's a welcome feature and it should be fairly easy to at least have a global limit.

This is where we save the file and we specifically set the size limit to None, then we save the file size in the attachments table.

https://github.com/dani-garcia/bitwarden_rs/blob/baf7d1be4e9a7e70de99c6005b09571bd17ff6f9/src/api/core/ciphers.rs#L677

To put a global user limit we'd have to add all the sizes of that users attachments and then set that `size_limit` to `LIMIT - (sum of all attachments sizes)`. 
If we set 1Gb limit per user and I have 3 100Mb files, then the limit for the next one would be 700Mb for example. Then when the limit is exceeded I assume we'd get a  SaveResult::Partial or a SaveResult::Error and we'd be done.

To have different limits for users we'd need to add a column to the users table and a way to change them in the admin page, but I think most people would be happy with a global user limit for now.

pdarcos:
Thanks @dani-garcia !

Much appreciated

dani-garcia:
Okay so I've implemented the size limit functionality, with two limits, one for users and another for organizations, so depending where the cipher is located it can be one or the other, and setting it to zero disables uploading at all. The options are `USER_ATTACHMENT_LIMIT` and `ORG_ATTACHMENT_LIMIT` and they use KB.

The commit: https://github.com/dani-garcia/bitwarden_rs/commit/325039c31695ac981da3b88dbbe6c6f40c6a180d

PR #866, opened 2020-02-16
Allow disabling attachment uploading

mprasil:
This adds an configuration option to disable uploading attachments in
cases where running out of space on server is a concern for example.

This was raised in #865 and seems like worthwhile addition.

### DIFF
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -631,8 +631,7 @@ fn share_cipher_by_uuid(
     }
 }
 
-#[post("/ciphers/<uuid>/attachment", format = "multipart/form-data", data = "<data>")]
-fn post_attachment(
+fn _post_attachment(
     uuid: String,
     data: Data,
     content_type: &ContentType,
@@ -704,6 +703,23 @@ fn post_attachment(
     Ok(Json(cipher.to_json(&headers.host, &headers.user.uuid, &conn)))
 }
 
+#[post("/ciphers/<uuid>/attachment", format = "multipart/form-data", data = "<data>")]
+fn post_attachment(
+    uuid: String,
+    data: Data,
+    content_type: &ContentType,
+    headers: Headers,
+    conn: DbConn,
+    nt: Notify,
+) -> JsonResult {
+    if CONFIG.disable_attachments() {
+        err!("Uploading attachments is not allowed on server")
+    }
+
+    _post_attachment(uuid, data, content_type, headers, conn, nt)
+}
+
+
 #[post("/ciphers/<uuid>/attachment-admin", format = "multipart/form-data", data = "<data>")]
 fn post_attachment_admin(
     uuid: String,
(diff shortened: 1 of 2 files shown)


dani-garcia:
Two things, why create another method instead of keeping it as it is?

Also I think i'd rather combine this with a user size limit implementation (as mentioned in #807), that way you can just set it to zero and we avoid adding multiple similar config options.

mprasil:
> why create another method instead of keeping it as it is?

So that you can still allow posting "new" attachment when it's being shared, which is technically creating new attachment while deleting the old - used in `/ciphers/<uuid>/attachment/<attachment_id>/share` endpoint.

As for having this defined as attachment limit, I feel like that is going to be quite expensive operation if you want to check the per user quota. The global quota might be faster to check, but for that one I'm kinda thinking that actually pointing `attachments_folder` to a directory with filesystem quota (or just partition dedicated to attachments) will give you this functionality on system level where it probably makes most sense. (and you can do it right now without code change to `bitwarden_rs`)

But I don't feel strongly either way. If you'd rather implement the limits, feel free to close this.

dani-garcia:
Okay the share thing makes sense!

About the file size, we have the size saved in the database, so it should be a fairly straight forward query:
```sql
select sum(file_size) from ciphers join attachments
where ciphers.uuid == attachments.cipher_uuid
and user_uuid == <user_id>
```

We only need to replace the `user_uuid == <user_id>` when putting the attachment in an organization cipher but that should be easy to handle.

You're right that a global quota might be handled better from outside bitwarden_rs though.

mprasil:
Ah I wasn't aware we store the size. Should have checked first. The per-user quota still does not guarantee that someone won't create a ton of users with a ton of attachments and I'm not sure how organization-owned files will be accounted.

I guess what I'm trying to say is that in my opinion for most people it's either about storing as much attachments as they can where filesystem is already providing the limit or they don't want attachments at all where simple true/false setting might make sense.

But if you really rather figure out a way to implement the quota instead, please close this, I don't want to cause another dangling PR. :smile: Just figured I'd implement that with couple minutes to spare in case it's useful. (I do understand that extra option needs to be maintained long term, so absolutely do close this if you're leaning towards quota functionality)

dani-garcia:
Okay so I've implemented the size limit functionality, with two limits, one for users and another for organizations, so depending where the cipher is located it can be one or the other, and setting it to zero disables uploading at all. The options are `USER_ATTACHMENT_LIMIT` and `ORG_ATTACHMENT_LIMIT` and they use KB.

About the creating a ton of users concern, I assume anyone setting these limits already has limited signups some other way so I think it should be fine.


The commit: https://github.com/dani-garcia/bitwarden_rs/commit/325039c31695ac981da3b88dbbe6c6f40c6a180d

Thanks anyway!

mprasil:
That's great news! As for the `ORG_ATTACHMENT_LIMIT`, there's still issue that user can spawn a ton of organizations just for himself and go over the limit that way, not sure what would be the best way to resolve that.

Anyway, setting the limits to `0` will serve the same purpose as this PR tried to achieve, so in that context, it's great.

Issue #767, opened 2019-12-13
Bitwarden_rs  Support for Subpaths，how to config   (caddy) ?


bigbighill:
docker-compose.yml

version: '3'
services:
  bitwarden:
    image:  bitwardenrs/server:latest
    container_name: bitwarden
    restart: always
    volumes:
      - ./data:/data
      - /etc/localtime:/etc/localtime
      - /var/log:/var/log
    ports:
      - "127.0.0.1:9990:80"
      - "127.0.0.1:9991:3012"
    environment:
      SIGNUPS_ALLOWED: "true"
      DOMAIN: "https://www.xxxx.com/subpath"
      DATABASE_URL: "/data/bitwarden.db"
      ROCKET_WORKERS: "10"
      WEB_VAULT_ENABLED: "true"
      ADMIN_TOKEN: "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
   
 
  caddy:
    image: abiosoft/caddy
    container_name: caddy
    restart: always
    volumes:
      - ./Caddyfile:/etc/Caddyfile:ro
      - ./Caddycerts:/root/.caddy
      - ./www:/var/www
      - /etc/localtime:/etc/localtime
    ports:
      - "127.0.0.1:2015:2015"
      - 443:443
      - 80:80
    environment:
      ACME_AGREE: "true"              # agree to Let's Encrypt Subscriber Agreement
      ARGS:  "version='1.0.4'"

Caddyfile
https://www.xxxx.com {
	tls asdf@gmail.com
	root /var/www
	index index.html
	gzip
	header / {
        # Enable HTTP Strict Transport Security (HSTS)
        Strict-Transport-Security "max-age=31536000;"
        # Enable cross-site filter (XSS) and tell browser to block detected attacks
        X-XSS-Protection "1; mode=block"
        # Disallow the site to be rendered within a frame (clickjacking protection)
        X-Frame-Options "DENY"
        # Prevent search engines from indexing (optional)
        X-Robots-Tag "none"
    }
        proxy /subpath/notifications/hub/negotiate bitwarden:80 {
		transparent
	}
	proxy /subpath notifications/hub bitwarden:3012 {
		websocket
	}	
	proxy /subpath bitwarden:80 {
		transparent
	}
}

https://www.xxxx.com/subpath      404 error

https://github.com/bitwarden/server/issues/277  now support subpaths ,how to config?

thanks

[...]

Issue #528, opened 2019-07-11
Support for Subpaths


waphire:
Is it possible to run Bitwarden in a reverse proxy setup on a subpath instead of a subdomain?

I've searched around and have only found this: https://github.com/bitwarden/server/issues/277
Is it not a supported scenario for bitwarden_rs too?

dani-garcia:
This isn't a supported feature, but it can be made to work with some changes both to the web vault and the server code. That does mean recompiling the server and the web vault, so I would recommend using a subdomain.

There are some old patches available here: https://github.com/dani-garcia/bitwarden_rs/issues/241#issuecomment-436376497.

Issue #488, opened 2019-05-23
Admin does not support running on sub-path


LaszloHont:
Could bitwarden_rs use relative URLs?

I don't want the world to know about https://bitwarden.example.com/ so I am using https://secure.example.com/bitwarden/. This is not findable in the SSL observatory logs, and privacy is why I am self-hosting.

I have tried starting bitwarden_rs with DOMAIN=https://secure.example.com/bitwarden/ but when I visit https://secure.example.com/bitwarden/admin/ and submit the token I get a 404 from /admin/

Thanks.

dani-garcia:
Running on a subdomain won't work unless you apply the patches from https://github.com/dani-garcia/bitwarden_rs/issues/241#issuecomment-436376497.

If the reason you are doing it is exclusively privacy, i would recommend using a less obvious subdomain, like just secure.example.com, or issuing a wildcard certificate which might not make the exact subdomain findable in those logs (I'm not sure if that's how it works, so this might not work).

As far as I know there isn't much we can do here, so I'll close this.

gerroon:
@dani-garcia is there any reason why that patch did not make to the trunk? I personally would like to run it under a subfolder as well but I am not a coder and it seems to me that keeping up the patch with the docker updates quite a bit of involvement if you will.

dani-garcia:
Mostly because it requires to hardcode the subfolder directly into the web vault files, which means every person would need to make a specific build for their own subfolder, so we can't distribute a generic version.

gerroon:
I see, that maks sense. I hope this will be somehow possible in the future

PR #868, opened 2020-02-19
Add backend support for alternate base dir (subdir/subpath) hosting

### LINKED ISSUE #241: Run locally at suburl
I'm running bitwarden_rs locally (i.e., not in a docker container) and using Apache as reverse proxy via a suburl from my main web-facing domain name.  The setup was comparatively easy -- just adding the url prefix to the "mount" commands in main.rs got me 98% of the way there.  I then had to customize web-vault source in a couple of places -- the baseUrl in jslib/src/services/api.service.ts, and the urls for api and identity in src/app/services/services.module.ts.

At that point I pretty much had a usable system.  But immediately post-login there was a post request sent to "hub/negotiate" instead of to "<urlprefix>/hub/negotiate"   As far as I can tell this call is coming from line 20 of api/notifications.rs.  But I can't figure out why it's not respecting the previously defined url prefix from the mount command on line 57 of main.rs [in my setup now = .mount("/<urlprefix>/notifications", api::notifications_routes())]

I'm working around it for now by just using apache to redirect /hub/negotiate to <urlprefix>/hub/negotiate.  But I'd love to figure out how to solve the problem internally to bitwarden_rs.

Thanks in advance for any advice.  And thanks a million for a great project.

jjlin:
To use this, include a path in the `DOMAIN` URL, e.g.:

* `DOMAIN=https://example.com/custom-path`
* `DOMAIN=https://example.com/multiple/levels/are/ok`

Resolves #241, #488, #528, #767, and maybe others when combined with
dani-garcia/bw_web_builds#11 (web vault support).

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -16,11 +16,11 @@ const JWT_ALGORITHM: Algorithm = Algorithm::RS256;
 lazy_static! {
     pub static ref DEFAULT_VALIDITY: Duration = Duration::hours(2);
     static ref JWT_HEADER: Header = Header::new(JWT_ALGORITHM);
-    pub static ref JWT_LOGIN_ISSUER: String = format!("{}|login", CONFIG.domain());
-    pub static ref JWT_INVITE_ISSUER: String = format!("{}|invite", CONFIG.domain());
-    pub static ref JWT_DELETE_ISSUER: String = format!("{}|delete", CONFIG.domain());
-    pub static ref JWT_VERIFYEMAIL_ISSUER: String = format!("{}|verifyemail", CONFIG.domain());
-    pub static ref JWT_ADMIN_ISSUER: String = format!("{}|admin", CONFIG.domain());
+    pub static ref JWT_LOGIN_ISSUER: String = format!("{}|login", CONFIG.domain_origin());
+    pub static ref JWT_INVITE_ISSUER: String = format!("{}|invite", CONFIG.domain_origin());
+    pub static ref JWT_DELETE_ISSUER: String = format!("{}|delete", CONFIG.domain_origin());
+    pub static ref JWT_VERIFYEMAIL_ISSUER: String = format!("{}|verifyemail", CONFIG.domain_origin());
+    pub static ref JWT_ADMIN_ISSUER: String = format!("{}|admin", CONFIG.domain_origin());
     static ref PRIVATE_RSA_KEY: Vec<u8> = match read_file(&CONFIG.private_rsa_key()) {
         Ok(key) => key,
         Err(e) => panic!("Error loading private RSA Key.\n Error: {}", e),
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,6 +1,8 @@
 use std::process::exit;
 use std::sync::RwLock;
 
+use reqwest::Url;
+
 use crate::error::Error;
 use crate::util::{get_env, get_env_bool};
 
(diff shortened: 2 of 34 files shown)


jjlin:
BTW, this is my first time writing any code in Rust, so feel free to let me know if I'm doing anything weird. I'll also comment on a few specific changes a bit later.

src/static/templates/admin/base.hbs
```diff
@@ -6,10 +6,10 @@
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
     <title>Bitwarden_rs Admin Panel</title>
 
-    <link rel="stylesheet" href="/bwrs_static/bootstrap.css" />
```
jjlin: These paths all need to be relative to work with a different base dir.

src/static/templates/email/invite_accepted.hbs
```diff
@@ -3,6 +3,6 @@ Invitation accepted
 <html>
 <p>
     Your invitation for <b>{{email}}</b> to join <b>{{org_name}}</b> was accepted.
-    Please <a href="{{url}}">log in</a> to the bitwarden_rs server and confirm them from the organization management page.
+    Please <a href="{{url}}/">log in</a> to the bitwarden_rs server and confirm them from the organization management page.
```
jjlin: I added a `/` after all bare `{{url}}` links because, for reasons I don't quite understand, if the base URL is `https://example.com/my/path`, then visiting the web vault at `https://example.com/my/path` results in rendering issues, while `https://example.com/my/path/` (with trailing `/`) works fine.

src/config.rs
```diff
@@ -240,6 +242,10 @@ make_config! {
         domain:                 String, true,   def,    "http://localhost".to_string();
         /// Domain Set |> Indicates if the domain is set by the admin. Otherwise the default will be used.
         domain_set:             bool,   false,  def,    false;
```
jjlin: It seems like `domain_set` could be made `auto` as well, but maybe there's some subtlety I'm missing, so I left it as is.

src/auth.rs
```diff
@@ -16,11 +16,11 @@ const JWT_ALGORITHM: Algorithm = Algorithm::RS256;
 lazy_static! {
     pub static ref DEFAULT_VALIDITY: Duration = Duration::hours(2);
     static ref JWT_HEADER: Header = Header::new(JWT_ALGORITHM);
-    pub static ref JWT_LOGIN_ISSUER: String = format!("{}|login", CONFIG.domain());
-    pub static ref JWT_INVITE_ISSUER: String = format!("{}|invite", CONFIG.domain());
-    pub static ref JWT_DELETE_ISSUER: String = format!("{}|delete", CONFIG.domain());
-    pub static ref JWT_VERIFYEMAIL_ISSUER: String = format!("{}|verifyemail", CONFIG.domain());
-    pub static ref JWT_ADMIN_ISSUER: String = format!("{}|admin", CONFIG.domain());
+    pub static ref JWT_LOGIN_ISSUER: String = format!("{}|login", CONFIG.domain_origin());
```
jjlin: I changed this to `domain_origin()` instead of `domain()` to avoid JWT encode/decode issues in case the installation is moved to a different base dir. Note that in the usual case where the backend is hosted at the root of the domain, `domain_origin()` and `domain()` should be identical.

jjlin:
I've also added docs on the wiki: [Using an alternate base dir](https://github.com/dani-garcia/bitwarden_rs/wiki/Using-an-alternate-base-dir)

src/config.rs
dani-garcia: Yeah, domain_set is a bit special cased here: 
https://github.com/dani-garcia/bitwarden_rs/blob/0a3008e7538c1a29b8b5a6439004e89465fce391/src/config.rs#L107-L111

src/static/templates/admin/base.hbs
dani-garcia: I've changed it to instead use an absolute path with the domain_path variable to avoid the / at the end issue that you mention for the emails

src/static/templates/admin/page.hbs
```diff
@@ -1,6 +1,6 @@
 <main class="container">
     <div id="users-block" class="my-3 p-3 bg-white rounded shadow">
-        <h6 class="border-bottom pb-2 mb-0">Registered Users</h6>
+        <h6 class="border-bottom pb-2 mb-0">Registered Users {{urlpath}}</h6>
```
jjlin: Was it intentional for this `{{urlpath}}` to be here?

dani-garcia: Oops, no that was me checking that the urlpath was set to the correct value, forgot to remove it!

PR #876, opened 2020-02-25
Make panics logable (as warn)

BlackDex:
panic!()'s only appear on stderr, this makes tracking down some strange
issues harder with the usage of docker since stderr does not get logged
into the bitwarden.log file. This change logs the message to stdout and
the logfile when activated.


BlackDex:
Hmm, backtrace doesn't seem to be available when using it like this.

dani-garcia:
Maybe we can do something like the log_panic crate:
https://docs.rs/crate/log-panics/2.0.0/source/src/lib.rs


Which is using the backtrace crate:
https://crates.io/crates/backtrace

There's also support for backtrace in the std but nightly only for now:
https://doc.rust-lang.org/std/backtrace/struct.Backtrace.html

BlackDex:
Ok, i have updated the code now and used the log-panic example and integrated it into bitwarden_rs it self.

This does mean that with all panics the backtrace gets logged, but in my opinion, this could be a good thing especially since this project is still kinda beta, and all panics could be an issue and need fixing. With this enabled, we can always ask users to send them.

dani-garcia:
I think logging a panic trace should be fine, we should be striving for 0 panics in code anyway. Thanks!

PR #879, opened 2020-02-26
Added SMTP test button in the admin gui

BlackDex:
- Added a test button for checking the e-mail settings.
- Fixed a bug with the _post JavaScript function:
  A function was overwriten with a variable and errors were not handled
correctly like a 500 for example.

### DIFF
--- a/src/static/templates/admin/page.hbs
+++ b/src/static/templates/admin/page.hbs
@@ -71,6 +71,17 @@
                 them to avoid confusion. This does not apply to the read-only section, which can only be set through the
                 environment.
             </div>
+
+            <div id="smtp-test-form-block" class="align-items-center mb-3 text-white-50 bg-secondary">
+                <h6 class="mb-0 text-white">SMTP Test</h6>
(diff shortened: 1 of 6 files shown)


src/static/templates/admin/page.hbs
```diff
@@ -260,6 +282,15 @@
             "Error inviting user", data);
         return false;
     }
+    function smtpTest() {
+        inv = document.getElementById("smtp-test-email");
+        data = JSON.stringify({ "email": inv.value });
+        inv.value = "";
+        _post("/admin/test/smtp/",
```
jjlin: You should rebase on the latest `master`; there's a `{{urlpath}}` needed here now.

BlackDex: Meh, i thought i had done a git pull!

BlackDex: @jjlin fixed that :)

dani-garcia:
Nice thanks!

PR #888, opened 2020-03-02
Added command line flags for help and version

zethra:
I added help and version cli flags

Signed-off-by: zethra <benaagoldberg@gmail.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -16,6 +16,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ansi_term"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ee49baf6cb617b853aa8d93bf420db2383fab46d314482ca2803b40d5fde979b"
+dependencies = [
+ "winapi 0.3.8",
+]
+
+[[package]]
 name = "arrayvec"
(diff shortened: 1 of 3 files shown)


dani-garcia:
Looks good to me, I assume this auto generates the --help flag by itself?
I'd change the short flag of --version to a lowercase v, which I think is the most common option.

jjlin:
`-V` is pretty common too, especially when `-v` is taken for verbose.

zethra:
@dani-garcia Yes, help is auto generated.  As far as the version flag, I was thinking along the same lines as @jjlin but I can change it if you want.

dani-garcia:
Well I don't think we are going to be supporting a verbose flag when we have explicit log level options, so I think it's better to keep them all lowercase for the sake of consistency

mqus:
Regarding the version, wouldn't it be possible to use `env!("CARGO_PKG_VERSION")` as a backup, provided that the version in cargo.toml is kept up-to-date? (It's currently still at 1.0.0 though)

dani-garcia:
It's an option, yeah, but it's another thing to keep track of before a release and if I ever forget to do it the release will end up with an incorrect version, which is worse than no version at all, I'd say.

dani-garcia:
thanks, @zethra!

PR #900, opened 2020-03-12
Fix escaping of SMTP_AUTH_MECHANISM in .env.template

sleweke:
Correctly escapes the quotation marks of the `SMTP_AUTH_MECHANISM` variable in `.env.template`. 
The file is sourced by the shell and swallows the simple quotation marks. 
However, we need a proper JSON value, that is, it has to have quotation marks.
Otherwise, we get the error
```
thread '<unnamed>' panicked at 'Failure to parse mechanism. Is it proper Json? Eg. `"Plain"` not `Plain`', src/mail.rs:52:22
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This is fixed by adding escaped quotation marks.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -187,7 +187,7 @@
 # SMTP_SSL=true
 # SMTP_USERNAME=username
 # SMTP_PASSWORD=password
-# SMTP_AUTH_MECHANISM="Plain"
+# SMTP_AUTH_MECHANISM="\"Plain\""
 # SMTP_TIMEOUT=15
 
 # vim: syntax=ini


sleweke:
I agree. That would be a much better solution.
Unfortunately, I'm not a rust developer and this exceeds my skill.
So, someone else will have to step up in order to do this properly.

dani-garcia:
Okay I fixed this so that the config admits the value with and without quotes, and also allows to ignore the first capital leter and use all lowercase.
https://github.com/dani-garcia/bitwarden_rs/commit/afd9f4e278a6a8fa178e30153cce725518893871

Thanks @sleweke for bringing up this issue!

PR #901, opened 2020-03-12
Use opportunistic TLS in SMTP connections

sleweke:
If SSL is disabled, the SMTP `ClientSecurity` of the `lettre` crate defaults to `None`, that is, an insecure connection. 
This is changed to `Opportunistic`, which uses TLS if available. 
If TLS is not available, an insecure connection is used (i.e., this change is backward compatible).


dani-garcia:
Thanks!

PR #910, opened 2020-03-16
Fix alpine build with openssl crate as default

ymage:
Make openssl crate as default (non feature-flipped)
Make static build


dani-garcia:
We already have openssl as default instead of as a feature, but I guess I forgot to remove the cfg!, so you can remove that line from the main.rs entirely:
https://github.com/dani-garcia/bitwarden_rs/blob/master/Cargo.toml#L111

The static builds are a nice idea!

PR #927, opened 2020-03-25
Update healthcheck script to handle alternate base dir

### LINKED ISSUE #925: Healthchek failing if 'domain' is set in config.json
### Subject of the issue
healthcheck.sh script does not aware of 'domain' setting and HEALTHCHECK will fail if subfolder is used.

### Your environment

* Bitwarden_rs version: 1.14.1

* Install method: Docker

### Proposed Fix
Temporary I use this healthcheck.sh. It works for my setting but I as I am not familiar with the server internals, am not sure it will work for all possible variants for defining domain.

My domain definition from config.json. I do not use DOMAIN variable

  "domain" : "https://my.server/bw_subfolder/",

```
#!/usr/bin/env sh

DM="$(cat /data/config.json | grep domain | sed -n 's|.*http.*/\(.*\)/.*|\1|p')"
if [ ! -z "$DM" ]
then
   DM="/$DM"
fi

if [ -z "$ROCKET_TLS"]
then
  curl --fail http://localhost:${ROCKET_PORT:-"80"}$DM/alive || exit 1
else
  curl --insecure --fail https://localhost:${ROCKET_PORT:-"80"}$DM/alive || exit 1
fi
```

jjlin:
Fixes #925

### DIFF
--- a/docker/healthcheck.sh
+++ b/docker/healthcheck.sh
@@ -1,8 +1,52 @@
-#!/usr/bin/env sh
+#!/bin/sh
 
-if [ -z "$ROCKET_TLS"]
-then
-  curl --fail http://localhost:${ROCKET_PORT:-"80"}/alive || exit 1
-else
-  curl --insecure --fail https://localhost:${ROCKET_PORT:-"80"}/alive || exit 1
+# Use the value of the corresponding env var (if present),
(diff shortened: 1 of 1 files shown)


dani-garcia:
Thanks!

PR #939, opened 2020-03-31
Fix attachment size limit calculation

### LINKED ISSUE #937: Attachements always exceed Size Limit
### Subject of the issue

Attachements wont be uploaded with Error "Attachment size limit exceeded with this file: SizeLimit" even with SizeLimit beeing set above the file SIze.

### Your environment

* Bitwarden_rs version: 1.14.1-91dd1947

* Install method: Docker-Image
* Clients used: Web Vault, Windows Client
* Reverse proxy and version: 
* Version of mysql/postgresql: mariadb 10.4
* Other relevant information: 

### Steps to reproduce

Set per User attachement Limit to 1000KB. Tried it in the admin page as well as in the docker-compose. Tried to upload files way below this limit, which throws the error above.

### Expected behaviour

The attachement should be abloaded.

### Actual behaviour

Uploading the attachement gets rejected by the error "Attachment size limit exceeded with this file: SizeLimit".

### Relevant logs

[2020-03-30 12:52:36][multipart::server][INFO] Multipart::with_boundary(_, "----WebKitFormBoundaryzCinABGOv5R6D0d0")
[2020-03-30 12:52:36][error][ERROR] Attachment size limit exceeded with this file: SizeLimit
[2020-03-30 12:52:36][response][INFO] POST /api/ciphers/<uuid>/attachment multipart/form-data (post_attachment) => 400 Bad Request

jjlin:
The config values (in KB) need to be converted to bytes when comparing
against total attachment sizes.

Fixes #937.

### DIFF
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -665,8 +665,8 @@ fn post_attachment(
     let size_limit = if let Some(ref user_uuid) = cipher.user_uuid {
         match CONFIG.user_attachment_limit() {
             Some(0) => err_discard!("Attachments are disabled", data),
-            Some(limit) => {
-                let left = limit - Attachment::size_by_user(user_uuid, &conn);
+            Some(limit_kb) => {
+                let left = (limit_kb * 1024) - Attachment::size_by_user(user_uuid, &conn);
                 if left <= 0 {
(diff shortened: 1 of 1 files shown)


dani-garcia:
Hmm, I think I made the change and tested it in bytes initially and then documented it in kilobytes, thanks for noticing!

PR #941, opened 2020-03-31
Added claims to the Headers to verify user type.

BlackDex:
-- NOTE: Let me know what you think of this.
-- It could be handy, but it can also be a pain if we really need to update the claims.
-- Somewhere i think we should invalidate the security stamp if we change the user rights because that is the safest way, and it would force the claims to refresh.

The claims contains the current rights of which orgs the user belongs
to and what rights it has. This is usefull for requests which do not
provide any organization ID in the requests, but we need it to verify if
that request is allowed to be executed. This saves complex database
calls since it is already in the claims.

There is one downside. If the user level has been changed, or is added
to a new org, the claims are not updated instantly. A user should log
out and back in to receive the correct claims/rights.

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -4,7 +4,6 @@
 use crate::util::read_file;
 use chrono::{Duration, Utc};
 use once_cell::sync::Lazy;
-use num_traits::FromPrimitive;
 
 use jsonwebtoken::{self, Algorithm, Header, EncodingKey, DecodingKey};
 use serde::de::DeserializeOwned;
@@ -76,7 +75,7 @@ pub fn decode_admin(token: &str) -> Result<AdminJWTClaims, Error> {
     decode_jwt(token, JWT_ADMIN_ISSUER.to_string())
 }
 
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Debug, Serialize, Deserialize, Clone)]
 pub struct LoginJWTClaims {
     // Not before
     pub nbf: i64,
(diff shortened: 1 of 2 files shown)


src/auth.rs
```diff
@@ -273,9 +309,10 @@ impl<'a, 'r> FromRequest<'a, 'r> for Headers {
             Ok(claims) => claims,
             Err(_) => err_handler!("Invalid claim"),
         };
+        let claim = claims.clone();
 
-        let device_uuid = claims.device;
-        let user_uuid = claims.sub;
+        let device_uuid = claim.device;
+        let user_uuid = claim.sub;
```
dani-garcia, line 315: If we are only using the cloned object here, I think it would be better to just clone these two fields, instead of the whole struct, or even better if we could get away with no clones, maybe removing the fields or just storing a reference.

src/auth.rs
```diff
@@ -336,13 +374,19 @@ impl<'a, 'r> FromRequest<'a, 'r> for OrgHeaders {
...
+                        // }
+
                         let conn = match request.guard::<DbConn>() {
                             Outcome::Success(conn) => conn,
                             _ => err_handler!("Error getting DB"),
                         };
 
                         let user = headers.user;
-                        let org_user = match UserOrganization::find_by_user_and_org(&user.uuid, &org_id, &conn) {
+                        match UserOrganization::find_by_user_and_org(&user.uuid, &org_id, &conn) {
```
dani-garcia, line 389: If we keep this db query here, it doesn't make much sense to check the org level against the claims because we already have it here and this one is instant not like the claims

BlackDex:
Thx for the review.
I think that we then should leave it as is for now.
And that were needed we just do the extra checking within the needed functions.
This since there are for now no performance issues regarding the database, and i don't think it will be a bottleneck for a while either. 

This will only become useful and nice if we are able to trigger a token_refresh during sync, which is done several times during the session.

So for now, i will close this PR.

PR #949, opened 2020-04-05
Introduce db connection max pool size configuration

Koisell:
How many connections used in the pool size could be dynamic in order to
adapt to any environment. This value can now be set by using
DB_CONNECTION_POOL_MAX_SIZE.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -333,6 +333,9 @@ make_config! {
         /// Log level
         log_level:              String, false,  def,    "Info".to_string();
 
+        /// Database connection pool max size.
+        db_connection_pool_max_size: u32,true,  def,    10;
+
         /// Enable DB WAL |> Turning this off might lead to worse performance, but might help if using bitwarden_rs on some exotic filesystems,
         /// that do not support WAL. Please make sure you read project wiki on the topic before changing this setting.
         enable_db_wal:          bool,   false,  def,    true;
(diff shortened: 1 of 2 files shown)


dani-garcia:
Have you found any scenario where the server performance is limited to the number of connections? I'd assume that the web server will be the main limiter for now, at least until we get a proper async version.

Koisell:
I faced the situation in which I needed the pool to be smaller in order to allow proper rolling update since r2d2 doesn't properly end the session and I can't modify the maximum number of connection allowed by my database.

BlackDex:
I Do wonder how this will help with a rolling update (of the database server i guess).
If r2ds/diesel has just 2 instead of 20 connections they still are connected. If the connection breaks, it should just reinitialize it.

Could you explain it a bit more in detail in how limiting affects your use case?

Koisell:
I will to explain what happened to me as much as I can.

I use a mySQL managed database. I recently reduced my plan, so the database allow only 15 parallels connexions. When running a rolling upgrade I have 20 parallels connexions and so r2d2 never start.

While playing a bit with the code and my setup I also realized r2d2 doesn't close connexions when you stop the application and by default on mySQL the wait_timeout parameters is set to [8 hours](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_wait_timeout) which mean even with a database with 25 connexions I should wait 8 hours between two updates (ok that's unprobable in production but possible in some dev/test environments). Of course I can connect to the database to manually kill  connexions.

To me this is mainly my problem. I assume people should design systems at proper scale and with proper tooling to handle incidents  but since not everybody can't do that for various reason I summit this PR to make the connexion pool more flexible.

PS: I'm also looking to make a PR to r2d2 to provide a close function but I don't know how it will end.

Koisell:
> Not to favor anyone, but this PR has conflicts and since there is a new PR ( #1108 ) which has this solved already maybe best to close this one?
I agree and close this PR

PR #956, opened 2020-04-08
Fix Duo auth failure with non-lowercased email addresses

### LINKED ISSUE #955: DUO 2FA - login fails if email is not lowercase
### Subject of the issue
When logging in with DUO 2FA configured, login fails if email entered is not all lowercase.

### Your environment

* Bitwarden_rs version: 
1.14.1-843604c9
* Install method: official docker image: bitwardenrs/server:latest
* Clients used: Web (Chrome)
* Reverse proxy and version: Traefik 1.7.21
* Version of mysql/postgresql: n/a
* Other relevant information: 

### Steps to reproduce
1. start up a bitwarden_rs container
2. create an account with the following email: Test@test.com
3. log in, and set up 2FA via DUO normally
4. log out
5. log in with the following email (note the capital T): Test@test.com
6. perform DUO 2FA
7. observe that login FAILED
8. go back to login screen, and log in with the following email (note the lowercase t): test@test.com
9. perform DUO 2FA
10. observe that login SUCCEEDS, and you are now in your vault

### Expected behaviour
From what I understand, email shouldn't be required to be all lowercase. Without DUO, we are able to login using mixed uppercase and lowercase email. But with DUO enabled, we are forced to use lowercase email.

### Actual behaviour
With DUO enabled, users cannot login using mixed case emails. Users are forced to enter lower case email, even though the user registered with mixed case emails initially.

### Relevant logs
AJAX call received HTTP 400 error:
```
POST /identity/connect/token response:
{"ErrorModel":{"Message":"Error validating duo authentication","Object":"error"},"Message":"","Object":"error","ValidationErrors":{"":["Error validating duo authentication"]},"error":"","error_description":""}
```
docker log output:
```
[2020-04-07 19:48:38][request][INFO] POST /api/accounts/prelogin
[2020-04-07 19:48:38][response][INFO] POST /api/accounts/prelogin (prelogin) => 200 OK
[2020-04-07 19:48:38][request][INFO] POST /identity/connect/token
[2020-04-07 19:48:38][error][ERROR] 2FA token not provided
[2020-04-07 19:48:38][response][INFO] POST /identity/connect/token (login) => 400 Bad Request
[2020-04-07 19:48:51][request][INFO] POST /identity/connect/token
[2020-04-07 19:48:51][error][ERROR] Error validating duo authentication
[2020-04-07 19:48:51][response][INFO] POST /identity/connect/token (login) => 400 Bad Request
```

jjlin:
Fixes #955.

### DIFF
--- a/src/api/core/two_factor/duo.rs
+++ b/src/api/core/two_factor/duo.rs
@@ -21,9 +21,9 @@ pub fn routes() -> Vec<Route> {
 
 #[derive(Serialize, Deserialize)]
 struct DuoData {
-    host: String,
-    ik: String,
-    sk: String,
+    host: String, // Duo API hostname
+    ik: String,   // integration key
+    sk: String,   // secret key
 }
 
 impl DuoData {
@@ -190,6 +190,7 @@ fn duo_api_request(method: &str, path: &str, params: &str, data: &DuoData) -> Em
     use reqwest::{header::*, Method, blocking::Client};
     use std::str::FromStr;
 
+    // https://duo.com/docs/authapi#api-details
     let url = format!("https://{}{}", &data.host, path);
     let date = Utc::now().to_rfc2822();
     let username = &data.ik;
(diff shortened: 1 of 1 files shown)


jjlin:
This appears to be the only such problem, since `username` isn't passed directly much. Another option would've been to lowercase `username` immediately at https://github.com/dani-garcia/bitwarden_rs/blob/843604c9e7f755c6d401938640b6bc9fa0886f1e/src/api/identity.rs#L392 but I decided against that in case there's something that actually does rely on a case-sensitive email at some point.

dani-garcia:
Thanks!

PR #957, opened 2020-04-09
Domain whitelist cleanup and fixes

jjlin:
* Make `SIGNUPS_DOMAINS_WHITELIST` override the `SIGNUPS_ALLOWED` setting.
  Otherwise, a common pitfall is to set `SIGNUPS_DOMAINS_WHITELIST` without
  realizing that `SIGNUPS_ALLOWED=false` must also be set.

* Whitespace is now accepted in `SIGNUPS_DOMAINS_WHITELIST`. That is,
  `foo.com, bar.com` is now equivalent to `foo.com,bar.com`.

* Add validation on `SIGNUPS_DOMAINS_WHITELIST`. For example, `foo.com,`
  is rejected as containing an empty token.

* Make org owner invitations respect the email domain whitelist.
  This closes a loophole where org owners can invite new users from any domain.

### DIFF
--- a/src/api/core/accounts.rs
+++ b/src/api/core/accounts.rs
@@ -68,7 +68,7 @@ fn register(data: JsonUpcase<RegisterData>, conn: DbConn) -> EmptyResult {
     let mut user = match User::find_by_mail(&data.Email, &conn) {
         Some(user) => {
             if !user.password_hash.is_empty() {
-                if CONFIG.signups_allowed() {
+                if CONFIG.is_signup_allowed(&data.Email) {
                     err!("User already exists")
                 } else {
                     err!("Registration not allowed or user already exists")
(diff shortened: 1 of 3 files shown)


jjlin:
Just a note to clarify why it makes sense for `SIGNUPS_DOMAINS_WHITELIST` to override `SIGNUPS_ALLOWED`. Currently, I think with the combination of these two settings, there are two reasonable ways to interpret them:

1. If `SIGNUPS_ALLOWED=false` and `SIGNUPS_DOMAINS_WHITELIST` is set, this means "signups are not allowed, except from this list of domains". It appears this was the interpretation that was intended when the whitelist was originally implemented.
2. If `SIGNUPS_ALLOWED=true` and `SIGNUPS_DOMAINS_WHITELIST` is set, this means "signups are allowed, but only from this list of domains".

In both of these interpretations, `SIGNUPS_ALLOWED` is effectively ignored/overridden.

tomuta:
I like this, but maybe we could go another route and have only one setting, like `SIGNUPS_ALLOWED` with values of `true`, `false`, or a list of domains?

jjlin:
The simplicity of that is appealing, but it's a much more backward incompatible change. If we went that route, I'd propose calling it something like `SIGNUP_DOMAINS` with an empty value meaning no signups allowed, `*` meaning all signups allowed, or a list of specific domains.

dani-garcia:
The idea of a unified setting is nice, and might be something we want to do for a possible future version 2.0 were we get to reestructure the whole configuration thing, but as it is we'd have to keep backwards compatibility with the current `SIGNUPS_DOMAINS_WHITELIST` setting, which might make it more confusing.

jjlin:
Actually, I think I need to push one more commit to this.

jjlin:
@dani-garcia Any design/implementation objections to this, or are you just still looking at it?

dani-garcia:
Ah, I left it open planning to give it a more in-depth look and then just forgot about it!

Went through it now and looks good to me.

PR #959, opened 2020-04-09
Add RPM package

alexpdp7:
This is a WIP, I'm interested in creating a CentOS 8 package for bitwarden_rs that installs a systemd user service using SQLite and provide a COPR dnf/yum repo.

(My thought is that from reading the documentation, I see that bitwarden_rs is meant to be single-user? Having a systemd user service would mean that you could have every user in a system run their own bitwarden_rs easily, but maybe I got this part wrong).

There is no reason this wouldn't work for other CentOS/RHEL/Fedora versions.

The package "works" right now, but it's running with `WEB_VAULT_ENABLED=false`, it only supports SQLite.

It's also pushing to a COPR repo in my fork. To test it:

```
$ sudo dnf copr enable koalillo/bitwarden_rs 
$ sudo dnf install bitwarden_rs
$ curl http://localhost:8000/
```

This is in part a RPM/COPR learning exercise, but I think it would be a cool idea to mainline RPM support. If we merge something like this, it should be easy to set up CI to push new packages to a COPR repo for every release automatically. People then can install it with the two commands above and have it automatically patched with dnf-automatic.

I'm happy to fix more stuff. If you guys don't want to mainline this (I know maintaining more things is never easy), I'm happy to keep this in an external repo.

Also note that at this point:

* Fedora/CentOS apparently won't accept nightly-based Rust software in their main repos
* I think this PR manages to evade best practices in all steps of the way. It works, but it's not "good"

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
-name = "vaultwarden"
-version = "1.0.0"
+name = "bitwarden_rs"
+version = "1.24.0"
 authors = ["Daniel García <dani-garcia@users.noreply.github.com>"]
 edition = "2021"
 rust-version = "1.60"
@@ -11,6 +11,7 @@ readme = "README.md"
 license = "GPL-3.0-only"
 publish = false
 build = "build.rs"
+description = "foo"
 
 [features]
 # Empty to keep compatibility, prefer to set USE_SYSLOG=true
@@ -152,3 +153,12 @@ rocket_contrib = { git = 'https://github.com/SergioBenitez/Rocket', rev = '263e3
 # In particular, `cron` has since implemented parsing of some common syntax
 # that wasn't previously supported (https://github.com/zslayton/cron/pull/64).
 job_scheduler = { git = 'https://github.com/jjlin/job_scheduler', rev = 'ee023418dbba2bfe1e30a5fd7d937f9e33739806' }
+
+[package.metadata.rpm.cargo]
+buildflags = ["--release", "--features=sqlite"]
+
+[package.metadata.rpm.targets]
+bitwarden_rs = { path = "/usr/bin/bitwarden_rs" }
+
+[package.metadata.rpm.files]
+"bitwarden_rs.service" = { path = "/usr/lib/systemd/system/bitwarden_rs.service" }
--- a/.rpm/bitwarden_rs.service
+++ b/.rpm/bitwarden_rs.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=bitwarden_rs
+After=network.target
+
+[Service]
+ExecStart=/usr/bin/bitwarden_rs
+Environment=WEB_VAULT_ENABLED=false
+Environment=DATA_FOLDER=/var/lib/bitwarden_rs
+User=bitwarden_rs
+Group=bitwarden_rs
+
+[Install]
+WantedBy=multi-user.target
--- a/.rpm/bitwarden_rs.spec
+++ b/.rpm/bitwarden_rs.spec
@@ -1,0 +1,54 @@
+%define __spec_install_post %{nil}
+%define __os_install_post %{_dbpath}/brp-compress
+%define debug_package %{nil}
+
+Name: bitwarden_rs
+Summary: foo
+Version: @@VERSION@@
+Release: @@RELEASE@@
+License: GPLv3
+Group: Applications/System
+Source0: %{name}-%{version}.tar.gz
+
+BuildRoot: %{_tmppath}/%{name}-%{version}-%{release}-root
+BuildRequires: systemd
+
+Requires(pre): shadow-utils
+Requires(post): systemd
+Requires(preun): systemd
+Requires(postun): systemd
+
+%description
+%{summary}
+
+%prep
+%setup -q
+
+%install
+rm -rf %{buildroot}
+mkdir -p %{buildroot}
+cp -a * %{buildroot}
+mkdir -p %{buildroot}%{_localstatedir}/lib/bitwarden_rs/
+
+%clean
+rm -rf %{buildroot}
+
+%systemd_post bitwarden_rs.service
+
+%preun
+%systemd_preun bitwarden_rs.service
+
+%postun
+%systemd_postun_with_restart bitwarden_rs.service
+
+%files
+%defattr(-,root,root,-)
+%{_bindir}/*
+%{_unitdir}/bitwarden_rs.service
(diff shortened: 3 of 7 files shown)


mqus:
Thank you for your effort! Did you see https://github.com/dani-garcia/bitwarden_rs/wiki/Available-packages already?

> (My thought is that from reading the documentation, I see that bitwarden_rs is meant to be single-user? Having a systemd user service would mean that you could have every user in a system run their own bitwarden_rs easily, but maybe I got this part wrong).

No, bitwarden_rs is a server application which can store the passwords of multiple users, which access them by using a client like the Web vault, a mobile app or the desktop application. Bitwarden_rs should be able to serve at least tens of users, even if you use sqlite as a database backend and host it on a raspberry pi. You probably can provide it as a user service anyway, but then configuring the clients will be hard, because the server will need a different port for each user.

I package bitwarden_rs for Archlinux and already made a single-instance per machine config for systemd which you can find [in the wiki](https://github.com/dani-garcia/bitwarden_rs/wiki/Setup-as-a-systemd-service). I'm also using tmpfiles and sysusers for setting up the data directory und the user of bitwarden_rs. The files for the archlinux package can be found (here)[https://aur.archlinux.org/cgit/aur.git/tree/?h=bitwarden_rs].

If you have any questions feel free to ask. If you look at the issues here it seems like everyone is running a variant of docker but I'm a baremetal guy :)

alexpdp7:
Yeah, I was confused by the message everywhere that "single user functionality is supported".

My big question here is if the maintainers are willing to set up automated RPM builds on releases. If they do, I'm happy to work to adapt https://github.com/MrMEEE/bitwarden_rs_rpm to work on CI.

Otherwise, I'll set up a parallel project (or work with @MrMEEE) to perform out-of-tree separate automated builds (which is completely fine). This PR should serve as a proof-of-concept that an in-tree build can work and is "developer-friendly".

alexpdp7:
I had some time to work on this today and turned the package into a proper multi-user package, using SQLite.

The PR builds the RPM and pushes it to COPR, so it can be installed in two commands. If you enable the systemd service, things will just work. (Right now I'm pushing to a personal COPR at https://copr.fedorainfracloud.org/coprs/koalillo/bitwarden_rs/ ).

It should be simple to tie this into the repo's CI to automatically build packages on release. I'd like to propose doing so :)

The difference with @MrMEEE's package is that this one doesn't require an external database, but MrMEEE's includes the web interface.

MrMEEE:
Hmmm.. great job @alexpdp7 maybe we should work on integrating the web-ui builds and the option of using an external db into your project.. I'm not very good at travis, but I think we should be able to get a good result :)

alexpdp7:
Thanks @MrMEEE , I have created two issues in my fork to represent these features:

https://github.com/alexpdp7/bitwarden_rs/issues/2
https://github.com/alexpdp7/bitwarden_rs/issues/1

I have no idea right now how to do either (this is like my third RPM, and it's the most complex one so far).

The thing about providing a script which uses Docker is that it should be reasonably easy to integrate this into any CI. It seems that releases are now made using GitHub actions:

https://github.com/dani-garcia/bitwarden_rs/blob/master/.github/workflows/workspace.yml

, it wouldn't be difficult to run my script as part of that when there is a release and push to COPR (you need to add the COPR secrets to GitHub and change the .yml, I did something similar for another project:

https://github.com/alexpdp7/barrier/commit/dbdabbf06c45f56c60c36fa933cb807c7f310909

. It should be pretty easy to do that in this project, however I'd like to get some feedback from the maintainers before doing that (if they are not willing to merge the changes, it's useless).

BlackDex:
@alexpdp7 
I am in favor of closing this PR since it states that it is still a WIP for one.
Also, i don't know if we want to have distro specific code/building in this repo (@dani-garcia ?).

A mention into the wiki regarding a location where to get these, or how to build this your self would in my opinion be enough.

alexpdp7:
Hi,

I've edited the PR description as it was outdated and removed the draft status. I have been using the package for a while and it works. I just updated the package to 1.17.0.

I'm happy to remove the PR and keep my package isolated, but it'd be much nicer to merge into the main repository and set up automated builds- it should be possible to set up things so an RPM package is built automatically each time the project is released (perhaps even nightly RPMs).

Cheers,

Álex

Cargo.toml
```diff
@@ -9,6 +9,7 @@ readme = "README.md"
 license = "GPL-3.0-only"
 publish = false
 build = "build.rs"
+description = "foo"
```
williamdes, line 12: Todo

BlackDex:
I'm going to close this PR since we do not want to maintain several different packaging formats for this project for now.
Our main source will be docker container images which we will support.

We do like that people create packages for Vaultwarden for these other OS'es, but we can't maintain those our self.

If you want to provide these package we still encourage you to do so, but mention this on the Third-Party wiki here: https://github.com/dani-garcia/vaultwarden/wiki/Third-party-packages

alexpdp7:
No worries. The packages are already listed in the wiki.

Thanks for your honest answer- if there's no chance of upstreaming, it's actually easier to maintain packaging as a separate repo instead of a branch.

I wish you would reconsider (I've been maintaining this for two years- Vaultwarden has been completely stable all this time, so packaging has required no maintenance other than bumping the versions. This means that maintaining the RPM package upstreamed would probably reduce the effort required to keep the package up to date.

However, I understand the decision and again, I appreciate having feedback.

williamdes:
I also think it is better in a separate repository like we do in Debian with https://salsa.debian.org
I think there is some similar host for RPM somewhere

alexpdp7:
Well, it's a complex topic. Ideally, packaging should be maintained within a distro, with the mechanisms provided by the distro. So in Fedora/EL case, you would have a separate packaging repository in git.centos.org or src.fedoraproject.org, and the package would be included in Fedora/EL. However, it is relatively hard to get a package in a distro- IMHO this depends a lot on the "desirability" of the package [and I think Vaultwarden is relatively popular). In this case, the packaging is outside the main software repo and in a separate repo (most likely hosted by the distro).

If a package does not become part of a distro, then I think *if the developers of the software are willing to do it*, to maintain the packaging in the main repo. This means that when you make a release, you can automate releasing the package too. I have done this, and it's very nice.

However, it's understandable that the devs don't want to do this. In this case, your only option is keeping the packaging separate. How to automate this depends a lot on the distro and how you want to do things (there are cross-packaging things like using OBS).

In this case, I'm already using COPR (a Fedora-provided package builder), so I might try to have a repo on src.fedoraproject.org.

PR #978, opened 2020-04-30
Delete the call to the map file

AltiUP:
The file bootstrap.css.map is missing, the reference can be deleted.
Or else you have to add the file.


dani-garcia:
Thanks!

PR #987, opened 2020-05-07
GlobalEquivalentDomains updates from upstream bitwarden

theycallmesteve:
Updates the global equivalent domain values from bitwarden.  It also fixes some whitespace inconsistencies that I noticed.


dani-garcia:
Thanks!

PR #990, opened 2020-05-08
Extend collection response models

theycallmesteve:
Extend collection response models to include collection, collectionDetails, and collectionGroupDetails

Edit: This fixes read only collections being listed in add/edit cipher for an organization.  It also sets the ExternalId to "Not supported in bitwarden_rs" for collectionGroupDetails, which will be displayed edit collection.

### DIFF
--- a/src/db/models/collection.rs
+++ b/src/db/models/collection.rs
@@ -23,13 +23,23 @@ impl Collection {
         }
     }
 
-    pub fn to_json(&self) -> Value {
-        json!({
+    pub fn to_json(&self, user_uuid: &str, conn: &DbConn, resp_model: &str) -> Value {
+        let mut json_object = json!({
             "Id": self.uuid,
             "OrganizationId": self.org_uuid,
             "Name": self.name,
-            "Object": "collection",
-        })
+            "Object": resp_model,
+            "ExternalId": "", //NOT_IMPLEMENTED
+        });
+
+        if resp_model == "collectionDetails" {
+            json_object["ReadOnly"] = json!(!self.is_writable_by_user(&user_uuid, &conn));
+        } else if resp_model == "collectionGroupDetails" {
+            json_object["ExternalId"] = json!("Not supported in bitwarden_rs"); //NOT_IMPLEMENTED
+            json_object["Groups"] = json!([]); //NOT_IMPLEMENTED
+        }
+
+        json_object
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/db/models/collection.rs
```diff
@@ -23,13 +23,23 @@ impl Collection {
         }
     }
 
-    pub fn to_json(&self) -> Value {
-        json!({
+    pub fn to_json(&self, user_uuid: &str, conn: &DbConn, resp_model: &str) -> Value {
```
dani-garcia, line 26: This user_uuid should be an Option<&str>, to avoid using "".

Also, I'd rather not have magic strings as the resp_model parameter, you can either create an enum and implement to_string for it:
```rust
pub enum ModelType {
    collection,
    collectionGroupDetails,
}

impl ToString for ModelType {
    fn to_string(&self) -> String {
        match self {
            Self::collection => "collection".into(),
            Self::collectionGroupDetails => "collectionGroupDetails".into()
        }
    }
}

fn main() {
    println!("{}", ModelType::collectionGroupDetails.to_string());
}

```

Or in this case, you could get away with a boolean to decide if it's collectionDetails or collectionGroupDetails (because collection is only used without the user id).

theycallmesteve: I will make the change for user_uuid  to this PR and change both PR's to include the enum/to_string changes.

If you wouldn't mind though, I'd like to make sure I'm not missing something.  I assume avoiding magic strings and using an enum/to_string is for better code readability and for easier maintenance.

What is the reason to avoid using ""?  Without function overloading, we trade "" for None and use Some(x) on the rest of the calls.  Is there some type safety issue that isn't apparent to me?

Again, I'm just trying to understand.  I'm happy to make the changes, regardless of the reason, in order to further the cause.

dani-garcia: The magic strings is for two things: one, to avoid typo errors, like writing "colectionGroupDetails" which wouldn't work (only one l). It also allows any future readers to just open the definition from their editor to check what other options are available.

The reason for avoiding an empty string is mostly to better show what the code means, "" could either mean no value present, like it does now or it could just mean a value is present, and the value is an empty string, differenciating betweeen None and Some("") allows to separate those two meanings, in this case it isn't a big deal but imagine for example a function that is used to change some config value, when we send no value, it might make sense to use a default instead, so does "" mean use a default, or literally use the string ""? Hard to know without reading documentation, but None and Some("") make it explicit.

And don't worry about asking any questions in the future! 😄

PR #998, opened 2020-05-18
Fixing bad width in 2FA email template

frdescam:
On mobile apps (tested with protonmail and K-9 mail) the 2FA email was shifted to the left because of some width style of 600px that was good for other emails but not for this one, this one needed a width of 410px :
before :
Screenshot_20200518-134242_ProtonMail
after :
Screenshot_20200518-134215_ProtonMail
before :
Screenshot_20200518-134319_Mail
after :
Screenshot_20200518-134309_Mail

Nothing has changed for webmails but of course I haven't tested every mailclient


dani-garcia:
I think all the emails are using the same template more or less, is this the only one where this happens? Do your changes still look correctly when used in a desktop system?

frdescam:
Yes all emails are using the same template, the problem is that the width of the container is of 600px but the text inside this email is only of 410px, this is the only one that is too short so the problem does not happens in other emails.
Yes it looks correct on desktop webmails from protonmail and nextcloud, I didn't test more

After thinking a lot about that, i'm wondering, did I use the good method to fix that? may be it would be cleaner to change in every template the width of the contained box to `max content`

I'm sorry i'm not very experimented with css :/ Do you want me to make a new pull request?

frdescam:
Ok, I did it, I changed the width to `max-content`, it will allow in the future to change templates text without breaking mobile displaying, is that ok for you?

dani-garcia:
Okay, I tested and it looks good to me thanks!

PR #1004, opened 2020-05-24
KB -> KiB for attachements size

frdescam:
Changing KB, MB, GB, TB, PB to KiB, MiB, GiB, TiB, PiB as the value is calculated using 1024B as 1KB

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -227,7 +227,7 @@ pub fn delete_file(path: &str) -> IOResult<()> {
     res
 }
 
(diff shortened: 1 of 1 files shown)


dani-garcia:
I think I mostly agree with what was mentioned about it being possibly more confusing, I realize that using KB is technically incorrect, but at the same time the size we display is the encrypted size which is probably going to be a little different to the unencrypted size that the users expect, so we won't be showing 100% accurate figures either way.

Thanks anyway for bringing it up!

frdescam:
OK! No problem, this was just an idea.

PR #1007, opened 2020-05-25
Admin interface restyle

BlackDex:
Some changes done to the admin interface.

Main changes:
- Splitted up settings and users into two separate pages.
- Added Organizations and Diagnostics pages.
- Added verified shield when the e-mail address has been verified.
- Added the amount of personal items in the database to the users overview.

Items which can be added later:
- Amount of cipher items accessible for a user, not only his personal items.
- Amount of users per Org
- Version update check in the diagnostics overview.
- Copy/Pasteable runtime config which has sensitive data changed or removed for support questions either on the forum or github issues.
- Option to delete Orgs and all its passwords (when there are no members anymore).
- Etc....


src/api/admin.rs
```diff
@@ -195,13 +241,29 @@ fn logout(mut cookies: Cookies) -> Result<Redirect, ()> {
...
     Ok(Json(Value::Array(users_json)))
 }
 
+#[get("/users/overview")]
+fn users_overview(_token: AdminToken, conn: DbConn) -> ApiResult<Html<String>> {
+    let users = User::get_all(&conn);
+    let users_json: Vec<Value> = users.iter()
+    .map(|u| {
+        let mut usr = u.to_json(&conn);
+        if let Some(ciphers) = Cipher::count_owned_by_user(usr["Id"].as_str().unwrap(), &conn) {
```
dani-garcia, line 257: Easier this way?

```suggestion
        if let Some(ciphers) = Cipher::count_owned_by_user(&u.uuid, &conn) {
```

src/api/admin.rs
```diff
@@ -242,6 +304,50 @@ fn update_revision_users(_token: AdminToken, conn: DbConn) -> EmptyResult {
...
+    use crate::util::read_file_string;
+
+    let vault_version_path = format!("{}/{}", CONFIG.web_vault_folder(), "version.json");
+    let vault_version_str = read_file_string(&vault_version_path)?;
+    let web_vault_version: WebVaultVersion = serde_json::from_str(&vault_version_str)?;
+
+    let github_ips = ("github.com", 0).to_socket_addrs().map(|mut i| i.next());
+    let dns_resolved = match github_ips {
+        Ok(Some(a)) => a.ip().to_string() ,
+        _ => "0.0.0.0".to_string(),
```
dani-garcia, line 335: If we don't actually resolve, maybe it's better to just directly say "Couldn't resolve domain name"?

Also this might return multiple addresses in the case of a IPv6 capable network no? Might be worth it to iterate and show them all, though it's not very important.

src/static/templates/admin/diagnostics.hbs
```diff
@@ -0,0 +1,58 @@
...
+                        <i class="fa fa-check text-success d-none" id="dns-success" title="DNS Resolving works!"></i>
+                        <i class="fa fa-warning text-warning d-none" id="dns-warning" title="DNS Resolving failed. Please fix."></i>
+                    </dd>
+
+                    <dt class="col-sm-5">Date & Time (UTC)</dt>
+                    <dd class="col-sm-7">
+                        <span id="time-server" class="d-block"><b>Server:</b> {{diagnostics.server_time}}</span>
+                        <span id="time-browser" class="d-block"><b>Browser:</b> <span id="time-browser-string"></span></span>
+                        <i class="fa fa-check text-success d-none" id="time-success" title="Time settings are correct."></i>
+                        <i class="fa fa-warning text-warning d-none" id="time-warning" title="Time settings seem incorrect."></i>
```
dani-garcia, line 41: Are these here for a future update? If so it's fine to leave them but otherwise we should probably remove them if they do nothing.

dani-garcia:
Also try to squash your commits if you can, to avoid all those merge commits.

BlackDex:
Thx for the feedback. Some nice beginner mistakes there ;). Will fix them. And while I'm at it i will try to fix the fa-check/warnings that was an oversight.

Also good point on the resolving to make that just a text when an error occurred.

BlackDex:
@dani-garcia I did some git magic and some further small changes. All is now i one commit.

BlackDex:
@dani-garcia I just also added a version check to the diagnostic page.

src/api/admin.rs
```diff
@@ -242,6 +304,93 @@ fn update_revision_users(_token: AdminToken, conn: DbConn) -> EmptyResult {
...
+fn get_github_api(url: &str) -> Result<Value, Error> {
+    use reqwest::{header::USER_AGENT, blocking::Client};
+    let github_api = Client::builder().build()?;
+
+    let res = github_api
+        .get(url)
+        .header(USER_AGENT, "Bitwarden_RS")
+        .send()?;
+
+    // If we get a 404, return a 404, it means no breached accounts
```
dani-garcia, line 331: I think this comment is not applicable anymore :)

BlackDex: @dani-garcia, i rewrote history again ;).

dani-garcia:
Thanks @BlackDex!

Issue #968, opened 2020-04-16
Dockerfiles run services as the root user


xhalo32:
### Subject of the issue
Shouldn't use the root user to run the services inside the containers.

https://medium.com/@mccode/processes-in-containers-should-not-run-as-root-2feae3f0df3b

jjlin:
As noted in https://github.com/dani-garcia/bitwarden_rs/pull/1014#issuecomment-637965744, this probably isn't a good default for bitwarden_rs, but I've now added documentation on how to run as non-root:

https://github.com/dani-garcia/bitwarden_rs/wiki/Hardening-Guide#run-as-a-non-root-user

jostyee:
@jjlin I've got a permission error after adding user settings to my docker-compose.yml:

```
[2020-06-11 15:37:54][bitwarden_rs][ERROR] Launch error Bind(
    Io(
        Os {
            code: 13,
            kind: PermissionDenied,
            message: "Permission denied",
        },
    ),
)
```

Stampede10343:
> As noted in [#1014 (comment)](https://github.com/dani-garcia/bitwarden_rs/pull/1014#issuecomment-637965744), this probably isn't a good default for bitwarden_rs, but I've now added documentation on how to run as non-root:
> 
> https://github.com/dani-garcia/bitwarden_rs/wiki/Hardening-Guide#run-as-a-non-root-user

Should this remove the [listing here](https://github.com/dani-garcia/bitwarden_rs/wiki/Running-the-server-with-non-root-user), since its basically the same, and better suited in the Hardening Guide, IMO.

jjlin:
@Stampede10343 I didn't notice that page, but I think it has a few additional things that would be good to mention. I'll merge it into the hardening guide page tonight. Thanks for calling attention to that.

jjlin:
It's updated now.

BlackDex:
Thx for the updated wiki, closing this issue.

PR #1014, opened 2020-06-02
Run bitwarden-rs as non-root user

Stampede10343:
Updated the Dockerfiles to switch to UID 999 before running. Because
non-root users cannot bind to ports below 1024, I also updated the
ROCKET_PORT to 8080.

This should address #968

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,9 +35,10 @@ Pull the docker image and mount a volume from the host for persistent storage:
 
 ```sh
 docker pull bitwardenrs/server:latest
-docker run -d --name bitwarden -v /bw-data/:/data/ -p 80:80 bitwardenrs/server:latest
+docker run -d --name bitwarden -v /bw-data/:/data/ -p 8080:80 bitwardenrs/server:latest
 ```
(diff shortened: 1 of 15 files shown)


jjlin:
Unfortunately, this change will probably break many (most?) existing bitwarden_rs instances.

It's certainly better to run with less privileges, all else being equal. However, the article cited in #968 doesn't address the fact that bind-mounting data into a container running as non-root can significantly complicate ownership/permission management of that data, which will generate a lot more questions/issues from users who aren't necessarily well-versed on this. Balanced against this, I think the marginal security benefit of running as non-root probably isn't a good tradeoff for bitwarden_rs, especially since it's written in a memory-safe language.

I think it would be better just to document how interested users can harden their configuration. There's an existing [wiki page](https://github.com/dani-garcia/bitwarden_rs/wiki/Hardening-Guide) that could use help.

Stampede10343:
Agreed, its a bit tough on existing users, especially when you aren't solid on linux permissions, I'll take a look at the wiki page and maybe do a write up on how to run as non-root for security conscious users.

PR #1029, opened 2020-06-10
Multi-arch image support

### LINKED ISSUE #1011: Docker images list wrong architecture
### Subject of the issue
All of your docker images/tags are uploaded with the OS/Arch set to Linux/amd64. This leads to "docker pull bitwardenrs/server" or "FROM bitwardenrs/server" always pulling the tag "latest" even on  other hardwares/architectures.

### Steps to reproduce
Run "docker pull bitwardenrs/server" on a Raspberry Pi (ARMv7).

### Expected behaviour
Docker should automatically pull the tag raspberry.

### Actual behaviour
Docker pulls tag latest and running the image fails.

jjlin:
This implementation uses the existing Dockerfiles, and just creates a manifest list that references each arch-specific image. If the Docker tag being built begins with a version number (`x.y.z`), it also creates an additional manifest list (`latest` or `alpine`) that is a synonym for `x.y.z` or `x.y.z-alpine`. This should make it easier for users to track the latest released version.

Fixes #1011.


hooks/push
```diff
@@ -0,0 +1,96 @@
...
+
+    # Push the manifest list.
+    docker manifest push --purge ${manifest_list}
+done
+
+# Avoid logging credentials and tokens.
+set +ex
+
+# Delete the arch-specific tags, if credentials for doing so are available.
+# Note that `DOCKER`_PASSWORD must be the actual user password. Passing a JWT
```
dani-garcia, line 61: I just checked this just in case (because I have 2FA enabled in docker hub), and you can generate access tokens from the docker hub page that work in place of the password:
image

The URL is https://hub.docker.com/settings/security

jjlin: Yeah, those are the access tokens that won't work (or at least it doesn't for me).

jjlin: BTW, just amended the commit to fix the `DOCKER_PASSWORD` quoting.

dani-garcia: Ah that sucks if that's the case, I just tried the login procedure and a basic tags list but that doesn't require authentication now that I see.

I'll see if I get to review and test this sometime this week, I haven't had a lot of time lately to dedicate, but this seems like a great quality of life thing.

jjlin: Thanks. Regarding the access token issue, I opened https://github.com/docker/hub-feedback/issues/2006.

jjlin: Looks like personal access tokens aren't going to be supported anytime soon, so if you don't want to remove 2FA from your Docker Hub account, maybe creating a non-2FA bot/machine account with access to that repo could be a workaround.

jjlin:
Rebased to fix conflicts.

jjlin:
Addressed your comments and rebased. The edit repository permission is what I would guess too, but I've only tested with a personal account.

PR #1044, opened 2020-07-04
Allow postgres:// in DATABASE_URL

ArmaanT:
This PR adjusts the config validation so that `DATABASE_URL`s of the form `postgres://...` are valid when using the PostgreSQL database backend


dani-garcia:
Thanks!

Issue #1045, opened 2020-07-05
TLS failures using self hosted email server


causalityloop:
### Subject of the issue
bitwarden cannot use configured email server because the cert received, from the mail server, is not trusted. Adding ability to install respective root/subordinate CA's will allow bitwarden to trust the cert

### Your environment

* Bitwarden_rs version: 1.15.1-b34d5482

* Install method: docker compose sourcing image at bitwardenrs/server:latest
* Clients used: 
* Reverse proxy and version: 
* Version of mysql/postgresql: 
* Other relevant information: 

### Steps to reproduce

Used the following related smpt settings
```
extra_hosts:
  - "mail.codeloft.tech:<email_server_ip>"
environment:
  - SMTP_HOST=<email_server_ip>
  - SMTP_FROM=<user>
  - SMTP_PORT=565
  - SMTP_SSL=true
  - SMTP_EXPLICIT_TLS=true
  - SMTP_USERNAME=<user>
  - SMTP_PASSWORD=<password>

```

### Expected behaviour

Emails are delivered

### Actual behaviour

Emails are not delivered. 

### Relevant logs

The bitwarden log contains:

```
bitwardenrs  | [2020-07-05 12:51:08][bitwarden_rs::api::core::accounts][ERROR] Error sending welcome email: SmtpError.
bitwardenrs  | [CAUSE] Io(
bitwardenrs  |     Custom {
bitwardenrs  |         kind: Other,
bitwardenrs  |         error: Failure(
bitwardenrs  |             Ssl(
bitwardenrs  |                 Error {
bitwardenrs  |                     code: ErrorCode(
bitwardenrs  |                         1,
bitwardenrs  |                     ),
bitwardenrs  |                     cause: Some(
bitwardenrs  |                         Ssl(
bitwardenrs  |                             ErrorStack(
bitwardenrs  |                                 [
bitwardenrs  |                                     Error {
bitwardenrs  |                                         code: 337047686,
bitwardenrs  |                                         library: "SSL routines",
bitwardenrs  |                                         function: "tls_process_server_certificate",
bitwardenrs  |                                         reason: "certificate verify failed",
bitwardenrs  |                                         file: "../ssl/statem/statem_clnt.c",
bitwardenrs  |                                         line: 1915,
bitwardenrs  |                                     },
bitwardenrs  |                                 ],
bitwardenrs  |                             ),
bitwardenrs  |                         ),
bitwardenrs  |                     ),
[...]

PR #1048, opened 2020-07-05
Add startup script to support init operations

jjlin:
This is useful for making local customizations upon container start. To use
this feature, mount a script into the container as `/etc/bitwarden_rs.sh`
and/or a directory of scripts as `/etc/bitwarden_rs.d`. In the latter case,
only files with an `.sh` extension are sourced, so files with other
extensions (e.g., data/config files) can reside in the same dir.

Note that the init scripts are run each time the container starts (not just
the first time), so these scripts should be idempotent.

### DIFF
--- a/docker/Dockerfile.j2
+++ b/docker/Dockerfile.j2
@@ -288,9 +288,10 @@ COPY --from=build app/target/release/bitwarden_rs .
 {% endif %}
 
 COPY docker/healthcheck.sh /healthcheck.sh
+COPY docker/start.sh /start.sh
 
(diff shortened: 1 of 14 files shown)


jjlin:
Fixes #1045.

dani-garcia:
Thanks! We should probably document this in the wiki somewhere too.

jjlin:
> Thanks! We should probably document this in the wiki somewhere too.

Sure, I added a section at https://github.com/dani-garcia/bitwarden_rs/wiki/Starting-a-Container#customizing-container-startup.

PR #1067, opened 2020-07-23
Add config option for log timestamp format

jjlin:


jjlin:
@dani-garcia I think it would be nice to log with milliseconds by default, which could be useful for checking request/response time differences more easily. I left it with the current format since some people might rely on that, but if you're fine with including ms, I can add it.

dani-garcia:
Yeah sounds reasonable to change it to milliseconds, I assume most people parsing logs are using it for Fail2Ban or similar purposes, and at least in the wiki example we only match against the message not the entire line, so hopefully most people did the same.

PR #1069, opened 2020-07-24
Skip cleanup of `arm32v6` arch-specific tags

### LINKED ISSUE #1064: Exit Code 139 with new docker multiarch on Pi Zero W
I just wanted to try out the new server:latest multiarch docker image on my Pi Zero W. Unfortunately bitwarden_rs doesn't start anymore and ends with an error code 139. I guess it could be due to my outdated docker-ce version (18.06.3) on raspbian stretch which I can't update due to a bug, otherwise docker won't run on my Pi Zero anymore. There is already a long issue (https://github.com/moby/moby/issues/38175). Apparently it hasn't been fixed yet. Or could it be something else?
What can I do now to get the latest updates? The current armv6 image in the docker hub seems to support only amd64. Therefore I am currently stucked to my local image. 

EDIT: Probably this is the reason: https://github.com/MichaIng/DietPi/issues/3258#issuecomment-563445349 
So RBPi1/Zero always need their own image /builds.

jjlin:
Fixes #1064.

### DIFF
--- a/hooks/push
+++ b/hooks/push
@@ -88,6 +88,12 @@ jwt=$(curl -sS -X POST \
 repo="${DOCKER_REPO#*/}"
 
 for arch in ${arches[@]}; do
+    # Don't delete the `arm32v6` tag; Docker can't seem to properly
+    # auto-select that image on Armv6 platforms like Raspberry Pi 1 and Zero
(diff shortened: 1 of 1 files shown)


dani-garcia:
Thanks!

PR #1076, opened 2020-07-26
Fix soft delete notifications

jjlin:
A soft-deleted entry should now show up in the trash folder immediately
(previously, an extra sync was required).


jjlin:
Ref: https://github.com/bitwarden/server/blob/229478ad/src/Core/Services/Implementations/CipherService.cs#L703

dani-garcia:
Thanks, didn't notice it!

PR #1082, opened 2020-08-03
Added klm/flyingblue and eventbrite domains to global domains

sebthom:


dani-garcia:
Closing this in favor of #1083, to keep it as close as possible to upstream.

PR #1083, opened 2020-08-04
Add a script to auto-generate the global equivalent domains JSON file

jjlin:
The script works by reading the relevant files from the upstream Bitwarden
source repo and generating a matching JSON file. It could potentially be
integrated into the build/release process, but for now it can be run manually
as needed.


jjlin:
Setting to draft pending some upstream changes to the Eventbrite list.

PR #1096, opened 2020-08-13
Add a note that settings in .env can be overridden

mqus:
Just a small attempt to make it more clear that the value of settings in the `.env` file may be ignored in some cases.

I'm not sure on the exact wording, please tell me if you have other suggestions :)


dani-garcia:
Yeah that seems good enough for me, thanks!

PR #1108, opened 2020-08-21
Add DATABASE_MAX_CONNS config setting

rfwatson:
Tiny change to allow the database connection pool size to be configured externally (useful for those of us deploying to databases with limited connection count allowances)

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -20,6 +20,10 @@
 ## - https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING
 # DATABASE_URL=postgresql://user:password@host[:port]/database_name
 
+## Database max connections
+## Define the size of the connection pool used for connecting to the database.
+# DATABASE_MAX_CONNS=10
+
 ## Individual folders, these override %DATA_FOLDER%
 # RSA_KEY_FILENAME=data/rsa_key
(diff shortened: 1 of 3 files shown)


Ornias1993:
This one is actually passing the checks and the other one isn't ;)

whyvra:
Would love to see this PR merged in. When I tried to deploy bitwarden_rs today, it filled up all of my pgpool connections. I find it kinda crazy that it needs 10 simultaneous connections to the database.

+1

BlackDex:
this PR has some conflicts so a simple merge is not possible.

rfwatson:
I’ll fix the conflicts later if I have time

BlackDex:
Well, there arn't any conflicts it seems.

But i do have one remark. It would be best if there is some validation done on the input value.
r2d2 states that if the value is 0 it will panic. I would like to prevent that from happening at all.

So i think at least a check that is is 1 or higher, maybe even a max, because i wouldn't want to have 1000 connections to my database by accident.

rfwatson:
Updated in https://github.com/dani-garcia/bitwarden_rs/commit/b9daa59e5d05f1050e6ba4199352953493a57dfb

BlackDex:
LGTM, and since there seems to be quite some request about this i think it is fine by me.

BlackDex (approved):

Issue #1113, opened 2020-08-23
Clarification - Does this project support multiple users?


tkambler:
I've read some sources that say this project does support multiple users. Other sources seem to indicate that it does not. Could you please clarify?

mqus:
To get this out of the way: This project does of course support multiple users. How would you expect things like ["Organizations"](https://github.com/dani-garcia/bitwarden_rs#features`) to work otherwise? I haven't heard about large userbases (>1000) yet but everything below that should work flawlessly, even on less powerful hardware.

More importantly (I just noticed you closed this issue :wink: ): why did you think bitwarden_rs does not support multiple users? Maybe we could improve our documentation.

tkambler:
One of the first things you see in the documentation is a message that seems to indicate that `bitwarden_rs` only supports a single user account.

![single](https://user-images.githubusercontent.com/4448053/91192543-a9302900-e6bb-11ea-8b0c-4c1394af3937.jpg)

mqus:
Thank you, I can see how this might lead to misunderstandings.

bokkoman:
Ok, so there is multi user support. But do you still need to have an official licence to be able to support more than 2? Because the free bitwarden only supports 2 users, after that it is $3/$5 a user. Does vaultwarden rely on this? 
Or can we have "unlimited" users with vaultwarden without any licence or payment?

BlackDex:
Vaultwarden does not support the licenses model of Bitwarden.
You do not need any license for all the features we have.
Vaultwarden does not have all the features that Bitwarden has, most of them are linked to organizational features like group support or SSO for example.

All single-user features are there, including all the features which you need to have a license for when using Bitwarden.

PR #1119, opened 2020-08-28
Clarify multiuser capabilities

mqus:
Maybe prevents #1113 and similar issues from popping up again in the future.

As always, please advise on the wording, I don't really know what was meant with "Single user functionality" in the first place.


dani-garcia:
I think it was meant to indicate all the functionality tied to a single user account, as opposed to an organization which made sense when the organization support was more barebones, but now I'd probably just remove the line.

PR #1124, opened 2020-08-31
Fixing MsgPack headers type and support mobile SignalR

aaxdev:
The type of the field "headers" in MsgPack structure seems to be a Map instead of an Array.
This cause crashes with the package SignalR C# client, but not with the Typescript one used by desktop/browser clients.

The package SignalR is also checking if the client is a browser, and will set a header value or a query parameter according the case ([see here](https://github.com/dotnet/aspnetcore/blob/7a65454466e0f17b7879794205e3887471f5839e/src/SignalR/clients/csharp/Http.Connections.Client/src/Internal/WebSocketsTransport.cs#L127)).

So, a dedicated function to resolve any of them seemed better.

Now notifications should works with any SignalR/MsgPack packages, i'm needing this to continue to implement notifications on mobile app.

Thanks @BlackDex for helping about research and for the code improvement, i'm new with the rust lang.


BlackDex (approved):

dani-garcia (approved):

dani-garcia:
Thanks!

PR #1125, opened 2020-09-01
Hide ciphers from non-selected collections for org owners/admins

### LINKED ISSUE #1123: Organization items shows up in My Vault even when no access
### Subject of the issue

If you disable the collection from a user EVEN owner, then the  items of the organization should not show up in "My Vault". The owner must go to the organization's vault to see the items
### Your environment

* Bitwarden_rs version: 1.16.3

* Install method:  Docker latest/postgresql
* Clients used: 
* Reverse proxy and version: 
* Version of mysql/postgresql: postgresql
* Other relevant information: 

### Steps to reproduce

N/A

### Expected behaviour

Organization Vault details should not show up in "My Vault" for owners, if the the checkbox for the collection has not been enabled to them.. If they still need to access the information, they can go to the organization vault.

### Actual behaviour

make 2 owners.. then for yourself..  set it to : This user can access only the selected collections.
and remove access to the collection. Once you do this the collection should not show up in your vault and you should have to go to the organizations vault

jjlin:
If org owners/admins set their org access to only include selected
collections, then ciphers from non-selected collections shouldn't
appear in "My Vault". This matches the upstream behavior.

Fixes #1123.

### DIFF
--- a/src/db/models/cipher.rs
+++ b/src/db/models/cipher.rs
@@ -382,39 +382,58 @@ impl Cipher {
         }}
     }
 
-    // Find all ciphers accessible to user
-    pub fn find_by_user(user_uuid: &str, conn: &DbConn) -> Vec<Self> {
+    // Find all ciphers accessible or visible to the specified user.
+    //
+    // "Accessible" means the user has read access to the cipher, either via
+    // direct ownership or via collection access.
+    //
+    // "Visible" usually means the same as accessible, except when an org
+    // owner/admin sets their account to have access to only selected
+    // collections in the org (presumably because they aren't interested in
+    // the other collections in the org). In this case, if `visible_only` is
+    // true, then the non-interesting ciphers will not be returned. As a
+    // result, those ciphers will not appear in "My Vault" for the org
+    // owner/admin, but they can still be accessed via the org vault view.
+    pub fn find_by_user(user_uuid: &str, visible_only: bool, conn: &DbConn) -> Vec<Self> {
         db_run! {conn: {
-            ciphers::table
-                .left_join(users_organizations::table.on(
-                    ciphers::organization_uuid.eq(users_organizations::org_uuid.nullable()).and(
-                        users_organizations::user_uuid.eq(user_uuid).and(
-                            users_organizations::status.eq(UserOrgStatus::Confirmed as i32)
-                        )
-                    )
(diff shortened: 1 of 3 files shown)


src/db/models/collection.rs
```diff
@@ -208,21 +208,20 @@ impl Collection {
...
-                            .inner_join(collections::table)
-                            .filter(users_collections::collection_uuid.eq(&self.uuid))
-                            .filter(users_collections::user_uuid.eq(&user_uuid))
-                            .filter(users_collections::read_only.eq(false))
-                            .select(collections::all_columns)
-                            .first::<CollectionDb>(conn)
-                            .ok()
-                            .is_some() // Read only or no access to collection
-                    }}
+                if user_org.has_full_access() {
```
jjlin, line 211: This is needed for an org admin/owner to be able to modify a collection when they haven't selected "access all" for themselves.

src/db/models/collection.rs
```diff
@@ -208,21 +208,20 @@ impl Collection {
         match UserOrganization::find_by_user_and_org(&user_uuid, &self.org_uuid, &conn) {
             None => false, // Not in Org
             Some(user_org) => {
-                if user_org.access_all {
-                    true
-                } else {
-                    db_run! { conn: {
-                        users_collections::table
-                            .inner_join(collections::table)
```
jjlin, line 216: Nothing in the `collections` table is actually used/returned, so this join isn't needed.

src/db/models/cipher.rs
```diff
@@ -382,39 +382,58 @@ impl Cipher {
         }}
     }
 
-    // Find all ciphers accessible to user
-    pub fn find_by_user(user_uuid: &str, conn: &DbConn) -> Vec<Self> {
+    // Find all ciphers accessible or visible to the specified user.
+    //
+    // "Accessible" means the user has read access to the cipher, either via
```
jjlin, line 387: I left the option for returning by accessibility open, but Looking at the upstream Bitwarden code, I didn't see any API that actually does that. For reference, the code for returning ciphers for a user is in https://github.com/bitwarden/server/blob/master/src/Sql/dbo/Functions/UserCipherDetails.sql.

hexicans (approved):
Good.

PR #1143, opened 2020-09-14
Format some common Lettre errors a bit simpler

BlackDex:
Currently when for example using the admin interface to send out a test e-mail just
returns `SmtpError`. This is not very helpful. What i have done.

- Match some common Lettre errors to return the error message.
- Other errors will just be passed on as before.

Some small other changes:
- Fixed a clippy warning about using clone().
- Fixed a typo where Lettere was spelled with one t.

### DIFF
--- a/src/mail.rs
+++ b/src/mail.rs
@@ -54,7 +54,7 @@ fn mailer() -> SmtpTransport {
             for wanted_mechanism in mechanism.split(',') {
                 for m in &allowed_mechanisms {
                     if m.to_string().to_lowercase() == wanted_mechanism.trim_matches(|c| c == '"' || c == '\'' || c == ' ').to_lowercase() {
-                        selected_mechanisms.push(m.clone());
+                        selected_mechanisms.push(*m);
                     }
                 }
             };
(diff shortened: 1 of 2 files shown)


src/mail.rs
```diff
@@ -330,6 +330,24 @@ fn send_email(address: &str, subject: &str, body_html: &str, body_text: &str) ->
...
         .multipart(alternative)?;
 
-    let _ = mailer().send(&email)?;
-    Ok(())
+    match mailer().send(&email) {
+        Ok(_) => Ok(()),
+        // Match some common errors and make them more user friendly
+        Err(e) => match &e {
+            lettre::transport::smtp::Error::Client(_) => {
+                err!(format!("SMTP Client error: {}", e));
```
dani-garcia, line 338: Could the internal parameters from these errors be added directly to the error message instead? Seems like printing the wrapper error won't provide more info for any cases other than the fallback

BlackDex:
Without this, the errors returned via json for example will only show SmtpError amd nothing more.

dani-garcia:
Right but I mean do something like this instead
```rust
lettre::transport::smtp::Error::Client(x) => {
                err!(format!("SMTP Client error: {}", x));
```

Didn't check what those values are so might not make sense at all.

BlackDex:
Ah! Ok, yea i think that won't be an issue, ill change that in a few and force push it.

BlackDex:
I will test them anyway.

BlackDex:
Changed the code now.
I was also able to shorten the Transient and Permanent error messages :).

PR #1146, opened 2020-09-19
Enhanced user and orgs tables in admin view.

BlackDex:
As requested here: https://bitwardenrs.discourse.group/t/searchable-user-list-on-admin-panel/299

- Changed the table layout a bit.
- Added functions to the tables:
  + Search
  + Sort
  + Paginate


dani-garcia:
Sounds good to me but let's swap the js and css files to the non minified versions, the gains from minifying them are negligible when talking about the hardly used admin page, plus it will make it considerably easier to check their integrity.

BlackDex:
That is a good one, didn't thought about it that way.
Changed the files and code to match the non min versions.

I did however changed jquery to the non-compressed slim one, which work also just fine and does saves a lot.

PR #1151, opened 2020-09-21
alpine arm dockerfile

dr-bonez:
also adds catatonit for sigterm handling and strips the binary

### DIFF
--- a/docker/Dockerfile.j2
+++ b/docker/Dockerfile.j2
@@ -9,7 +9,7 @@
 {%     set package_arch_name = "" %}
 {%   elif "arm32v7" in target_file %}
 {%     set build_stage_base_image = "messense/rust-musl-cross:armv7-musleabihf" %}
-{%     set runtime_stage_base_image = "cmosh/alpine-arm" %}
+{%     set runtime_stage_base_image = "cmosh/alpine-arm:3.6" %}
 {%     set package_arch_name = "" %}
 {%   endif %}
 {% elif "amd64" in target_file %}
--- a/docker/Dockerfile.j2
+++ b/docker/Dockerfile.j2
@@ -3,9 +3,15 @@
 
 {% set build_stage_base_image = "rust:1.45" %}
 {% if "alpine" in target_file %}
-{%   set build_stage_base_image = "clux/muslrust:nightly-2020-07-09" %}
-{%   set runtime_stage_base_image = "alpine:3.12" %}
-{%   set package_arch_name = "" %}
+{%   if "amd64" in target_file %}
+{%     set build_stage_base_image = "clux/muslrust:nightly-2020-07-09" %}
+{%     set runtime_stage_base_image = "alpine:3.12" %}
+{%     set package_arch_name = "" %}
+{%   elif "arm32v7" in target_file %}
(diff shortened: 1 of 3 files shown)


docker/Dockerfile.j2
```diff
@@ -228,6 +248,9 @@ RUN apt-get update && apt-get install -y \
 {%   endif %}
     && rm -rf /var/lib/apt/lists/*
 {% endif %}
+{% if "alpine" in target_file and "arm32v7" in target_file %}
+RUN apk add --no-cache -X http://dl-cdn.alpinelinux.org/alpine/edge/testing catatonit
```
jjlin, line 252: I don't think this should be added, as it special-cases the Alpine arm32v7 build for no obvious reason. If you really want this, you can use `docker run --init ...`.

dr-bonez: I talked to @dani-garcia about this in matrix, he said to try it for only this build target until signal handling is managed in the binary itself.

docker/Dockerfile.j2
```diff
@@ -3,9 +3,15 @@
...
-{%   set build_stage_base_image = "clux/muslrust:nightly-2020-07-09" %}
-{%   set runtime_stage_base_image = "alpine:3.12" %}
-{%   set package_arch_name = "" %}
+{%   if "amd64" in target_file %}
+{%     set build_stage_base_image = "clux/muslrust:nightly-2020-07-09" %}
+{%     set runtime_stage_base_image = "alpine:3.12" %}
+{%     set package_arch_name = "" %}
+{%   elif "arm32v7" in target_file %}
+{%     set build_stage_base_image = "messense/rust-musl-cross:armv7-musleabihf" %}
+{%     set runtime_stage_base_image = "cmosh/alpine-arm" %}
```
dani-garcia, line 12:
```suggestion
{%     set runtime_stage_base_image = "cmosh/alpine-arm:3.6" %}
```

Let's add the version number like the others to try to make the images as reproducible as possible

dani-garcia (changes requested):
Sounds good to me, add the arm arch to this file so the images get built by docker too:
https://github.com/dani-garcia/bitwarden_rs/blob/master/hooks/arches.sh#L24

About the init thing, ultimately I want it handled by the binary and that's happening when async rocket gets released, but I don't know when that's happening. My initial idea was to just add it to this one as a test then potentially have something similar in the others, just a small init to handle the shutdown signal, but I can see how a short term solution like that might not be worth it when you can do the --init command like @jjlin says. Maybe it could be handled by the start script (https://github.com/dani-garcia/bitwarden_rs/blob/master/docker/start.sh)?

dr-bonez:
We could use `tini` instead of `catatonit` since evidently it is what docker recommends, but imo the callee shouldn't need to have knowledge as to whether the container has a proper init system in order to call it correctly, so I'm not a fan of the `--init` solution.

dani-garcia:
At the end of the day, I'm surprised we haven't gotten more complaints about the way we handle sigterm, and I believe we should add something similar to the rest of the images, I think testing it in this new image is a good way to see how well it could work on the rest, and we can change them once we see this one working out (or we can leave them as is until rocket 0.5 is released and handles this)

PR #1160, opened 2020-09-25
Add `vendored_openssl` feature.

eduardosm:
This feature enables the `vendored` feature from the `openssl` crate and build a statically linked version of openssl.

With this feature, I am able to build a fully static binary with [cross](https://github.com/rust-embedded/cross):
```
cross +nightly build --target x86_64-unknown-linux-musl --features "vendored_openssl sqlite" --release
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1553,6 +1553,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "77af24da69f9d9341038eba93a073b1fdaaa1b788221b00a69bce9e762cb32de"
 
 [[package]]
+name = "openssl-src"
+version = "111.11.0+1.1.1h"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


BlackDex:
What is the reason you want a statically build sqlite version using musl?

eduardosm:
> What is the reason you want a statically build sqlite version using musl?

For the same reason you may want a statically built version of any other executable.

A possible use is to build a binary that runs on any distribution without symbol versioning issues.

BlackDex:
Sorry, i ment more, did you also checked if this works with postgresql and mysql, since we build multi db for a while now

eduardosm:
mysql and postgresql depend on the mysqlclient-sys and pq-sys, which do not offer vendoring the C library, so it wouldn't work with mysql or postgresql unless you manage to get a musl static build of the required C libraries.

dani-garcia:
Yeah I don't expect mysql or postgresql libs to be able to be compiled statically (at least not easily), so it might not make much sense to allow to use this feature with either the postgres or mysql features, but then again it doesn't break anything to have openssl statically compiled and the others dinamically compiled if the user wants it like that.

Issue #1161, opened 2020-09-26
Store built binaries in the release assets


GerardoGR:
For convenience it would be nice to have the pre-built binaries in the Github release assets.

I've seen that there is **Release** github action for uploading the asset, but this doesn't get triggered:

```
    - name: Release
      uses: Shopify/upload-to-release@1.0.0
      if: startsWith(github.ref, 'refs/tags/')
      with:
        name: bitwarden_rs-${{ matrix.db-backend }}-${{ matrix.target }}${{ matrix.ext }}
        path: target/${{ matrix.target }}/release/bitwarden_rs${{ matrix.ext }}
        repo-token: ${{ secrets.GITHUB_TOKEN }}
```

PR #1154, opened 2020-09-24
Add bitwarden_rs helm-chart

Ornias1993:
Due to much community request:
A Helm-Chart (based on the work from GissiLabs), that includes all possible options for bitwarden_rs.

To be sure it doesn't need too much babysitting: it has it's own CI already setup and ready to go.
CI tests both pushes and pullrequests. It does both a Lint test and actual install test.

Releases for the charts would be on every commit pushed to master, with [release chart] in the title.
(this prevents accidental releases overwriting the current version and such.)

Future TODO's:
- Add a chart-version check to the CI and update if chart version is updated

**Requirements:**
Needs github pages to be setup and linked to a branch called gh-pages. (before being merged)
No actual website will be published, just a webspot to serve the packaged helmcharts.

**TLDR:**
- Adds a "contrib" directory for community contributions (that are not directly integrated into bitwarden_rs)
- Adds a helm-chart for the project
- Adds Test and Releasing CI for the helm chart
- Some cleanup

Co-authored-by: gissilabs
Signed-off-by: Kjeld Schouten-Lebbing <kjeld@schouten-lebbing.nl>

PR #1170, opened 2020-10-04
Uploading binaries to pre-release on tag creation

GerardoGR:
I would like to have the binaries compiled for the released version in the release assets section. This PR introduces that and also creates a pre-release on tag creation. This a proposal to solve an issue I created in GitHub #1161 and asked in the matrix channel.

### DIFF
--- a/.github/workflows/workspace.yml
+++ b/.github/workflows/workspace.yml
@@ -4,6 +4,10 @@ on:
   push:
     paths-ignore:
       - "**.md"
+    branches:
+      - "*"
+    tags:
+      - "*"
   #pull_request:
   #  paths-ignore:
   #    - "**.md"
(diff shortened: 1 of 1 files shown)


Ornias1993:
Yes this would be very nice and I could also hook into this when we get #1154 merged. :)

.github/workflows/workspace.yml
```diff
@@ -139,10 +143,66 @@ jobs:
...
+  upload-release-assets:
+    name: Upload release assets
+    if: startsWith(github.ref, 'refs/tags/')
+    needs: [build, prepare-release]
+    strategy:
+      fail-fast: false
+      # TODO: YAML Anchors could reduce the matrix repetition but they are not
+      # supported https://github.community/t/support-for-yaml-anchors/16128
+      matrix: 
+        db-backend: [sqlite, mysql, postgresql]
```
Ornias1993, line 174: I think the use of matrix is fine, not super efficient but it isn't running that often anyway :)

.github/workflows/workspace.yml
```diff
@@ -139,10 +143,66 @@ jobs:
...
+        include:
+          - target: x86_64-unknown-linux-gnu
+            os: ubuntu-latest
+            ext:
+          # - target: x86_64-unknown-linux-musl
+          #   os: ubuntu-latest
+          #   ext:
+          # - target: x86_64-apple-darwin
+          #   os: macOS-latest
+          #   ext:
```
Ornias1993, line 189: On a later date we should look into packaging this into a nice dmg or something.
But thats a bit offtopic.

.github/workflows/workspace.yml
```diff
@@ -4,6 +4,10 @@ on:
   push:
     paths-ignore:
       - "**.md"
+    branches:
+      - "*"
```
Ornias1993, line 8: I don't think branches* is needed with tags*

Ornias1993 (approved):
Generally fine, one slight footnote.

BlackDex:
The issue i have with this is that what is build isn't a static binary.
You need the correct library versions installed on your system else this will not work, or cause unexpected issues.
This is where the docker containers are great in by providing these exact same libraries as where the binary is build with.

If we want to provide this, i think we should mention which OS was used to build it on.
So, having a musl static build version for any OS would get a clear go from my side, but in my experience this is only going to cause confusion and weird issue we will get reported here.

I don't know how @dani-garcia or @jjlin think about this, but for me a musl version would be ok, dynamic ones no.

jjlin:
I agree that for this to be broadly useful, the binaries would have to be statically linked. Or there would have to be separate builds uploaded for each distro/version being targeted, which could get a bit messy.

BlackDex:
@GerardoGR, if you are able to get static binaries build using gitlab actions i would be happy to say yes to this.
But for now, i think we should close this PR until it is possible to have static builds which would work on almost all Linux distributions.

GerardoGR:
@BlackDex I will have a look on how to build the static binaries. If it is fine for you, I will close this PR in the meanwhile.

PR #1171, opened 2020-10-05
Fixed building mysql, postgresql and sqlite3 for arm

### LINKED ISSUE #530: MySQL images for Alpine and ARM devices not building correctly
Hey all,

I run successfully `bitwardenrs/server:latest`. I wanted to add MySQL and went on trying `bitwardenrs/server-mysql:latest` and `bitwardenrs/server-mysql:alpine`, to no avail.

The errors are `"standard_init_linux.go:207: exec user process caused "no such file or directory"` in the case of alpine, and in the other case I can't gather much because my log is truncated. I will try to find better log, but meanwhile in the case of `alpine` it seems that something was possibly renamed and forgot to update all the occurences

Thank you

### LINKED ISSUE #1066: bitwardenrs/server-mysql is not a multiarch image
As announced in the description for [Release 1.16](https://github.com/dani-garcia/bitwarden_rs/releases/tag/1.16.0), the image server-mysql and server-postgresql should be a multiarch image. This is not the case:

image

Am I misunderstanding this?

bitwardenrs/server is multiarch and runs fine on arm7

BlackDex:
With some apt/dpkg magic building multidb containers for arm versions
now also works. As long as the build stage and docker-image stage use
the same base (debian buster now) it should all work.

Resolves #530, resolves #1066

### DIFF
--- a/docker/Dockerfile.j2
+++ b/docker/Dockerfile.j2
@@ -35,11 +35,11 @@
 ####################### VAULT BUILD IMAGE  #######################
 {% set vault_image_hash = "sha256:e40228f94cead5e50af6575fb39850a002dad146dab6836e5da5663e6d214303" %}
 {% raw %}
-#  This hash is extracted from the docker web-vault builds and it's prefered over a simple tag because it's immutable.
+#  This hash is extracted from the docker web-vault builds and it's preferred over a simple tag because it's immutable.
 #  It can be viewed in multiple ways:
 #  - From the https://hub.docker.com/repository/docker/bitwardenrs/web-vault/tags page, click the tag name and the digest should be there.
 #  - From the console, with the following commands:
-#      docker pull bitwardenrs/web-vault:v2.16.0b
+#      docker pull bitwardenrs/web-vault:v2.16.1
 #      docker image inspect --format "{{.RepoDigests}}" bitwardenrs/web-vault:v2.16.1
 #
 #  - To do the opposite, and get the tag from the hash, you can do:
@@ -54,13 +54,9 @@ FROM {{ build_stage_base_image }} as build
 # Alpine only works on SQlite
 ARG DB=sqlite
 
-{% elif "amd64" in target_file %}
-# AMD64 supports all
-ARG DB=sqlite,mysql,postgresql
-
 {% else %}
-# ARM only supports SQLite for now
-ARG DB=sqlite
+# Debian-based builds support multidb
+ARG DB=sqlite,mysql,postgresql
 
 {% endif %}
 # Build time options to avoid dpkg warnings and help with reproducible builds.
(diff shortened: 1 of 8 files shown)


docker/Dockerfile.j2
```diff
@@ -75,14 +71,19 @@ ENV RUSTFLAGS='-C link-arg=-s'
...
         /etc/apt/sources.list.d/deb-src.list \
     && dpkg --add-architecture {{ package_arch_name }} \
     && apt-get update \
     && apt-get install -y \
         --no-install-recommends \
         libssl-dev{{ package_arch_prefix }} \
-        libc6-dev{{ package_arch_prefix }}
+        libc6-dev{{ package_arch_prefix }} \
+        libpq5{{ package_arch_prefix }} \
+        libpq-dev \
```
jjlin, line 84: Does it not work if you use `{{ package_arch_prefix }}` here? That's what provides the `libpq.so` symlink missing below.

docker/Dockerfile.j2
```diff
@@ -54,13 +54,9 @@ FROM {{ build_stage_base_image }} as build
...
 ARG DB=sqlite
 
-{% elif "amd64" in target_file %}
-# AMD64 supports all
-ARG DB=sqlite,mysql,postgresql
-
 {% else %}
-# ARM only supports SQLite for now
-ARG DB=sqlite
+# AMD64 and ARM both support multidb
```
jjlin, line 58: This might be better described as `Debian-based builds support multidb`

docker/Dockerfile.j2
```diff
@@ -75,14 +71,19 @@ ENV RUSTFLAGS='-C link-arg=-s'
 
 {% elif "arm" in target_file %}
 # Install required build libs for {{ package_arch_name }} architecture.
+# To compile both mysql and posgresql we need some extra packages for both host arch and target arch
```
jjlin, line 74: `posgresql` is misspelled

docker/Dockerfile.j2
BlackDex: No, unfortunately not.
I actually need both, and diesel-rs needs the host-arch version of pg_config for some reason to work.
Since libpq5 has the correct library, but not the symlink this was the only solution.

docker/Dockerfile.j2
BlackDex: Indeed, i will adapt it.

docker/Dockerfile.j2
BlackDex: Woeps.

docker/Dockerfile.j2
BlackDex: I have added some extra comment at the line where i create the symlink.

BlackDex:
@jjlin Changed your findings, and updated a comment line to clarify the reason of the symlink.

dani-garcia:
Okay, LGTM but gonna need to modify this file as well:

https://github.com/dani-garcia/bitwarden_rs/blob/master/hooks/arches.sh#L12-L19

That entire block is unneeded now, I think

BlackDex:
@dani-garcia indeed, missed that one. Fixed now.

PR #1172, opened 2020-10-06
Updated .env.template with missing config values.

BlackDex:
Several values were missing from the .env.template file.
Updated this with all the values which were missing.


sgissi:
@BlackDex I assume these variables are already working on 1.16.3? If so, I will update the Helm chart.

BlackDex:
@sgissi these are already working for a few versions back but never were added to the template.

PR #1173, opened 2020-10-06
Set DB Max pool size via environment variables

CollectiveMakers:
Tiny changes to set DB maximum pool size via environment variable (DB_POOL_MAX_SIZE). Usefull on small test infrastructures with low connection limits to database

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -350,6 +350,9 @@ make_config! {
         /// Max database connection retries |> Number of times to retry the database connection during startup, with 1 second between each retry, set to 0 to retry indefinitely
         db_connection_retries:  u32,    false,  def,    15;
 
+        /// Max database pool size |> Number of simultaneous connections to the database
+        db_pool_max_size:       u32,    true,   def,    10;
+        
         /// Bypass admin page security (Know the risks!) |> Disables the Admin Token for the admin page so you may use your own auth in-front
         disable_admin_token:    bool,   true,   def,    false;
 
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -44,14 +44,14 @@ macro_rules! generate_connections {
             pub fn from_config() -> Result<Self, Error> {
                 let url = CONFIG.database_url();
                 let conn_type = DbConnType::from_url(&url)?;
-
+                let max_size = CONFIG.db_pool_max_size();
                 match conn_type { $(
                     DbConnType::$name => {
                         #[cfg($name)]
                         {
                             paste::paste!{ [< $name _migrations >]::run_migrations()?; }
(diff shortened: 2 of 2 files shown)


BlackDex:
Those docker changes i can't approve since these should be the defaults. You can already change them via `-e PORT` etc... During docker run.

What is the actual reason this needs to be in here?

CollectiveMakers:
First reason is to allow direct use of dockerfiles on Cloud platforms using same environment variables at all stages (build and run)
Second reason is that binding tcp port numbers lower than 1024 doesn’t allow to run processes as non root user (which is recommended in a security perspective)


Cordialement.

Romain LORENZINI
________________________________
De : Mathijs van Veluw <notifications@github.com>
Envoyé : Tuesday, October 6, 2020 9:06:40 PM
À : dani-garcia/bitwarden_rs <bitwarden_rs@noreply.github.com>
Cc : Romain LORENZINI <romain@cmakers.io>; Author <author@noreply.github.com>
Objet : Re: [dani-garcia/bitwarden_rs] Set TCP Ports and DB Max pool size via environment variables (#1173)


Those docker changes i can't approve since these should be the defaults. You can already change them via -e PORT etc... During docker run.

What is the actual reason this needs to be in here?

BlackDex:
Ah, that seems plausible usage indeed.
Well the defaults won't change so I think it this will not harm anything else.

jjlin:
This doesn't let you the same variables during build and run phases. For example, you would use `docker build --build-arg PORT=8080` to build an image where `ROCKET_PORT` defaults to 8080, but `docker run -e ROCKET_PORT=8081` to override it when starting a new container (`docker run -e PORT=8081` wouldn't work).

I think something like
```
ARG ROCKET_PORT=80
ENV ROCKET_PORT=$ROCKET_PORT
```
would be better.

Also, there are already a couple of open PRs for the DB pool size. That's a separate issue and should be a separate PR.

BlackDex:
Agreed on the separate PR's indeed.

And the var should be the same indeed on all levels.

CollectiveMakers:
Ok changed this PR as a dedicated one for BD pool size and I will create another one for TCP Ports.
I saw the other PR related to DB Pool size but I think this one suits more existing code structure.

BlackDex:
That looks nice, but i think we also need some input validation.
For example, it must be 1 or higher. 0 will cause a panic.
And maybe a max amount, but that could be arbitrary.

jjlin:
This looks like pretty much the same implementation as #1108

BlackDex:
Going to close this one since it's a duplicate of #1108, and that one has some extra validity checks in it.

Thanks though for the PR!

PR #1175, opened 2020-10-06
Set tcp ports via environment variables in Dockerfile

CollectiveMakers:
First reason is to allow direct use of dockerfiles on Cloud platforms using same environment variables at all stages (build and run)
Second reason is that binding tcp port numbers lower than 1024 doesn’t allow to run processes as non root user (which is recommended in a security perspective)


jjlin:
I think you'll also need `ENV WEBSOCKET_PORT=$WEBSOCKET_PORT`

BlackDex:
Indeed, else the default isn't there anymore

jjlin:
The default would still be there, but `docker build --build-arg WEBSOCKET_PORT=1234` wouldn't change the default from 3012 to 1234

BlackDex:
Indeed, it was only an arg yes. That will not pass on to the container.

BlackDex:
Regarding the binding on a lower port, that is where `ROCKET_PORT` is for so that you can change this using the `ENV` values.
I actually do not see the benefits of having the image built with this _default_ changed and still having the ENV value used to set the port's anyway.

Also, there are many _Read Only_ configuration options which still need to be set using either a `.env` file or `-e ENV_VAR`.
If you are stating to want to have them the same across build and run, then you should probably add them all during build also.

Maybe i am a bit harsh on this, but i think it doesn't have any real added value by just using `ROCKET_PORT` and `WEBSOCKET_PORT` during the `docker run` command if it was build with other values or not and still use `-e ROCKET_PORT=` with the same value in the end.

PR #1178, opened 2020-10-08
Updated the azure-pipelines.yml for multidb

BlackDex:
Updated the azure-pipelines.yml to build multidb now.
- Updated to Ubuntu 18.04 (Closer matches the docker builds)
- Added some really needed apt packages to be sure that they are
  installed
- Now run cargo test using all database backeds in one go.


dani-garcia:
LGTM, should we change the Github Actions as well?
https://github.com/dani-garcia/bitwarden_rs/blob/master/.github/workflows/workspace.yml

PR #1212, opened 2020-10-31
Added error handling during dotenv loading

### LINKED ISSUE #1196: Handle spaces in unquoted env variables
### Subject of the issue
Spaces in an unquoted env variable value will stop processing every following env var which may result in various errors.

### Your environment
* Bitwarden_rs version: 1.17.0
* Install method: Custom Docker Image

### Steps to reproduce
Add a space to an env var value:

```
SMTP_FROM_NAME=Bitwarden RS                
SMTP_USERNAME=relay@example.com       
SMTP_PASSWORD=1234567                       
SMTP_AUTH_MECHANISM=plain                   
SMTP_TIMEOUT=30
```

Check the loaded values in the admin dashboard. All settings after `SMTP_FROM_NAME` will not be applied anymore. Quoting the value will fix the issue e.g. `SMTP_FROM_NAME="Bitwarden RS"`.

### Expected behaviour
Some kind of improved error handling. Add a variable parsing error message to the log, or fix it automatically and strip off everything after a space. Maybe we could still have a warning in the logs that var `EXAMPLE` was stripped.

BlackDex:
Some issue people report are because of misconfiguration or bad .env
files. To mittigate this i added error handling for this.

- Panic/Quit on a LineParse error, which indicates bad .env file format.
- Emits a info message when there is no .env file found.
- Emits a warning message when there is a .env file, but not no permissions.
- Emits a warning on every other message not specifically catched.

This partially fixes #1196 by reporting the line regarding spaces.
Solving that ticket by allowing spaces would be difficult from within
bitwarden_rs it self.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -53,7 +53,32 @@ macro_rules! make_config {
 
         impl ConfigBuilder {
             fn from_env() -> Self {
-                dotenv::from_path(".env").ok();
+                match dotenv::from_path(".env") {
+                    Ok(_) => (),
+                    Err(e) => match e {
+                        dotenv::Error::LineParse(msg, pos) => {
+                            panic!("Error loading the .env file:\nNear {:?} on position {}\nPlease fix and restart!\n", msg, pos);
+                        },
+                        dotenv::Error::Io(ioerr) => match ioerr.kind() {
+                            std::io::ErrorKind::NotFound => {
+                                println!("[INFO] No .env file found.\n");
+                                ()
+                            },
+                            std::io::ErrorKind::PermissionDenied => {
+                                println!("[WARNING] Permission Denied while trying to read the .env file!\n");
+                                ()
+                            },
+                            _ => {
+                                println!("[WARNING] Reading the .env file failed:\n{:?}\n", ioerr);
+                                ()
+                            }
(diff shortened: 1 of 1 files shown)


dani-garcia:
Somehow this simple PR passed by me unnoticed lol.

Related upstream issue: https://github.com/dotenv-rs/dotenv/issues/38.

Checking the dotenv error format, I think we should handle them differently: https://github.com/dotenv-rs/dotenv/blob/master/dotenv/src/errors.rs#L9

An IO error is probably caused by a non existing file and a warning there is fine, but in case of a parse error, i think we should error out and quit, because that's always a mistake by the admin.

BlackDex:
You are right indeed. That would be the best case. I just didn't want to break active configs which could have an issue like using a `;` in the beginnen of a line as comment even though it does break the dotenv loading any further passed that line.

But they are faulty so panic/error out would be better.
I will change it so it will error out on those specific errors.

Could take a while, I'm on holiday now :)

BlackDex:
@dani-garcia i have updated the PR to reflect your comments.
This also gives the users a bit more useful information :).

dani-garcia:
Looking great!

PR #1220, opened 2020-11-10
Return 404 instead of fallback icon

jameshurst:
Returning a 404 not found error instead of the fallback icon allows clients to use their own icon instead (which may be higher quality than the fallback icon). This behaviour matches that of the official icons server as well (https://github.com/bitwarden/server/blob/master/src/Icons/Controllers/IconsController.cs#L89-L93).


src/api/icons.rs
```diff
@@ -190,37 +188,37 @@ fn check_icon_domain_is_blacklisted(domain: &str) -> bool {
...
+            None
         }
     }
 }
 
 fn get_cached_icon(path: &str) -> Option<Vec<u8>> {
     // Check for expiration of negatively cached copy
     if icon_is_negcached(path) {
-        return Some(FALLBACK_ICON.to_vec());
+        return None;
```
jjlin, line 221: I think returning `None` here effectively breaks negative caching under the current `get_icon()` logic

jameshurst: Ah yep you're right. The `icon_is_negcached` check could be moved up to `get_icon` which would keep the expected behaviour.

BlackDex:
Seems good to me.

BlackDex (approved):

dani-garcia:
Thanks!

PR #1223, opened 2020-11-12
Support Manager role

### LINKED ISSUE #1136: Org managers can't create or manage collections
According to https://bitwarden.com/help/article/user-types-access-control/#user-types, org managers should be able to:

* Only administer a collection it is associated with by an Owner or Admin
* Access and manage assigned collections in an organization
* Create new collections and modify the assigned collections
* Set user access for assigned collections

Attempting to create a new collection via the web vault results in

Screenshot 2020-09-12 003122

and log messages

```
[2020-09-12 00:29:38.993][request][INFO] POST /api/organizations/3c8806a6-2359-4c6c-b769-3fe52dab91bb/collections
[2020-09-12 00:29:38.995][auth][ERROR] Unauthorized Error: You need to be Admin or Owner to call this endpoint
[2020-09-12 00:29:38.995][response][INFO] POST /api/organizations/<org_id>/collections (post_organization_collections) => 401 Unauthorized
```
Attempting to manage an existing collection via the web vault results in the `Your login session has expired` UI message, and log messages
```
[2020-09-12 00:31:57.174][request][INFO] GET /api/organizations/3c8806a6-2359-4c6c-b769-3fe52dab91bb/collections/320a942c-d232-4273-96e5-d2c872f8d1d8/details
[2020-09-12 00:31:57.175][auth][ERROR] Unauthorized Error: You need to be Admin or Owner to call this endpoint
[2020-09-12 00:31:57.175][response][INFO] GET /api/organizations/<org_id>/collections/<coll_id>/details (get_org_collection_detail) => 401 Unauthorized
```

(This issue was first reported at https://bitwardenrs.discourse.group/t/manager-can-not-create-a-collection/287.)

mvalois:
Wanna fix #1136
This is my first time with Rust. Feel free to ask for changes (especially for error handling).
Maybe the function `check_manager_in_collection` can be moved to a better spot.

### DIFF
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -5,7 +5,7 @@ use serde_json::Value;
 
 use crate::{
     api::{EmptyResult, JsonResult, JsonUpcase, JsonUpcaseVec, Notify, NumberOrString, PasswordData, UpdateType},
-    auth::{decode_invite, AdminHeaders, Headers, OwnerHeaders},
+    auth::{decode_invite, AdminHeaders, Headers, OwnerHeaders, ManagerHeaders},
     db::{models::*, DbConn},
     mail, CONFIG,
 };
@@ -217,7 +217,7 @@ fn get_org_collections(org_id: String, _headers: AdminHeaders, conn: DbConn) -> 
 #[post("/organizations/<org_id>/collections", data = "<data>")]
 fn post_organization_collections(
     org_id: String,
-    _headers: AdminHeaders,
+    headers: ManagerHeaders,
     data: JsonUpcase<NewCollectionData>,
     conn: DbConn,
 ) -> JsonResult {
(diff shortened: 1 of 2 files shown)


BlackDex:
This is just a copy of the recently closed PR #1222 which noted that there probably should be a better way to do this.
And i think there is. These checks should not be done per function. That is where the ManagerHeader should be for.
So that check should end-up somewhere in `auth.rs` in my opinion. And emit a error there if the user doesn't have the right permissions. Though this can be tricky if we do not have all the information.

mvalois:
You mean it is possible not to add any extra line in `collections.rs` except changing `headers` to `ManagerHeaders` ?

BlackDex:
Yes, there is. You probably need to do these checks in auth.rs, and when thinking about it and checking your changes, there are also some cases where you have access as a Manager, but there is no collection_id provided. But those items you want to have checked before they enter the function at all i think.

Also, i see some other issues like always saving the CollectionUser, which should not be done when the user as access to all collections. And i'm missing some function, like changing a collection name.

BlackDex:
Going to close this PR as #1242 has been merged :).
Thanks for the PR, and making it clear this was a needed feature.

mvalois:
Thank you for your time!

PR #1229, opened 2020-11-18
Email fixes

### LINKED ISSUE #1077: Apple SMTP certificate validation error
### Subject of the issue
Apple SMTP server cannot be used because of certificate validation error.

### Your environment
* Bitwarden_rs version: 1.15.1-1e950c7d
* Install method: docker

### Steps to reproduce
Used the settings as described here (with and without TLS) https://support.apple.com/en-us/HT202304
Not much room for error.
### Relevant logs
Screenshot 2020-08-01 at 16 14 14

### LINKED ISSUE #1207: improve the sending of emails
### improve the sending of emails

I am using bitwarden_rs with my email server to send notification. Currently my rspamd adds the spam header because the mail had no message id. A value of 6 is marked as spam in the default rspamd configuration.
Is there a way to improve this behavior?

```
MISSING_MID (2.5)                         Message id is missing
HTML_SHORT_LINK_IMG_1 (2)                 Short html part (0..1K) with a link to an image
MIME_BASE64_TEXT_BOGUS (1)                Has text part encoded in base64 that does not contain any 8bit characters
URI_COUNT_ODD (1) [3]                     Odd number of URIs in multipart/alternative message
MIME_GOOD (-0.1) [multipart/alternative]  Known Content Type
MIME_BASE64_TEXT (0.1)                    Has Text Part encoded in base64
```

BlackDex:
### Updated email processing.
- Added an option to enable smtp debugging via SMTP_DEBUG. This will trigger a trace of the smtp commands sent/received to/from the mail server. Useful when troubleshooting.
- Added two options to ignore invalid certificates which either do not match at all, or only doesn't match the hostname. ( fixes #1077 )
- Added Message-ID Header, which triggered spamfilters when absent. ( fixes #1207 )
- Updated lettre to the latest alpha.4 version.

### Fixed plain/text email format
plain/text emails should not contain html elements like <p> <a> etc..
This triggers some spamfilters and increases the spam score.

Also added the github link into the text only emails since this also triggers spamfilters to increase the score since the url/link count is different between the multipart messages.

Tested by using https://www.mail-tester.com/ to validate the spamscore. (Including the missing Message-ID Header mentioned above)

### Updated admin settings page.
- Added check if settings are changed but not saved when sending test email.
- Added some styling to emphasize some risks settings.
- Fixed alignment of elements when the label has multiple lines.


dani-garcia:
Thanks!

PR #1235, opened 2020-11-22
Allow setting SMTP auth password from file via SMTP_PASSWORD_FILE

### LINKED ISSUE #1194: Docker Secret on SMTP_PASSWORD
Is it possible to get a docker secret feature on (SMTP_PASSWORD=supersecretpassword) so it's not exposed?

for example: SMTP_PASSWORD_FILE=/run/secrets/supersecretpassword

Thanks!

janost:
Fixes #1194.

This change allows to user to inject SMTP password from a file through the `SMTP_PASSWORD_FILE` variable.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,3 +1,4 @@
+use std::fs;
 use std::process::exit;
 use std::sync::RwLock;
 
@@ -430,6 +431,10 @@ make_config! {
         smtp_username:                 String, true,   option;
         /// Password
         smtp_password:                 Pass,   true,   option;
+        /// Password file
+        smtp_password_file:            String, true,   option;
+        /// Field to store the actual SMTP password
+        smtp_password_used:            String, false,  gen,      |c| decide_smtp_password(&c.smtp_password, &c.smtp_password_file);
         /// SMTP Auth mechanism |> Defaults for SSL is "Plain" and "Login" and nothing for Non-SSL connections. Possible values: ["Plain", "Login", "Xoauth2"]. Multiple options need to be separated by a comma ','.
         smtp_auth_mechanism:           String, true,   option;
         /// SMTP connection timeout |> Number of seconds when to stop trying to connect to the SMTP server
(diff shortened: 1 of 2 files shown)


BlackDex:
Hello @janost  I do not like the extra config setting with the _file.
For one because there are multiple other settings which could be preferred to be in a file and that would add more complexity.

And it looks like there is no error anymore when there is no password set at all anymore. Only if both are set.

I don't know what @dani-garcia thinks about this.
But maybe it's is an idea to use a prefix it something like `file://` detect that and then load the string from that file. That could be used for more then only smtp, and it does not add an extra _file setting. Though we probably still need to use something like the _used unfortunately.

These were just some quick thoughts.

janost:
Hey @BlackDex thanks for the review, valid points. I'll look into them as soon as I can.

dani-garcia:
I'd like to avoid adding extra individual alternatives, but potentially we could modify the config loading to add an extra _FILE variable for each option in a generic way, and handle all the same way here:

https://github.com/dani-garcia/bitwarden_rs/blob/5379329ef71ba4b600eb028d721806a16e62535f/src/util.rs#L286-L302

BlackDex:
> I'd like to avoid adding extra individual alternatives, but potentially we could modify the config loading to add an extra _FILE variable for each option in a generic way, and handle all the same way here:
> 
> https://github.com/dani-garcia/bitwarden_rs/blob/5379329ef71ba4b600eb028d721806a16e62535f/src/util.rs#L286-L302

Oe, that is a nice one indeed. That would also make the docker-secrets feature available when enabled for all items :).

BlackDex:
We only need to think of what to show in the admin settings and if we even want to allow these settings to be viewed/changed via the web-interface at all because it's all in a file. It would be more sane to have a reload function or something to re-read those _FILE configs.

janost:
Better implementation can be found in #1244.

Closing.

PR #1236, opened 2020-11-23
feat: add email bcc option

RainshawGao:
As some mail service providers do not store mails sent by smtp into 'Sent' folder automatically, and some other providers such as Gmail do so. This option can make the mails bcc to the specific address, and is a compromise solution for those providers which can not store the smtp mails.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -426,6 +426,10 @@ make_config! {
         smtp_from:                     String, true,   def,     String::new();
         /// From Name
         smtp_from_name:                String, true,   def,     "Bitwarden_RS".to_string();
+        /// Enable BCC
+        smtp_bcc:                      bool,   true,   def,     false;
+        /// BCC Address
+        smtp_bcc_address:              String, true,   option;
         /// Username
         smtp_username:                 String, true,   option;
         /// Password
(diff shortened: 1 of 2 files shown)


BlackDex:
I have a bit of mixed feelings with this.
It can also be used to easily snoop password hints if this settings could be changed somehow (and you do not have access to the database).

I Also wonder why you want to have to store these sent emails at all, since they are only intended for the the receiver and not the sender to be kept somewhere.

If  you really want this, i would suggest to use something like postfix's always_bcc function or something.
http://www.postfix.org/postconf.5.html#always_bcc

I think this is not something we want to add to bitwarden_rs it self.

RainshawGao:
I know what you mean. Regarding security issues, I think it is the responsibility of the administrator to manage various passwords and configuration files to prevent leakage of password hints. The main purpose of saving and sending letters is to unify the behaviors of various service providers. Adding this option does not mean it must be enabled, the administrator can decide for himself.
I don't know where to use always_bcc, is it in docker? Or on the host? I did not find the /etc/postfix/main.cf file in the docker container

dani-garcia:
I'll leave the final decision on whether to merge this to @BlackDex, because I'm not very well versed in email stuff in the end, but at least code wise, if we were to merge this, I'd remove the smtp_bcc option and just leave the smtp_bcc_address one.

BlackDex:
I agree with @dani-garcia regarding that extra enable option.

But, my other concern is still that we should not try to "fix" other smtp servers which do not save sent items into a sent folder.
Also, i want to try and keep stuff like this to match bitwarden upstream it self, it doesn't support this (except for there own test environment).

Regarding the `always_bcc` that is a feature of the postfix mail server, that is not within the docker image of bitwarden_rs.
It was more an option to use if you setup your own mail server (even just for forwarding).


One other thing i do find interesting is the `idna` stuff. Have you found any cases where this was broken with just using To also?
Because, if that is the case, we should fix that at least.

RainshawGao:
using bcc is the main idea that receiver do not know about the mail has been delivered to the admin mailbox for the purpose of logging mail.
As you say, it is differ from the upstream, so i would like to keep it in my own branch, and this pr will be closed, thank you!

PR #1242, opened 2020-11-27
Adding Manager Role support

### LINKED ISSUE #1136: Org managers can't create or manage collections

BlackDex:
This has been requested a few times (#1136 & #246 & forum), and there already were two
(1:1 duplicate) PR's (#1222 & #1223) which needed some changes and no
followups or further comments unfortunally.

This PR adds two auth headers.
- ManagerHeaders
  Checks if the user-type is Manager or higher and if the manager is
part of that collection or not.
- ManagerHeadersLoose
  Check if the user-type is Manager or higher, but does not check if the
user is part of the collection, needed for a few features like
retreiving all the users of an org.

I think this is the safest way to implement this instead of having to
check this within every function which needs this manually.

Also some extra checks if a manager has access to all collections or
just a selection.

fixes #1136

### DIFF
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -5,7 +5,7 @@ use serde_json::Value;
 
 use crate::{
     api::{EmptyResult, JsonResult, JsonUpcase, JsonUpcaseVec, Notify, NumberOrString, PasswordData, UpdateType},
-    auth::{decode_invite, AdminHeaders, Headers, OwnerHeaders},
+    auth::{decode_invite, AdminHeaders, Headers, OwnerHeaders, ManagerHeaders, ManagerHeadersLoose},
     db::{models::*, DbConn},
     mail, CONFIG,
 };
@@ -217,7 +217,7 @@ fn get_org_collections(org_id: String, _headers: AdminHeaders, conn: DbConn) -> 
 #[post("/organizations/<org_id>/collections", data = "<data>")]
 fn post_organization_collections(
     org_id: String,
-    _headers: AdminHeaders,
+    headers: ManagerHeadersLoose,
     data: JsonUpcase<NewCollectionData>,
     conn: DbConn,
 ) -> JsonResult {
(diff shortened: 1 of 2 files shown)


src/api/core/organizations.rs
```diff
@@ -228,17 +228,30 @@ fn post_organization_collections(
         None => err!("Can't find organization details"),
     };
 
+    // Get the user_organization record so that we can check if the user has access to all collections.
+    let user_org = match UserOrganization::find_by_user_and_org(&headers.user.uuid, &org_id, &conn) {
```
dani-garcia, line 232: This is something that we already get in the Headers fromrequest impl, maybe we should store it too?

BlackDex: That is a good idea. I will improve this pr. Maybe also remove some subsequent same request in the functions.

mvalois:
> which needed some changes and no followups or further comments unfortunately.

Thank you for taking time to implement this need. I tried all my best, but my lack of knowledge of Rust made me rely on more experimented users. I'm building it to try it on my own, and will give you feedback.

BlackDex:
> > which needed some changes and no followups or further comments unfortunately.
> 
> Thank you for taking time to implement this need. I tried all my best, but my lack of knowledge of Rust made me rely on more experimented users. I'm building it to try it on my own, and will give you feedback.

No problem, i saw the need was there, and had some extra time left.

src/api/core/organizations.rs
BlackDex: Fixed this, i passed this on, saves a DB call :).
I didn't passed this change on to all the other functions, because i think that is out of scope for this change, and could cause other issues.

BlackDex:
@dani-garcia Thx for the note about that query already being done. Changes are applied.

blacklist-arcc:
Thanks for the great support!

PR #1243, opened 2020-11-28
Fix Key Rotation during password change.

### LINKED ISSUE #1240: Error after changing Bitwarden_rs master password and encryption
**Error after changing Bitwarden_rs master password and encryption**

Hi,

today I changed the master password of my Bitwarden_rs and I also checked the box to change the encryption.
As Bitwarden_rs announced before I was logged off after that of course. But beside the note that I was logged off, there were two red error boxes under the log out message on the upper right side of my screen as well. Both just said that an error occured. Just one sentence. No error code or additional information. Is this OK and part of the routine or not?

I tried to log in with the new master password and everything works fine as it was before. But still I am a little bit concerned that maybe something with the encryption is not right and that there could be a security issue with my password encryption. Could this be the case and how can I find out?

My environment:
My Bitwarden_rs installation runs on a Raspberry Pi 3 B in a Docker container (installed via image that I got from Docker) and with a traefik reverse proxy.
My Bitwarden_rs version is 1.17.0.
I use Bitwarden_rs with the Andoid App as well as the Chrome extension. But when the error occured I was logged in via my DDNS domain on Chrome.

I reproduced the change of the master password one more time (again with the checked box for the encryption) and the two erros again occured.

In summary I need to know if these errors are normal or not - and if not: What could be the reason and what needs to be done to make sure that my password are safe and ecrypted.

Thank you!

BlackDex:
When ticking the 'Also rotate my account's encryption key' box, the key
rotated ciphers are posted after the change of password.

During the password change the security stamp was reseted which made
the posted key's return an invalid auth. This reset is needed to prevent other clients from still being able to read/write.

This fixes this by adding a new database column which stores a stamp exception which includes the allowed route and the current security stamp before it gets reseted.
When the security stamp check fails it will check if there is a stamp exception and tries to match the route and security stamp.

Currently it only allows for one exception. But if needed we could expand it by using a `Vec<UserStampException>` and change the functions accordingly.

fixes #1240

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -215,12 +215,10 @@ pub fn generate_admin_claims() -> AdminJWTClaims {
 //
 // Bearer token authentication
 //
-use rocket::{
-    request::{FromRequest, Request, Outcome},
-};
+use rocket::request::{FromRequest, Outcome, Request};
 
 use crate::db::{
-    models::{Device, User, UserOrgStatus, UserOrgType, UserOrganization, CollectionUser},
+    models::{CollectionUser, Device, User, UserOrgStatus, UserOrgType, UserOrganization, UserStampException},
     DbConn,
 };
 
@@ -298,7 +296,25 @@ impl<'a, 'r> FromRequest<'a, 'r> for Headers {
         };
 
         if user.security_stamp != claims.sstamp {
-            err_handler!("Invalid security stamp")
+            if let Some(stamp_exception) = user
+                .stamp_exception
+                .as_deref()
+                .and_then(|s| serde_json::from_str::<UserStampException>(s).ok())
+            {
+                let current_route = match request.route().and_then(|r| r.name) {
+                    Some(name) => name,
+                    _ => err_handler!("Error getting current route for stamp exception"),
+                };
+
+                // Check if both match, if not this route is not allowed with the current security stamp.
+                if stamp_exception.route != current_route {
+                    err_handler!("Invalid security stamp: Current route and exception route do not match")
+                } else if stamp_exception.security_stamp != claims.sstamp {
+                    err_handler!("Invalid security stamp for matched stamp exception")
+                }
+            } else {
+                err_handler!("Invalid security stamp")
+            }
         }
 
         Outcome::Success(Headers { host, device, user })
@@ -423,10 +439,6 @@ impl Into<Headers> for AdminHeaders {
     }
 }
 
-
-
-
-
 // col_id is usually the forth param ("/organizations/<org_id>/collections/<col_id>")
 // But there cloud be cases where it is located in a query value.
 // First check the param, if this is not a valid uuid, we will try the query value.
@@ -478,7 +490,7 @@ impl<'a, 'r> FromRequest<'a, 'r> for ManagerHeaders {
                                     None => err_handler!("The current user isn't a manager for this collection"),
                                 }
                             }
-                        },
+                        }
                         _ => err_handler!("Error getting the collection id"),
                     }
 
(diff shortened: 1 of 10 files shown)


dani-garcia:
Would this not leave the security_stamp unchanged when doing a password change?

BlackDex:
@dani-garcia, yes, but there is no other way unfortunately. I can't detect if there is going to be a key rotate after. Also, the web-vault does a logout when performing this action.

dani-garcia:
Right, the problem would be mostly other clients that are already logged in, which might start to have issues, the official server does the same but sends a logout notification, which we can't do for some clients.

Maybe when we receive the password change we can set the security stamp to a static invalid value, like `_ONLY_FOR_KDF_CHANGE_`, and only accept it during the kdf change? Then any other endpoint will fail and force a login, which would reset the security stamp to a valid value.

BlackDex:
@dani-garcia i have changed it now so it will only allow the specified route and still matches the previous security stamp.
This way it is still secure, but only the route we know is coming next is allowed.

BlackDex:
Ow, never mind, this is not working either, it will prevent all other normal logins when there is NO key rotate!
I have to fix that first.

BlackDex:
@dani-garcia, if you have any bright idea's on how to solve this¸ that would be great.

Just changing it (to either `_ONLY_FOR_KDF_CHANGE_` or something else) kinda like i did, does not work if you do not check the `key rotate` checkbox, because, all logins after that will not match. Also, there is nothing (yet) which does a security_reset on login.
I also can't just reset the key on every other request not  matching, because it can of course be possible that an other request (from maybe an other client) could reset this.

So the only thing i can think of right now is by checking if it is a special security-stamp and do a reset during the `/connect/token` aka `fn login` when it's a password login or maybe in `fn _password_login`, and the special token is set.

What do you think?

BlackDex:
Just a random thought before i forget them.

What if i do create a new stamp, and allow the new one to be used for everything else except the special one. But still use the previous one for validation of the allowed route. That still needs the stamp to be reset i think else that pipe check would trigger all the time.

BlackDex:
@dani-garcia I have made some changes, and tested this out.
Please let me know what you think if it or if you have any other idea :).

dani-garcia:
I like where this is going, but maybe we should consider creating another database field instead of trying to give the security_stamp field multiple uses.

BlackDex:
> I like where this is going, but maybe we should consider creating another database field instead of trying to give the security_stamp field multiple uses.

Yea, maybe that is best.
This is a kinda hacky way indeed.

Ill see what i can do to have that fixed :)

BlackDex:
@dani-garcia, I think i have a nice solution now.

src/auth.rs
```diff
@@ -298,7 +298,31 @@ impl<'a, 'r> FromRequest<'a, 'r> for Headers {
         };
 
         if user.security_stamp != claims.sstamp {
-            err_handler!("Invalid security stamp")
+            if let Some(stamp_exception_json) = &user.stamp_exception {
+                if let Ok(stamp_exception) = serde_json::from_str::<UserStampException>(&stamp_exception_json) {
+                    // Get the requested route to match with the exception route.
+                    let current_route = match request.route() {
+                        Some(route) => {
```
dani-garcia, line 305: This match could probably be simplified using `request.route().and_then(|r| r.name)`, which keeps us from having too many nested ifs.

Potentially you can also change the following above, but that might not look as nice:
`user.stamp_exception.as_deref().and_then(|s| serde_json::from_str::<UserStampException>(s).ok())`

BlackDex: Thanks for the pointers, i wasn't really happy with this part either.
Will adjust it right away.

BlackDex: I did changed both and with some fmt, it doesn't look that bad :).

PR #1244, opened 2020-11-29
Read config vars from files

### LINKED ISSUE #1194: Docker Secret on SMTP_PASSWORD

janost:
Fixes #1194.

This is a better implementation of the change previously posted in #1235.

This change allows the user to read configuration variable values from arbitrary files by using the `_FILE` suffix in environment variables.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -283,20 +283,37 @@ where
 
 use std::env;
 
+pub fn get_env_str_value(key: &str) -> Option<String>
+{
+    let key_file = format!("{}_FILE", key);
+    let value_from_env = env::var(key);
+    let value_file = env::var(&key_file);
+
+    match (value_from_env, value_file) {
+        (Ok(_), Ok(_)) => panic!("You should not define both {} and {}!", key, key_file),
+        (Ok(v_env), Err(_)) => Some(v_env),
+        (Err(_), Ok(v_file)) => match fs::read_to_string(v_file) {
+            Ok(content) => Some(content.trim().to_string()),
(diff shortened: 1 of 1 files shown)


dani-garcia:
Looks good to me, I also mostly agree with @BlackDex said here: https://github.com/dani-garcia/bitwarden_rs/pull/1235#issuecomment-735000791.

Reloading is a nice feature but the main problem is there are some settings that aren't reloadable on the fly, and until rocket async we don't have a way to restart the server itself, so it will mean it will work for some settings and not others.

If a user is loading the config from a file, it probably won't make much sense to let them set the value in the admin page config, because they probably want to control where and how the value is stored. At the same time, that's something they can themselves look for and be careful around it.

Also, maybe in this case we should completely disable the option to see the secrets from the admin page?

We should consider what we are going to do about that before the next release, or at least leave it as an undocumented feature until we have all those corners fleshed out. 

Thanks!

ahmaddxb:
i tried doing this but it didnt work

`    environment:
      - ADMIN_TOKEN_FILE=/run/secrets/vaultwarden_admin_token`

BlackDex:
@ahmaddxb , if you ever used the `/admin` interface to save settings, those env/file items are being overwritten.

ahmaddxb:
I got it work i was missing this, 

```
    secrets:
      - vaultwarden_admin_token
```

PR #1245, opened 2020-11-30
Show last active it on admin users page

janost:
Feature request from #246.

This change implements saving last login date and time for users to the database and adds this information to the admin panel's user list. I also added the user creation date to the admin page.

### DIFF
--- a/src/db/models/device.rs
+++ b/src/db/models/device.rs
@@ -178,4 +178,15 @@ impl Device {
                 .from_db()
         }}
     }
+
+    pub fn find_latest_active_by_user(user_uuid: &str, conn: &DbConn) -> Option<Self> {
+        db_run! { conn: {
+            devices::table
+                .filter(devices::user_uuid.eq(user_uuid))
+                .order(devices::updated_at.desc())
+                .first::<DeviceDb>(conn)
+                .ok()
+                .from_db()
+        }}
+    }
 }
--- a/src/db/models/user.rs
+++ b/src/db/models/user.rs
@@ -288,6 +288,13 @@ impl User {
             users::table.load::<UserDb>(conn).expect("Error loading users").from_db()
         }}
     }
+
+    pub fn last_active(&self, conn: &DbConn) -> Option<NaiveDateTime> {
+        match Device::find_latest_active_by_user(&self.uuid, conn) {
+            Some(device) => Some(device.updated_at),
+            None => None
+        }        
+    }
 }
 
 impl Invitation {
--- a/src/static/templates/admin/users.hbs
+++ b/src/static/templates/admin/users.hbs
@@ -21,6 +21,8 @@
                             <div class="float-left">
                                 <strong>{{Name}}</strong>
                                 <span class="d-block">{{Email}}</span>
+                                <span class="d-block">Created at: {{created_at}}</span>
+                                <span class="d-block">Last active: {{last_active}}</span>
                                 <span class="d-block">
                                     {{#if TwoFactorEnabled}}
                                         <span class="badge badge-success mr-2" title="2FA is enabled">2FA</span>
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -297,6 +297,11 @@ fn users_overview(_token: AdminToken, conn: DbConn) -> ApiResult<Html<String>> {
             usr["cipher_count"] = json!(Cipher::count_owned_by_user(&u.uuid, &conn));
             usr["attachment_count"] = json!(Attachment::count_by_user(&u.uuid, &conn));
             usr["attachment_size"] = json!(get_display_size(Attachment::size_by_user(&u.uuid, &conn) as i32));
+            usr["created_at"] = json!(&u.created_at.format("%Y-%m-%d %H:%M:%S").to_string());
+            usr["last_active"] = match u.last_active(&conn) {
+                Some(timestamp) => json!(timestamp.format("%Y-%m-%d %H:%M:%S").to_string()),
+                None => json!("Never")
+            };
             usr
     }).collect();
 


jjlin:
I don't think it's necessary to add a new `last_login` column. The device `updated_at` timestamp is already updated on each login, so you could just get that instead, e.g.
```sql
SELECT d.updated_at
FROM devices d INNER JOIN users u ON d.user_uuid = u.uuid
WHERE u.email = 'user@example.com'
ORDER BY d.updated_at DESC LIMIT 1
```

BlackDex:
I was thinking about this my self a few weeks back. And i think there is an other way without notifying the database.

I think you can use the `device` table for this. It has an updated at field. With that you can show when last login and with which device.

janost:
Thanks for the review @jjlin, @BlackDex!

If I understand the API correctly, what you're proposing is not functionally equivalent to this implementation.
The current `last_login` field only gets updated if a user logs in with username and password.
The `updated_at` field in the device model gets updated with a new timestamp even if the application is just locked, for example the user uses biometrics to unlock the Android app, since it requests a refresh token.

I'm fine with either implementation, we just need to agree on the term "Last login".

jjlin:
As an admin, I think I would want to know when a user was "last active" or "last online". The `updated_at` field is updated whenever the user logs in fresh, or if they're already logged in, whenever the cached token is expiring "soon" (in the next hour, AFAIK) and needs refreshing. The token refresh can happen even if the user simply opens the app, but doesn't actually unlock it. I think `updated_at` is a reasonable first approximation for last activity; ideally, a user's last sync, last cipher written, etc. could be factored in as well, but these aren't tracked currently.

IMO at least, last login (the current implementation) is not very useful since users can log in once via the browser extension or mobile app and then stay logged in indefinitely. So you could have a very active user whose "last login" is months or even years ago.

BlackDex:
I my self was thinking about a more detailed overview when for example clicking on a user, and show all active devices. Also with the ability to deauthorize per device. But for the main user overview i agree with @jjlin .

janost:
@jjlin @BlackDex You're right, I'm convinced. :+1: 
I've changed the implementation, please check it out.

src/static/templates/admin/users.hbs
```diff
@@ -21,6 +21,8 @@
                             <div class="float-left">
                                 <strong>{{Name}}</strong>
                                 <span class="d-block">{{Email}}</span>
+                                <span class="d-block">Created at: {{created_at}}</span>
+                                <span class="d-block">Last login: {{last_login}}</span>
```
jjlin, line 25: I think this would be more accurately labeled `Last active` or similar.

src/db/models/device.rs
```diff
@@ -178,4 +178,15 @@ impl Device {
                 .from_db()
         }}
     }
+
+    pub fn find_latest_active_by_user(user_uuid: &str, conn: &DbConn) -> Option<Self> {
```
jjlin, line 182: I'd suggest instead having a `last_active` method for `User`, which might later be extended to consider more than just device token updates.

janost:
@jjlin @BlackDex I changed the implementation according to your feedback.

src/db/models/user.rs
```diff
@@ -288,6 +288,14 @@ impl User {
             users::table.load::<UserDb>(conn).expect("Error loading users").from_db()
         }}
     }
+
+    pub fn last_active(&self, conn: &DbConn) -> Option<NaiveDateTime> {
+        let last_device = Device::find_latest_active_by_user(self.uuid.as_str(), conn);
```
jjlin, line 293: `&self.uuid` would be more efficient. Looks good otherwise.

janost: Thanks, I missed that. Fixed.

BlackDex:
Looking at it visually it's take up a lot of space. Also very much on a mobile (small screen) device.
I'm thinking about how to show this a bit nicer.

BlackDex:
I have something like this as the first row of the user:
```html
                    <tr>
                        <td colspan="5">
                            <span>Created at: {{created_at}}</span> | <span>Last active: {{last_active}}</span>
                        </td>
                    </tr>
```
With the following at every `<td>` in the next `<tr>`:
```html
 class="border-top-0"
```

Which then looks like this:
image

I'm also not quite sure about this approach.

jjlin:
It might be useful to sort on these values sometimes, so maybe they could each go under a separate column, formatted like
```
2020-01-01
12:34:56
```
I can't imagine how mobile devices would ever display this much data nicely, so that may be a lost cause anyway.

BlackDex:
@jjlin The table is responsive, so it kinda looks ok, you have to scroll left/right to see all.
I do have a lot of orgs with this test user, but it's of course possible.

Maybe we need to create some kind of overlay when there are more then x orgs a user is a member of.
That would create some space. And sorting on date-time is maybe already active. I have to check that JS-Library.

BlackDex:
What about using popover for the orgs when there are more then x?
https://thednp.github.io/bootstrap.native/#popoverExamples

BlackDex:
or, maybe just a collapse. That would be better for the searchbox i think.

BlackDex:
I think that Datatables can sort dates by default especially when using the ISO Format.

BlackDex:
This example i found looks ok for the orgs: https://jsfiddle.net/g4q5vp2f/1/ 
WIth that in place, we can just shrink that column, and add two new ones for last active and created

janost:
@BlackDex What is your opinion on merging this and working on the UI changes in a separate issue?
I feel like it is out of scope of this feature implementation and it might need further discussion.

src/db/models/user.rs
```diff
@@ -288,6 +288,14 @@ impl User {
             users::table.load::<UserDb>(conn).expect("Error loading users").from_db()
         }}
     }
+
+    pub fn last_active(&self, conn: &DbConn) -> Option<NaiveDateTime> {
+        let last_device = Device::find_latest_active_by_user(&self.uuid, conn);
+        match last_device {
```
BlackDex, line 294: I would shorten this and remove the `let last_device`, something like this.
```rust
match Device::find_latest_active_by_user(&self.uuid, conn) {
```

BlackDex (changes requested):
I think that is a good idea.
I'm also comfortable to merge this, since it's not something which could break other items.

I have only one remark left see my suggestion.

BlackDex:
Also, would be nice if you can merge these commit's all into one :).

janost:
@BlackDex Done! :+1:

BlackDex (approved):

dani-garcia:
Nice, thanks!

omueller:
Many thanks for this new very useful feature !     A way to sort the date by "last active" would also be great, and/or a way to export all data as CSV (then we can sort over the terminal or in a spreadsheet program).

BlackDex:
I'm planning to do a bit of a restyle of that specific interface, that should make sorting by date possible.
Don't know when it will be done.

omueller:
Great, thanks for the update @BlackDex!

PR #1247, opened 2020-11-30
Implement admin ability to enable/disable users

janost:
Feature request from #246.

This change implements the following:
- New bool flag on the `users` table: `enabled` with default value of `true`.
- 2 new admin endpoints:
  - `/admin/users/<uuid>/enable`
  - `/admin/users/<uuid>/disable`
- Check during login if the user is disabled and refuse login
- Add `Disabled` badge and enable/disable user links on admin page


BlackDex:
I Haven't tested it, but a quick look, and seems oke i think.

dani-garcia:
Okay had to solve the merge conflict but looks good!

N1c093:
I don't know if this is the right place to ask my question. I just saw, that this PR was merged some minutes ago.

Will it be possible to use this feature with fail2ban? For example: disable an account for 10 Minutes after 5 wrong login attemps.

dani-garcia:
You could potentially do it yes, but it might be a bit of work, you'd need to first login to get the admin session token, then obtain the user ID, then call the new enable/disable API endpoint.

PR #1252, opened 2020-12-03
Updated dependencies and Dockerfiles

BlackDex:
- Updated crates
- Updated rust-toolchain
- Updated Dockerfile to use latest rust 1.48 version
- Updated AMD64 Alpine to use same version as rust-toolchain and support PostgreSQL.
- Updated Rocket to the commit right before they updated hyper.
  Until that update there were some crates updated and some small fixes.
  After that build fails and we probably need to make some changes (which is probably something already done in the async branch)


BlackDex:
btw, i tested the docker builds, they all work.

dani-garcia:
Yeah any more rocket updates should wait until 0.5 is stable, or we find a major bug that requires updating to the current master. Eventually I'd want to not use any git dependencies and then we can even publish to crates.io

PR #1257, opened 2020-12-06
Validate cipher updates with revision date

### LINKED ISSUE #780: [BUG] Change data in the web overwritten password saved from Chrome extension
### Subject of the issue
(Note: I'm not sure if this is a bug in the official Chrome extension or in the bitwraden_rs project. But I'm gonna report it here first.)

**This is a potential data loss bug that can be easily reproduced, so would appreciate if this can be prioritized.**

### Your environment
* Bitwarden_rs version: 1.13.0-4cec502f 
* Clients used: Chrome extension
* Reverse proxy and version: Pomerium (probably not related)
* Other relevant information: 
 
### Steps to reproduce
1. Add a new entry in the vault. In my case, I created an entry with password `simple` with login URl `https://example.com`.
2. In the web, go to "Tools" then "Weak Passwords Report", you should find `example.com` in the list as `Very Weak`.
3. You click on that entry, you should be at the "EDIT ITEM" page. Keep it open.
4. Now click on the Chrome extension icon, find `example.com`, modify the password to `complex`, save. At this time, you should see "Password history: $NUM" in the Chrome extension.
5. To make sure the server received this change, in the extension, go to "Settings" -> "Sync" -> "Sync Vault Now".
6. Now go back to the web, the "EDIT ITEM" page, modify anything on the form (e.g. change username to `foo`) then save.
7. Go back to vault in the web, search example.com and check the password

### Expected behaviour
The username of the entry should be `foo` and the password should be `complex`, with at least one password history with value `simple`.

### Actual behaviour
The username is `foo` as expected, the password is `simple`, there's no password history at all. I.e. the password that I saved from the browser extension is gone.

FWIW: this is actually happening to me when I go through weak passwords report one by one and modify the password in a different tab, and accidentally saved the form with modified data in the web UI.

jjlin:
Prevent clients from updating a cipher if the local copy is stale.
Validation is only performed when the client provides its last known
revision date; this date isn't provided when using older clients,
or when the operation doesn't involve updating an existing cipher.

Fixes #780.

### DIFF
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -1,6 +1,7 @@
 use std::collections::{HashMap, HashSet};
 use std::path::Path;
 
+use chrono::{NaiveDateTime, Utc};
 use rocket::{http::ContentType, request::Form, Data, Route};
 use rocket_contrib::json::Json;
 use serde_json::Value;
@@ -194,6 +195,14 @@ pub struct CipherData {
     #[serde(rename = "Attachments")]
     _Attachments: Option<Value>, // Unused, contains map of {id: filename}
     Attachments2: Option<HashMap<String, Attachments2Data>>,
+
+    // The revision datetime (in ISO 8601 format) of the client's local copy
+    // of the cipher. This is used to prevent a client from updating a cipher
+    // when it doesn't have the latest version, as that can result in data
+    // loss. It's not an error when no value is provided; this can happen
+    // when using older client versions, or if the operation doesn't involve
+    // updating an existing cipher.
+    LastKnownRevisionDate: Option<String>,
 }
 
 #[derive(Deserialize, Debug)]
@@ -238,6 +247,17 @@ pub fn update_cipher_from_data(
     nt: &Notify,
     ut: UpdateType,
 ) -> EmptyResult {
+    // Check that the client isn't updating an existing cipher with stale data.
+    if let Some(dt) = data.LastKnownRevisionDate {
+        match NaiveDateTime::parse_from_str(&dt, "%+") { // ISO 8601 format
+            Err(err) =>
+                warn!("Error parsing LastKnownRevisionDate '{}': {}", dt, err),
+            Ok(dt) if cipher.updated_at.signed_duration_since(dt).num_seconds() > 1 =>
+                err!("The client copy of this cipher is out of date. Resync the client and try again."),
+            Ok(_) => (),
+        }
+    }
+
     if cipher.organization_uuid.is_some() && cipher.organization_uuid != data.OrganizationId {
         err!("Organization mismatch. Please resync the client before updating the cipher")
     }
@@ -1030,7 +1050,7 @@ fn _delete_cipher_by_uuid(uuid: &str, headers: &Headers, conn: &DbConn, soft_del
     }
 
     if soft_delete {
-        cipher.deleted_at = Some(chrono::Utc::now().naive_utc());
+        cipher.deleted_at = Some(Utc::now().naive_utc());
         cipher.save(&conn)?;
         nt.send_cipher_update(UpdateType::CipherUpdate, &cipher, &cipher.update_users_revision(&conn));
     } else {


jjlin:
This corresponds to changes in bitwarden/server#994. This feature hasn't been released upstream yet, but I've tested it with a build of their latest web vault code.

BlackDex:
Looks good. I think this will resolve #780 right? At least when the clients are updated.

jjlin:
More or less. There's still a 1 second window where the server would accept two conflicting writes, but that's not too likely. I think that window could be reduced to as small as 1 millisecond, but I'd have to see what time resolution the other clients provide.

src/api/core/ciphers.rs
```diff
@@ -238,6 +247,20 @@ pub fn update_cipher_from_data(
     nt: &Notify,
     ut: UpdateType,
 ) -> EmptyResult {
+    // Check that the client isn't updating an existing cipher with stale data.
+    if data.LastKnownRevisionDate.is_some() {
+        let dt = data.LastKnownRevisionDate.unwrap();
+        let res = NaiveDateTime::parse_from_str(&dt, "%+"); // ISO 8601 format
+        if res.is_ok() {
+            let dt = res.unwrap();
+            if cipher.updated_at.signed_duration_since(dt).num_seconds() > 1 {
```
jjlin, line 256: BTW, the web vault seems to provide a datetime with millisecond resolution, whereas bitwarden_rs seems to maintain nanosecond resolution, so in principle this slop factor could be reduced to as little as 1 ms, assuming the other clients also provide millisecond resolution.

src/api/core/ciphers.rs
```diff
@@ -238,6 +247,20 @@ pub fn update_cipher_from_data(
...
+                err!("The client copy of this cipher is out of date. Resync the client and try again.")
+            }
+        } else {
+            warn!("error parsing LastKnownRevisionDate '{}': {}", dt, res.unwrap_err());
+        }
+    }
```
cynecx, line 262: This is more idiomatic rust code:

```suggestion
    if let Some(dt) = &data.LastKnownRevisionDate {
        match NaiveDateTime::parse_from_str(dt, "%+") {
            Ok(dt) if cipher.updated_at.signed_duration_since(dt).num_seconds() > 1 => {
                err!("The client copy of this cipher is out of date. Resync the client and try again.");
            }
            Ok(_) => (),
            Err(err) => warn!("error parsing LastKnownRevisionDate '{}': {}", dt, err),
        }
    }
```

src/api/core/ciphers.rs
```diff
@@ -1,6 +1,7 @@
 use std::collections::{HashMap, HashSet};
 use std::path::Path;
 
+use chrono::{NaiveDateTime, Utc};
```
cynecx, line 4: `Utc` is unused.

BlackDex: @cynecx, i think you have to check again ;). See the `if soft_delete` part at the bottom of the page.

cynecx: Sorry, my bad. I've only applied the changes above so rustc put out a diagnostic.

src/api/core/ciphers.rs
jjlin: Thanks, I wasn't aware of conditionals on match arms. I amended to use a variant of your suggestion.

jjlin:
Actually, it turns out client support is already included in v2.17.1 of the web vault, but hasn't made it into a mobile client release yet. So this PR would be functional in conjunction with dani-garcia/bw_web_builds#21.

src/api/core/ciphers.rs
dani-garcia: Yeah I think the chances of a modification in a 1 second timespan are slim enough that 1 second will do fine.

dani-garcia:
The web vault with the changes is building now, once it's done I'll change the dockerfiles to point to it, but in the meantime this can be merged already.

PR #1262, opened 2020-12-08
Updated icon downloading.

BlackDex:
- Added more checks to prevent panics (Removed unwrap)
- Try do download from base domain or add www when the provided domain
  fails
- Added some more domain validation checks to prevent errors
- Added the ICON_BLACKLIST_REGEX to a Lazy Static HashMap which
  speeds-up the checks!
- Validate the Regex before starting/config change.
- Some cleanups
- Disabled some noisy debugging from 2 crates.

### DIFF
--- a/src/api/icons.rs
+++ b/src/api/icons.rs
@@ -1,7 +1,9 @@
 use std::{
+    collections::HashMap,
     fs::{create_dir_all, remove_file, symlink_metadata, File},
     io::prelude::*,
     net::{IpAddr, ToSocketAddrs},
+    sync::RwLock,
     time::{Duration, SystemTime},
 };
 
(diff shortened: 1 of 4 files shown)


src/api/icons.rs
```diff
@@ -177,8 +196,25 @@ fn check_icon_domain_is_blacklisted(domain: &str) -> bool {
     // Skip the regex check if the previous one is true already
     if !is_blacklisted {
         if let Some(blacklist) = CONFIG.icon_blacklist_regex() {
-            let regex = Regex::new(&blacklist).expect("Valid Regex");
-            if regex.is_match(&domain) {
+            // Generate the regex just once, since this can be costly depending upon the complexity of the regex
+            if !ICON_BLACKLIST_REGEX.read().unwrap().contains_key(&blacklist) {
```
dani-garcia, line 200: To avoid doing read().unwrap() twice for each call, I'd probably do something like:

```rust
if let Some(blacklist) = CONFIG.icon_blacklist_regex() {
    let mut regex_hashmap = ICON_BLACKLIST_REGEX.read().unwrap();

    // Use the pre-generate Regex stored in a Lazy HashMap if there's one, otherwise generate it
    let regex = if let Some(regex) = regex_hashmap.get(&blacklist) {
        regex
    
    } else {
        // Can't hold a read and write lock simultaneously, so drop the read
        drop(regex_hashmap);
        
        // Generate the regex just once, since this can be costly depending upon the complexity of the regex
        let blacklist_regex = Regex::new(&blacklist).unwrap();
        ICON_BLACKLIST_REGEX
            .write()
            .unwrap()
            .insert(blacklist.to_string(), blacklist_regex);

        regex_hashmap = ICON_BLACKLIST_REGEX.read().unwrap();
        regex_hashmap.get(&blacklist).unwrap()
    };

    if regex.is_match(&domain) {
        warn!("Blacklisted domain: {:#?} matched {:#?}", domain, blacklist);
        is_blacklisted = true;
    }
}
```

PR #1265, opened 2020-12-10
Fix stale data check failure when cloning a cipher

jjlin:


src/api/core/ciphers.rs
```diff
@@ -212,22 +212,35 @@ pub struct Attachments2Data {
     Key: String,
 }
 
+/// Called when an org admin clones an org cipher.
 #[post("/ciphers/admin", data = "<data>")]
 fn post_ciphers_admin(data: JsonUpcase<ShareCipherData>, headers: Headers, conn: DbConn, nt: Notify) -> JsonResult {
-    let data: ShareCipherData = data.into_inner().data;
+    post_ciphers_create(data, headers, conn, nt)
```
jjlin, line 218: Moved the main code into `post_ciphers_create` since that path is much more commonly used.

BlackDex (approved):

PR #1299, opened 2021-01-07
Admin interface WIP

BlackDex:
A WIP.
- Check if the DOMAIN is configured correctly.
- Created a start on a copy/paste-able json string of the config which has passwords and other sensitive values masked.


src/config.rs
```diff
@@ -196,6 +202,32 @@ macro_rules! make_config {
...
+            pub fn get_support_json(&self) -> serde_json::Value {
+                let cfg = {
+                    let inner = &self.inner.read().unwrap();
+                    inner.config.clone()
+                };
+
+                let mut support_json = serde_json::Map::new();
+                let privacy_config = vec!["domain", "domain_origin", "domain_path", "helo_name", "signups_domains_whitelist", "smtp_from", "smtp_host", "smtp_username"];
+
+                $($(
```
BlackDex, line 215: @dani-garcia i'm not very happy with this part, although i'm not sure on how to optimize/prettify this.
The issue mostly is the different types here, and that $ty matches the internal Option<> type and not Option.
So i need to do some weird hokes-spokus to get neat values.

Do you have any other suggestions on how to generate this?
The macros and types are messing up my concentration hehe.

dani-garcia: Yeah macros don't really lead to pretty code haha, I made an alternative more in line with the rest of the macro code, what do you think?


```rust
// Made this const and moved it up with the Regex
const PRIVACY_CONFIG: &[&str] = &["domain", "domain_origin", "domain_path", "helo_name", "signups_domains_whitelist", "smtp_from", "smtp_host", "smtp_username"];
           
            ...

            pub fn get_support_json(&self) -> serde_json::Value {
                let cfg = {
                    let inner = &self.inner.read().unwrap();
                    inner.config.clone()
                };

                json!({ $($(
                    stringify!($name): make_config! { @supportstr $name, cfg.$name, $ty, $none_action },
                )+)+ })
            }
        }
    };

    // Support string print
    ( @supportstr $name:ident, $value:expr, Pass, option ) => { $value.as_ref().map(|_| String::from("***")) }; // Optional pass, we map to an Option<String> with "***"
    ( @supportstr $name:ident, $value:expr, Pass, $none_action:ident ) => { String::from("***") }; // Required pass, we return "***"
    ( @supportstr $name:ident, $value:expr, $ty:ty, option ) => { // Optional other value, we return as is or convert to string to apply the privacy config
        if PRIVACY_CONFIG.contains(&stringify!($name)) {
            json!($value.as_ref().map(|x| PRIVACY_REGEX.replace_all(&x.to_string(), "*").to_string()))
        } else {
            json!($value)
        }
    };
    ( @supportstr $name:ident, $value:expr, $ty:ty, $none_action:ident ) => { // Required other value, we return as is or convert to string to apply the privacy config
       if PRIVACY_CONFIG.contains(&stringify!($name)) {
            json!(PRIVACY_REGEX.replace_all(&$value.to_string(), "*").to_string())
        } else {
            json!($value)
        }
    };
```

Note that I only checked that it compiles, but the result might end up slightly different.

Issue #905, opened 2020-03-14
Bitwarden does not react on SIGTERM (v1.13.1 sqlite)


mkilchhofer:
### Subject of the issue
At least the sqlite flavor of bitwarden_rs does not react on SIGTERM. I already wondered why evicting kubernetes nodes takes longer when bitwarden_rs runs on this node.
I now determined that the binary `/bitwarden_rs` inside the sqlite flavored docker image does not react on a SIGTERM signal.

### Your environment

* Bitwarden_rs version: 1.13.1

* Install method: Docker image (sqlite flavor)
* Clients used: n/a
* Reverse proxy and version: n/a
* Version of mysql/postgresql: n/a
* Other relevant information: n/a

### Steps to reproduce / Actual behaviour

Terminate the docker Image takes long and the exist status it not 0:
```bash
[root@localhost ~]# time podman stop bitwardenrs 
153e82e8be39928e69f61500a895a26d166041351f9d9076f00bbadc10ac41f4

real	0m10.252s
user	0m0.079s
sys	0m0.075s

[root@localhost ~]# podman ps -a 
CONTAINER ID  IMAGE                                COMMAND        CREATED         STATUS                       PORTS                   NAMES
153e82e8be39  docker.io/bitwardenrs/server:1.13.1  /bitwarden_rs  16 minutes ago  Exited (137) 50 seconds ago                          bitwardenrs
```
Another attempt to check if the binary handles SIGTERM:

1. Start the docker image again (`podman start bitwardenrs`)
2. Start a bash shell inside the container context (`podman exec -ti bitwardenrs bash`)
3. Update apt and install `procps` package
```bash
root@153e82e8be39:/# apt update && apt install -y procps
```
4. Send SIGTERM to the running binary:
```bash
root@153e82e8be39:/# ps -ef
UID          PID    PPID  C STIME TTY          TIME CMD
root           1       0  0 21:33 ?        00:00:00 /bitwarden_rs
root          24       0  0 21:34 pts/0    00:00:00 bash
root         521      24  0 21:35 pts/0    00:00:00 ps -ef

root@153e82e8be39:/# kill -SIGTERM 1
root@153e82e8be39:/# 
```
Nothing happens, I expect, that bitwarden starts to teardown.
### Expected behaviour

Bitwarden_rs binary implements a signal handler and handles SIGTERM correctly

### Relevant logs

```bash
[root@localhost ~]# podman logs -f -t bitwardenrs
[...]

PR #1321, opened 2021-01-22
Improve shutdown behavior (on kubernetes and allow CTRL+C)

mkilchhofer:
**Description of the change**

@mfriedenhagen pointed at an init process in this issue here: https://github.com/dani-garcia/bitwarden_rs/issues/905#issuecomment-753454442
First I thought that this would not help, since Rocket does not explicitly handly any signal sent to it. But it seems that adding an init process like `dumb-init` still improves the shutdown behavior. According to the [documentation](https://github.com/Yelp/dumb-init/blob/master/README.md) of dumb-init this might be the reason why:
> The Linux kernel applies special signal handling to processes which run as PID 1.

**Benefits**

Especially running bitwarden inside kubernetes resulted, that evicting a node which runs bitwarden takes a long time becase we need to wait until kublet kills bitwarden hardly. With this improvement, the container terminates as fast as all other workload.

**Possible drawbacks**

No known drawbacks

**Applicable issues**

- #905

### DIFF
--- a/docker/Dockerfile.j2
+++ b/docker/Dockerfile.j2
@@ -204,6 +204,7 @@ RUN [ "cross-build-start" ]
 RUN apk add --no-cache \
         openssl \
         curl \
+        dumb-init \
 {%   if "sqlite" in features %}
         sqlite \
 {%   endif %}
@@ -220,13 +221,11 @@ RUN apt-get update && apt-get install -y \
     openssl \
     ca-certificates \
     curl \
+    dumb-init \
     sqlite3 \
     libmariadb-dev-compat \
     libpq5 \
     && rm -rf /var/lib/apt/lists/*
-{% endif %}
-{% if "alpine" in target_file and "armv7" in target_file %}
-RUN apk add --no-cache -X http://dl-cdn.alpinelinux.org/alpine/edge/community catatonit
 {% endif %}
 
 RUN mkdir /data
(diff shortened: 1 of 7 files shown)


mkilchhofer:
I did not add dumb-init to the alpine variant of armv7, becase it seems already using another init process called [catatonit](https://github.com/openSUSE/catatonit) from openSUSE:
https://github.com/dani-garcia/bitwarden_rs/blob/master/docker/armv7/Dockerfile.alpine#L105

BlackDex:
@mkilchhofer
First, thanks for your contribution.

All these docker files are generated via the Makefile script by using a jinja2 template using the Dockerfile.j2.
So editing these Dockerfiles directly is not the correct way.

Also, i see that dumb-init also is available on Alpine, so maybe it would be good to change and update that one too so that we would have just one way of working for this.

mkilchhofer:
Ahh! :bulb: Okay, I will have a look into the Jinja template

mkilchhofer:
Seems easy to also update the Jinja and run `make all` :)
PR is updated.

Do we have tests for all architectures?

BlackDex:
LGTM, i will try and test some stuff, i have an RPI2B my self.

BlackDex (approved):
Seems good to me, i have tested it on amd64 and armv7 both Alpine and Debian versions work.
I can now Ctrl+C to stop the container when running it interactively.

dani-garcia:
Great, thanks! This will have to do until we update to rocket 0.5

PR #1329, opened 2021-01-25
JSON Response updates and small fixes

### LINKED ISSUE #971: Organization permission issues
### Your environment

* Bitwarden_rs version: 

* Install method: 
* Clients used: 
* Reverse proxy and version: 
* Other relevant information: 

### Steps to reproduce

Invite and Confirm a user with USER TYPE="User" and ACCESS CONTROL="This user can access only the selected collections." and select no collections.

### Expected behaviour

Read Only user cannot add item to the organization.

### Actual behaviour

Log in as the user and add an item, either from the organization or set the ownership to the organization.  The COLLETIONS will show "There are no collections to list.", but pressing save will create an item in the organization in the "Unassigned" collection.

Additionally, if the user is added to a collection with "Read Only" persmissions, this collection will show in COLLECTIONS when adding an item.  If the Read Only Collection is selected, then pressing save will generate a "No rights to modify the collection" error, however an item will be created in the USER's vault with null values.

These behaviors are not present in the official release.

BlackDex:
Updated several json response models.
Also fixed a few small bugs.

ciphers.rs:
  - post_ciphers_create:
    * Prevent cipher creation to organization without a collection.
  - update_cipher_from_data:
    * ~~Fixed removal of user_uuid which prevent user-owned shared-cipher to be not editable anymore when set to read-only.~~
    * Cleanup the json_data by removing the `Response` key/values from several objects.
  - delete_all:
    * Do not delete all Collections during the Purge of an Organization (same as upstream).

cipher.rs:
  - Cipher::to_json:
    * Updated json response to match upstream.
    * Return empty json object if there is no type_data instead of values which should not be set for the type_data.

organizations.rs:
  * Added two new endpoints to prevent Javascript errors regarding tax

organization.rs:
  - Organization::to_json:
    * Updated response model to match upstream
  - UserOrganization::to_json:
    * Updated response model to match upstream

collection.rs:
  - Collection::{to_json, to_json_details}:
    * Updated the json response model, and added a detailed version used during the sync
  - hide_passwords_for_user:
    * Added this function to return if the passwords should be hidden or not for the user at the specific collection (used by `to_json_details`)

Update 1: Some small changes after comments from @jjlin.
Update 2: Fixed vault purge by user to make sure the cipher is not part of an organization.

Resolves #971
Closes #990, Closes #991


src/api/core/ciphers.rs
```diff
@@ -225,6 +227,11 @@ fn post_ciphers_admin(data: JsonUpcase<ShareCipherData>, headers: Headers, conn:
 fn post_ciphers_create(data: JsonUpcase<ShareCipherData>, headers: Headers, conn: DbConn, nt: Notify) -> JsonResult {
     let mut data: ShareCipherData = data.into_inner().data;
 
+    // Check if there are one more more collections selected and err if not before creating an empty cipher.
+    if data.CollectionIds.is_empty() {
```
dani-garcia, line 231: The comment mentions this is also used when cloning ciphers, so I assume collectionsIds will be empty in that case without being an error.

BlackDex: That is something i have to check, that is why it's still a WIP :).
I also created this draft-pr to show that i'm working on this.

BlackDex: Thanks for the notice  btw, after i tested this i needed to add an extra check.

BlackDex:
btw, i have tested this also with the latest mobile Andriod client (and beta), as also with version v2.0.6 (Jun 16, 2019), they both worked as it should.

BlackDex:
Think about this again, i haven't tested the cli version. Maybe i should do that also.

BlackDex:
cli works fine too :)

src/db/models/collection.rs
```diff
@@ -49,12 +49,21 @@ impl Collection {
...
             "OrganizationId": self.org_uuid,
             "Name": self.name,
             "Object": "collection",
         })
     }
+
+    pub fn to_json_details(&self, user_uuid: &str, conn: &DbConn) -> Value {
+        let mut json_object = self.to_json();
+        json_object["Object"] = json!("collectionDetails");
+        json_object["ReadOnly"] = json!(!self.is_writable_by_user(user_uuid, conn));
```
jjlin, line 63: I think `is_writable_by_user()` also needs a little rewriting. It currently uses `user_org.access_all` when it should probably use `user_org.has_full_access()`. Also, it could be optimized a bit using the row count as in `hide_passwords_for_user` rather than selecting all columns.

src/db/models/organization.rs
```diff
@@ -147,16 +147,20 @@ impl Organization {
...
+            "MaxCollections": 10, // The value doesn't matter, we don't check server-side
             "MaxStorageGb": 10, // The value doesn't matter, we don't check server-side
             "Use2fa": true,
             "UseDirectory": false,
             "UseEvents": false,
             "UseGroups": false,
             "UseTotp": true,
             "UsePolicies": true,
+            "UseSso": false, // We do not support SSO
+            "SelfHost": true,
```
jjlin, line 162: Related to `/organizations/<org_id>/tax`, does returning this as `true` mean that the client won't try to access that endpoint?

src/api/core/ciphers.rs
```diff
@@ -323,7 +331,7 @@ pub fn update_cipher_from_data(
                     || cipher.is_write_accessible_to_user(&headers.user.uuid, &conn)
                 {
                     cipher.organization_uuid = Some(org_id);
-                    cipher.user_uuid = None;
+                    // cipher.user_uuid = None;
```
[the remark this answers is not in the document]
BlackDex: It should be deleted.

src/db/models/organization.rs
```diff
@@ -285,8 +290,26 @@ impl UserOrganization {
             "UseGroups": false,
             "UseTotp": true,
             "UsePolicies": true,
-            "UseApi": false,
+            "UseApi": false, // not supported by us
             "SelfHost": true,
+            "SsoBound": false, // We do not support SSO
+            "UseSso": false, // We do not support SSO
+            "UseBusinessPortal": false, // Disable BusinessPortal Button
```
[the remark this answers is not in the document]
BlackDex: I added this because the business portal is outside of the web-vault. It for one revives the button which we also hide via css. I'll put an extra comment there.
But we need to implement this business portal at some point.

src/db/models/cipher.rs
```diff
@@ -155,6 +165,12 @@ impl Cipher {
             "ViewPassword": !hide_passwords,
 
             "PasswordHistory": password_history_json,
+
+            // All Cipher types are included by default as null, but only the matching one will be populated
+            "Login": null,
+            "Card": null,
+            "Identity": null,
+            "SecureNote": null,
```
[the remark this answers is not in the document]
BlackDex: Sound logical

src/db/models/collection.rs
BlackDex: Yesterday i saw something else also.
Read only and hide passwords are also effective for owners and admins on upstream. Except when you view the ciphers via the organization tab. So the logic indeed looks different on multiple parts.
I think both access_all and has_full_access arn't fully correct

src/api/core/organizations.rs
```diff
@@ -1006,6 +1008,13 @@ fn put_policy(org_id: String, pol_type: i32, data: Json<PolicyData>, _headers: A
     Ok(Json(policy.to_json()))
 }
 
+#[allow(unused_variables)]
+#[get("/organizations/<org_id>/tax")]
+fn get_organization_tax(org_id: String, _headers: Headers, _conn: DbConn) -> EmptyResult {
+    // Prevent a 404 error, which also causes Javascript errors.
+    err!("Only allowed when not self hosted.")
```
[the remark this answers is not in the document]
BlackDex: It is always called, doesn't matter which type of hosting it is. Upstream checks this server-side. So too prevent 404's causing errors adding this is enough. Upstream uses the same response btw.

src/db/models/organization.rs
BlackDex: I'll double-check but i think it doesn't matter.

src/db/models/collection.rs
jjlin: If that's how upstream works, I think it's debatable whether that's correct behavior. Given owners/admins can clone those ciphers or export the whole org, it seems like a pointless restriction.

src/api/core/ciphers.rs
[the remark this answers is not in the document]
jjlin: Personally, I'd be more comfortable keeping this line and adding `cipher.organization_uuid = None` to the other branch to ensure that the invariant that only `user_uuid` or `organization_uuid` is set can't be violated.

src/db/models/organization.rs
BlackDex: I double-checked it, and it doesn't matter if i put it on false or true, these endpoints are always called for.

src/api/core/ciphers.rs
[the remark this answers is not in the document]
BlackDex: The user_uuid should never be removed here.
This causes the ownership of a cipher to be removed.
As an example when removing the user_uuid:
- As member with only user access, and hide-passwords enabled.
- I will add a cipher to an org.
- Later on i need to change the password, i can't any more because i can't edit that field.

Upstream checks if the current user is the owner of the cipher and then does allow that user to edit his own added cipher.
We actually check for this same situation during a cipher update, but always fail because the user_uuid has been removed.

src/db/models/collection.rs
BlackDex: @jjlin `is_writable_by_user()` already is using `user_org.has_full_access()` so that is fine i think.

src/api/core/ciphers.rs
```diff
@@ -366,6 +374,19 @@ pub fn update_cipher_from_data(
         }
     }
 
+    // Cleanup cipher data, like removing the 'Response' key.
```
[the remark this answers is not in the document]
BlackDex: I really have no idea, i have tried to check the sources of web, server and jslib, but it's too much data, and frankly if i would look deeper that would be a waist of my time :).

I think it is something caused by some Javascript library they use. I only know that it does not need to be stored or returned. Upstream is specifically selecting which fields they use/store, we just take the whole `data` key/value and store that without checking the validity of it mostly. That is why they do not store it, or remove it, because they simply ignore it when parsing the POST/PUT requests.

BlackDex:
@jjlin I have updated the PR and added a few more extra comments and some other small updates.

src/api/core/ciphers.rs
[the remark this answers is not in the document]
jjlin: I don't understand what you mean. A cipher is owned by either a user or an org, never both at once. So if you set `cipher.organization_uuid` to some value, it's now owned by that org, and `cipher.user_uuid` should be cleared, right?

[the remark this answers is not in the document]
BlackDex: Hmm, i have to update the delete query also now i think of it.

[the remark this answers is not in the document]
jjlin: I haven't tested upstream's behavior in this case, but if true, this is really unintuitive, and I haven't found any documentation of this distinction. I suspect there are various places in bitwarden_rs code that assume user/org ownership is mutually exclusive, and having both uuids set could result in erroneous behavior. I would suggest shelving this "feature" until there's more time to audit the entire codebase at least. There should also be a lot more comments clarifying this point.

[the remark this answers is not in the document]
BlackDex: Fixed vault purge by user now also.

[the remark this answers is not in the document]
BlackDex: Hmm, well maybe you are correct regarding holding this part off.
I will re-add the user_uuid = null again.

We do a lot of changes to a cipher, and save it between functions.
So, maybe we indeed need to audit the whole chain.

src/db/models/collection.rs
jjlin: Unless I'm missing something, I don't believe `Collection::is_writable_by_user()` uses `has_full_access()`.

BlackDex: @jjlin Than i think you are missing something, or you mean something totally different then what i think you are revering to:
https://github.com/dani-garcia/bitwarden_rs/blob/4628e4519de043aae343b4d0a1e2f110bd40aa8a/src/db/models/collection.rs#L222

jjlin: Sigh, that's my own code too. Yeah, you're right, I had an old tag checked out to look at something else.

PR #1333, opened 2021-01-27
Fix collection access issues for owner/admin users

### LINKED ISSUE #1307: Owners and Admins should be able to hide collections from there main vault overview/filter.
### Subject of the issue
Owners and Admins of Organizations should be able to hide collections from there _Filters section_ as upstream calls it.
You are able to uncheck the checkboxes when managing peoples even on owners and admins when you select the `This user can access only the selected collections.`. This doesn't prevent a user with owner or admin level privileges to still access them, but it should hide those collections from the main vault overview.

Also hiding and read-only checkboxes seem to work. But only when you are not viewing those items via the organization page.

### Your environment
* Using Docker: false
* Bitwarden_rs version: v1.18.0
* Web-vault version: v2.17.1
* DNS Check: true
* Time Check: true
* Domain Check: false
* Database type: SQLite
* Clients used: web-vault
* Reverse proxy and version: n/a
* Other relevant information: n/a

### Steps to reproduce
1. Go to the organization.
2. Click on Manage > People
3. Select a user, make him admin or owner
4. Select the option: This user can access only the selected collections. 
5. Check and uncheck some collections.
6. Using that user go to your main vault overview, and that user should only see the checked collections.
7. Also, that user should only be able to view passwords, or prevent to update shared items opened from his main vault overview. But should be able to do that when opened via the organizations overview.

### Expected behaviour
Only see the selected collections.

### Actual behaviour
Seeing all collections

Also see: https://bitwarden.com/help/article/user-types-access-control/#access-control
Where it states this behavior

jjlin:
The implementation of the `Manager` user type (#1242) introduced a regression
whereby owner/admin users are incorrectly denied access to certain collection
APIs if their access control for collections isn't set to "access all".

Owner/admin users should always have full access to collection APIs, per
https://bitwarden.com/help/article/user-types-access-control/#access-control:

> Assigning Admins and Owners to Collections via Access Control will only
> impact which Collections appear readily in the Filters section of their
> Vault. Admins and Owners will always be able to access "un-assigned"
> Collections via the Organization view.

Fixes #1307.

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -330,9 +330,9 @@ pub struct OrgHeaders {
     pub org_id: String,
 }
 
-// org_id is usually the second param ("/organizations/<org_id>")
-// But there are cases where it is located in a query value.
-// First check the param, if this is not a valid uuid, we will try the query value.
+// org_id is usually the second path param ("/organizations/<org_id>"),
+// but there are cases where it is a query value.
+// First check the path, if this is not a valid uuid, try the query values.
 fn get_org_id(request: &Request) -> Option<String> {
     if let Some(Ok(org_id)) = request.get_param::<String>(1) {
         if uuid::Uuid::parse_str(&org_id).is_ok() {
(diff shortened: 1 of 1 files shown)


BlackDex (approved):
Ah, thx for the fix.

PR #1341, opened 2021-01-31
Updated dependencies and small mail fixes

BlackDex:
- Updated rust nightly
- Updated depenencies
- Removed unicode support for regex (less dependencies)
- Fixed dependency and nightly changes/deprications
- Some mail changes for less spam point triggering


src/crypto.rs
```diff
@@ -67,7 +67,7 @@ pub fn generate_token(token_size: u32) -> Result<String, Error> {
     // token of fixed width, left-padding with 0 as needed.
     use rand::{thread_rng, Rng};
     let mut rng = thread_rng();
-    let number: u64 = rng.gen_range(low, high);
+    let number: u64 = rng.gen_range(low..=high);
```
jjlin, line 70: This should be `low..high`

BlackDex: Ah, indeed, else we could get 1 number which would be to high.

src/config.rs
```diff
@@ -557,6 +557,10 @@ fn validate_config(cfg: &ConfigItems) -> Result<(), Error> {
             err!("Both `SMTP_HOST` and `SMTP_FROM` need to be set for email support")
         }
 
+        if !cfg.smtp_from.contains('@') {
+            err!("SMTP_FROM does not contain a mandatory @ sign")
```
dani-garcia, line 561: Hmm, just realized that this is going to cause issues because _enable_smtp is enabled by default, so this will break for anyone who doesn't use email.

BlackDex: Ah, darn, overlooked that.

dani-garcia: Docker hub was being particularly slow and no builds finished before I stopped them, so no harm done. I would just add a check for host.is_some() and that should be good.

BlackDex: Isn't it better to check for `cfg.smtp_from.is_some() &&` ?

BlackDex: Though, if i do not configure host, but do configure smtp_from, i get the message that i need to configure both.

BlackDex: So, while it would be better to check if there is a value set, it would not have broken stuff.

dani-garcia: smtp_from uses a default value of an empty string right now, so you'd have to change it to optional, though that seems fine to me. And i think the cases where someone configures _from but not _host are practically negligible. Either both should be set or neither should be set.

BlackDex: That check is already there.

dani-garcia: Well we can either do `if  cfg.smtp_host.is_some() && !cfg.smtp_from.contains('@') {`

Or  `if  !cfg.smtp_host.is_empty() && !cfg.smtp_from.contains('@') {`

They have both the same results because we make sure that they are both equal in the check before:

```
if cfg.smtp_host.is_some() == cfg.smtp_from.is_empty() {
            err!("Both `SMTP_HOST` and `SMTP_FROM` need to be set for email support")
        }
```

BlackDex: New PR with the fix #1356

PR #1353, opened 2021-02-03
Extra features for admin interface.

### LINKED ISSUE #936: Impossible to delete user if it is the only owner of an organization
### Subject of the issue
If a user exists who created an organization that they are the only owner of, it is impossible to delete them.

### Your environment

* Bitwarden_rs version: 1.14.1-5471088e

* Clients used: admin page
* Reverse proxy and version: 
* Other relevant information: 

### Steps to reproduce

1. Start with admin_token set
2. Create a user, and create an organization with that user
3. Navigate to /admin/ and attempt to delete the user

### Expected behaviour
It should be possible to delete this user, or at least prevent future logins somehow. Alternatively, users should be prevented from creating organizations.

### Actual behaviour
The user could not be deleted with the following error message:

> Error deleting user
> Can't delete last owner

### Relevant logs

```
[2020-03-28 23:08:43][request][INFO] POST /admin/users/<uuid>/delete
[2020-03-28 23:08:43][error][ERROR] Can't delete last owner
[2020-03-28 23:08:43][response][INFO] POST /admin/users/<uuid>/delete (delete_user) => 400 Bad Request
```

BlackDex:
- Able to modify the user type per organization
- Able to remove a whole organization
- Added podman detection
- Only show web-vault update when not running a containerized bitwarden_rs

Resolves #936


src/static/templates/admin/diagnostics.hbs
```diff
@@ -253,10 +268,14 @@
         supportString += "* DNS Check: " + dnsCheck + "\n";
         supportString += "* Time Check: " + timeCheck + "\n";
         supportString += "* Domain Configuration Check: " + domainCheck + "\n";
+        supportString += "* HTTPS Check: " + httpsCheck + "\n";
         supportString += "* Database type: {{ diagnostics.db_type }}\n";
         {{#case diagnostics.db_type "MySQL" "PostgreSQL"}}
         supportString += "* Database version: [PLEASE PROVIDE DATABASE VERSION]\n";
```
dani-garcia, line 274: Is this something we could potentially just query from the database? Something like this, google says it might work:
```rust
diesel::sql_query("select version();").load(&conn);
```
Not necessasily something to do right now, but might be easier than asking people to find it by themselves.

BlackDex: Indeed. That could be a good addition. For the next round of admin updates :)

PR #1356, opened 2021-02-03
Fixed small buggy in validation

BlackDex:


BlackDex:
@dani-garcia this should fix it.
Let docker run ;)

PR #1413, opened 2021-02-19
Remove unnecessary allocations

paolobarbolini:
Removes unnecessary clones or allocations from the mail module.

### DIFF
--- a/src/mail.rs
+++ b/src/mail.rs
@@ -58,7 +58,7 @@ fn mailer() -> SmtpTransport {
 
     let smtp_client = match CONFIG.smtp_auth_mechanism() {
         Some(mechanism) => {
-            let allowed_mechanisms = vec![SmtpAuthMechanism::Plain, SmtpAuthMechanism::Login, SmtpAuthMechanism::Xoauth2];
+            let allowed_mechanisms = [SmtpAuthMechanism::Plain, SmtpAuthMechanism::Login, SmtpAuthMechanism::Xoauth2];
             let mut selected_mechanisms = vec![];
             for wanted_mechanism in mechanism.split(',') {
                 for m in &allowed_mechanisms {
@@ -115,7 +115,7 @@ pub fn send_password_hint(address: &str, hint: Option<String>) -> EmptyResult {
 
     let (subject, body_html, body_text) = get_text(template_name, json!({ "hint": hint, "url": CONFIG.domain() }))?;
 
-    send_email(address, &subject, &body_html, &body_text)
+    send_email(address, &subject, body_html, body_text)
 }
 
 pub fn send_delete_account(address: &str, uuid: &str) -> EmptyResult {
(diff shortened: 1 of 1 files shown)


BlackDex:
Hello @paolobarbolini, thx for the contribution.
The `vec!` removal i understand, that array indeed doesn't change size at runtime.

But changing &str to String not. &str is a pointer to the original, while String is a new copy if I'm correct. Also &str is mostly used for functions, also to indicate it will not transform the string, so actually i think it's now creating allocations instead of removing them.

paolobarbolini:
> &str is a pointer to the original, while String is a new copy if I'm correct. Also &str is mostly used for functions, also to indicate it will not transform the string, so actually i think it's now creating allocations instead of removing them

[`String` is a wrapper around `Vec<u8>`](https://doc.rust-lang.org/1.50.0/src/alloc/string.rs.html#278-280). `&str` is very similar to `&[u8]`. With that said `Vec` is a dynamically allocated linear piece of memory, while `&[u8]` is a pointer to some memory, which may belong to a number of different things, including [`Vec`](https://doc.rust-lang.org/1.50.0/std/vec/struct.Vec.html#impl-AsRef%3C%5BT%5D%3E) (or `String`).

Your current implementation gets a `String` from the template renderer, gets a string slice (basically a pointer) to the memory allocated by `String` and give it to `send_email`. However lettre either wants a `Vec<u8>` or a `String` as it's `body`, not a `&str`, so you call `String::from(body_html)` (`body_html` was a `&str`).

---

`String::from(some_reference_to_a_str)` is implemented like this in the std:

```rust
impl From<&str> for String {
  fn from(s: &str> -> String {
    let vec = Vec<u8>::from(s);
    // SAFETY: vec is guaranteed to be valid utf-8, since it was built from a `&str`
    unsafe { String::from_utf8_unchecked(vec) }
  }
}

impl Clone for Vec<u8> {
  fn clone(&self) -> Vec<u8> {
    Vec::from(self.as_slice())
  }
}

impl From<&[u8]> for Vec<u8> {
  fn from(b: &[u8]) -> Vec<u8> {
    let mut vec = Vec::new();
    // this is making a new allocation!!!
    vec.extend_from_slice(b);
    vec
  }
}
```

I rewrote this code here because the actual functionality is split into multiple functions across multiple files in the std. This however is an actual pretty good representation of what's actually happening under the hood, at least for `String` and `Vec<u8>`.

---

At this point you'll probably realize that doing `String::from` removes an heap allocation, since you could have directly given it the original `String`.

> Also &str is mostly used for functions

Unless the function needs a `String`, then passing a `String` is usually better, since it potentially avoids having to make a new heap allocation to create a new copy of `String`. This is also explained in the [Rust API guidelines - Caller decides where to copy and place data (C-CALLER-CONTROL)](https://rust-lang.github.io/api-guidelines/flexibility.html#c-caller-control)

dani-garcia:
Right, @paolobarbolini is right that we allocate the string, then pass a reference and allocate a new string when calling String::from, there's no reason to pass a reference when the String isn't going to be used again.

Maybe we can swap the allowed_mechanisms array to also be `const`? Not sure if the compiler is smart enough to do it by itself.

paolobarbolini:
> Maybe we can swap the allowed_mechanisms array to also be `const`? Not sure if the compiler is smart enough to do it by itself.

I'm not sure it would make any difference

dani-garcia:
Yeah, even if it does it's gonna be like two cpu instructions at most, no need to bother. Thanks!

PR #1444, opened 2021-02-27
Remove `md5.js` dependency

jjlin:
Switch to the built-in WebCrypto APIs for computing identicon hashes.


src/static/templates/admin/base.hbs
```diff
@@ -20,16 +20,24 @@
...
     <script>
         function reload() { window.location.reload(); }
         function msg(text, reload_page = true) {
             text && alert(text);
             reload_page && reload();
         }
-        function identicon(email) {
-            const data = new Identicon(md5(email), { size: 48, format: 'svg' });
+        async function sha256(message) {
+            // https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest
```
williamdes, line 31: https://caniuse.com/mdn-api_subtlecrypto_digest

williamdes (approved):

BlackDex:
This does cause those icons to not work on http. Could be a good indicator for some people.

But have you tried what happens without https? Does the page still load? I haven't checked my self yet.

jjlin:
AFAICT, the admin page still works, except for identicons not loading, and an error message in the console:
```
Uncaught (in promise) TypeError: Cannot read property 'digest' of undefined
    at sha256 (overview:33)
    at identicon (overview:39)
    at overview:361
    at overview:363
```

BlackDex (approved):

PR #1447, opened 2021-02-27
Allow only bug report issues

jjlin:
Remove templates for other issue types, directing them to the forum instead.


jjlin:
It would probably be good to have some topic templates on the Discourse forum (https://meta.discourse.org/t/what-are-topic-templates/38295), but I don't think I have permissions to set those up.

BlackDex:
Ill check the templates at the forum and see if i can do that.

BlackDex:
Maybe you can point people to the `Generate support string` at the diagnostics page? That would help us a lot, and sensitive items should be hidden already.

BlackDex:
I'm not able the see this feature with my access rights, or it doesn't work on a mobile. Will check via a normal browser tomorrow, but i think only @dani-garcia has access to that on discourse.

jjlin:
> Maybe you can point people to the `Generate support string` at the diagnostics page? That would help us a lot, and sensitive items should be hidden already.

Added

BlackDex (approved):

BlackDex:
I hope this helps :)

PR #1463, opened 2021-03-05
Add a dot in find command.

std2main:
Add a dot indicting current directory to search by find. 

find in mac won't work without the dot


williamdes (approved):
`./` would be more obvious :thinking:

std2main:
Sounds good to me, Updated with './'

PR #1469, opened 2021-03-07
CORS fixes

jjlin:
* The Safari extension apparently now uses the origin `file://` and expects
  that to be returned (see bitwarden/browser#1311, bitwarden/server#800).

* The `Access-Control-Allow-Origin` header was reflecting the value of the
  `Origin` header without checking whether the origin was actually allowed.
  This effectively allows any origin to interact with the server, which
  defeats the purpose of CORS.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -48,10 +48,16 @@ impl CORS {
         }
     }
 
-    fn valid_url(url: String) -> String {
-        match url.as_ref() {
-            "file://" => "*".to_string(),
-            _ => url,
+    // Check a request's `Origin` header against the list of allowed origins.
+    // If a match exists, return it. Otherwise, return None.
+    fn get_allowed_origin(headers: &HeaderMap) -> Option<String> {
+        let origin = CORS::get_header(headers, "Origin");
+        let domain_origin = CONFIG.domain_origin();
+        let safari_extension_origin = "file://";
+        if origin == domain_origin || origin == safari_extension_origin {
+            Some(origin)
+        } else {
+            None
         }
     }
 }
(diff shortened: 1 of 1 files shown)


jjlin:
See also: https://bitwardenrs.discourse.group/t/cant-login-with-safari-plugin-access-control-allow-origin/503

I'm not able to test on Safari, but hopefully that part works.

williamdes (approved):

BlackDex:
Ill check for safari if that works. Let you know.

BlackDex:
Strangely i do not have issues using safari and the latest version of bitwarden. Though i do see an update of safari (now using v14.0), i will update to v14.3 and see what happens there.

BlackDex:
So, i'm not seeing any issues with the old code, i also do not see any issues with the new code using Safari and the latest Bitwarden on MacOS.

So, i do not think it will solve the issues these people have.
Besides that, this change may still be good as it looks more sane then just allowing anything with CORS if someone sends `file://`.

jjlin:
Did you log out of the extension completely? One specific (potential) issue with the previous code was it returned `*` for the `file://` origin, and clients are not supposed to accept `*` for credentialed requests.

But in any case, if you can't reproduce the problem with the old code, then evidently it doesn't affect you for some reason, so on that basis, I don't think it's valid to say it won't help other people who do have this problem.

BlackDex:
I do not say it won't help other people, i just say i do not think it will solve there issues, and therefore asked the people who reported it to provide some more details.

Not being able to reproduce could mean there is something else causing the issue.

I even needed to configure the domain where the extension needed to login. Soo, yea, fully logged out, and as i said, tested both the current master and your patch. Both work fine, so either way would probably work, unless something in between the client and bitwarden_rs is changing the request/response.

And besides that, this patch does look a bit more secure then it was before.

BlackDex:
It seemed that it was user error regarding this issue, but i think this can be merged also.
Ill leave that up-to @dani-garcia to do, to not build a new docker image just yet.

BlackDex:
@jjlin Maybe you wan't to rebase this your self? Or can we just press the `Update branch` button?

BlackDex (approved):

PR #1487, opened 2021-03-15
Send access check fixes

jjlin:
Adjust checks for max access count, expiration date, and deletion date.
The date checks aren't that important, but the access count check
currently allows one more access than it should.


williamdes (approved):

BlackDex (approved):

BlackDex:
Yea, that looks like the right thing to do indeed.
I saw that danig mentioned he needed to change some MySQL migration issue, lets not merge just yet.
Then danig can do this together with his fixes.

PR #1499, opened 2021-03-16
Use minified scripts

RealOrangeOne:
Use minified versions of CSS and JS libraries. This has a few benefits:

- Reduces the size of the resulting binary, and thus docker container
- Reduces page load size
- Reduces the chance of a new contributor trying to edit part of the library themselves directly (albeit a small chance)
- Reduce the amount of JS in the repository (GitHub ignores minified files, so should show this repository as being mostly rust!)

I've not changed the version of any file, and each commit contains where it was downloaded from, should someone want to verify it's not been modified. I couldn't find a minified version of identicons, so left as-is.


BlackDex:
Thx for the contribution.
This has been discussed very briefly here already.
https://github.com/dani-garcia/bitwarden_rs/pull/1146#issuecomment-695348372

RealOrangeOne:
> make it considerably easier to check their integrity.

Not sure I agree here. The files exist on the remote, so all you do is change which one you're comparing to. 

The gains are negligible, yes, but they do exist.

jjlin:
It's not so much checking integrity against the original minified source as being able to audit the code as is in the repo, which is difficult to do if it's minified. For a piece of software where security is critical, on balance, I think using the original source is better, especially when this code isn't frequently loaded anyway.

williamdes:
> It's not so much checking integrity against the original minified source as being able to audit the code as is in the repo, which is difficult to do if it's minified. For a piece of software where security is critical, on balance, I think using the original source is better, especially when this code isn't frequently loaded anyway.

I agree

RealOrangeOne:
> audit the code as is in the repo

Audits to this project shouldn't be auditing the scripts themselves, that should be done by looking at the actual source. What's in these minified files is already bundled code. If you want to audit these scripts, you audit where they came from. The same is true for crates.io dependencies, or npm if this were using a full JS build pipeline. 

This is definitely the wrong place to be doing JS audits. You confirm that the file is as downloaded from the original vendor, and then you can audit the vendors source and processes etc, to confirm _that_ doesn't have any security issues, and then their dependencies, and their dependencies etc etc.

jjlin:
How do I know the minified versions provided by upstream haven't been tampered with? I don't track their build process, and might not be able to even if I wanted to. If you wanted to do the minification as part of the bitwarden_rs build, I'd be more in favor of that.

BlackDex:
Still, the admin interface isn't used that much, and besides that those files are cached by the browser. And i do like the readability of at least the css files.

If these files were part of a high traffic website i would jump to this, since it does save traffic, could maybe load a bit quicker.

Since the performance and size gains arn't that much, I'm not for merging this my self.

dani-garcia:
Agree with the comments here, the gains are quite small for something as low traffic as this, most of the size gains could probably be achieved by enabling gzip/brotli compression on the server, which would also benefit the web vault (which downloads about 5MB of JS when not gzipped, 1 MB when gzipped).

PR #1529, opened 2021-03-23
Return generic message when Send not available

mprasil:
This should help avoid leaking information about (non)existence of Send
and be more in line with what official server returns.

### DIFF
--- a/src/api/core/sends.rs
+++ b/src/api/core/sends.rs
@@ -13,6 +13,8 @@ use crate::{
     CONFIG,
 };
 
+const SEND_INACCESSIBLE_MSG: &str = "Send does not exist or is no longer available";
+
 pub fn routes() -> Vec<rocket::Route> {
     routes![
         post_send,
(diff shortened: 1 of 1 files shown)


BlackDex (approved):

jjlin:
Personally, I think it's more useful to keep the existing error messages. The user is going to be denied access in any case, but the more specific error message could at least help them (or the sender) understand why. IMO, the usual enumeration arguments don't apply here since everything is randomly generated.

But if everyone wants to go forward with a single error message, that should at least be a constant instead of being duplicated however many times.

williamdes:
> Personally, I think it's more useful to keep the existing error messages. The user is going to be denied access in any case, but the more specific error message could at least help them (or the sender) understand why. IMO, the usual enumeration arguments don't apply here since everything is randomly generated.
> 
> But if everyone wants to go forward with a single error message, that should at least be a constant instead of being duplicated however many times.

I agree with you: I think it's more useful to keep the existing error messages

mprasil:
> or the sender

The UI on the sender side [very clearly indicates](https://bitwarden.com/help/article/send-faqs/#q-what-do-the-icons-next-to-my-sends-indicate) all of the possible failure states. So I don't think the extra error messages benefit the sender in any way.

The Bitwarden documentation very clearly defines the [expected expiration behavior for both sender and receiver](https://bitwarden.com/help/article/send-lifespan/#expiration-behavior) and while bitwarden_rs does not always try to have identical behavior to upstream, this behavior is contrary to upstream documentation in [multiple places](https://bitwarden.com/help/article/send-lifespan/#maximum-access-count-behavior), which can be surprising to some users that will rely on the deniability of the link existence once it expired. I personally don't need that sort of privacy, but I don't think it's okay to assume, this will be okay for all users.

> that should at least be a constant instead of being duplicated however many times

Good point, will fix that.

PR #1546, opened 2021-03-27
Run Clippy

RealOrangeOne:
Run `clippy` over the codebase, and fix the issues

https://github.com/rust-lang/rust-clippy

The diff is large, but the code is now nice and rust-y, and fixes some `unwrap`s, unnecessary code and cleaner conditionals.

I intend to also do a PR for `cargo fmt`, but best get this one in first so it's easier to review.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -82,16 +82,26 @@ jobs:
         with:
           profile: minimal
           target: ${{ matrix.target-triple }}
+          components: clippy
       # End Uses the rust-toolchain file to determine version
 
 
-      # Run cargo tests (In release mode to speed up cargo build afterwards)
+      # Run cargo tests (In release mode to speed up future builds)
       - name: '`cargo test --release --features ${{ matrix.features }} --target ${{ matrix.target-triple }}`'
         uses: actions-rs/cargo@v1
         with:
           command: test
           args: --release --features ${{ matrix.features }} --target ${{ matrix.target-triple }}
       # End Run cargo tests
+
+
+      # Run cargo clippy (In release mode to speed up future builds)
+      - name: '`cargo clippy --release --features ${{ matrix.features }} --target ${{ matrix.target-triple }}`'
+        uses: actions-rs/cargo@v1
+        with:
+          command: clippy
+          args: --release --features ${{ matrix.features }} --target ${{ matrix.target-triple }}
+      # End Run cargo clippy
 
 
       # Build the binary
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,15 @@ on:
       - "docker/**"
       - "hooks/**"
       - "tools/**"
+  pull_request:
+    # Ignore when there are only changes done too one of these paths
+    paths-ignore:
+      - "**.md"
+      - "**.txt"
+      - "azure-pipelines.yml"
+      - "docker/**"
+      - "hooks/**"
+      - "tools/**"
 
 jobs:
   build:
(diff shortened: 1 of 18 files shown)


src/util.rs
```diff
@@ -62,25 +62,25 @@ impl CORS {
     }
 }
 
-impl Fairing for CORS {
+impl Fairing for Cors {
     fn info(&self) -> Info {
         Info {
-            name: "CORS",
+            name: "Cors",
```
RealOrangeOne, line 68: Is this change correct? Might need to be as before to be used downstream.

.github/workflows/build.yml
```diff
@@ -94,6 +104,15 @@ jobs:
       # End Run cargo tests
 
 
+      # Run cargo clippy (In release mode to speed up future builds)
+      - name: '`cargo clippy --release --features ${{ matrix.features }} --target ${{ matrix.target-triple }}`'
```
dani-garcia, line 108: Would this fail the builds in the case of clippy warnings? Because if so I'd rather ignore it and just review them by hand later.

src/db/models/organization.rs
```diff
@@ -90,17 +90,11 @@ impl PartialOrd<i32> for UserOrgType {
     }
 
     fn gt(&self, other: &i32) -> bool {
-        match self.partial_cmp(other) {
-            Some(Ordering::Less) | Some(Ordering::Equal) => false,
-            _ => true,
-        }
+        !matches!(self.partial_cmp(other), Some(Ordering::Less) | Some(Ordering::Equal))
```
dani-garcia, line 93:
```suggestion
        matches!(self.partial_cmp(other), Some(Ordering::Greater))
```

src/db/models/organization.rs
```diff
@@ -90,17 +90,11 @@ impl PartialOrd<i32> for UserOrgType {
...
     }
 
     fn ge(&self, other: &i32) -> bool {
-        match self.partial_cmp(other) {
-            Some(Ordering::Less) => false,
-            _ => true,
-        }
+        !matches!(self.partial_cmp(other), Some(Ordering::Less))
```
dani-garcia, line 97:
```suggestion
        matches!(self.partial_cmp(other), Some(Ordering::Greater) | Some(Ordering::Equal))
```

src/util.rs
dani-garcia: That should be just informational, I don't expect anyone to depend on that value for anything, it should be fine

dani-garcia (commented):
Some comments, mostly to avoid the ! before the matches! which is easily missed

.github/workflows/build.yml
BlackDex: Not sure if it will fail, but it does take extra time.
Maybe there is some way to let it ignore, but that would defeat the purpose i think.

Not having it enabled by default would be my choice.
Checking this more often by hand would be good for now.

BlackDex:
Also, regarding the fmt, would be better of we had some basic rules for that instead of the defaults. Now we only have an exception for the width, which i think could be a bit wider.

.github/workflows/build.yml
RealOrangeOne: > but it does take extra time

It does, but it reuses much of the information from the previous test compile

Not running it on CI would mean things would end up back in the position they are now. If people are adding new code, it's worth making sure it's still up to standard, even if it does add a few seconds to the build.

Looks like it won't fail the build on "warnings", but I do believe it'll put some useful comments on the diff showing where the issues are so they can be reviewed and fixed manually. Alternatively I can escalate warnings to errors, which I think to be useful should probably be done.

PR #1548, opened 2021-03-27
Updated diagnostics page

BlackDex:
- Added reverse proxy check
- Better deffinition of internet proxy
- Added SQL Server version detection

### DIFF
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -16,7 +16,7 @@ use crate::{
     api::{ApiResult, EmptyResult, JsonResult, NumberOrString},
     auth::{decode_admin, encode_jwt, generate_admin_claims, ClientIp},
     config::ConfigBuilder,
-    db::{backup_database, models::*, DbConn, DbConnType},
+    db::{backup_database, get_sql_server_version, models::*, DbConn, DbConnType},
     error::{Error, MapResult},
     mail,
     util::{format_naive_datetime_local, get_display_size, is_running_in_docker},
@@ -93,6 +93,27 @@ impl<'a, 'r> FromRequest<'a, 'r> for Referer {
 
     fn from_request(request: &'a Request<'r>) -> request::Outcome<Self, Self::Error> {
         Outcome::Success(Referer(request.headers().get_one("Referer").map(str::to_string)))
+    }
+}
+
+#[derive(Debug)]
+struct IpHeader(Option<String>);
+
+impl<'a, 'r> FromRequest<'a, 'r> for IpHeader {
+    type Error = ();
+
+    fn from_request(req: &'a Request<'r>) -> Outcome<Self, Self::Error> {
+        if req.headers().get_one(&CONFIG.ip_header()).is_some() {
+            Outcome::Success(IpHeader(Some(CONFIG.ip_header())))
+        } else if req.headers().get_one("X-Client-IP").is_some() {
+            Outcome::Success(IpHeader(Some(String::from("X-Client-IP"))))
+        } else if req.headers().get_one("X-Real-IP").is_some() {
+            Outcome::Success(IpHeader(Some(String::from("X-Real-IP"))))
+        } else if req.headers().get_one("X-Forwarded-For").is_some() {
+            Outcome::Success(IpHeader(Some(String::from("X-Forwarded-For"))))
+        } else {
+            Outcome::Success(IpHeader(None))
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/db/mod.rs
```diff
@@ -125,16 +125,34 @@ macro_rules! db_run {
...
         }
     };
+
+    // Same for all dbs
+    ( @raw $conn:ident: $body:block ) => {
+        db_run! { @raw $conn: sqlite, mysql, postgresql $body }
+    };
+
+    // Different code for each db
+    ( @raw $conn:ident: $( $($db:ident),+ $body:block )+ ) => {
```
dani-garcia, line 145: Why create a separate macro variant, the only difference I see with the above ones are the imports, and they shouldn't affect anything.

src/db/mod.rs
```diff
@@ -222,6 +240,40 @@ pub fn backup_database() -> Result<(), Error> {
...
+}
+
+/// Get the SQL Server version
+pub fn get_sql_server_version(conn: &DbConn) -> String {
+    db_run! {@raw conn:
+        postgresql, mysql {
+            match diesel::sql_query("SELECT version() AS version;").get_result::<SqlVersion>(conn).ok() {
+                Some(v) => {
+                    v.version
+                    // info!("V: {:#?}", v.version);
```
dani-garcia, line 258: These comments should probably be removed, also is there a need for the struct? Can diesel retrieve just that one column with something like `.get_result::<String>(..)`? (Or maybe using a tuple `.get_result::<(String)>(..)`?)

src/db/mod.rs
BlackDex: The `#[allow(unused)] use [<__ $db _model>]::*;` will break, because there is no database model for these kind of queries.

dani-garcia: But it should just import the existing models just fine, no? You simply won't have a use for them but that's fine.

BlackDex: It won't compile.
```
error[E0432]: unresolved import `__postgresql_model`
   --> src/db/mod.rs:130:46
    |
130 |                           #[allow(unused)] use [<__ $db _model>]::*;
    |                                                ^^^^^^^^^^^^^^^^^ use of undeclared crate or module `__postgresql_model`
```

src/db/mod.rs
BlackDex: That was the first thing i did. But also that will not compile.
I think the issue is that there is no schema available for the output, and thus we need to force the usage of a custom struct.

src/db/mod.rs
dani-garcia: Oh right! Those are created next to the structs with the other macro, not on the db module itself, yeah you are right!

src/db/mod.rs
BlackDex: Regarding the comments. those need to be removed indeed. Man, looked over the diff's a few time and still this gets into the commit.

dani-garcia: Hmm, I doubt that relies on the schema for the output, I don't think diesel parses the query string (not like sqlx does at least), but yea it can stay as it is, it was mostly curiosity on my part.

BlackDex: If i find a way to simplify it i will :).

src/api/admin.rs
```diff
@@ -96,6 +96,29 @@ impl<'a, 'r> FromRequest<'a, 'r> for Referer {
     }
 }
 
+struct IpHeader {
+    pub exists: bool,
```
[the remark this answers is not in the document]
BlackDex: I had been working with other languages to much haha. Much better option to use option.

williamdes (approved):

src/api/admin.rs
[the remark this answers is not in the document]
BlackDex: Is this a better option?

BlackDex:
Made some changes, let me know if this is better :).

I Also tried some other stuff to get the query working without the struct, but i'm just not seeing it.

src/db/mod.rs
BlackDex: I Asked at the diesel-rs channel, this was there answer:

> Georg Semmler
> @weiznich
> Mar 28 09:09
> @BlackDex sql_query requires always a custom struct that implements QueryableByName. Instead of using sql_query you could define your version function via sql_function! and use the typed query builder (so diesel::select(version()) which allows you to select directly into a String.
> That's because we can check at compile time that the later variant returns exactly one column/field, while we don't know that for the first one.

So, i think in this case it's easier to have this struct there. I will move that struct to within the function, since it's not used outside. Will do that in a new PR.

PR #1550, opened 2021-03-28
Cargo fmt

RealOrangeOne:
Run `cargo fmt` on the codebase, and add it to CI

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -89,7 +89,7 @@ jobs:
         with:
           profile: minimal
           target: ${{ matrix.target-triple }}
-          components: clippy
+          components: clippy, rustfmt
       # End Uses the rust-toolchain file to determine version
 
 
@@ -109,6 +109,15 @@ jobs:
           command: clippy
           args: --release --features ${{ matrix.features }} --target ${{ matrix.target-triple }}
       # End Run cargo clippy
+
+
+      # Run cargo fmt
+      - name: '`cargo fmt`'
+        uses: actions-rs/cargo@v1
+        with:
+          command: fmt
+          args: --all -- --check
+      # End Run cargo fmt
 
 
       # Build the binary
--- a/build.rs
+++ b/build.rs
@@ -1,7 +1,7 @@
-use std::process::Command;
 use std::env;
+use std::process::Command;
 
-fn main() { 
+fn main() {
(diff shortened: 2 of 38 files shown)


williamdes (approved):

dani-garcia:
I think we should try to tweak rustfmt's style a bit, for example, this change:
```rust
let steps: i64 = if CONFIG.authenticator_disable_time_drift() { 0 } else { 1 };

let steps: i64 = if CONFIG.authenticator_disable_time_drift() {
    0
} else {
    1
};
```
Looks worse afterwards.

Then sometimes it decides to not split lines at all, like this:
```
let json = json!({"page_content": "admin/login", "version": VERSION, "error": msg, "urlpath": CONFIG.domain_path()});
let json =
    json!({"page_content": "admin/login", "version": VERSION, "error": msg, "urlpath": CONFIG.domain_path()});
```

Which I assume is because of the macro, but i'd rather that change not be done.

There are others too, particularly with long strings as function parameters, where it moves the string to a new line and it barely reduces the width, in those cases it might be better to split the string and escape the newline with \

I don't know if those things are configurable in rustfmt, if they aren't I'll just have to deal with it 😄

RealOrangeOne:
@dani-garcia I've re-applied to updated `master`, and manually reviewed and re-flowed a couple lines manually.

BlackDex:
I did some testing, and i think i have a nice config. The only thing  missing is a new feature so that chains are not being put on one line as we now have a lot of those separated on a new-line, and using fmt now will compress those to one line.

In v2.x they have this, and i see that there is a PR open for a backport of this feature: https://github.com/rust-lang/rustfmt/pull/4782 , which i would really like to see merged, and then we could use that to have it more inline with how our code looks right now.

These is btw my tested settings
```toml
version = "Two"
edition = "2018"
max_width = 225
newline_style = "Unix"
use_small_heuristics = "Off"
struct_lit_single_line = false
overflow_delimited_expr = true
```

RealOrangeOne:
> max_width = 225
Seems excessive?

I'll try running with that config, see what the diff looks like.

BlackDex:
> > max_width = 225
> > Seems excessive?
> 
> I'll try running with that config, see what the diff looks like.

I checked with `wc -L` which files had long lines, and that seems what we are comfortable with now.

RealOrangeOne:
Turns out there's already a `rustfmt.toml` file in the repo, with a length of 120. Given that's a far more sane number, and what the rest of this PR was formatted with, I left it. 225 was making lines far longer than we'd probably want.

PR #1575, opened 2021-04-01
Just ignore scripts

RealOrangeOne:
Nothing else in `src/static` is vendored external scripts, so just ignore these.

This also fixes the glob, which previously wasn't matching anything


BlackDex (approved):

RealOrangeOne:
image

Looks like that did it :tada:

PR #1579, opened 2021-04-03
Add support for auto-deleting trashed items

jjlin:
Upstream will soon auto-delete trashed items after 30 days, but some people use the trash as an archive folder, so to avoid unexpected data loss, this implementation requires the user to explicitly enable auto-deletion.

This PR also adds a generic job scheduler, and rewrites deletion of old sends to use it.

### DIFF
--- a/src/db/models/cipher.rs
+++ b/src/db/models/cipher.rs
@@ -1,5 +1,7 @@
-use chrono::{NaiveDateTime, Utc};
+use chrono::{Duration, NaiveDateTime, Utc};
 use serde_json::Value;
+
+use crate::CONFIG;
 
 use super::{
     Attachment,
@@ -271,6 +273,17 @@ impl Cipher {
         Ok(())
     }
 
+    /// Purge all ciphers that are old enough to be auto-deleted.
+    pub fn purge_trash(conn: &DbConn) {
+        if let Some(auto_delete_days) = CONFIG.trash_auto_delete_days() {
+            let now = Utc::now().naive_utc();
+            let dt = now - Duration::days(auto_delete_days);
+            for cipher in Self::find_deleted_before(&dt, conn) {
+                cipher.delete(&conn).ok();
+            }
+        }
+    }
+
     pub fn move_to_folder(&self, folder_uuid: Option<String>, user_uuid: &str, conn: &DbConn) -> EmptyResult {
         User::update_uuid_revision(user_uuid, conn);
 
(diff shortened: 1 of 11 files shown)


williamdes (approved):

src/main.rs
```diff
@@ -301,7 +305,7 @@ fn check_web_vault() {
     }
 }
 
-fn launch_rocket(extra_debug: bool) {
+fn create_db_pool() -> db::DbPool {
     let pool = match util::retry_db(db::DbPool::from_config, CONFIG.db_connection_retries()) {
```
RealOrangeOne, line 309: Any reason not to just return `pool` directly?

src/db/models/send.rs
```diff
@@ -271,4 +272,13 @@ impl Send {
                 .load::<SendDb>(conn).expect("Error loading sends").from_db()
         }}
     }
+
+    pub fn find_by_past_deletion_date(conn: &DbConn) -> Vec<Self> {
```
RealOrangeOne, line 276: This method name is quite a mouthful. Perhaps `find_expired`?

src/db/models/send.rs
```diff
@@ -205,6 +205,13 @@ impl Send {
         }}
     }
 
+    /// Purge all sends that are past their deletion date.
+    pub fn purge(conn: &DbConn) {
+        for send in Self::find_by_past_deletion_date(&conn) {
```
RealOrangeOne, line 210: Rather than filtering and iterating, it'd probably be more efficient (both in code and runtime) to do a `DELETE ... WHERE` query instead?

src/db/models/cipher.rs
```diff
@@ -511,6 +524,15 @@ impl Cipher {
         }}
     }
 
+    /// Find all ciphers that were deleted earlier than the specified datetime.
+    pub fn find_deletion_candidates(dt: &NaiveDateTime, conn: &DbConn) -> Vec<Self> {
```
RealOrangeOne, line 528: Perhaps something like `find_deleted_before`? At which point the name becomes enough to remove the comment

src/db/models/cipher.rs
```diff
@@ -271,6 +273,17 @@ impl Cipher {
         Ok(())
     }
 
+    /// Purge all ciphers that are old enough to be auto-deleted.
+    pub fn purge_trash(conn: &DbConn) {
```
RealOrangeOne, line 277: Perhaps this logic should live in the scheduled task instead, rather than on the model?

src/main.rs
jjlin: Apparently not, updated.

src/db/models/send.rs
jjlin: Expired has a different meaning in the terminology of Sends. It doesn't need to be public at the moment, but I don't think it matters that much.

src/db/models/cipher.rs
jjlin: Thanks, I like this name better. Self-documenting names are not a replacement for doc comments, though.

src/db/models/send.rs
jjlin: This might work in other cases, but if you follow the logic, it doesn't work here.

src/db/models/cipher.rs
jjlin: I think it makes more sense for this logic to be kept closer to the model, or at least the API implementation.

PR #1583, opened 2021-04-03
Updated icon fetching.

BlackDex:
- Added image type checking, and prevent downloading non images.
  We didn't checked this before, which could in turn could allow someone
to download an arbitrary file.
- This also prevents SVG images from being used, while they work on the
  web-vault and desktop client, they didn't on the mobile versions.
- Because of this image type checking we can return a valid file type
  instead of only 'x-icon' (which is still used as a fallback).
- Prevent rel values with `icon-mask`, these are not valid favicons.

### DIFF
--- a/src/api/icons.rs
+++ b/src/api/icons.rs
@@ -37,6 +37,7 @@ static CLIENT: Lazy<Client> = Lazy::new(|| {
 
 // Build Regex only once since this takes a lot of time.
 static ICON_REL_REGEX: Lazy<Regex> = Lazy::new(|| Regex::new(r"(?i)icon$|apple.*icon").unwrap());
+static ICON_REL_BLACKLIST: Lazy<Regex> = Lazy::new(|| Regex::new(r"(?i)mask-icon").unwrap());
 static ICON_SIZE_REGEX: Lazy<Regex> = Lazy::new(|| Regex::new(r"(?x)(\d+)\D*(\d+)").unwrap());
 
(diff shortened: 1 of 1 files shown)


williamdes (approved):

src/api/icons.rs
```diff
@@ -650,7 +671,37 @@ fn save_icon(path: &str, icon: &[u8]) {
...
+            Some("jpeg")
+        } else if source_2bytes == BMP_HEADER.to_vec() {
+            Some("bmp")
+        } else {
+            None
+        }
+    }
+
+    None
+}
```
paolobarbolini, line 707: This could use [subslice patterns](https://blog.rust-lang.org/2020/03/12/Rust-1.42.html#subslice-patterns)

```suggestion
fn get_icon_type(bytes: &[u8]) -> Option<&'static str> {
    match bytes {
        [137, 80, 78, 71, ..] => Some("png"),
        [0, 0, 1, 0, ..] => Some("x-icon"),
        [82, 73, 70, 70, ..] => Some("webp"),
        [255, 216, 255, ..] => Some("jpeg"),
        [66, 77, ..] => Some("bmp"),
        _ => None
    }
}
```

BlackDex: I tried that the hard way and it didn't worked, thx for the input. Will change it in a few!

BlackDex: Updated code. Thanks for the help.

PR #1584, opened 2021-04-04
Some admin interface updates.

BlackDex:
- Fixed bug when web-vault is disabled.
- Updated sql-server version check to be simpler thx to @weiznich ( https://github.com/dani-garcia/bitwarden_rs/pull/1548#discussion_r604767196 )
- Use `VACUUM INTO` to create a SQLite backup instead of using the external sqlite3 application.
  - This also removes the dependancy of having the sqlite3 packages installed on the final image unnecessary, and thus removed it.
- Updated backup filename to also have the current time.
- Add specific bitwarden_rs web-vault version check (to match letter patched versions)
  Will work when https://github.com/dani-garcia/bw_web_builds/pull/33 is build (But still works without it also).

### DIFF
--- a/docker/Dockerfile.j2
+++ b/docker/Dockerfile.j2
@@ -215,9 +215,6 @@ RUN apk add --no-cache \
         openssl \
         curl \
         dumb-init \
-{%   if "sqlite" in features %}
-        sqlite \
-{%   endif %}
 {%   if "mysql" in features %}
         mariadb-connector-c \
 {%   endif %}
@@ -232,7 +229,6 @@ RUN apt-get update && apt-get install -y \
     ca-certificates \
     curl \
     dumb-init \
-    sqlite3 \
     libmariadb-dev-compat \
     libpq5 \
     && rm -rf /var/lib/apt/lists/*
--- a/docker/amd64/Dockerfile
+++ b/docker/amd64/Dockerfile
@@ -86,7 +86,6 @@ RUN apt-get update && apt-get install -y \
     ca-certificates \
     curl \
     dumb-init \
-    sqlite3 \
     libmariadb-dev-compat \
     libpq5 \
     && rm -rf /var/lib/apt/lists/*
--- a/docker/amd64/Dockerfile.alpine
+++ b/docker/amd64/Dockerfile.alpine
@@ -82,7 +82,6 @@ RUN apk add --no-cache \
         openssl \
         curl \
         dumb-init \
-        sqlite \
         postgresql-libs \
         ca-certificates
 
--- a/docker/arm64/Dockerfile
+++ b/docker/arm64/Dockerfile
@@ -129,7 +129,6 @@ RUN apt-get update && apt-get install -y \
     ca-certificates \
     curl \
     dumb-init \
-    sqlite3 \
     libmariadb-dev-compat \
     libpq5 \
     && rm -rf /var/lib/apt/lists/*
--- a/docker/armv6/Dockerfile
+++ b/docker/armv6/Dockerfile
@@ -129,7 +129,6 @@ RUN apt-get update && apt-get install -y \
     ca-certificates \
     curl \
     dumb-init \
-    sqlite3 \
     libmariadb-dev-compat \
     libpq5 \
     && rm -rf /var/lib/apt/lists/*
--- a/docker/armv7/Dockerfile
+++ b/docker/armv7/Dockerfile
@@ -129,7 +129,6 @@ RUN apt-get update && apt-get install -y \
     ca-certificates \
     curl \
     dumb-init \
-    sqlite3 \
     libmariadb-dev-compat \
     libpq5 \
     && rm -rf /var/lib/apt/lists/*
--- a/docker/armv7/Dockerfile.alpine
+++ b/docker/armv7/Dockerfile.alpine
@@ -86,7 +86,6 @@ RUN apk add --no-cache \
         openssl \
         curl \
         dumb-init \
-        sqlite \
         ca-certificates
 
 RUN mkdir /data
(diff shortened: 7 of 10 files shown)


williamdes (approved):

jjlin:
While I continue to think the SQLite backup feature is ill-advised, using `VACUUM INTO` should at least allow removing the `sqlite3` binary to slim down the Docker images.

BlackDex:
That only works on the alpine image. Since the debian based still needs the shared library, unless we can install that separately.

jjlin:
I'm pretty sure SQLite is statically linked into the bitwarden_rs binary in either case, so it should be fine.

BlackDex:
I think it isn't on my dev environment. I'll have to double check.

BlackDex:
> While I continue to think the SQLite backup feature is ill-advised, using `VACUUM INTO` should at least allow removing the `sqlite3` binary to slim down the Docker images.

@jjlin 
What is your suggestion then to backup? Why shouldn't there be an easy way for people to create a backup for sqlite? Creating a backup for MySQL and Postgresql is a different thing, and normally done via other tools. But for sqlite, there is no other way then this, and it's quick and easy.

BlackDex:
> I'm pretty sure SQLite is statically linked into the bitwarden_rs binary in either case, so it should be fine.

I checked it out, seems it does indeed embed it.
On my dev environment it doesn't because i have the dev files installed and it some how lets ldd show that it uses that.
So, i think i can remove it indeed in the end image.

BlackDex:
@jjlin Modified the Dockerfile template to remove the sqlite package, as it indeed isn't needed anymore.

jjlin:
> > While I continue to think the SQLite backup feature is ill-advised, using `VACUUM INTO` should at least allow removing the `sqlite3` binary to slim down the Docker images.
> 
> @jjlin
> What is your suggestion then to backup? Why shouldn't there be an easy way for people to create a backup for sqlite? Creating a backup for MySQL and Postgresql is a different thing, and normally done via other tools. But for sqlite, there is no other way then this, and it's quick and easy.

We've discussed this briefly in Matrix chat before, but to document it here, my main concern is that it facilitates poor backup practices:

1. It doesn't do a complete backup (including attachments and config).
2. It suggests that occasional manual backups are a reasonable backup strategy.
3. It somewhat encourages people to just leave their backups on the same storage media as their original data.

I think we can agree that experienced operators are not going to use this feature -- they're going to read https://github.com/dani-garcia/bitwarden_rs/wiki/Backing-up-your-vault and set up automated backups of some sort.
So the only users are going to be inexperienced ones, and they're exactly the ones who won't realize the issues above. Instead, they'll think that this backup method is totally fine -- after all, it's what the developers built into the software, and they must know best.

Personally, I would lean towards dropping this backup feature entirely, as it doesn't add a lot of value. It's a one-line CLI command, and even if you did decide to do manual backups, you're hopefully going to be in the CLI to copy that file somewhere else anyway. But maybe it would be more useful to people running on Synology or something, I don't know.

Otherwise, at minimum, I think there should be some wording to make it clear that this feature doesn't produce a complete backup, and a link to the wiki page on backups.

Some other enhancements might include:

1. Actually making a complete backup (tarball with database backup, attachments, and config), to address issue (1).
2. Perhaps using the job scheduler to provide built-in scheduled backups, to address issue (2).

Issue (3) is still a problem, though. But overall, I have no idea how many people even use this backup feature and whether these enhancements would be worthwhile.

dani-garcia:
I think I mostly agree that the backup feature is a bit incomplete, and doing a more complete backup automation system is probably out of scope considering it's something that can be done externally in an easy way if you know what you are doing, but I think we should have some easy way to create and possibly automate complete backups, even if it's just a pre-made script in the wiki where users can input their bitwarden installation and a backup folder and it makes a complete backup for them.

pepa65:
> I think we can agree that experienced operators are not going to use this feature -- they're going to read https://github.com/dani-garcia/bitwarden_rs/wiki/Backing-up-your-vault and set up automated backups of some sort.

"This feature" refers to the ability to Vacuum Into for making the sqlite backup, which is great, and obviates the need to include a separate sqlite3 binary in the docker image. And if you read that link about backing up your vault, it talks about the database backups extensively, so I don't see any problem with this. That page itself could refer to an improved & automated backup script, great! But the feature we want.

jjlin:
Yes, I wrote the page on backups. You can `VACUUM INTO` a backup file on the Docker host as well, as part of an automated process. The point is that backups should be complete and automated, which this feature doesn't help with.

BlackDex:
I mostly agree regarding the backups. But i find that this app provides the usage of sqlite as default. And sqlite is serverless, which means embedded into the application in this case. If it was external i would totally agree with all the statements.

pepa65:
> backups should be complete and automated

So how can you have complete and automated backups without this feature??

jjlin:
There are several examples at https://github.com/dani-garcia/bitwarden_rs/wiki/Backing-up-your-vault#examples

BlackDex:
Using the sqlite3 binary on the host somewhere which does the magic.

pepa65:
OK, sorry for all the noise, I think I get it. @jjlin prefers to use the external sqlite3 backup command because that links in with other commands to do a full backup, so he has no use for a backup command from within the app.
But it can be a useful feature to take a snapshot of the database, even if the attachments are not backed up at the same time.

BlackDex:
There are a dozen ways to create backups.
I make snapshots of the sqlite database using the sqlite3 tool every 5 hours. The whole volume/folder of the data is synced to my nextcloud every 30min (which has versioning and trash features). And nextcloud is backedup externally.

PR #1587, opened 2021-04-06
Allow outbound requests to go via a proxy

RealOrangeOne:
This allows a couple things:

- Outbound request code is now more consistent, and clients are constructed in the same place (icons is _slightly_ different, but might fix that too)
- Outbound requests can be forwarded through a proxy, to hide the location of their server. Because the icons API is unauthenticated, anyone can force Bitwarden_RS to make a request.

### DIFF
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -3,7 +3,7 @@ use serde::de::DeserializeOwned;
 use serde_json::Value;
 use std::{env, time::Duration};
 
-use reqwest::{blocking::Client, header::USER_AGENT};
+
 use rocket::{
     http::{Cookie, Cookies, SameSite},
     request::{self, FlashMessage, Form, FromRequest, Outcome, Request},
@@ -19,7 +19,7 @@ use crate::{
     db::{backup_database, get_sql_server_version, models::*, DbConn, DbConnType},
     error::{Error, MapResult},
     mail,
-    util::{format_naive_datetime_local, get_display_size, is_running_in_docker},
+    util::{format_naive_datetime_local, get_display_size, is_running_in_docker, get_reqwest_client},
     CONFIG,
 };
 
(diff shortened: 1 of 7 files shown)


BlackDex:
If I'm correct this is already working.
You should be able to use HTTP_PROXY/HTTPS_PROXY already.

RealOrangeOne:
> If I'm correct this is already working.
> You should be able to use HTTP_PROXY/HTTPS_PROXY already.

Not according to my testing, or at least not with a socks5 proxy.

BlackDex:
> > If I'm correct this is already working.
> > You should be able to use HTTP_PROXY/HTTPS_PROXY already.
> 
> Not according to my testing, or at least not with a socks5 proxy.

I just did a test, and only adding "socks" to the features is enough, so that would be a go for me :).

I do not want to add the HTTP_PROXY and HTTPS_PROXY to our config logic, as it only adds more stuff we do not need to handle our self's, since reqwest does this already.

Also, having just one location to create a reqwest client/builder is a nice item.

src/api/admin.rs
```diff
@@ -469,24 +469,22 @@ struct GitCommit {
...
 }
 
 fn has_http_access() -> bool {
-    let http_access = Client::builder().build().unwrap();
+    let http_access = get_reqwest_client();
 
     match http_access
         .head("https://github.com/dani-garcia/bitwarden_rs")
         .timeout(Duration::from_secs(10))
-        .header(USER_AGENT, "Bitwarden_RS")
```
williamdes, line 489: why removing this ?

BlackDex: Because that is done in the function in util.rs

williamdes: Okay, thank you :)

src/util.rs
```diff
@@ -522,3 +522,25 @@ where
...
+
+use reqwest::{blocking::{Client, ClientBuilder}, header, Proxy};
+
+pub fn get_reqwest_client() -> Client {
+    get_reqwest_client_builder().build().expect("Failed to build client")
+}
+
+pub fn get_reqwest_client_builder() -> ClientBuilder {
+    let mut headers = header::HeaderMap::new();
+    headers.insert(header::USER_AGENT, header::HeaderValue::from_static("Bitwarden_RS"));
```
williamdes, line 534: that would be a good time to add a new ENV for the user agent string ?

williamdes (approved):

src/util.rs
BlackDex: No, that is not something we will make variable in my opinion

RealOrangeOne:
> I just did a test, and only adding "socks" to the features is enough, so that would be a go for me :).

Once i'd sent this message, I had a feeling it was the `feature` which would change the support. I'll remove the custom configuration in that case and let `reqwest`'s internals deal with it.

BlackDex:
> > If I'm correct this is already working.
> > You should be able to use HTTP_PROXY/HTTPS_PROXY already.
> 
> Not according to my testing, or at least not with a socks5 proxy.

I just did a test, and only adding "socks" to the features is enough, so that would be a go for me :).

I do not want to add the HTTP_PROXY and HTTPS_PROXY to our config logic, as it only adds more stuff we do not need to handle our self's, since reqwest does this already.

Also, having just one location to create a reqwest client/builder is a nice item.

PR #1602, opened 2021-04-10
Warn that the SQLite backup feature doesn't produce a complete backup

jjlin:
Also add a link to the wiki page on backups.


BlackDex (approved):

BlackDex:
That was a great idea.

williamdes (approved):

PR #1604, opened 2021-04-12
Two-step login organization policy enforcement - Resolves dani-garcia/bitwarden_rs#981

olivierIllogika:
* a user without 2fa trying to join a 2fa org will fail, but user gets an email to enable 2fa
* a user disabling 2fa will be removed from 2fa orgs; user gets an email for each org
* an org enabling 2fa policy will remove users without 2fa; users get an email


pepa65:
Must users really be removed from orgs when they don't have the enforced 2fa? They won't be able to login, but once they set up 2fa, they should be able to -- I would think. I would prefer:

- User without 2fa can't join a 2fa org, but will get an email
- User disabling 2fa will get an email for each 2fa org
- Users without 2fs in an org enforcing 2fa will get an email

williamdes (approved):

olivierIllogika:
> Must users really be removed from orgs when they don't have the enforced 2fa? They won't be able to login, but once they set up 2fa, they should be able to -- I would think. I would prefer:
> 
> * User without 2fa can't join a 2fa org, but will get an email
> * User disabling 2fa will get an email for each 2fa org
> * Users without 2fs in an org enforcing 2fa will get an email

The proposed changes are certainly a harder security stance than what you suggest. As an organization admin, I **require** non-2fa users to loose access if I enforce 2fa. I think what you need is the creation of a new policy that _encourages_ 2fa.

For more details, here's the [official documentation on the policy](https://bitwarden.com/help/article/policies/#two-step-login) and the [changes from upstream](https://github.com/bitwarden/server/commit/81424a8526e4a22561597c98ae43fb8f2948596a); here's the longer [discussion on the bitwarden community forums](https://community.bitwarden.com/t/force-2fa-for-organizations/467/).

pepa65:
Maybe I don't understand how it works, but I think if an org starts enforcing 2fa, you must have 2fa, otherwise you cannot login. So non-2fa users automatically would lose access. But they are still members of your org (even if they can't access without 2fa), and once they get 2fa, they should be able to get access as they did before.
If not, users would need to be re-enrolled, while their identities have already been established and approved. In our organisation we encourage 2fa, but it is not (yet) mandatory. But once we start enforcing it, I would just like non-2fa people to (temporarily) loose access, but not to lose all connections to the org. That could lead to a lot of work for the admins (and the users) and delay the introduction of enforced 2fa in many organisations.

olivierIllogika:
@pepa65 I understand your argument and it makes total sense. Ideally, users without 2fa would get a status like "missing 2fa" preventing org access without being removed.

My understanding is that the project's priority is following upstream. My priority was getting enforceable 2fa policy for my own organization; I'm just sharing the result.

Maybe this discussion should be taken to the bitwarden community forums as a feature request or improvement on the current implementation. If the web-vault is updated to support the flow you're suggesting, I'll gladly make the backend modifications.

pepa65:
If that is how upstream operates, I understand we would by default follow that, it just doesn't make good sense to me. We would be in trouble if we would start enforcing 2fa, instantly losing most registered users. Everybody would be unhappy...

cksapp:
Just a thought, but would it be possible as a resolution to this concern be placing users back into the "Invited" status until they enable 2FA, and once enabling 2FA and being able to accept the invite to the organization be placed back into the "Accepted" status.
This also provides an opportunity to vault admins to verify fingerprint phrase again after enabling 2FA?

bokkabonga:
> 
> 
> Just a thought, but would it be possible as a resolution to this concern be placing users back into the "Invited" status until they enable 2FA, and once enabling 2FA and being able to accept the invite to the organization be placed back into the "Accepted" status.
> This also provides an opportunity to vault admins to verify fingerprint phrase again after enabling 2FA?
> 
> Edit: Not requesting this directly, will probably make an official feature request in their forums. Just food for thought in this case.

I really like this Idea as it makes a lot of sense. But when placing the users back into the invited status, this would most likely automatically send a new invitation mail, correct?

However i believe, that this is not how upstream works so we should try to stay close to the upstream project by completely removing the users from the org.

PR #1641, opened 2021-04-29
making SMTP force TLS description in admin panel clearer

jonathanmmm:
fixes https://github.com/dani-garcia/vaultwarden/discussions/1639

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -482,7 +482,7 @@ make_config! {
         smtp_host:                     String, true,   option;
         /// Enable Secure SMTP |> (Explicit) - Enabling this by default would use STARTTLS (Standard ports 587 or 25)
         smtp_ssl:                      bool,   true,   def,     true;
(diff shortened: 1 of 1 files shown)


BlackDex:
Thx for your contribution. But i think i will change the logic of that will not break other config. If that is not possible, then this comment could be useful.

jonathanmmm:
Ok, if I only checked force TLS I got an error, if this is true for everyone, then no configs can be destroyed, as they wouldn't work.

jonathanmmm:
Yes, you have to make both set to true still be working, but optional and in the future only one has to be set.
Maybe in the admin panel this can be archieved through a 3 choose button style (either no encryption, implicit or explicit), how this is solved via config file, I don't know right now. Maybe through a SMTP_security_level from 0 to 2 and still accept old flags set.

BlackDex:
I have changed the logic in my PR #1670.

jonathanmmm:
Great, ok, I closed my PR

PR #1653, opened 2021-05-01
Add support for password reprompt

jjlin:
Upstream PR: https://github.com/bitwarden/server/pull/1269


williamdes (approved):

dani-garcia:
It doesn't seem this has support in any of the clients other than jslib yet, so I'll wait until we have something to test it on before merging in case they are any more changes.

jjlin:
Sure, there's no hurry.

PR #1663, opened 2021-05-03
admin: Return newly-created user in invite_user

dongcarl:
```
Instead of having the caller dig through /admin/users for the right one,
just return the user upon creation.
```

### DIFF
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -12,7 +12,7 @@ use rocket::{
 use rocket_contrib::json::Json;
 
 use crate::{
-    api::{ApiResult, EmptyResult, NumberOrString},
+    api::{ApiResult, EmptyResult, JsonResult, NumberOrString},
     auth::{decode_admin, encode_jwt, generate_admin_claims, ClientIp},
     config::ConfigBuilder,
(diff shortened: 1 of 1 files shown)


williamdes (approved):

BlackDex (approved):
Doesn't seem to break the normal working. Just keep in mind that if something fails, like sending out the e-mail, you won't get the JSON back with the user data, but that user will be created.

We may need look at that scenario, but that has nothing to do with this PR.
This is fine by me.

dongcarl:
> Just keep in mind that if something fails, like sending out the e-mail, you won't get the JSON back with the user data, but that user will be created.

Hmmm... What if we move `user.save(&conn)?;` to right before the `Ok(Json(user.to_json(&conn)))`?

BlackDex:
> > Just keep in mind that if something fails, like sending out the e-mail, you won't get the JSON back with the user data, but that user will be created.
> 
> Hmmm... What if we move `user.save(&conn)?;` to right before the `Ok(Json(user.to_json(&conn)))`?

That would work indeed :).
Feel free to add that this this PR :).

dongcarl:
>  Feel free to add it to this this PR :).

Added! Also added an endpoint to get info for a particular user (specified by UUID).

PR #1679, opened 2021-05-08
Updated Pipelines and fixed new Hadolints

BlackDex:
- Removed azure-pipelines
- Updated gh-actions to run `cargo test` per db feature
- Fail on warnings by adding `RUSTFLAGS` env
- Updated Dockerfile to fix some new hadolint warnings


BlackDex:
@dani-garcia You probably need to disable the azure-pipeline with this PR since i removed that file.

williamdes (approved):

PR #1702, opened 2021-05-16
Updated icon fetching and crates.

BlackDex:
- Updated some crates
- Updated icon fetching code:
  + Use a cookie jar and set Max-Age to 2 minutes for all cookies
  + Locate the base href tag to fix some locations
  + Changed User-Agent (Helps on some sites to get HTML instead of JS)
  + Reduced HTML code limit from 512KB to 384KB
  + Allow some large icons higer-up in the sort
  + Allow GIF images
  + Ignore cookie_store and hyper::client debug messages

### DIFF
--- a/src/api/icons.rs
+++ b/src/api/icons.rs
@@ -3,14 +3,14 @@ use std::{
     fs::{create_dir_all, remove_file, symlink_metadata, File},
     io::prelude::*,
     net::{IpAddr, ToSocketAddrs},
-    sync::RwLock,
+    sync::{Arc, RwLock},
     time::{Duration, SystemTime},
 };
 
 use once_cell::sync::Lazy;
 use regex::Regex;
-use reqwest::{blocking::Client, blocking::Response, header, Url};
-use rocket::{http::ContentType, http::Cookie, response::Content, Route};
+use reqwest::{blocking::Client, blocking::Response, header};
+use rocket::{http::ContentType, response::Content, Route};
 
 use crate::{
     error::Error,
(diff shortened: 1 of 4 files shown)


src/api/icons.rs
```diff
@@ -25,19 +25,17 @@ pub fn routes() -> Vec<Route> {
 static CLIENT: Lazy<Client> = Lazy::new(|| {
     // Generate the default headers
     let mut default_headers = header::HeaderMap::new();
-    default_headers.insert(header::USER_AGENT, header::HeaderValue::from_static("Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_4) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/13.1.1 Safari/605.1.15"));
-    default_headers.insert(header::ACCEPT_LANGUAGE, header::HeaderValue::from_static("en-US,en;q=0.8"));
+    default_headers
+        .insert(header::USER_AGENT, header::HeaderValue::from_static("Links (2.22; Linux X86_64; GNU C; text)"));
```
dani-garcia, line 29: I'm surpised this doesn't break more sites with generic "your browser is not supported" messages or redirects.

BlackDex: I was too, but apparently some sites do not serve javascript redirects in these cases. Besides that most just serve there normal page.

I tested it with all the sites i have my self and from reports from people having issues. They all seem to work besides a few with CloudFlare protection, basic auth, javascript redirects or JavaScript icon url changers.

williamdes (approved):

PR #1706, opened 2021-05-17
Add `TRASH_AUTO_DELETE_DAYS` to .env.template

jjlin:
This should have been added as part of #1579.


BlackDex (approved):

BlackDex:
Well, github doesn't seem to work that well. There should be no need for a check, but it still forces this before i can merge it seems.

williamdes (approved):

PR #1736, opened 2021-05-29
Clarify Rocket env var defaults

### LINKED ISSUE #1735: [Documentation] What's the default value of `ROCKET_PORT`?
According to `.env.tamplate`, it seems to be `8000`:

https://github.com/dani-garcia/vaultwarden/blob/7d5186e40aa7950df8f03e1df727e2390e25c77c/.env.template#L2

https://github.com/dani-garcia/vaultwarden/blob/7d5186e40aa7950df8f03e1df727e2390e25c77c/.env.template#L264

But `docker/healthcheck.sh` says it's `80`:

https://github.com/dani-garcia/vaultwarden/blob/7d5186e40aa7950df8f03e1df727e2390e25c77c/docker/healthcheck.sh#L3-L6

jjlin:
Resolves #1735.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -258,10 +258,11 @@
 ## In any case, if a code has been used it can not be used again, also codes which predates it will be invalid.
 # AUTHENTICATOR_DISABLE_TIME_DRIFT=false
 
-## Rocket specific settings, check Rocket documentation to learn more
-# ROCKET_ENV=staging
-# ROCKET_ADDRESS=0.0.0.0 # Enable this to test mobile app
-# ROCKET_PORT=8000
+## Rocket specific settings
+## See https://rocket.rs/v0.4/guide/configuration/ for more details.
+# ROCKET_ADDRESS=0.0.0.0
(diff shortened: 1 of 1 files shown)


dani-garcia:
In the docker image we force ROCKET_ENV=staging and ROCKET_PORT=80, while running the binary by itself would give  ROCKET_ENV=dev and ROCKET_PORT=8000 by default. I'm partial to including the docker defaults as it's the most official usage method, but maybe it should be mentioned that those values are only applicable to the docker image?

jjlin:
An extra note for `ROCKET_PORT` is fine with me, but I'm not sure `ROCKET_ENV` even needs to be in `.env.template`, as the vast majority of users aren't going to use it or even understand what it is. There are many other potential Rocket env vars not listed in `.env.template` anyway, so advanced users who need it can figure it out from the docs.

Also, according to the docs, `ROCKET_ENV` would default to `production` for release builds, but I'm not sure why `ROCKET_ENV` is even set in the Dockerfile as the Docker image doesn't include any env-specific configuration AFAIK.

BlackDex (approved):

PR #1761, opened 2021-06-09
Update dependencies

jjlin:
Notably, update `diesel` to 1.4.7 and `libsqlite3-sys` to 0.22.2 to pick up
the fix for CVE-2021-20227 added in SQLite 3.34.1.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.15.1"
+version = "0.15.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "03345e98af8f3d786b6d9f656ccfa6ac316d954e92bc4841f0bba20789d5fb5a"
+checksum = "e7a2e47a1fbe209ee101dd6d61285226744c6c8d3c21c8dc878ba6cb9f467f3a"
 dependencies = [
  "gimli",
 ]
@@ -74,9 +74,9 @@ checksum = "cdb031dd78e28731d87d56cc8ffef4a8f36ca26c38fe2de700543e627f8a464a"
 
 [[package]]
 name = "backtrace"
-version = "0.3.59"
+version = "0.3.60"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4717cfcbfaa661a0fd48f8453951837ae7e8f81e481fbb136e3202d72805a744"
+checksum = "b7815ea54e4d821e791162e078acbebfd6d8c8939cd559c9335dceb1c8ca7282"
 dependencies = [
  "addr2line",
  "cc",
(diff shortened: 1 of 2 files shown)


BlackDex:
Isn't it maybe a good idea to update the rest of the packages also?
Like do an `cargo update` maybe and update all the other dependencies, maybe not handlebars right now, because that will take some other modifications i think.

jjlin:
I did run `cargo update` -- there are many other changes in `Cargo.lock`. I'm just calling out a notable change that would not have happened on its own with just `cargo update`.

BlackDex:
Isn't it then a good thing to also update the `Cargo.toml` to represent the same versions as are now in the `Cargo.lock`?

jjlin:
I wouldn't mind, but `cargo update` doesn't do that AFAIK (see e.g., rust-lang/rust#57443). I haven't looked into whether there are other tools that make that change in an automated way, but I'd rather not do it manually.

dani-garcia:
I usually use `cargo upgrade --dry-run` (from [cargo-edit](https://crates.io/crates/cargo-edit)) as an easy way to check for semver incompatible dependency updates, and for bumping the versions in the Cargo.toml. Not sure if it's good practice or not, simply something I've been doing myself

jjlin:
Ah, good to know. I'm personally okay with the default `cargo update` version selection behavior, and it looks like following that with `cargo upgrade --to-lockfile` would then update `Cargo.toml` to match what was updated in `Cargo.lock`.

Running `cargo upgrade` alone seems to actually downgrade a bunch of crates, presumably as a consequence of `rocket v0.5.0-dev -> v0.4.7`? Even adding `--allow-prerelease` doesn't seem to change this. So how are you guys getting this to work?

```
$ cargo upgrade --dry-run
    Updating 'https://github.com/rust-lang/crates.io-index' index
Starting dry run. Changes will not be saved.
vaultwarden:
    Upgrading lettre v0.10.0-rc.1 -> v0.10.0-alpha.5
    Upgrading regex v1.5.4 -> v1.4.3
    Upgrading yubico v0.10.0 -> v0.9.2
    Upgrading openssl v0.10.34 -> v0.10.32
    Upgrading cookie_store v0.15.0 -> v0.12.0
    Upgrading diesel v1.4.7 -> v1.4.5
    Upgrading handlebars v3.5.5 -> v3.5.2
    Upgrading idna v0.2.3 -> v0.2.1
    Upgrading syslog v4.0.1 -> v5.0.0
    Upgrading libsqlite3-sys v0.22.2 -> v0.20.1
    Upgrading serde v1.0.126 -> v1.0.123
    Upgrading rocket_contrib v0.5.0-dev -> v0.4.7
    Upgrading pico-args v0.4.1 -> v0.4.0
    Upgrading backtrace v0.3.59 -> v0.3.56
    Upgrading time v0.2.26 -> v0.2.25
    Upgrading reqwest v0.11.3 -> v0.11.0
    Upgrading tracing v0.1.26 -> v0.1.23
    Upgrading cookie v0.15.0 -> v0.14.3
    Upgrading paste v1.0.5 -> v1.0.4
    Upgrading url v2.2.2 -> v2.2.0
    Upgrading serde_json v1.0.64 -> v1.0.62
    Upgrading once_cell v1.7.2 -> v1.5.2
    Upgrading rocket v0.5.0-dev -> v0.4.7
```

PR #1789, opened 2021-06-22
Only create private key in pem format if it doesn't already exist

Bobonium:
The original logic checks if the private and public key are available in the `.der` format, but ignores the private key in `.pem` format.

This change does not have any impact on the already working setups, it merely enables the User to only provide a single private key in `.pem` format.

My use case is simple. I deploy vaultwarden on kubernetes through terraform and I'm unable to create private keys in `.der` format directly. By simply checking for an existing `.pem` key first, I can simply mount the key that I've created with terraform and on startup the application will automatically convert it to the correct format.

Since the `CONFIG` provides methods to get the key names I also switched to using those instead of generating the names.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -263,17 +263,19 @@ fn check_rsa_keys() {
 
         info!("OpenSSL detected, creating keys...");
 
-        let key = CONFIG.rsa_key_filename();
+        let pem = CONFIG.private_rsa_key_pem();
+        let priv_der = CONFIG.private_rsa_key();
+        let pub_der = CONFIG.public_rsa_key();
 
-        let pem = format!("{}.pem", key);
-        let priv_der = format!("{}.der", key);
-        let pub_der = format!("{}.pub.der", key);
+        let mut success = true;
 
-        let mut success = Command::new("openssl")
-            .args(&["genrsa", "-out", &pem])
-            .status()
-            .expect("Failed to create private pem file")
-            .success();
+        if !util::file_exists(&pem) {
+            success = Command::new("openssl")
+                .args(&["genrsa", "-out", &pem])
+                .status()
+                .expect("Failed to create private pem file")
+                .success();
+        }
 
         success &= Command::new("openssl")
             .args(&["rsa", "-in", &pem, "-outform", "DER", "-out", &priv_der])


BlackDex:
May i ask which kind of certs you want to provide here? The location you modified is only used for the JWT to generate tokens for login etc. 
These files either get auto generated if they do not exist or just read of they are, and they should be stored on a volume of shared storage.

Bobonium:
The current logic only checks for the key in the `.der` format before automatically generating first the private key in `.pem` and then using that key to generate the `.der` format for public and private keys

In my use case I only want to provide the private key in `.pem` and let the application generate the `.der` files on startup
Using the same `.pem` key on multiple startups will of course generate the same `.der` files, as it's still the same private key, merely in another file format.

The reason why I do have this specific use case is https://registry.terraform.io/providers/hashicorp/tls/latest/docs/resources/private_key#private_key_pem to be precise. If the application can't simply handle the format conversion without key generation, then I'd be left with either of three options:

Option 1: build a hacky solution that depends on a local openssl installation to generate the `.der` file format and store that in a secret in Kubernetes and mount from there

Option 2:  completely ignore this, and get a new JWT on each startup, which invalidates existing sessions (this it what I currenly do for now)

 Option 3: create a persistent volume and store the data on there. This is the worst approach as persistence in kubernetes introduces a lot of potential problems. Best example would be that I could only do rolling upgrades without downtime on a deployment if the underlying PVC would be ReadWriteMany, which are a nightmare to begin with.

BlackDex:
Also, i think there is an other option. We could backport https://github.com/dani-garcia/vaultwarden/blob/436d8860594f76bc8d01571edf4519cf9ce0485b/src/main.rs#L244 i think that could solve your issue.

Bobonium:
> Also, i think there is an other option. We could backport
> 
> https://github.com/dani-garcia/vaultwarden/blob/436d8860594f76bc8d01571edf4519cf9ce0485b/src/main.rs#L244
> i think that could solve your issue.

that specifically would not solve the problem. because of https://github.com/dani-garcia/vaultwarden/blob/436d8860594f76bc8d01571edf4519cf9ce0485b/src/main.rs#L249-L255


To simplify the problem:
There's 3 files being worked with in vaultwarden here. The `priv.pem`, the `priv.der` and the `pub.der`
`priv.pem` is generated first and from that the `priv.der` and `priv.pub` are generated

The current logic only checks if `priv.der and `pub.der` exist, so it would work if you'd simple provide the private and public keys in the `.der` format and do not put the `.pem` format into the directory.

What currently does not work is to put specifically only the `priv.pem` file into the dir. The application will then fail to start, if it can't overwrite this file

The code that you provided would still have the exact problem because it would not allow to put only provide the `.pem` file format format

BlackDex:
Correct me if im wrong, but terraform will output both priv and pub pem files, which are the only files read (or created if missing) in the async branch. You should be able to provide both right?

In any case, if something is going to be changed here, i want it to not use the openssl binary at all, and more match the async branch using the library.

williamdes (approved):
I am not sure If I understood well the debate but, I think pem is easier than der
And for the code side, checking if is does not already exists is a good solution

That said the code probably needs one shot of `cargofmt` (space before equals sign)

BlackDex:
> I am not sure If I understood well the debate but, I think pem is easier than der
> And for the code side, checking if is does not already exists is a good solution

That is why i think the async branched solution would be enough, it only uses pem, no der.

Bobonium:
> I am not sure If I understood well the debate but, I think pem is easier than der
> And for the code side, checking if is does not already exists is a good solution
> 
> That said the code probably needs one shot of `cargofmt` (space before equals sign)

I just updated the branch after running `cargo fmt`



> > I am not sure If I understood well the debate but, I think pem is easier than der
> > And for the code side, checking if is does not already exists is a good solution
> 
> That is why i think the async branched solution would be enough, it only uses pem, no der.

Ah sorry about that, I was mistaken earlier. I had a quick look at the code that you linked and thought that it also creates a pem format and converts it over. After having another proper look at it right now, you are in fact correct. The Async branch code would resolve the same problem. Sorry for any confusion this might've caused

dani-garcia:
I've backported the change from the async branch that @BlackDex mentions, which I think should fix this issue, as I plan to refactor that branch anyway.

Now we don't use any der files, the only file that is really needed is the private pem file, as long as that exists, the public pem will be derived from it. If neither exist they will both be created instead.

https://github.com/dani-garcia/vaultwarden/blob/46e0f3c43a81ce9411612c152e414162a9c220ac/src/main.rs#L255-L274

Thanks for your PR!

Issue #411, opened 2019-02-23
Docker: Unable to run as non-root user


rcdailey:
Using the `latest` tag as of today (2/23/2019), I'm unable to launch the docker image provided by @mprasil. I get this error after `docker-compose run`:

```
bitwarden | [2019-02-23 11:33:55][launch][INFO] Configured for staging.
bitwarden | [2019-02-23 11:33:55][launch_][INFO] address: 0.0.0.0
bitwarden | [2019-02-23 11:33:55][launch_][INFO] port: 80
bitwarden | [2019-02-23 11:33:55][launch_][INFO] log: normal
bitwarden | [2019-02-23 11:33:55][launch_][INFO] workers: 10
bitwarden | [2019-02-23 11:33:55][launch_][INFO] secret key: generated
bitwarden | [2019-02-23 11:33:55][launch_][INFO] limits: forms = 32KiB, json* = 10MiB
bitwarden | [2019-02-23 11:33:55][launch_][INFO] keep-alive: 5s
bitwarden | [2019-02-23 11:33:55][launch_][INFO] tls: disabled
bitwarden | [2019-02-23 11:33:55][rocket::fairing::fairings][INFO] Fairings:
bitwarden | [2019-02-23 11:33:55][_][INFO] 1 response: Application Headers
bitwarden | [2019-02-23 11:33:55][bitwarden_rs][ERROR] Launch error Bind(Io(Os { code: 13, kind: PermissionDenied, message: "Permission denied" }))
```

My `docker-compose.yml` is as follows:

```yml
version: '3.7'

services:
  app:
    image: mprasil/bitwarden
    container_name: bitwarden
    user: $UID:$GID
    networks:
      - reverse_proxy
    volumes:
      - ./data:/data
      - /etc/timezone:/etc/timezone:ro
    environment:
      - TZ=America/Chicago

networks:
  reverse_proxy:
    external: 'true'
    name: reverse_proxy
```

And yes, the permissions of my `./data` directory are set properly. The same user referred to by `UID` and `GID` also owns the `./data` directory.

mprasil:
Please see #287. You're definitely missing environment variable `ROCKET_PORT`set to something above `1024` (only root can bind to first `1024` ports) and it seems like you also might need to add `DATA_FOLDER` set to `/data`.

Hope that helps, please continue any discussion under #287 to avoid duplicates.

rcdailey:
[...]

Issue #287, opened 2018-12-13
[Docker] Unable to connect to database when using non-root user


JuniorJPDJ:
I wanted to use non-root user and container didn't start.
```
bitwarden_1  | thread 'main' panicked at 'Can't conect to DB: BadConnection("Unable to open the database file")', src/libcore/result.rs:1009:5
```

dani-garcia:
I'm not very knowledgeable in docker, but I imagine this is because the non-root user doesn't have permission in the data folder.

bassarf:
As far as I know, and as far as my 2 minute google research goes, this is a docker thing.
The docker deamon runs as root. But on the docker website there is a manual for non-root users to manage docker: https://docs.docker.com/install/linux/linux-postinstall/

IMO this issue should be closed because it is not specific to bitwarden_rs but rather being a docker thing

JuniorJPDJ:
@bassarf that issue is not about starting container as non-root user, it's about dropping privileges of bitwarden_rs running in docker container to other (non-root) user.
Still, thank you for your interest, my description was misleading.

Sohalt:
You will have to set the correct permissions on the `/data` volume. In addition you need to adjust the `ROCKET_PORT` to something above 1024, as everything below is a privileged port and can only be opened by root.

p3lim:
I had an issue similar to this: `[bitwarden_rs][ERROR] Error creating database directory`
I solved it by setting `DATA_FOLDER` to a subdirectory of the `/data` directory/volume, e.g. `/data/data`.

mprasil:
I've documented the procedure in [wiki page](https://github.com/dani-garcia/bitwarden_rs/wiki/Running-docker-container-with-non-root-user), so I believe this is resolved. I ran the container as `nobody` following the wiki, but feel free to re-open or open new issue if you encounter any problems.

PR #1838, opened 2021-07-05
Expose $ROCKET_PORT instead of 80

waschinski:
EXPOSE in the Dockerfiles is hard-coded to 80 instead using the defined rocket port. This is causing problems e.g. when setting up a reverse proxy via Traefik using labels as Traefik will use the ports set by the EXPOSE command. Current workaround is to provide the rocket port to the service using an extra label.

### DIFF
--- a/docker/Dockerfile.j2
+++ b/docker/Dockerfile.j2
@@ -239,7 +239,7 @@ RUN [ "cross-build-end" ]
 {% endif %}
 
 VOLUME /data
-EXPOSE 80
+EXPOSE $ROCKET_PORT
 EXPOSE 3012
 
 # Copies the files from the context (Rocket.toml file and web-vault)
--- a/docker/amd64/Dockerfile
+++ b/docker/amd64/Dockerfile
@@ -94,7 +94,7 @@ RUN mkdir /data \
 
 
 VOLUME /data
-EXPOSE 80
+EXPOSE $ROCKET_PORT
 EXPOSE 3012
 
 # Copies the files from the context (Rocket.toml file and web-vault)
--- a/docker/amd64/Dockerfile.alpine
+++ b/docker/amd64/Dockerfile.alpine
@@ -89,7 +89,7 @@ RUN mkdir /data \
 
 
 VOLUME /data
-EXPOSE 80
+EXPOSE $ROCKET_PORT
 EXPOSE 3012
 
 # Copies the files from the context (Rocket.toml file and web-vault)
--- a/docker/arm64/Dockerfile
+++ b/docker/arm64/Dockerfile
@@ -134,7 +134,7 @@ RUN mkdir /data \
 RUN [ "cross-build-end" ]
 
 VOLUME /data
-EXPOSE 80
+EXPOSE $ROCKET_PORT
 EXPOSE 3012
 
 # Copies the files from the context (Rocket.toml file and web-vault)
--- a/docker/armv6/Dockerfile
+++ b/docker/armv6/Dockerfile
@@ -134,7 +134,7 @@ RUN mkdir /data \
 RUN [ "cross-build-end" ]
 
 VOLUME /data
-EXPOSE 80
+EXPOSE $ROCKET_PORT
 EXPOSE 3012
 
 # Copies the files from the context (Rocket.toml file and web-vault)
--- a/docker/armv7/Dockerfile
+++ b/docker/armv7/Dockerfile
@@ -134,7 +134,7 @@ RUN mkdir /data \
 RUN [ "cross-build-end" ]
 
 VOLUME /data
-EXPOSE 80
+EXPOSE $ROCKET_PORT
 EXPOSE 3012
 
 # Copies the files from the context (Rocket.toml file and web-vault)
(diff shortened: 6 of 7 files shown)


williamdes (approved):

RealOrangeOne:
My understanding of this is that it'll only actually expose whatever's set as `$ROCKET_PORT` at container build time, rather than at runtime? Have you tested this working? 

Additionally, if you're running this in docker behind Traefik, does the port it's running on really matter? [see my config](https://github.com/RealOrangeOne/infrastructure/blob/master/ansible/roles/vaultwarden/files/docker-compose.yml)

waschinski:
> My understanding of this is that it'll only actually expose whatever's set as `$ROCKET_PORT` at container build time, rather than at runtime? Have you tested this working?
> 
> Additionally, if you're running this in docker behind Traefik, does the port it's running on really matter? [see my config](https://github.com/RealOrangeOne/infrastructure/blob/master/ansible/roles/vaultwarden/files/docker-compose.yml?rgh-link-date=2021-07-05T14%3A47%3A43Z)

The `EXPOSE` instruction just informs Docker that the container listens on that specified network port at runtime. Traefik's [port detection](https://doc.traefik.io/traefik/providers/docker/#port-detection) is using the Docker API to get that information for configuration discovery.

You are right though - I was just reading the documentation again - and this will not solve the issue having to set up a certain label like you did in your configuration. Because:
> If a container exposes multiple ports [...] then you must manually specify which port Traefik should use for communication by using the label `traefik.http.services.<service_name>.loadbalancer.server.port`".

Anyway, exposing port 80 while actually listening to a different port is still simply wrong, isn't it?

RealOrangeOne:
> exposing port 80 while actually listening to a different port is still simply wrong, isn't it?

It's not really that simple - What you're after isn't really possible with Docker. 

In your case, you either force Traefik to listen on your custom port, or stop specifying the custom port altogether. Either would work, either are correct, although personally i'd err towards defining less and just having vaultwarden listen on port 80.

waschinski:
> > exposing port 80 while actually listening to a different port is still simply wrong, isn't it?
> 
> It's not really that simple - What you're after isn't really possible with Docker.

All I did was changing the server port to 8080 and I am also using this in my docker port configuration. So technically it is possible even if it may be unnecessary in my use case, you are absolutely right on that. Though I still think the port in use should also be exposed then.

But I have no hard feelings and I now see that the use case is not really there, so if it's deemed to be unnecessary simply close the PR. :)

> In your case, you either force Traefik to listen on your custom port, or stop specifying the custom port altogether. Either would work, either are correct, although personally i'd err towards defining less and just having vaultwarden listen on port 80.

I went with the former even though I actually don't know why I did so after all and will revert that. Thanks for pointing this out.

waschinski:
So while switching back to port 80 and getting errors I remembered why I did change the port after all. [Using a non-root user you _have_ to change the port to something other than 80.](https://github.com/dani-garcia/vaultwarden/issues/287#issuecomment-460013030)

RealOrangeOne:
Mine runs absolutely fine running as a non-root user on port 80. Have you definitely tested it doesn't work?

waschinski:
Yes, it was throwing an error. Pretty sure it was the same as in this (pretty old) [issue](https://github.com/dani-garcia/vaultwarden/issues/411). And I just checked, it's even mentioned in the [Wiki](https://github.com/dani-garcia/vaultwarden/wiki/Hardening-Guide#run-as-a-non-root-user).

BlackDex:
The EXPOSE instruction only tells the user which are the default ports configured and if the default is used and someone uses the `-P` flag on docker It will actually expose those ports. If you do not define the `-P` or `-p 8080:80` flags, nothing will be mapped no ports will be exposed what so ever.

So, this change has no effect unless someone or some software uses these values to do the mapping, and then it will only be one on one.

Now, i don't think it will break anything probably, it doesn't make it more clear to the users which are the default ports which could be exposed.

waschinski:
So we agree this change will at least fix using `-P` and won't break other uses.

In addition, when looking at the [documentation](https://github.com/dani-garcia/vaultwarden/wiki/Starting-a-Container) it even says `The service will be exposed on host-port 80 or 8080.` referring to the `Podman as non-root` example which simply is _not_ the case right now. Isn't this PR fixing this after all?

jjlin:
No, this doesn't fix `-P`.

`$ROCKET_PORT` in `EXPOSE $ROCKET_PORT` is not resolved dynamically when the container is created. `EXPOSE` instructions are evaluated at image build time, so this has exactly the same effect as `EXPOSE 80` since `ROCKET_PORT` has the value 80 in the Dockerfile.

You can easily verify for yourself that after `docker run -e ROCKET_PORT=8080 ...`, `docker ps` still lists `80/tcp, 3012/tcp` under the `PORTS` column.

waschinski:
> `$ROCKET_PORT` in `EXPOSE $ROCKET_PORT` is not resolved dynamically when the container is created. `EXPOSE` instructions are evaluated at image build time, so this has exactly the same effect as `EXPOSE 80` since `ROCKET_PORT` has the value 80 in the Dockerfile.

Now I finally got what @RealOrangeOne meant right away. Sorry it took a while and thanks for taking the time explaining it again.

So in order to have the port exposed "dynamically" a build argument would have to be used/passed at build time which would then result in a separate image being built for exactly that defined port - not very dynamic after all. :/

PR #1839, opened 2021-07-05
Updated attachment limit descriptions

### LINKED ISSUE #1818: Attachment size limit displayed to user does not match the limit imposed in the Vaultwarden admin panel
This upload will fail and will return the shown error due to the size restriction set by the admin using the admin panel. It is set to 50MB in this example.

BlackDex:
The user and org attachment limit use `size` as wording while it should
have been `storage` since it isn't per attachment, but the sum of all attachments.

- Changed the wording in the config/env
- Changed the wording of the error messages.

Resolves #1818

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -194,11 +194,13 @@
 ## Name shown in the invitation emails that don't come from a specific organization
 # INVITATION_ORG_NAME=Vaultwarden
 
-## Per-organization attachment limit (KB)
-## Limit in kilobytes for an organization attachments, once the limit is exceeded it won't be possible to upload more
+## Per-organization attachment storage limit (KB)
+## Max kilobytes of attachment storage allowed per organization.
+## When this limit is reached, organization members will not be allowed to upload further attachments for ciphers owned by that organization.
 # ORG_ATTACHMENT_LIMIT=
-## Per-user attachment limit (KB).
-## Limit in kilobytes for a users attachments, once the limit is exceeded it won't be possible to upload more
+## Per-user attachment storage limit (KB)
+## Max kilobytes of attachment storage allowed per user.
+## When this limit is reached, the user will not be allowed to upload further attachments.
 # USER_ATTACHMENT_LIMIT=
 
 ## Number of days to wait before auto-deleting a trashed item.
--- a/src/config.rs
+++ b/src/config.rs
@@ -356,9 +356,9 @@ make_config! {
         /// HIBP Api Key |> HaveIBeenPwned API Key, request it here: https://haveibeenpwned.com/API/Key
         hibp_api_key:           Pass,   true,   option;
 
-        /// Per-user attachment limit (KB) |> Limit in kilobytes for a users attachments, once the limit is exceeded it won't be possible to upload more
+        /// Per-user attachment storage limit (KB) |> Max kilobytes of attachment storage allowed per user. When this limit is reached, the user will not be allowed to upload further attachments.
         user_attachment_limit:  i64,    true,   option;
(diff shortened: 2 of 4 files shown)


.env.template
```diff
@@ -194,11 +194,11 @@
 ## Name shown in the invitation emails that don't come from a specific organization
 # INVITATION_ORG_NAME=Vaultwarden
 
-## Per-organization attachment limit (KB)
-## Limit in kilobytes for an organization attachments, once the limit is exceeded it won't be possible to upload more
+## Per org attachment storage limit (KB)
```
williamdes, line 197: I would argue organization is better not abbreviated for non English native users

williamdes (approved):

src/config.rs
```diff
@@ -356,9 +356,9 @@ make_config! {
         /// HIBP Api Key |> HaveIBeenPwned API Key, request it here: https://haveibeenpwned.com/API/Key
         hibp_api_key:           Pass,   true,   option;
 
-        /// Per-user attachment limit (KB) |> Limit in kilobytes for a users attachments, once the limit is exceeded it won't be possible to upload more
+        /// Per user attachment storage limit (KB) |> Limit in kilobytes of the maximum allowed per user attachments storage, when the limit exceeds it will not be possible to upload anymore
```
jjlin, line 359:
```suggestion
        /// Per-user attachment storage limit (KB) |> Max kilobytes of attachment storage allowed per user. When this limit is reached, the user will not be allowed to upload further attachments.
```

src/config.rs
```diff
@@ -356,9 +356,9 @@ make_config! {
...
         user_attachment_limit:  i64,    true,   option;
-        /// Per-organization attachment limit (KB) |> Limit in kilobytes for an organization attachments, once the limit is exceeded it won't be possible to upload more
+        /// Per org attachment storage limit (KB) |> Limit in kilobytes of the maximum allowed per user attachments storage, when the limit exceeds it will not be possible to upload anymore
```
jjlin, line 361:
```suggestion
        /// Per-organization attachment storage limit (KB) |> Max kilobytes of attachment storage allowed per org. When this limit is reached, org members will not be allowed to upload further attachments for ciphers owned by that org.
```

.env.template
```diff
@@ -194,11 +194,11 @@
...
+## Limit in kilobytes of the maximum allowed per user attachments storage, when the limit exceeds it will not be possible to upload anymore
```
jjlin, line 198:
```suggestion
## Per-organization attachment storage limit (KB)
## Max kilobytes of attachment storage allowed per org.
## When this limit is reached, org members will not be allowed to upload further attachments for ciphers owned by that org.
```

.env.template
```diff
@@ -194,11 +194,11 @@
...
 # ORG_ATTACHMENT_LIMIT=
-## Per-user attachment limit (KB).
-## Limit in kilobytes for a users attachments, once the limit is exceeded it won't be possible to upload more
+## Per user attachment storage limit (KB)
+## Limit in kilobytes of the maximum allowed per user attachments storage, when the limit exceeds it will not be possible to upload anymore
```
jjlin, line 201:
```suggestion
## Per-user attachment storage limit (KB)
## Max kilobytes of attachment storage allowed per user.
## When this limit is reached, the user will not be allowed to upload further attachments.
```

.env.template
BlackDex: It is used within a lot of comments. Also, we do not support other languages then English (yet).

BlackDex:
@jjlin Thanks for the grammar/spell check.

PR #1845, opened 2021-07-07
Add Edge's frame-ancestors

umireon:
Edge's frame-ancestors are required for Edge extension to do WebAuthn.

スクリーンショット 2021-07-08 0 41 26

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -29,7 +29,7 @@ impl Fairing for AppHeaders {
         res.set_raw_header("X-Content-Type-Options", "nosniff");
         res.set_raw_header("X-XSS-Protection", "1; mode=block");
         let csp = format!(
-            "frame-ancestors 'self' chrome-extension://nngceckbapebfimnlniiiahkandclblb moz-extension://* {};",
+            "frame-ancestors 'self' chrome-extension://nngceckbapebfimnlniiiahkandclblb chrome-extension://jbkfoedolllekgbhcbcoahefnbanhhlh moz-extension://* {};",
             CONFIG.allowed_iframe_ancestors()
         );
         res.set_raw_header("Content-Security-Policy", csp);
--- a/src/util.rs
+++ b/src/util.rs
@@ -31,7 +31,7 @@ impl Fairing for AppHeaders {
         let csp = format!(
             // Chrome Web Store: https://chrome.google.com/webstore/detail/bitwarden-free-password-m/nngceckbapebfimnlniiiahkandclblb
             // Edge Add-ons: https://microsoftedge.microsoft.com/addons/detail/bitwarden-free-password/jbkfoedolllekgbhcbcoahefnbanhhlh?hl=en-US
-            // Firefox Browser Add-ons: https://addons.mozilla.org/ja/firefox/addon/bitwarden-password-manager/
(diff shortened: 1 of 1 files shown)


umireon:
`chrome-extension://*` also works.

BlackDex:
> `chrome-extension://*` also works.

Which would mean you allow every extension out there.

dani-garcia:
I think the mozilla one was autogenerated for each client? Some kind of security measure to keep extensions from messing with each other or something like that. Otherwise I would have included it explicitly.

src/util.rs
```diff
@@ -29,7 +29,7 @@ impl Fairing for AppHeaders {
         res.set_raw_header("X-Content-Type-Options", "nosniff");
         res.set_raw_header("X-XSS-Protection", "1; mode=block");
         let csp = format!(
-            "frame-ancestors 'self' chrome-extension://nngceckbapebfimnlniiiahkandclblb moz-extension://* {};",
+            "frame-ancestors 'self' chrome-extension://nngceckbapebfimnlniiiahkandclblb chrome-extension://jbkfoedolllekgbhcbcoahefnbanhhlh moz-extension://* {};",
```
williamdes, line 32: it would be nice to have a comment above, linking to webstores for each extension listed

umireon: Added.

src/util.rs
```diff
@@ -29,7 +29,10 @@ impl Fairing for AppHeaders {
         res.set_raw_header("X-Content-Type-Options", "nosniff");
         res.set_raw_header("X-XSS-Protection", "1; mode=block");
         let csp = format!(
-            "frame-ancestors 'self' chrome-extension://nngceckbapebfimnlniiiahkandclblb moz-extension://* {};",
+            // Chrome Web Store: https://chrome.google.com/webstore/detail/bitwarden-free-password-m/nngceckbapebfimnlniiiahkandclblb
+            // Edge Add-ons: https://microsoftedge.microsoft.com/addons/detail/bitwarden-free-password/jbkfoedolllekgbhcbcoahefnbanhhlh?hl=en-US
+            // Firefox Browser Add-ons: https://addons.mozilla.org/ja/firefox/addon/bitwarden-password-manager/
```
williamdes, line 34:
```suggestion
            // Firefox Browser Add-ons: https://addons.mozilla.org/en-US/firefox/addon/bitwarden-password-manager/
```

williamdes (approved):
Thank you !
I added a fix for one link to use the en instead of the ja version

src/util.rs
umireon: Oh sorry.

umireon:
The bitwarden/server does not seem to set frame-ancestors in CSP.
https://github.com/bitwarden/server/blob/7e127cec131d418f54827ea5cd666131abb5503a/util/Setup/Configuration.cs#L75-L83

Not setting frame-ancestors allows anything.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors

I cannot confirm this on the real Bitwarden endpoint because I don't have any paid subscriptions on Bitwarden.

However, I don't think the frame-ancestors should be removed from the vaultwarden implementation.

BlackDex (approved):

PR #1881, opened 2021-08-03
Fix sends notification format and add endpoints to get all sends or one send by id

fabianthdev:
This small pull request fixes the notification format for `SyncSendCreate`, `SyncSendUpdate` and `SyncSendDelete`notification types.

With the correct notification format, the sends in the web vault and other bitwarden clients are automatically updated when the user creates, updates or deletes a send in another session. To allow the clients to update specific sends, this PR also includes new endpoints for getting all sends or one specific send by uuid.

I've verified that this works with the web vault and the bitwarden mobile client.

### DIFF
--- a/src/api/core/sends.rs
+++ b/src/api/core/sends.rs
@@ -18,6 +18,8 @@ const SEND_INACCESSIBLE_MSG: &str = "Send does not exist or is no longer availab
 
 pub fn routes() -> Vec<rocket::Route> {
     routes![
+        get_sends,
+        get_send,
         post_send,
         post_send_file,
         post_access,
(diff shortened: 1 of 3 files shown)


williamdes (approved):

fabianthdev:
I've noticed that the nightly build has failed. Looking at the error messages, I can see that the first two errors occur in the file `src/api/core/sends.rs` in lines I didn't touch in this PR (line 170 and 213). The third error is thrown in file `src/api/db/models/send.rs` due to an empty None statement (line ). I've changed that line and will update the PR in a minute. As I don't know Rust particularly well, I don't know what to do about the first two errors.

BlackDex:
@fabianthdev i have checked this PR yesterday quickly, but it looks like the send files doesn't get updated. It does work great on Text items, but not on files.

I also see that there is a sends v2 api which we do not yet support, but that doesn't matter, since the clients switch to the previous version.

But if you could look at the Send Files part why that isn't being pushed via the websocket updates, that would be great.
Overall, nice job on this PR :).

jjlin:
@fabianthdev If you want to work on the v2 API for Sends, it should be very similar to the v2 API for attachments (#1730, #1922). Otherwise I'll get around to it at some point.

PR #1886, opened 2021-08-06
Adding tzdata to Alpine container

mrckndt:
To be able to set a timezone inside a Alpine container with the env variable TZ the tzdata package is needed. Otherwise only UTC will be set. In my case fail2ban (which is running on the host with Europe/Berlin) is complaining about a too large time deviation.


williamdes (approved):

pepa65:
Saves doing it manually or adding it to the custom patch! `apk add tzdata` is indeed correct.

BlackDex (approved):
Seems to be not working indeed on the Alpine image.
Thanks for the contribution,

PR #1925, opened 2021-08-20
Fix syncing with Bitwarden Desktop v1.28.0

### LINKED ISSUE #1924: windows exe client 1.28.0 cannot sync with self host server.
image

but if I use exe client of version 1.27.1, everything is ok.

server info:

docker image: vaultwarden/server:latest

BlackDex:
Syncing with the latest desktop client (v1.28.0) fails because it expects some json key/values to be there.

This PR adds those key/value pairs.

Resolves #1924

### DIFF
--- a/src/db/models/user.rs
+++ b/src/db/models/user.rs
@@ -210,7 +210,10 @@ impl User {
             "PrivateKey": self.private_key,
             "SecurityStamp": self.security_stamp,
             "Organizations": orgs_json,
-            "Object": "profile"
+            "Providers": [],
+            "ProviderOrganizations": [],
+            "ForcePasswordReset": false,
+            "Object": "profile",
         })
     }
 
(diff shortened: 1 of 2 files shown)


src/db/models/user.rs
```diff
@@ -210,7 +210,10 @@ impl User {
             "PrivateKey": self.private_key,
             "SecurityStamp": self.security_stamp,
             "Organizations": orgs_json,
-            "Object": "profile"
+            "Object": "profile",
```
jjlin, line 213: The `Object` key should either be at the top or remain at the bottom. A trailing comma on the last entry would be good too.

jjlin (changes requested):

src/db/models/user.rs
BlackDex: Why is that? Doesn't matter where it is as long is it is there

williamdes (approved):

src/db/models/user.rs
jjlin: Because that's the object type key, which is common across all responses, and therefore "special" in some sense. Also, it's the convention throughout the codebase.

BlackDex: Well, i agree upon the rest of the code base. But in general, it doesn't matter where it is at, though a quick look at the top or bottom and seeing this specific item is nice.

BlackDex:
Even though they seem to have fixed this issue in v1.28.1, i think it is useful to have this added, since who knows what future updates will bring without these values.

PR #1928, opened 2021-08-22
Dependency updates

BlackDex:
Updated several dependencies and switch to different totp library.

- Switch oath with totp-lite
  oauth hasn't been updated in a long while and some dependencies could not be updated any more
  It now also validates a preseeding 0, as the previous library returned an int instead of a str which stripped a leading 0
- Updated rust to the current latest nightly (including build image)
- Updated bootstrap css and js
- Updated hadolint to latest version
- Updated default rust image from v1.53 to v1.54
- Updated new nightly build/clippy messages


vmvmpvm:
Why rust ‘nightly’?

BlackDex:
@vmvmpvm because rocket only works on nightly.
The async branch which uses the async rocket is using stable.

williamdes:
> @vmvmpvm because rocket only works on nightly.
> The async branch which uses the async rocket is using stable.

Yup and nope
The 0.5.0-rc version of rocket works fine, I migrated all my projects to it
And it uses stable
Let me know if you want me to migrate this project or want links to the migration commits I did

williamdes:
Ref: https://rocket.rs/master/guide/upgrading-from-0.4/
Ref: https://github.com/SergioBenitez/Rocket/discussions/1672

BlackDex:
Thanks for the links.

As i mentioned before, there already is a branch using stable and async. https://github.com/dani-garcia/vaultwarden/tree/async
It only is a little bit behind now on both new commits as crate versions.

PR #1950, opened 2021-08-29
Fix incorrect WebAuthn origin

jjlin:
This mainly affects users running Vaultwarden under a subpath.

Refs:

* https://github.com/kanidm/webauthn-rs/blob/b2cbb34/src/core.rs#L941-L948
* https://github.com/kanidm/webauthn-rs/blob/b2cbb34/src/core.rs#L316
* https://w3c.github.io/webauthn/#dictionary-client-data

### DIFF
--- a/src/api/core/two_factor/webauthn.rs
+++ b/src/api/core/two_factor/webauthn.rs
@@ -22,12 +22,14 @@ pub fn routes() -> Vec<Route> {
 
 struct WebauthnConfig {
     url: String,
+    origin: String,
     rpid: String,
 }
 
 impl WebauthnConfig {
     fn load() -> Webauthn<Self> {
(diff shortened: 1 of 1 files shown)


jjlin:
Issue reported by @synaptiko at https://github.com/dani-garcia/vaultwarden/discussions/1808#discussioncomment-1250957

BlackDex:
Not sure from the top of my head, but does this also fix the trailing / issues? So that it will not have a trailing / anymore?

jjlin:
> Not sure from the top of my head, but does this also fix the trailing / issues? So that it will not have a trailing / anymore?
> Also when someone adds two for example.

Yes, the origin extraction code should strip trailing `/`.

BlackDex (approved):

PR #1959, opened 2021-09-04
Update Debian base images to bullseye

jjlin:


BlackDex (approved):

BlackDex (changes requested):
I did approved it already, but if we are going to use bullseye as a runtime, we probably also want to use that for building. So maybe change the default rust:1.54 to rust:1.54-bullseye ?

jjlin:
> I did approved it already, but if we are going to use bullseye as a runtime, we probably also want to use that for building. So maybe change the default rust:1.54 to rust:1.54-bullseye ?

Sure, updated.

BlackDex (approved):

PR #1966, opened 2021-09-07
Fix issue when using uppercase chars in emails

### LINKED ISSUE #1963: Inviting users with capital letters in there mail address without mail enabled will fail to register
Inviting users with capital letters in there mail address without mail enabled will fail to register

see Matrix chat for more information

BlackDex:
In the case when SMTP is disabled and.
when inviting new users either via the admin interface or into an
organization and using uppercase letters, this would fail for those
users to be able to register since the checks which were done are
case-sensitive and never matched.

This PR fixes that issue by ensuring everything is lowercase.
Fixes #1963

### DIFF
--- a/src/api/core/accounts.rs
+++ b/src/api/core/accounts.rs
@@ -62,11 +62,12 @@ struct KeysData {
 #[post("/accounts/register", data = "<data>")]
 fn register(data: JsonUpcase<RegisterData>, conn: DbConn) -> EmptyResult {
     let data: RegisterData = data.into_inner().data;
+    let email = data.Email.to_lowercase();
 
-    let mut user = match User::find_by_mail(&data.Email, &conn) {
+    let mut user = match User::find_by_mail(&email, &conn) {
         Some(user) => {
             if !user.password_hash.is_empty() {
-                if CONFIG.is_signup_allowed(&data.Email) {
+                if CONFIG.is_signup_allowed(&email) {
                     err!("User already exists")
                 } else {
                     err!("Registration not allowed or user already exists")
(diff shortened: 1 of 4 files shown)


src/api/core/accounts.rs
```diff
@@ -62,11 +62,12 @@ struct KeysData {
 #[post("/accounts/register", data = "<data>")]
 fn register(data: JsonUpcase<RegisterData>, conn: DbConn) -> EmptyResult {
     let data: RegisterData = data.into_inner().data;
+    let reg_email = data.Email.to_lowercase();
```
jjlin, line 65: Any particular reason `reg_email` isn't just called `email`?

src/api/core/organizations.rs
```diff
@@ -539,19 +539,20 @@ fn send_invite(org_id: String, data: JsonUpcase<InviteData>, headers: AdminHeade
         err!("Only Owners can invite Managers, Admins or Owners")
     }
 
-    for email in data.Emails.iter() {
+    for mail in data.Emails.iter() {
```
jjlin, line 542: I think it would be cleaner to just call this `email` and have it shadowed by the `let email` below, instead of making up different names unnecessarily.

src/db/models/user.rs
```diff
@@ -349,7 +349,8 @@ impl User {
 }
 
 impl Invitation {
-    pub const fn new(email: String) -> Self {
+    pub fn new(mail: String) -> Self {
```
jjlin, line 352: Same here, the original `email` is fine

src/api/core/accounts.rs
BlackDex: No not really actually, ill change that indeed.

src/api/core/organizations.rs
BlackDex: Well, i wanted to follow the same line as is done with the new user fn. But i guess it can be changed there too.

src/db/models/user.rs
BlackDex: Same as above.

BlackDex:
@jjlin Fixed the remarks :)

PR #1978, opened 2021-09-16
Update dependencies

benarmstead:
Update cargo dependencies

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -236,15 +236,15 @@ dependencies = [
 
 [[package]]
 name = "bytes"
-version = "1.0.1"
+version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b700ce4376041dcd0a327fd0097c41095743c4c8af8887265942faf1100bd040"
+checksum = "c4872d67bab6358e59559027aa3b9157c53d9358c51423c17554809a8858e0f8"
 
 [[package]]
 name = "cc"
-version = "1.0.69"
+version = "1.0.70"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e70cc2f62c6ce1868963827bd677764c62d07c3d9a3e1fb1177ee1a9ab199eb2"
+checksum = "d26a6ce4b6a484fa3edb70f7efa6fc430fd2b87285fe8b84304fd0936faa0dc0"
 
 [[package]]
 name = "cfg-if"
@@ -345,7 +345,7 @@ dependencies = [
  "cookie 0.15.1",
  "idna 0.2.3",
  "log 0.4.14",
- "publicsuffix 2.1.0",
+ "publicsuffix 2.1.1",
  "serde",
  "serde_json",
  "time 0.2.27",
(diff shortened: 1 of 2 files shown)


BlackDex:
Thx for the contribution.
But i'm missing the Cargo.toml changes to represent the updated versions there.
While this isn't strictly necessary, it does give a better view on which versions are used by this project it self, instead of having to shift through both Cargo.toml and Cargo.lock.

I normally do a `cargo update --dry-run` and update the versions in the Cargo.toml file where needed and after that do a run without the `--dry-run` to update the Cargo.lock.

benarmstead:
Ok no problem, I will do that now and get back to you

benarmstead:
Think I have done it

BlackDex (approved):

jjlin:
Was there some specific issue addressed by these updates?

BlackDex:
What i saw was the OpenSSL libraries that were updated which were released because of CVE's (See: https://rustsec.org/advisories/RUSTSEC-2021-0098 & https://www.openssl.org/news/vulnerabilities.html)

jjlin:
Ok, it would be nice to have this in the commit or PR description instead of a vague message

BlackDex:
Well, the dep updates are sometimes also done without any specific descriptions besides, just updating it, and sometimes a small remark about updating code, but thats it..

But good item to think about for the future :).

jjlin:
That's more acceptable if done as part of the normal course of project maintenance by established maintainers, but my main point is I don't want to see random unjustified dep updates by unfamiliar people.

BlackDex:
Very valid point indeed.

BlackDex:
Also, that is where reviewing a PR comes in. If it's reviewed correctly then it shouldn't matter if it's from someone new or had done multiple PR's already. Dismissing a PR just because someone never pushed one before is not what we want i think.

jjlin:
I welcome quality contributions from anyone, new or old. But to be blunt, this is not a quality contribution. It has little to no value, because:

1. Dependency updates are going to get done in the normal course of development anyway, such as when someone makes a change that requires adding/removing/updating libraries, or when @dani-garcia prepares for a new release.

2. Dependency updates are not always an improvement. They can introduce regressions, and when the PR is coming from someone new, it's unknown if that person has done any testing on the result, or is even reasonably knowledgeable about how to do such testing.

So I don't think that PRs that update dependencies just for the sake of updating them should be accepted; there needs to be specific justification. For example, it's fine if there's a dependency that should be updated promptly because of a security issue, but that needs to be called out specifically. Otherwise, dependencies are updated all the time, so there could conceivably be "update dependencies" PRs coming in every few days, and that's not something I want to encourage.

Regarding your contention that the source of a PR doesn't matter if it's "reviewed correctly", I agree in theory. But in practice, time and attention are limited and not all PRs are reviewed as carefully as others. In particular, there's a tendency for PRs with a lot of generated changes (such as lock files, updates to JS libraries, etc.) to be reviewed less carefully, simply because it's tedious work. This is just human nature and won't change in the foreseeable future. Therefore, the source of a PR can matter in practice, and when possible, I prefer that PRs for changes like this to be done by people who are not new to the project.

BlackDex:
All valid points and i agree.

If it makes you feel a bit better, i had a local branch already with some updated libs and i noticed the same versions. And thus i tested it my self already. I only didn't noticed the cve before on the OpenSSL and therefore decided to merge.

I should have mentioned that in my comments I am sorry for the lack of my communication on that part. I will keep that in mind.

I am also glad you are make a point of this especially for software like Vaultwarden which handels sensitive information.

PR #1985, opened 2021-09-21
bump web version to v2.22.3

vivithecanine:
This updates the web interface to `v2.22.3` as it's currently a few versions behind at `v2.21.1`


BlackDex:
I think we are not going to merge this.
@dani-garcia is planning to enable the emergency-access feature in the web-vault and create a new release.
This web-vault version doesn't have that implemented yet, and he needs to release a new one first.

vivithecanine:
Understood, closing.

PR #1989, opened 2021-09-24
fix: Support no-data enterprise policies

domdomegg:
Boolean-toggle enterprise policies (like 'Two-Step Login' and 'Personal Ownership') don't provide a data attribute in the new version of the web client. This updates the backend to expect these to be optional.

Web change introduced in https://github.com/bitwarden/web/pull/1147 which added https://github.com/bitwarden/web/blob/2cbe023a38792ff70a2a4907f4748354bb4e0573/src/app/organizations/policies/base-policy.component.ts#L48-L50

### DIFF
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -1183,7 +1183,7 @@ struct PolicyData {
     enabled: bool,
     #[serde(rename = "type")]
     _type: i32,
-    data: Value,
+    data: Option<Value>,
 }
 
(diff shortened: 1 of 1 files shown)


BlackDex:
Indeed, same for `Disable Send`

williamdes (approved):

jjlin (approved):
Tested this and it seems to work fine. An empty `Option` JSON-serializes as the string `null`, so it works with the existing db schemas.

PR #1990, opened 2021-09-24
docs: Add basic contributing details to README

domdomegg:
As a new contributor (maybe just because I haven't used Rust before) I didn't easily see how to get started.

If this is too much for the README, happy just to have a link to the wiki page on building the binary which I only found after quite some digging (as I initially searched 'contributing' and 'developing').

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,16 @@ If you spot any bugs or crashes with vaultwarden itself, please [create an issue
 
 If you prefer to chat, we're usually hanging around at [#vaultwarden:matrix.org](https://matrix.to/#/#vaultwarden:matrix.org) room on Matrix. Feel free to join us!
 
+## Contributing
+To get started with development:
+
+1. [Install Rust](https://www.rust-lang.org/tools/install)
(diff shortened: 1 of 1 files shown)


BlackDex:
A link to the wiki would also be a good addition, because there it also describes what dependencies there are needed.

jjlin:
Instead of adding all this inline, I think it would be better to just link to an appropriate wiki page (updating or creating it as needed). In general, the README.md has not been that well-maintained since there is a much higher barrier to making changes.

williamdes (approved):

Ornias1993 (changes requested):
A github standard `CONTRUBUTING` file and both a wiki and `CONTRIBUTING` link in the readme seems to be beter suited for this...

BlackDex:
I'm going to close this PR. I agree with @jjlin that this better suited to be placed somewhere else like in a good wiki page, or maybe even the github standard CONTRUBUTING.

Feel free to create a new PR, or update the wiki :smile:

PR #1955, opened 2021-09-01
OIDC SSO Re: Issue #246

Sheap:
This branch adds the minimal required changes to allow OIDC SSO (at least with keycloak. There are many options/features of the process which I did not address, which may rule out other providers).

This is so far my biggest contact point with rust in a web server, as well as in implementing OIDC, so I wouldn't be surprised if there were places where I strayed from the path. But I have tested it, and at least it works. I'd welcome any feedback here. My greatest concern is that I'm currently ignoring the nonce. I believe it should be checked against, but I'm not sure where would be a sensible place to store it between generating it alongside the auth url, and consuming it when exchanging the code for an access token.

Note: There are associated changes required to the webvault here - I wasn't sure how I should link them, so to keep things simple for now, I've simply included a .patch file in this repo.

PR #1991, opened 2021-09-24
feature: Support single organization policy

domdomegg:
This adds back-end support for the [single organization policy](https://bitwarden.com/help/article/policies/#single-organization).

This is listed on [the wiki](https://github.com/dani-garcia/vaultwarden/wiki) as a feature in scope for Vaultwarden that a contribution would be welcome for.

NB: this is not already done by #1955, despite that PR enabling the enum for it (it only implements the RequireSso policy).


src/static/templates/email/send_single_org_removed_from_org.hbs
```diff
@@ -0,0 +1,7 @@
+Removed from {{{org_name}}}
+<!---------------->
+You have been removed from organization *{{org_name}}* because you a member of multiple organizations, but it requires you to be a member of only it.
```
williamdes, line 3: The "it" part is unclear

src/api/core/organizations.rs
```diff
@@ -774,6 +777,19 @@ fn accept_invite(_org_id: String, _org_user_id: String, data: JsonUpcase<AcceptD
     Ok(())
 }
 
+/// Enforces the single organization policy, if applicable.
+/// A non-owner/admin user belonging to an org with the single organization policy
+/// enabled isn't allowed to create new organizations or join other organizations
+///
+/// Ref: https://bitwarden.com/help/article/policies/#single-organization
+fn enforce_single_organization_policy(user_uuid: &String, conn: &DbConn) -> EmptyResult {
```
jjlin, line 785: Given that this function is called from a couple of places, it would probably be better located near the top of the file.

src/api/core/organizations.rs
```diff
@@ -102,6 +102,7 @@ fn create_organization(headers: Headers, data: JsonUpcase<OrgData>, conn: DbConn
     if !CONFIG.is_org_creation_allowed(&headers.user.email) {
         err!("User not allowed to create organizations")
     }
+    enforce_single_organization_policy(&headers.user.uuid, &conn)?;
```
jjlin, line 105: This currently returns an error message about joining an org, not creating one.

src/static/templates/email/send_single_org_removed_from_org.hbs
```diff
@@ -0,0 +1,7 @@
+Removed from {{{org_name}}}
```
jjlin, line 1: Consider keeping these messages the same as upstream if possible. The message you have below is missing a word anyway ("because you **are** a member").

src/api/core/organizations.rs
```diff
@@ -747,6 +748,8 @@ fn accept_invite(_org_id: String, _org_user_id: String, data: JsonUpcase<AcceptD
                     err!("You cannot join this organization until you enable two-step login on your user account.")
                 }
 
+                enforce_single_organization_policy(&user_org.user_uuid, &conn)?;
```
jjlin, line 751: Upstream actually does [two checks](https://github.com/bitwarden/server/blob/dac3b3e8939504959d94c2e2fd13789723ab2d57/src/Core/Services/Implementations/OrganizationService.cs#L1335-L1359) in the accept flow:

1. If the org the user is joining has the single org policy enabled and the user isn't joining as an owner/admin, then reject the acceptance if the user is already a member of any other org. This check isn't implemented currently AFAICT.

2. Otherwise, if the user is an accepted or confirmed member of an org with the single org policy enabled, then reject the acceptance. This is what the current check does, except it only checks for confirmed users instead of accepted users as well. That's sort of an edge case and maybe we can let it slide for now, given that some of our other policy implementations would have the same issue.

jjlin (changes requested):
You should probably review the relevant upstream PRs:

* https://github.com/bitwarden/server/pull/962
* https://github.com/bitwarden/server/pull/1171
* https://github.com/bitwarden/server/pull/1410

Email notification:

* https://github.com/bitwarden/server/blob/master/src/Core/MailTemplates/Handlebars/OrganizationUserRemovedForPolicySingleOrg.html.hbs

Error messages:

* https://github.com/bitwarden/server/blob/dac3b3e8939504959d94c2e2fd13789723ab2d57/src/Core/Services/Implementations/OrganizationService.cs#L653-L654
* https://github.com/bitwarden/server/blob/dac3b3e8939504959d94c2e2fd13789723ab2d57/src/Core/Services/Implementations/OrganizationService.cs#L1342-L1343
* https://github.com/bitwarden/server/blob/dac3b3e8939504959d94c2e2fd13789723ab2d57/src/Core/Services/Implementations/OrganizationService.cs#L1357-L1358

src/api/core/organizations.rs
domdomegg: I ended up inlining this function - given it's really just one very simple if statement and an error message (that needs to change depending on context) that seems to be the more sensible way. Also saw this other [PR review comment](https://github.com/dani-garcia/vaultwarden/pull/1955#discussion_r704718304) that suggested we prefer inlining very small functions.

src/api/core/organizations.rs
domdomegg: Good catch - have updated the error message to match upstream.

src/static/templates/email/send_single_org_removed_from_org.hbs
domdomegg: I've made this match upstream, seems like the sensible thing to do :+1:

src/api/core/organizations.rs
domdomegg: Correct, I missed check 1 and have now implemented it. One thing that might be worth noting is that like upstream, I have included invited org users (checks [allOrgUsers](https://github.com/bitwarden/server/blob/dac3b3e8939504959d94c2e2fd13789723ab2d57/src/Core/Services/Implementations/OrganizationService.cs#L1333) are [the same org](https://github.com/bitwarden/server/blob/dac3b3e8939504959d94c2e2fd13789723ab2d57/src/Core/Services/Implementations/OrganizationService.cs#L1340), and [the query doesn't filter out invites](https://github.com/bitwarden/server/blob/dac3b3e8939504959d94c2e2fd13789723ab2d57/src/Core/Repositories/EntityFramework/OrganizationUserRepository.cs#L237)).

On 2, yes this is an issue which is shared by many of the policy checks - I've renamed a couple methods to make it more explicit about what they're checking and added a TODO to check against accepted users too like upstream. I think this may be best in a separate PR as it'll probably require updating the way many of the policy checks are done.

Another thing to note is that because we auto-accept users when email is disabled without running through these checks, they are not enforced there. Again, this is a problem with all the checks but I'm not sure the most sensible way to go about resolving this?

src/api/core/organizations.rs
```diff
@@ -1219,6 +1248,29 @@ fn put_policy(
         }
     }
 
+    if pol_type_enum == OrgPolicyType::SingleOrg && data.enabled {
+        let org_list = UserOrganization::find_by_org(&org_id, &conn);
+
+        for user_org in org_list.into_iter() {
+            if user_org.atype < UserOrgType::Admin {
```
jjlin, line 1255: This should also filter out users in `Invited` state: https://github.com/bitwarden/server/blob/dac3b3e8939504959d94c2e2fd13789723ab2d57/src/Core/Services/Implementations/PolicyService.cs#L88

Those users will just get the appropriate error message when they try to accept.

src/api/core/organizations.rs
```diff
@@ -1219,6 +1248,29 @@ fn put_policy(
         }
     }
 
+    if pol_type_enum == OrgPolicyType::SingleOrg && data.enabled {
```
jjlin, line 1251: Some comments that summarize what this code is trying to do would be nice, but I think it would read more clearly with `org_list` renamed to `org_members` and `user_org` renamed to `member`.

src/api/core/organizations.rs
```diff
@@ -1219,6 +1248,29 @@ fn put_policy(
...
+        let org_list = UserOrganization::find_by_org(&org_id, &conn);
+
+        for user_org in org_list.into_iter() {
+            if user_org.atype < UserOrgType::Admin {
+                let is_member_of_another_org = UserOrganization::find_any_state_by_user(&user_org.user_uuid, &conn)
+                    .into_iter()
+                    .filter(|uo| uo.org_uuid != user_org.org_uuid)
```
jjlin, line 1258: This should also filter out users in `Invited` state: https://github.com/bitwarden/server/blob/dac3b3e8939504959d94c2e2fd13789723ab2d57/src/Core/Services/Implementations/PolicyService.cs#L112

Otherwise users will be dropped from this single org even if they didn't intend to join the other org they were invited to.

jjlin (changes requested):
I'd suggest rebasing and squashing at the end as well

src/api/core/organizations.rs
jjlin: AFAIK having no email service is not a configuration that's expected or designed for in upstream Bitwarden, so people who don't bother configuring it shouldn't expect things to work completely normally.

src/api/core/organizations.rs
domdomegg: Have updated.

FYI I think the 2FA organization policy also misses this. I can raise a PR to fix that if you'd like?

vaultwarden: https://github.com/dani-garcia/vaultwarden/blob/main/src/api/core/organizations.rs#L1210
upstream (bitwarden/server): https://github.com/bitwarden/server/blob/dac3b3e8939504959d94c2e2fd13789723ab2d57/src/Core/Services/Implementations/PolicyService.cs#L93

src/api/core/organizations.rs
domdomegg: Good feedback, have renamed variables as suggested and added explanatory comments

src/api/core/organizations.rs
domdomegg: Good catch, fixed :+1:

src/api/core/organizations.rs
jjlin: The 2FA org policy isn't my code anyway, but if there's something that can be fixed or improved, feel free.

jjlin (approved):

domdomegg:
Thanks @jjlin for your detailed, timely and valuable feedback on this. And thanks @dani-garcia for merging :raised_hands:

PR #2031, opened 2021-10-08
Update OpenSSL via an updated clux build image.

BlackDex:
Recently the LetsEncrypt DST certificate has expired.
Older versions of OpenSSL like v1.0.x have issues using this certificate.

Recently clux has updated his image to support OpenSSL v1.1.1[a-z].
This solves issues with those certificates.

This issues was disscused on Matrix.

### DIFF
--- a/docker/Dockerfile.j2
+++ b/docker/Dockerfile.j2
@@ -6,7 +6,7 @@
 {% set build_stage_base_image = "rust:1.55-buster" %}
 {% if "alpine" in target_file %}
 {%   if "amd64" in target_file %}
-{%     set build_stage_base_image = "clux/muslrust:nightly-2021-08-22" %}
+{%     set build_stage_base_image = "clux/muslrust:nightly-2021-10-06" %}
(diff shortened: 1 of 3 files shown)


Rihcus:
Have similar issues. Oddly the desktop app (Mac OS), browser extension (Chrome and Firefox nightly), and web portal (Firefox and Chrome) work fine. 

The main issues I have run into is the android and ios apps, and safari for ios (iOS 14).

Rihcus:
> > Have similar issues. Oddly the desktop app (Mac OS), browser extension (Chrome and Firefox nightly), and web portal (Firefox and Chrome) work fine.
> > The main issues I have run into is the android and ios apps, and safari for ios (iOS 14).
> 
> That probably has nothing to do with the server side of this if you are using a reverse proxy. If not, then it could be, but unlikely i think. But it probably is something on the client side maybe together with the certificate from LetsEncrypt having the expired DST Cert.

Does the lets encrypt android and ios app use the OS's root certs or does bitwarden embed root certs in the app directly and bypass system root certs?

PR #2073, opened 2021-10-29
Fix conflict resolution logic for `read_only` and `hide_passwords` flags

### LINKED ISSUE #2072: user sharing item to another collection that has "Hide Passwords" cannot read the password any more
### Subject of the issue

If a user is in 2 collections and one of them has the setting "Hide Passwords" set, all passwords that are in both collections will not be readable by the user, even if the user shares a login item from the collection with no "Hide Passwords" set. 

### Deployment environment

* vaultwarden version: 1.23.0

* Install method: vaultwarden/server:1.23.0-alpine in Kubernetes

* Clients used:  web

* Reverse proxy and version:  NA

* MySQL/MariaDB or PostgreSQL version:  postgresql13-server-13.4 with Patroni HA (2 hosts Oraclelinux 8)

* Other relevant details: NA

### Steps to reproduce
Lets say we have:

organiztaions: org1
collections: team1, team2, team3
users: user1, user2, user3, ...
login items in team1: linux1, linux2
login items in team2: windows1, windows2
login items in team3: router1, router2
collection access user1 (owner): all
collection access user2 (user): team2
collection access user3 (user): team3

Lets say user2 (or any other user from team2) wants to add login item windows1 to collection team3. user1 adds user2 to team3 with the option "Hide Passwords" so that user2 could not see the passwords for team3. user2 adds windows1 to collection team3. windows1 now is visible in two collections team2 and team3. 

### Expected behaviour
user2 can still read the passowrd from item windows1. user2 has no permissions to remove any items from team3. user2 has no permissions to add any items from team3 to team2 (passwords for items added from team3 to team2 are not readable).

### Actual behaviour
user2 cannot view the password of windows1 any more, it is grayed out (user2 can remove windows1 from team3 and the password becomes visible again) user2 can also now add any other item from team3 to team2 and remove any item from team3. The passwords will not be possible to read, but user3 (or other users from team3) would not see any items in team3 any more.

### Troubleshooting data

Used the web client with multiple users to test.

jjlin:
For one of these flags to be in effect for a cipher, upstream requires all of
(rather than any of) the collections the cipher is in to have that flag set.

Also, some of the logic for loading access restrictions was wrong. I think
that only malicious clients that also had knowledge of the UUIDs of ciphers
they didn't have access to would have been able to take advantage of that.

Fixes #2072.

### DIFF
--- a/src/db/models/cipher.rs
+++ b/src/db/models/cipher.rs
@@ -343,36 +343,39 @@ impl Cipher {
         db_run! {conn: {
             // Check whether this cipher is in any collections accessible to the
             // user. If so, retrieve the access flags for each collection.
-            let query = ciphers::table
+            let rows = ciphers::table
                 .filter(ciphers::uuid.eq(&self.uuid))
                 .inner_join(ciphers_collections::table.on(
                     ciphers::uuid.eq(ciphers_collections::cipher_uuid)))
                 .inner_join(users_collections::table.on(
                     ciphers_collections::collection_uuid.eq(users_collections::collection_uuid)
                         .and(users_collections::user_uuid.eq(user_uuid))))
-                .select((users_collections::read_only, users_collections::hide_passwords));
+                .select((users_collections::read_only, users_collections::hide_passwords))
+                .load::<(bool, bool)>(conn)
+                .expect("Error getting access restrictions");
 
-            // There's an edge case where a cipher can be in multiple collections
-            // with inconsistent access flags. For example, a cipher could be in
-            // one collection where the user has read-only access, but also in
-            // another collection where the user has read/write access. To handle
-            // this, we do a boolean OR of all values in each of the `read_only`
-            // and `hide_passwords` columns. This could ideally be done as part
-            // of the query, but Diesel doesn't support a max() or bool_or()
-            // function on booleans and this behavior isn't portable anyway.
-            if let Ok(vec) = query.load::<(bool, bool)>(conn) {
-                let mut read_only = false;
-                let mut hide_passwords = false;
-                for (ro, hp) in vec.iter() {
-                    read_only |= ro;
-                    hide_passwords |= hp;
-                }
-
-                Some((read_only, hide_passwords))
-            } else {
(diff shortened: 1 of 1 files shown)


perkons:
Thank you very much for the very fast fix and for your work! It is working now as expected.

One issue/exploit that I found now is that a user with "Hide Passwords" permissions in a collection can add an item from this collection to another collection with no "Hide Passwords" permissions and remove it from the one where he has "Hide Passwords". After that is done the user can view the password. So the "Hide Passwords" option becomes useless if not used together with "Read Only".

I personally think that the best solution here would be to limit permissions on a collection that has "Hide Passwords". I think the best solution that I can think of is to remove the Collections option under the Cog symbol when "Hide Passwords" is checked. So "Hide Passwords" would forbid the removal from items from the collection. If a feature request is needed I can create one, but I think that this can maybe be considered to be a bug?

I am Using vaultwarden/server:testing-alpine@sha256:3f9d68af0d3b86b941a1bbdb77b27e61188adfcfadc09b2d6aeca67ce923a84f

jjlin:
> One issue/exploit that I found now is that a user with "Hide Passwords" permissions in a collection can add an item from this collection to another collection with no "Hide Passwords" permissions and remove it from the one where he has "Hide Passwords". After that is done the user can view the password. So the "Hide Passwords" option becomes useless if not used together with "Read Only".

"Hide Passwords" is only intended to act as a hurdle for typical, unsophisticated users, not "clever" ones. Given it's only enforced on the client side, it's easy to disable for someone who knows what they're doing, especially in the web vault.

> I personally think that the best solution here would be to limit permissions on a collection that has "Hide Passwords". I think the best solution that I can think of is to remove the Collections option under the Cog symbol when "Hide Passwords" is checked. So "Hide Passwords" would forbid the removal from items from the collection. If a feature request is needed I can create one, but I think that this can maybe be considered to be a bug?

Vaultwarden doesn't implement custom client-side features, so you would have to make your case with upstream Bitwarden at https://community.bitwarden.com/c/feature-requests/5/.

PR #2103, opened 2021-11-16
Async/Awaited all db methods

BlackDex:
This is a rather large PR which updates the async branch to have all the
database methods as an async fn.

Some iter/map logic needed to be changed to a stream::iter().then(), but
besides that most changes were just adding async/await where needed.

Also updated the libraries to the latest available while at it.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,4 +1,6 @@
 #![forbid(unsafe_code)]
+// #![warn(rust_2018_idioms)]
+#![warn(rust_2021_compatibility)]
 #![cfg_attr(feature = "unstable", feature(ip))]
 // The recursion_limit is mainly triggered by the json!() macro.
 // The more key/value pairs there are the more recursion occurs.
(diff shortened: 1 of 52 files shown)


src/main.rs
```diff
@@ -1,12 +1,13 @@
...
+#![warn(rust_2018_idioms)]
+#![warn(rust_2021_compatibility)]
 #![cfg_attr(feature = "unstable", feature(ip))]
 // The recursion_limit is mainly triggered by the json!() macro.
 // The more key/value pairs there are the more recursion occurs.
 // We want to keep this as low as possible, but not higher then 128.
 // If you go above 128 it will cause rust-analyzer to fail,
 #![recursion_limit = "87"]
 
-extern crate openssl;
```
dani-garcia, line 9: Note that this is added because of #1019, not sure if it's needed still though.

BlackDex: I know, but i tested building all the containers and no issues anymore.

PR #2127, opened 2021-12-03
Added: Dependabot.yml

GoliathLabs:
Hello,
I have added a dependabot.yml configuration file. Dependabot will check for updates of the gh-actions used as well as the docker components and the cargo dependencies.


GoliathLabs:
@dani-garcia any feedback regarding this?

GoliathLabs:
That would make the work of @BlackDex also easier as it proposes updates of the cargo manifests on a regular basis

BlackDex:
First of all, thanks for the contribution.

Although it sounds nice, i think it will cause a lot of noise.
Dependabot will create a lot of PR's and we only update the libraries once in a while.
Also, some libraries we can't update to the latest version because of restrictions from other libraries like sqlite with diesel.
Same goes for the Dockerfiles actually, we do not want to have all the latest versions or releases for some of the builds, and Dependabot will keep nagging about this.

Regarding the gh-actions, i have watches set on the actions we use, and if needed i can update those.

Tnanaphon (approved):

GoliathLabs:
@BlackDex You can configure how many PRs Dependabot is allowed to create, thus reducing the "noise". As for the libraries that we can't update to the latest version: These can also be excluded from dependabot checks, the same goes for Docker files.

In my opinion Dependabot is a good tool to keep track of updates for dependencies. Regarding gh-actions, an update is proposed in the configured interval. Therefore, you don't have to keep an eye on it anymore.

For dependency updates, you can go through the PRs created by Dependabot and easily see what still needs to be done or what changes are included in the new version. I think that you should give Dependabot a chance and first see how the project runs with it. If it turns out that Dependabot is nothing for the Vaultwarden project, you can remove it again. However, I have noticed that Dependabot is helpful for many projects that I work on, because the updates can be processed gradually and you don't have to keep an eye on every release of a library you are using, but can focus on the development of the project itself.

Kind regards

GoliathLabs:
@BlackDex Take this for example: https://deps.rs/repo/github/dani-garcia/vaultwarden. deps.rs reports six outdated dependencies and one that is insecure. For these six dependabot would have already created PRs. And it would create one for chrono as soon as a new patched version is released. I for one can say that dependabot has saved me some work and time in my projects and more and more projects are using dependabot (e.g. etherpad, soon synapse (matrix server)). Even vaultwarden_ldap is now using it to keep track of dependency updates.

BlackDex:
Well, the issue is that some dependencies can't be updated, like ` libsqlite3-sys`.
The _Insecure_ one isn't even a `Chrono` issue if i'm correct its a third-party which is used, which is again also used by other libraries not marked as insecure.

Also, some updates need code changes before they even can be applied, and.. If i do the updates manually, or @dani-garcia does them, if I'm correct, we both do a `cargo upgrade --dry-run` first, check what can be updated and adjust where needed, test the updated version and push/pr/merge it.

Don't get me wrong, for a lot of projects it can probably help and provide warnings or notifications on when updates are needed for some packages. But in this case i think we need to exclude to much and we will probably never use the PR's generated by Dependabot, since we always want to test them out first, and bundle updates, and have the Cargo.lock and Cargo.toml both updated, something it doesn't do if I'm correct.

But in the end i think it is up-to @dani-garcia, but from my point of view i think it will create to much noise, and my self, i always want to see the changelogs of the main packages used for the project.

PR #2158, opened 2021-12-20
Add support for external icon services

jjlin:
If an external icon service is configured, icon requests return an HTTP redirect to the corresponding icon at the external service.

An external service may be useful for various reasons, such as if:

* The Vaultwarden instance has no external network connectivity.
* The Vaultwarden instance has trouble handling large bursts of icon requests.
* There are concerns that an attacker may probe the instance to try to detect whether icons for certain sites have been cached, which would suggest that the instance contains entries for those sites.
* The external icon service does a better job of providing icons than the built-in fetcher.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -129,10 +129,24 @@
 ## Number of times to retry the database connection during startup, with 1 second delay between each retry, set to 0 to retry indefinitely
 # DB_CONNECTION_RETRIES=15
 
+## Icon service
+## The predefined icon services are: internal, bitwarden, duckduckgo, google.
+## To specify a custom icon service, set a URL template with exactly one instance of `{}`,
+## which is replaced with the domain. For example: `https://icon.example.com/domain/{}`.
+##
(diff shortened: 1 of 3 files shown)


BlackDex:
Seem like a nice addition.
Though the internal fetcher works very well during my tests compared to all the others, it fetches the same or better.
But, it could help for instances which either have slow or no internet connection at all.

The burst part will mostly be solved when we switch to the async branch, at least during my tests it really helped.

BlackDex (approved):

jjlin:
> Seem like a nice addition. Though the internal fetcher works very well during my tests compared to all the others, it fetches the same or better. But, it could help for instances which either have slow or no internet connection at all.

Yes, internal seems to work as well as the others for the most part, though external might work better in some niche cases like #2019.

> The burst part will mostly be solved when we switch to the async branch, at least during my tests it really helped.

I expect so as well, though I also expect an external service could still be faster for people running on low-powered systems.

PR #2165, opened 2021-12-22
Basic ratelimit for user login (including 2FA) and admin login

dani-garcia:
Might need to adjust the default rate limiting values and explain their use better.

Also needs to document their use in the .env template.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -511,6 +511,16 @@ make_config! {
 
         /// Allowed iframe ancestors (Know the risks!) |> Allows other domains to embed the web vault into an iframe, useful for embedding into secure intranets
         allowed_iframe_ancestors: String, true, def,    String::new();
+
+        /// Seconds between login requests |> Number of seconds, on average, between login requests before rate limiting kicks in. Note that this applies to both the login and the 2FA, so it's recommended to allow a burst size of at least 2
+        login_ratelimit_seconds:       u64, false, def, 60;
+        /// Max burst size for login requests |> Allow a burst of requests of up to this size, while maintaining the average indicated by `login_ratelimit_seconds`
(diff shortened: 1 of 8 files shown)


BlackDex:
That looks very simple indeed. Will be a nice addition. Ill see if i can do some checks tomorrow :).

src/ratelimit.rs
```diff
@@ -0,0 +1,38 @@
...
+            err_code!("Too many login requests", 429);
+        }
+    }
+}
+
+pub fn check_limit_admin(ip: &IpAddr) -> Result<(), Error> {
+    match LIMITER_ADMIN.check_key(ip) {
+        Ok(_) => Ok(()),
+        Err(_e) => {
+            err_code!("Too many admin requests", 429);
```
williamdes, line 35: https://http.cat/

A bit of fun for everyone

williamdes (approved):

BlackDex (approved):

BlackDex (changes requested):
All looks fine, but i'm missing the new config values in the `.env.template` file :).

src/config.rs
```diff
@@ -511,6 +511,16 @@ make_config! {
 
         /// Allowed iframe ancestors (Know the risks!) |> Allows other domains to embed the web vault into an iframe, useful for embedding into secure intranets
         allowed_iframe_ancestors: String, true, def,    String::new();
+
+        /// Seconds between login requests |> Number of seconds, on average, between login requests before rate limiting kicks in. Note that this applies to both the login and the 2FA, so it's recommended to allow a burst size of at least 2
```
jjlin, line 515: It would probably be helpful to clarify that this limit is per IP.

jjlin (changes requested):

BlackDex (approved):

Issue #1453, opened 2021-03-01
Performance Issues


ncodee:
## Describe the Bug

When removing/adding entry on the web Vault, everything because super laggy, and unresponsive.
I have had the page completely freeze with a white screen, the only way to access the page again was to refresh it, which cancels out the current actions.

## Steps To Reproduce

Delete/edit any item on the web Vault.

## Expected Result

Be responsive, and not freezes the page.

## Actual Result

Becomes unresponsive, and freezes.

# Environment

* Vault Items: 2476
* Bitwarden_rs version: v1.19.0
* Web-vault version: v2.18.1
* Running within Docker: true
* Internet access: true
* Uses a proxy: false
* DNS Check: true
* Time Check: true
* Domain Configuration Check: true
* HTTPS Check: true
* Database type: SQLite
* Clients used: 
* Reverse proxy and version: 
* Other relevant information: 

### Config (Generated via diagnostics page)
```json
{
  "_duo_akey": null,
  "_enable_duo": false,
  "_enable_email_2fa": false,
  "_enable_smtp": true,
  "_enable_yubico": true,
  "_ip_header_enabled": true,
  "admin_token": "***",
  "allowed_iframe_ancestors": "",
  "attachments_folder": "/data/attachments",
  "authenticator_disable_time_drift": false,
  "data_folder": "/data",
  "database_max_conns": 10,
  "database_url": "/****/**.*******",
  "db_connection_retries": 15,
  "disable_2fa_remember": false,
  "disable_admin_token": false,
  "disable_icon_download": false,
  "domain": "*****://*****.*******.***",
  "domain_origin": "*****://*****.*******.***",
  "domain_path": "",
  "domain_set": true,
  "duo_host": null,
  "duo_ikey": null,
  "duo_skey": null,
  "email_attempts_limit": 3,
  "email_expiration_time": 600,
  "email_token_size": 6,
  "enable_db_wal": true,
  "extended_logging": true,
  "helo_name": null,
  "hibp_api_key": null,
  "icon_blacklist_non_global_ips": true,
  "icon_blacklist_regex": null,
  "icon_cache_folder": "/data/icon_cache",
  "icon_cache_negttl": 259200,
  "icon_cache_ttl": 2592000,
  "icon_download_timeout": 10,
  "invitation_org_name": "Bitwarden_RS",
  "invitations_allowed": true,
  "ip_header": "X-Real-IP",
  "log_file": null,
  "log_level": "Info",
[...]

PR #2182, opened 2021-12-28
Set `Expires` header when caching responses

RealOrangeOne:
Browsers are rather smart, but also dumb. This uses the `Expires` header
alongside `cache-control` to better prompt the browser to actually
cache.

Unfortunately, firefox still tries to "race" its own cache, in an
attempt to respond to requests faster, so still ends up making a bunch
of requests which could have been cached. Doesn't appear there's any way
around this.

See also https://github.com/dani-garcia/vaultwarden/issues/1453#issuecomment-1002115973

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -444,7 +444,7 @@ pub fn format_datetime_http(dt: &DateTime<Local>) -> String {
 
     // HACK: HTTP expects the date to always be GMT (UTC) rather than giving an
     // offset (which would always be 0 in UTC anyway)
-    return expiry_time.to_rfc2822().replace("+0000", "GMT");
+    expiry_time.to_rfc2822().replace("+0000", "GMT")
 }
 
 //
--- a/src/util.rs
+++ b/src/util.rs
@@ -11,9 +11,8 @@ use rocket::{
     Data, Request, Response, Rocket,
 };
 
-use httpdate::HttpDate;
 use std::thread::sleep;
-use std::time::{Duration, SystemTime};
+use std::time::Duration;
 
 use crate::CONFIG;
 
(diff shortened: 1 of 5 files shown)


williamdes (approved):

src/util.rs
```diff
@@ -99,29 +103,52 @@ impl Fairing for Cors {
...
+        }
     }
 
-    pub fn ttl(r: R, ttl: u64) -> Cached<R> {
-        Self(r, format!("public, immutable, max-age={}", ttl))
+    pub fn ttl(response: R, ttl: u64, is_immutable: bool) -> Cached<R> {
+        Self {
+            response,
+            is_immutable,
+            ttl: ttl,
```
dani-garcia, line 133:
```suggestion
            ttl,
```

Just fixing a clippy error, otherwise it looks good to me!

BlackDex:
Looks good indeed. But do we really need an extra crate for this??
Can't this also be done with Chrono maybe?

RealOrangeOne:
> do we really need an extra crate for this

It's already depended on by `hyper` to do the same thing, but I couldn't see an interface to get to the implementation we need which is exposed through `rocket`. I agree it's not ideal. I'll take another look and see what I can find.

RealOrangeOne:
@BlackDex I've written my own implementation, which seems to work as expected. I'm not entirely convinced it's better, as `httpdate` is already a dependency of `hyper`, do it doesn't bloat the binary. Unfortunately `hyper` doesn't expose any of `httpdate`'s functionality, so the only option is to either depend on it ourselves, or write a custom implementation.

williamdes (approved):
Very nice to not have one more dependency

dani-garcia:
Thanks, we can keep it this way, if it causes issues we can consider adding that dependency back.

PR #2188, opened 2021-12-30
Add config option to set the HTTP redirect code for external icons

jjlin:
The default code is 307 (temporary) to make it easier to test different icon services, but once a service has been decided on, users should ideally switch to using permanent redirects for cacheability.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -454,9 +454,14 @@ make_config! {
         /// To specify a custom icon service, set a URL template with exactly one instance of `{}`,
         /// which is replaced with the domain. For example: `https://icon.example.com/domain/{}`.
         /// `internal` refers to Vaultwarden's built-in icon fetching implementation. If an external
-        /// service is set, an icon request to Vaultwarden will return an HTTP 307 redirect to the
+        /// service is set, an icon request to Vaultwarden will return an HTTP redirect to the
         /// corresponding icon at the external service.
         icon_service:           String, false,  def,    "internal".to_string();
+        /// Icon redirect code |> The HTTP status code to use for redirects to an external icon service.
+        /// The supported codes are 307 (temporary) and 308 (permanent).
+        /// Temporary redirects are useful while testing different icon services, but once a service
+        /// has been decided on, consider using permanent redirects for cacheability.
+        icon_redirect_code:     u32,    true,   def,    307;
         /// Positive icon cache expiry |> Number of seconds to consider that an already cached icon is fresh. After this period, the icon will be redownloaded
         icon_cache_ttl:         u64,    true,   def,    2_592_000;
         /// Negative icon cache expiry |> Number of seconds before trying to download an icon that failed again.
@@ -691,6 +696,12 @@ fn validate_config(cfg: &ConfigItems) -> Result<(), Error> {
                 _ => err!(format!("Icon service URL `{}` has more than one placeholder \"{{}}\"", icon_service)),
             }
         }
+    }
+
+    // Check if the icon redirect code is valid
+    match cfg.icon_redirect_code {
+        307 | 308 => (),
+        _ => err!("Only HTTP 307/308 redirects are supported"),
     }
 
     Ok(())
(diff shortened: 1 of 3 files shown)


BlackDex:
I'm not sure, but won't a 301/308 cause browser to never recheck that specific url again? It will probably depend on the caching of the browser. But it could cause an issue that you can't easily switch to a different icon provide once activated?

jjlin:
> I'm not sure, but won't a 301/308 cause browser to never recheck that specific url again? It will probably depend on the caching of the browser. But it could cause an issue that you can't easily switch to a different icon provide once activated?

Yes, so as mentioned in the doc comments, configuring a 307 redirect is useful if you're testing different icon services. Clearing the browser cache should clear the 308 caching as well, so it shouldn't be that hard to get around.

It's certainly an option to make a 307 redirect the default to make it easier to test different icon services out of the box, but it probably means most users will never know or bother to set the 308 redirect in that case. I could very well be wrong, but I suspect most users interested in an external icon provider will just pick the one they perceive to be most trustworthy, see that it works, and move on.

So I'm not sure there's clearly a best default here. I don't feel that strongly about it though, so if people think 307 is a better default, that's fine with me.

src/config.rs
```diff
@@ -693,6 +697,12 @@ fn validate_config(cfg: &ConfigItems) -> Result<(), Error> {
         }
     }
 
+    // Check if the icon redirect code is valid
+    match cfg.icon_redirect_code {
+        307 | 308 => (),
+        _ => err!("Only HTTP 307/308 redirects are supported"),
```
RealOrangeOne, line 703: HTTP 301 and 302 are also valid redirect statuses

src/config.rs
```diff
@@ -454,9 +454,13 @@ make_config! {
...
         /// which is replaced with the domain. For example: `https://icon.example.com/domain/{}`.
         /// `internal` refers to Vaultwarden's built-in icon fetching implementation. If an external
-        /// service is set, an icon request to Vaultwarden will return an HTTP 307 redirect to the
+        /// service is set, an icon request to Vaultwarden will return an HTTP redirect to the
         /// corresponding icon at the external service.
         icon_service:           String, false,  def,    "internal".to_string();
+        /// Icon redirect code |> The HTTP status code to use for redirects to an external icon service.
+        /// The supported codes are 307 (temporary) and 308 (permanent).
+        /// Temporary redirects are mainly useful while testing different icon services.
+        icon_redirect_code:     u32,    true,   def,    308;
```
RealOrangeOne, line 463: IMO this should definitely be 307 by default. Avoid as much user annoyance by default. Given it's a simple redirect, performance shouldn't be degraded too much by it

src/config.rs
BlackDex: 307 and 308 are protocol fixed redirects. While in this case those will probably be fine, since icon requests are only `GET` and not `POST`.

Also, users which are still using an older OS/Browser could have issues with 308. Mostly Below Windows 10.

RealOrangeOne: Support for both [307](https://caniuse.com/mdn-http_status_307) and [308](https://caniuse.com/mdn-http_status_308) actually looks pretty good

src/config.rs
```diff
@@ -693,6 +697,12 @@ fn validate_config(cfg: &ConfigItems) -> Result<(), Error> {
         }
     }
 
+    // Check if the icon redirect code is valid
+    match cfg.icon_redirect_code {
```
RealOrangeOne, line 701: Suggestion: This could be written better as `if matches!(cfg.icon_redirect_code, 307 | 308) {`. Alternatively, defining the allowed status codes as a `const` array and using `.contains` could also be nice.

src/config.rs
jjlin: I'm aware 301/302 are valid, hence why the error message says "supported", not "valid". I don't see a great reason to implement 301/302 since there is wide support for 307/308 in non-ancient browsers, but they could easily be added later on if really needed.

src/config.rs
jjlin: That's fine, I'll change it to 307 as default.

src/config.rs
jjlin: The current form reads more clearly IMO, and will be familiar to more people.

BlackDex (approved):

PR #2195, opened 2021-12-31
Remove references to "bwrs"

RealOrangeOne:
Some things left over from the rebrand.

Requires https://github.com/dani-garcia/bw_web_builds/pull/58

Note: Because this renames the admin cookie, it will log users out of the admin. I don't think that's a huge concern.


dani-garcia:
We should, at least for a while, accept both BWRS_VERSION and VW_VERSION in the build script, I know some build scripts for Linux distros are using the BWRS_VERSION to apply the version number.

The admin logout isn't an issue as the cookie expires in 20 minutes anyway.

src/main.rs
```diff
@@ -1,4 +1,5 @@
 #![forbid(unsafe_code)]
+#![feature(const_option_ext)]
```
BlackDex, line 2: I do not think we want to have unstable features.
Especially since we will switch to the `async` branch which uses `stable` as the default in the near future, where this is not allowed.

BlackDex (commented):
I suggest to remove the `const_option_ext` to make it compatible with Rust `stable`

williamdes (approved):

build.rs
```diff
@@ -15,8 +15,8 @@ fn main() {
         "You need to enable one DB backend. To build with previous defaults do: cargo build --features sqlite"
     );
 
-    if let Ok(version) = env::var("BWRS_VERSION") {
-        println!("cargo:rustc-env=BWRS_VERSION={}", version);
+    if let Ok(version) = env::var("VW_VERSION") {
```
dani-garcia, line 18: I think the check for BWRS_VERSION should be done here instead of the get_version(), otherwise if you only define BWRS_VERSION it would fail this if and try to read the git info.

RealOrangeOne: Also removes some specialcasing from the resulting binary (and the hacks), which is always nice!

jjlin:
Maybe squash these commits? Or at least the ones that really have no value in being separate commits...

dani-garcia:
This is squash merged in https://github.com/dani-garcia/vaultwarden/commit/f4a9645b54f1c7361799c394360fdc079df9f502, but for some reason GitHub didn't catch it, I probably did something weird with the merge command. Thanks!

RealOrangeOne:
Looks like the merge commit you created didn't track the author correctly either, weird.

PR #2218, opened 2022-01-09
Add support for legacy HTTP 301/302 redirects for external icons

jjlin:
At least on Android, it seems the Bitwarden mobile client responds to
HTTP 307, but not to HTTP 308 for some reason.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -144,10 +144,11 @@
 
 ## Icon redirect code
 ## The HTTP status code to use for redirects to an external icon service.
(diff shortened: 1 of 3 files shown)


BlackDex:
Probably because Google abuses 308 for something else in Drive.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308

Maybe it has some connection, or the library used by Bitwarden doesn't support it properly.

BlackDex (approved):

PR #2255, opened 2022-01-23
Bump rust version to mitigate CVE-2022-21658

dscottboggs:
[CVE-2022-21658](https://www.cve.org/CVERecord?id=CVE-2022-21658) is a vulnerability in the Rust standard library. Updating to the latest rust nightly brings in the changes from Rust 1.58.1 which mitigate this issue.

See also [Rust#93110](https://github.com/rust-lang/rust/pull/93110). Crucially, this may [require further attention](https://github.com/rust-lang/rust/issues/93129).


Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "vaultwarden"
 version = "1.0.0"
 authors = ["Daniel García <dani-garcia@users.noreply.github.com>"]
 edition = "2021"
-rust-version = "1.57"
+rust-version = "1.58"
```
BlackDex, line 6: This actually should be 1.60 since the nightly version 2022-01-23 is that version.
And since this branch needs nightly i think it would be better to change it to v1.60.

docker/Dockerfile.j2
```diff
@@ -3,7 +3,7 @@
 # This file was generated using a Jinja2 template.
 # Please make your changes in `Dockerfile.j2` and then `make` the individual Dockerfiles.
 
-{% set build_stage_base_image = "rust:1.57-buster" %}
+{% set build_stage_base_image = "rust:1.58-buster" %}
 {% if "alpine" in target_file %}
 {%   if "amd64" in target_file %}
 {%     set build_stage_base_image = "blackdex/rust-musl:x86_64-musl-nightly-2021-12-25" %}
```
BlackDex, line 9: You should also change the nightly version here.
This prevents downloading the the rust toolchain during building, since it is already installed.
It should be done for all `build_stage_base_image` parameters.

BlackDex (changes requested):

PR #2315, opened 2022-02-16
Add support for custom .env file path

TinfoilSubmarine:
Hi! I'm working on packaging this for Void, and in writing the runit service, I came across an annoyance: since the `.env` file can't be in the current working directory, it won't find it and thus has to read from the environment variables directly. This *can* be done with something like:
```
#!/bin/sh
exec chpst -u _vaultwarden:_vaultwarden \
        env $(grep -v '^#' ${ENV_FILE:-/etc/vaultwarden.env} | xargs -d '\n') vaultwarden
```
But that is less than ideal IMO. This patch is a simple way to pass a single custom location where the env/config can be found.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -3,6 +3,11 @@
 ##
 ## Be aware that most of these settings will be overridden if they were changed
 ## in the admin interface. Those overrides are stored within DATA_FOLDER/config.json .
+##
+## By default, vaultwarden expects for this file to be named ".env" and located
+## in the current working directory. If this is not the case, the environment
+## variable ENV_FILE can be set to the location of this file prior to starting
+## vaultwarden.
 
 ## Main data folder
 # DATA_FOLDER=data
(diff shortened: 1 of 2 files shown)


williamdes (approved):

jjlin (changes requested):
There should also be some docs towards the top of https://github.com/dani-garcia/vaultwarden/blob/main/.env.template so people know this env var is available

TinfoilSubmarine:
I should mention that another solution to this is to cd to the directory containing the `.env` file (or a symlink to it) prior to executing `vaultwarden`. So my service file would look like:
```
ln -sf ${ENV_FILE:-/etc/vaultwarden.env} /var/lib/vaultwarden/.env
cd /var/lib/vaultwarden
exec chpst -u _vaultwarden:_vaultwarden vaultwarden
```
This works for my use case, but I'll leave this PR here in case it's seen as a benefit.

PR #2377, opened 2022-03-19
docs: add Helm installation steps

guerzon:
This PR adds steps to install using Helm.

This is also a shameless plug of my project and an invitation for Kubernetes users to test and provide feedback.

Cheers,
Lester


BlackDex:
I think the wiki is a better location for this instead of cluttering the README.
Please create a wiki article with this :).

guerzon:
> I think the wiki is a better location for this instead of cluttering the README. Please create a wiki article with this :).

Hi @BlackDex, thanks for the feedback. I have added a section to the [Deploment examples](https://github.com/dani-garcia/vaultwarden/wiki/Deployment-examples).

Issue #1402, opened 2021-02-18
N+1 query problem on api/sync


raphaelcoutu:
### Subject of the issue

Syncing/Loading ciphers with a remote PostgreSQL can be really slow (~ 50-60 seconds) due to a very large amount of SQL queries. The problem doesn't seem to affect users who are using SQLite or docker/local SQL databases due to no/minimal latency roundtrips. Browser extension could simply not sync.

### Your environment

* Install method: docker image
* Clients used: Web or Browser extension
* Bitwarden_rs version: v1.19.0
* Web-vault version: v2.18.1
* Running within Docker: true
* Internet access: true
* Uses a proxy: false
* DNS Check: true
* Time Check: true
* Domain Configuration Check: true
* HTTPS Check: false
* Database type: PostgreSQL
* Version of mysql/postgresql: 12.6
I have about ~300 ciphers in my Bitwarden wallet. My remote PostgreSQL server is about 30-50 ms from my local docker Bitwarden_rs setup (for testing purposes).

### Steps to reproduce

1. Creating a docker container with bitwarden_rs
2. Importing Bitwarden wallet (official self hosted) into bitwarden_rs
3. Loading vault home page

I used different setups to identity the problem. I have no problem with local SQLite or docker PostgreSQL.

I finally found the problem using a PostgreSQL log configuration in postgresql.conf. There were a lot of requests written for a single sync (2400 lines so I estimate about 1000 requests):
```
log_statement = 'all
log_directory = 'log'
log_filename = 'postgresql-%Y-%m-%d_%H%M%S.log'
logging_collector = on
```

Docker-compose used:
```yaml
version: '3'
services:
  # postgres:
  #   image: postgres:12
  #   restart: always
  #   volumes:
  #     - pgdata:/var/lib/postgresql/data
  #   environment:
  #     - POSTGRES_PASSWORD=admin
  #     - TZ=America/Montreal
  bitwarden:
    image: bitwardenrs/server
    restart: always
    volumes:
      - ./bw-data:/data
    ports:
      - 8080:80
      #- 3012:3012
    environment:
      - DOMAIN=http://localhost:8080
      - DATABASE_URL=postgresql://bitwardenrs:<<password>>@<<host>>:5432/bitwardenrs
      # (testing local) - DATABASE_URL=postgresql://postgres:admin@postgres
      - ENABLE_DB_WAL=false
      - SIGNUPS_ALLOWED=true
      #- WEBSOCKET_ENABLED=true
[...]

PR #2429, opened 2022-04-22
Improve sync speed and updated dep. versions

BlackDex:
Improved sync speed by resolving the N+1 query issues.
Solves #1402 and Solves #1453

With this change there is just one query done to retreive all the important data, and matching is done in-code/memory.
With a very large database the sync time went down about 3 times.

Also updated misc crates and Github Actions versions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,7 +46,7 @@ jobs:
     steps:
       # Checkout the repo
       - name: Checkout
-        uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.4.0
+        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # v3.0.2
       # End Checkout the repo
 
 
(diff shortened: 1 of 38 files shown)


manofthepeace:
Just wondering if this crate could be used https://crates.io/crates/tokio_cron_scheduler instead of a not so maintained version of the current job_scheduler. It is based on job_scheduler, and seem actively maintained for now at least.

BlackDex:
Small update.

I have fixed all the full-sync code to use the the `CipherSyncData`.
Depending on how many ciphers, collections and organizations a user has access to, with my tests I got a speedup of nearly 3x using this code.

I still checking for other optimizations to see if i can add those.

BlackDex:
Did some final checking and this looks like merge-ready.
Deleting multiple ciphers can still be optimized, but that currently is very entwined with functions calling other functions which could also call other functions.

There are also some optimizations which could be done with web-sockets, but also that need some more work and probably better for an other PR.

dani-garcia:
Thanks for your work @BlackDex!

bendem:
Just wanted to say that this is the greatest thing. Loading the home list went down from 4-8s to <0.5s. Best optimisation since the bread slicer!

PR #2433, opened 2022-04-24
Add `/api/{alive,now,version}` endpoints

jjlin:
The added endpoints work the same as in their upstream implementations.

Upstream also implements `/api/ip`. This seems to include the server's public
IP address (the one that should be hidden behind Cloudflare), which doesn't
seem like a great idea.

### DIFF
--- a/src/api/core/mod.rs
+++ b/src/api/core/mod.rs
@@ -12,8 +12,10 @@ pub use sends::purge_sends;
 pub use two_factor::send_incomplete_2fa_notifications;
 
 pub fn routes() -> Vec<Route> {
-    let mut mod_routes =
-        routes![clear_device_token, put_device_token, get_eq_domains, post_eq_domains, put_eq_domains, hibp_breach,];
+    let mut device_token_routes = routes![clear_device_token, put_device_token];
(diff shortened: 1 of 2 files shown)


BlackDex (approved):

BlackDex:
Yea, these are some nice additions. I agree on the IP endpoint. Even if you would hide the servers public endpoint, you could still figure out the forwarded-for IP's which might also be unwanted.

GoliathLabs:
@jjlin @BlackDex would it be an option that you could switch whether `/api/ip` should expose the IP or not? Since most users are probably not behind cloudflare and therefore have no disadvantage in providing the IP endpoint

hevisko:
Hmmm... What would /api/ip return if it's behind a cloudflare behind a caddy/nginx, ie. the vaultwarden's IP would be localhost, the caddy's IP is a rfc1918 IP, which sits behind a firewall that NATs the external IP to the caddy, that might an IP behind  cloudflare, so perhaps making the IP a trinary setting would be great:, like:
 0.0.0.0 - off, return 404
 anything from 0.0.0.1 to 255.255.255.254 - the value given (might be bogus)
 255.255.255.255 - the IP that vaultwarden is listening on

PR #2440, opened 2022-04-27
Database connection init

jjlin:
Adds support for database connection init statements. This is probably mainly useful for running connection-scoped pragma statements.

Also adds default connection-scoped pragmas for SQLite:

* [`PRAGMA busy_timeout = 5000`](https://www.sqlite.org/pragma.html#pragma_busy_timeout) tells SQLite to keep trying for up to 5000 ms when there is lock contention, rather than aborting immediately. This should hopefully prevent the vast majority of "database is locked" panics observed since the async transition, e.g.: #2436

* [`PRAGMA synchronous = NORMAL`](https://www.sqlite.org/pragma.html#pragma_synchronous) trades better performance for a small potential loss in durability (the default is `FULL`). The SQLite docs recommend `NORMAL` as "a good choice for most applications running in WAL mode".

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -515,10 +515,19 @@ make_config! {
         db_connection_retries:  u32,    false,  def,    15;
 
         /// Timeout when aquiring database connection
-        database_timeout:     u64,    false,  def,    30;
+        database_timeout:       u64,    false,  def,    30;
 
         /// Database connection pool size
         database_max_conns:     u32,    false,  def,    10;
+
+        /// SQLite connection init |> Statements to run when creating a new SQLite connection
+        sqlite_conn_init:       String, false,  def,    "".to_string();
+
+        /// MySQL connection init |> Statements to run when creating a new MySQL connection
+        mysql_conn_init:        String, false,  def,    "".to_string();
+
+        /// PostgreSQL connection init |> Statements to run when creating a new PostgreSQL connection
+        postgresql_conn_init:   String, false,  def,    "".to_string();
 
         /// Bypass admin page security (Know the risks!) |> Disables the Admin Token for the admin page so you may use your own auth in-front
         disable_admin_token:    bool,   true,   def,    false;
--- a/src/config.rs
+++ b/src/config.rs
@@ -521,7 +521,7 @@ make_config! {
         database_max_conns:     u32,    false,  def,    10;
 
         /// SQLite connection init |> Statements to run when creating a new SQLite connection
-        sqlite_conn_init:       String, false,  def,    "".to_string();
+        sqlite_conn_init:       String, false,  def,    "PRAGMA busy_timeout = 5000; PRAGMA synchronous = NORMAL;".to_string();
 
         /// MySQL connection init |> Statements to run when creating a new MySQL connection
         mysql_conn_init:        String, false,  def,    "".to_string();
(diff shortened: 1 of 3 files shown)


src/config.rs
```diff
@@ -515,11 +515,20 @@ make_config! {
...
 
         /// Timeout when aquiring database connection
-        database_timeout:     u64,    false,  def,    30;
+        database_timeout:       u64,    false,  def,    30;
 
         /// Database connection pool size
         database_max_conns:     u32,    false,  def,    10;
 
+        /// SQLite connection init |> Statements to run when creating a new SQLite connection
+        sqlite_conn_init:       String, false,  def,    "PRAGMA busy_timeout = 5000; PRAGMA synchronous = NORMAL;".to_string();
```
BlackDex, line 524: Do we really need multiple options?
Can't we just use one option, and if empty set some defaults determined by the connection type?
Since we can only have one database type at all per running application.

BlackDex (changes requested):

src/config.rs
jjlin: It could be done that way, but then it's not as clear to the user what the default is for each database type. Since pragmas are database-specific, you can also end up with cases where the user sets init statements for one database, then switches to another database and inadvertently carries over the incompatible init statements.

BlackDex: That kinda same goes for the `DATABASE_URL` parameter. While that does implicate the type of the database it self within the string, it still is one variable. If a user is going to switch from one db-type to an other, that does take some manual effort. And with that manual effort i think changing the `init` config should be part of it.

In reality, how often do people change there database type and would also have the init configured manually if we already provide a sane default.

Documenting the default per type, and stating it will use the default only for that type seems more logical to me.

I find adding 3 more configuration items which are mutually exclusive anyways based upon the database type is to much and only clutters the configuration more then it already is.

jjlin: As I alluded to earlier, the main advantage of separate config options is that the defaults are *self-documenting*. From the admin page, the user can literally see what the default would be, straight from the source of truth. (There are not really useful defaults you could have for `database_url` in the MySQL/PostgreSQL cases, so that's a key difference.)

With a single config option, the obvious default is the empty string, to indicate that the real default is to be determined elsewhere. This makes things more magical for the average user, who must now take on faith that the docs are up-to-date. To users who doesn't read the docs closely or undertand English well, it may even give the impression that the default is simply the empty string (i.e., no connection init).

As far as adding clutter goes, you either have extra lines to define config options along with their defaults, or extra lines documenting what the defaults are for each database type. I don't see a substantial difference there.

That said, I understand why having less config options "feels" cleaner -- I'm just pointing out why I don't see it as a clear win in this case. I also don't feel as strongly about it as you seem to, so I've changed it to the single config approach to move things forward.

BlackDex: Well, I'm always open for discussion about anything. And i might appear to be a bit harsh about this, but i just want to minimize the already bloated config options.

With the latest commit it looks ok for me :).
Very nice addition, since i do think some people could really benefit from this.
And with the async code, for SQLite it was needed in some situations.

Thx!

BlackDex (approved):

dani-garcia:
With this merged we can probably consider removing some of the uses of crate::util::retry() which I think was added to solve this "Database locked" issue as well.

BlackDex:
That depends. If this was only for sqlite, the probably yes.
It could also happen with mysql when deleting and adding ciphers are done at the same time and a lock is on a table of course.

PR #2455, opened 2022-05-08
Bump lettre to 0.10.0-rc.6

paolobarbolini:
Changes: https://github.com/lettre/lettre/compare/v0.10.0-rc.5...v0.10.0-rc.6


BlackDex:
This is also updated in #2429 already.
Not that it should hurt.

BlackDex (approved):

dani-garcia:
Lettre is updated with #2429 now, thanks for the work you put into it!

PR #2479, opened 2022-05-20
Migrate old ws crate to tungstenite

dani-garcia:
This new crate supports async, is more popular and is better supported. As another benefit, we dropped nearly 20 duplicate crates with older non-semver-compatible versions that were used by the ws crate.

I'm opening this as a PR to get some reviews and to avoid blocking a new release, which I'll hopefully be able to do this weekend.

### DIFF
--- a/src/api/notifications.rs
+++ b/src/api/notifications.rs
@@ -258,7 +258,7 @@ fn create_ping() -> Vec<u8> {
 }
 
 #[allow(dead_code)]
-#[derive(PartialEq)]
+#[derive(Eq, PartialEq)]
 pub enum UpdateType {
     CipherUpdate = 0,
     CipherCreate = 1,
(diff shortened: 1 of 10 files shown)


src/api/notifications.rs
```diff
@@ -416,27 +282,147 @@ pub enum UpdateType {
...
+    let (tx, mut rx) = tokio::sync::mpsc::channel(100);
+    users.map.entry(user_uuid.clone()).or_default().push((entry_uuid, tx));
+
+    let mut interval = tokio::time::interval(Duration::from_secs(15));
+    loop {
+        tokio::select! {
+            res = stream.next() =>  {
+                match res {
+                    Some(Ok(message)) => {
+                        //info!("RECEIVED {message:?}");
```
BlackDex, line 352: This can be removed i think :)

BlackDex:
Also, running a nightly clippy on the code mentions to add `Eq` besides `PartialEq` at line 261 for the `UpdateType` enum.
But that is not a big issue for now i think.

BlackDex (approved):

dani-garcia:
Merged in 
https://github.com/dani-garcia/vaultwarden/commit/54c78cf06daff8d9cc920825b38c8d9b711edbc9 - https://github.com/dani-garcia/vaultwarden/commit/e33b8fab34aa5fa9b6f3f3ca5c97f01c0a091847

PR #2523, opened 2022-06-01
Adding "UserEnabled" and "CreatedAt" member to the json output of a User

Lowaiz:
Adding "UserEnabled"  and "CreatedAt" member to the json output of a User in the `admin/users` and `admin/users/<ID>` web routes.

### DIFF
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -79,6 +79,7 @@ fn admin_disabled() -> &'static str {
 
 const COOKIE_NAME: &str = "VW_ADMIN";
 const ADMIN_PATH: &str = "/admin";
+const DT_FMT: &str = "%Y-%m-%d %H:%M:%S %Z";
 
 const BASE_TEMPLATE: &str = "admin/base";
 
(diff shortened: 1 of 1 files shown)


BlackDex (changes requested):
I'd rather not add those items there, since these are also used to return the user object to the Bitwarden clients.
It is better to add them to the generated json output within `admin.rs` instead.

Same goes for #2524, and in that case, i would probably suggest to merge them into one PR instead of two different ones.

BlackDex:
Since the rest of the json is CamelCase i think it is best to keep that the same in this case.

Lowaiz:
Here is the Camel Case !

BlackDex:
Could you squash the commits into one commit? Else it will clutter the git-history a bit.
You could do that with (from the top of my head)
```bash
git reset --soft HEAD~4
git commit -a
git push origin +add_disabled_member_to_json_user
```
That should reset the commits to the starting point you were, and leave the changes you made.
Then you can commit that again, and do a force push so the latest commit will be the only commit here :) .

Lowaiz:
Squash done and thank you for the commands.

src/api/admin.rs
```diff
@@ -307,10 +307,14 @@ fn logout(cookies: &CookieJar<'_>, referer: Referer) -> Redirect {
 
 #[get("/users")]
 async fn get_users_json(_token: AdminToken, conn: DbConn) -> Json<Value> {
+    const DT_FMT: &str = "%Y-%m-%d %H:%M:%S %Z";
```
BlackDex, line 310: I think it is better to move this `const` and all the others with the same value outside of the functions and place it near the other global `const` at line 80.

BlackDex (changes requested):

BlackDex:
The build fails because it's not formatted correctly.
I suggest to install `pre-commit` and use that.
See: https://pre-commit.com/

This repo already has the correct pre-commit config there, all you need to do after you installed it is run `pre-commit install`, then run the `--reset soft` again and do a commit, this will check everything which is needed for the PR to be valid.

BlackDex (approved):

PR #2529, opened 2022-06-03
Add Spanish version of USA healthcare.gov

drank-sinatra:
The site for USA health insurance has a totally separate base domain in Spanish

### DIFF
--- a/src/static/global_domains.json
+++ b/src/static/global_domains.json
@@ -328,6 +328,7 @@
     "Type": 33,
     "Domains": [
       "healthcare.gov",
+      "cuidadodesalud.gov",
       "cms.gov"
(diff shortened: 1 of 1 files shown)


BlackDex:
Thanks for the PR.

However we extract this information from the Upstream Bitwarden project using this tool: https://github.com/dani-garcia/vaultwarden/blob/main/tools/global_domains.py

It extract the data from this file: https://github.com/bitwarden/server/blob/master/src/Core/Utilities/StaticStore.cs

I suggest to create a PR there to have it updated for Bitwarden it self.
Also, since that would fix it for all users and all clients using either Bitwarden or Vaultwarden.

Issue #829, opened 2020-01-29
Verification mail sending error IDN Domain


Latschenharry:
Bitwardenrs is unable to send an verification mail to an IDN Domain.

### Your environment

* Bitwarden_rs version: 1.13.1-632d5526
* Install method: Docker Image
* Clients used: Webclient
* Other relevant information: Bitwardenrs is able to send a mail to an ACE converted Domain

### Steps to reproduce
Create an account with an IDN domain mail

### Expected behaviour
bitwardenrs is able to send the verification mail

### Actual behaviour
bitwardenrs is unable to send the verification mail

[bitwarden_rs::api::core::accounts][ERROR] Error sending delete account email: Error building email. invalid email address

dani-garcia:
I'm not sure about your terminology, when you say IDN domain do you mean using non-ascii characters? And by ACE converted do mean punycode? (Starting with "xn--").

If so, I haven't tested it but we could consider converting the email internally, to make sure the library accepts it.

Latschenharry:
Yes its exactly as described by you.

IDN=Internationalized Domain Name so for example:  überall-ist.de
ACE/punycode: xn--berall-ist-8db.de

An internal conversion would be very nice.

dani-garcia:
Okay, I've changed it to convert the domain to ascii now in https://github.com/dani-garcia/bitwarden_rs/commit/def174a517716967b80d5ecb8407772668bfd631, can you give it a try? if you are using the docker images, those might take some hours to build still.

Latschenharry:
Thank you very much for the fast help.
It works like it should.

PR #2532, opened 2022-06-04
Small email sending code improvements

paolobarbolini:
Changes building `Address` by going through the `FromStr` implementation. Also removes the manual `MultiPart` build implementation, which must have been a workaround for a bug in a previous version.

Depends on #2531

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3662,7 +3662,6 @@ dependencies = [
  "governor",
  "handlebars",
  "html5gum",
- "idna 0.2.3",
  "job_scheduler",
  "jsonwebtoken",
  "lettre",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -108,7 +108,6 @@ webauthn-rs = "0.3.2"
 url = "2.2.2"
 
 # Email libraries
-idna = "0.2.3" # Punycode conversion
 lettre = { version = "0.10.0-rc.6", features = ["smtp-transport", "builder", "serde", "native-tls", "hostname", "tracing"], default-features = false }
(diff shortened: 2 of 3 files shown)


BlackDex:
Thanks @paolobarbolini , looks nice.
Regarding the Punycode, i have no clue actually, i tested it without the conversion, and it seems to work just fine.
@dani-garcia do you remember why you added that maybe?

BlackDex (approved):

dani-garcia:
We changed it to use punycode because of a user having trouble using it with an internationalized domain: https://github.com/dani-garcia/vaultwarden/issues/829.

Not sure if that issue is applicable still.

BlackDex:
Looking at the changes done on the lettre side, i think this should be fixed now right @paolobarbolini ?

paolobarbolini:
> Looking at the changes done on the lettre side, i think this should be fixed now right @paolobarbolini ?

I'll have to test it somehow because I don't think we ever got any complaints about IDNs, so unless we happened to fix it in the middle of doing other changes we're still handling them wrong.

BlackDex:
@paolobarbolini Well you changed the email verifying now right? So that might have solved it right now?
I did some testing with several unicode domains, and it looks like not generating an issue.

paolobarbolini:
> @paolobarbolini Well you changed the email verifying now right? So that might have solved it right now? I did some testing with several unicode domains, and it looks like not generating an issue.

Ah right. The error is in the fact that verification was broken, not that we didn't encode the email properly 😅

paolobarbolini:
Changed. I didn't realize when the issue came up that vaultwarden wasn't even on lettre 0.10.0-alpha.1, so it was still using the ancient email builder :smile:

PR #2565, opened 2022-06-20
Allow FireFox relay in CSP.

BlackDex:
This PR is needed for https://github.com/dani-garcia/bw_web_builds/pull/71
Without this the web-vault will refuse to make calls to the FireFox Relay API.

Also fixed a small issue with the pre-commit config.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,8 +26,8 @@ repos:
       entry: cargo test
       language: system
       args: ["--features", "sqlite,mysql,postgresql,enable_mimalloc", "--"]
-      types_or: [file, rust]
-      files: (Cargo.toml|Cargo.lock)
+      types_or: [rust, file]
+      files: (Cargo.toml|Cargo.lock|.*\.rs$)
       pass_filenames: false
     - id: cargo-clippy
       name: cargo clippy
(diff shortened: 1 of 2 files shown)


GeekCornerGH (approved):

BlackDex:
I Updated the comments to give a bit more information, since this is security related, lets have some better info there.

Flash1232:
Just on a sidenote: AnonAddy (maybe also SimpleLogin) allows for self-hosted instances. Is there any chance this whitelist could be made configurable?

BlackDex:
> Just on a sidenote: AnonAddy (maybe also SimpleLogin) allows for self-hosted instances. Is there any chance this whitelist could be made configurable?

That is not an option for now, since Bitwarden does not support self-hosted instances of those services. They only allow the public services. There is no way to configure this in Bitwarden (As far as i could tell).

KainAlive:
Is Firefox Relay supported in Vaultwarden? I'm using the latest docker image and only get "AnonAddy", "Fastmail" and "SimpleLogin" as choices.

BlackDex:
> Is Firefox Relay supported in Vaultwarden? I'm using the latest docker image and only get "AnonAddy", "Fastmail" and "SimpleLogin" as choices.

That doesn't work for self hosted environments, only Bitwarden Cloud.

PR #2567, opened 2022-06-21
Fix hidden ciphers within organizational view.

### LINKED ISSUE #2566: Bug displaying organization elements that are in a folder
### Subject of the issue

Bug displaying organization elements that are in a folder (web vault version)

### Deployment environment

* Vaultwarden version: v1.25.0-887e320e
* Web-vault version: v2022.5.2 (also tested w/ v2022.05.0 from your last release web build)
* Running within Docker: false (Base: Debian)
* Uses a reverse proxy: true
* Internet access: true
* DNS Check: true
* Time Check: true
* HTTPS Check: true
* Database type: SQLite
* Database version: 3.35.4
* Clients used: Web, Android, Windows, Chrome, Firefox
* Reverse proxy and version: Nginx 1.21.6
* Other relevant information: 

### Steps to reproduce

1. Create a new item in an organization without assigning it a folder (the option should not be available in this case)
2. Go to the personal vault in "All vaults", find the new item and define a folder for it
(Or replace steps 1 and 2 by moving a personal item **that is in a folder** to an organization)
3. Go back to the "Organizations" tab, all items are displayed. But if you click in the left pane on "All items" or even select a collection for example, all the items in a folder are invisible (image 02).
4. The problem is not present as soon as I set the folder to "No folder".

### Expected behaviour / Actual behaviour

The organization items for which I have defined a folder should be displayed instead of disappearing.

Or it should be impossible to assign a folder to an item in the organization if it is not supposed to be possible, but the problem is only on the web version.

### Troubleshooting data

* OK - image01 : https://i.imgur.com/RhgwzXm.png

* Not OK - image02 : https://i.imgur.com/M89mq6h.png

Sorry for my English.

BlackDex:
This PR fixes #2566
If Organizational syncs returned a FolderId it would cause the web-vault
to hide the cipher because there is a FolderId set. Upstream seems to
not return FolderId and Favorite. When set to null/false it will behave
the same.

In this PR I have added a new CipherSyncType enum to select which type
of sync to execute, and return an empty list for both Folders and Favorites if this is for Orgs.
This also reduces the database load a bit since it will not execute those queries.

### DIFF
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -104,7 +104,7 @@ async fn sync(data: SyncData, headers: Headers, conn: DbConn) -> Json<Value> {
     // Get all ciphers which are visible by the user
     let ciphers = Cipher::find_by_user_visible(&headers.user.uuid, &conn).await;
 
-    let cipher_sync_data = CipherSyncData::new(&headers.user.uuid, &ciphers, &conn).await;
+    let cipher_sync_data = CipherSyncData::new(&headers.user.uuid, &ciphers, CipherSyncType::User, &conn).await;
 
     // Lets generate the ciphers_json using all the gathered info
     let ciphers_json: Vec<Value> = stream::iter(ciphers)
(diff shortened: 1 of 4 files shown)


GeekCornerGH (approved):
Looks fixed yeah :)

PR #2595, opened 2022-07-06
Address inconsistency v{version} with and without a v in the version with most recent updates. 

nneul:
Looks like it mainly just affects documentation of process, but it did result in some of the files having 'vv2022.5.02' in them. 

Any idea when the new vault will be merged to production/latest?


BlackDex:
I think there are a few more files which need to be adjusted.
Looks like you didn't run `make` within the `docker` directory which would have regenerated all the files.

williamdes (approved):

dani-garcia:
I've updated the missing ones on https://github.com/dani-garcia/vaultwarden/commit/93d4a12834b274f44e66bb6be801552a5b22dc27, thanks for the tip!

PR #2605, opened 2022-07-08
Fall back to move_copy_to  if persist_to fails while saving uploaded files.

ruifung:
This is to support scenarios where the attachments and sends folder are to be stored on a separate device from the tmp_folder (i.e. fuse-mounted S3 storage), due to having the tmp_dir on the same device being undesirable.

Example being fuse-mounted S3 storage with the reasoning that because S3 basically requires a copy+delete operations to rename files, it's inefficient to rename files on device, if it's even allowed.

### DIFF
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -17,7 +17,7 @@ use crate::{
     CONFIG,
 };
 
-use futures::{stream, stream::StreamExt};
+use futures::{stream, stream::StreamExt, TryFutureExt};
 
 pub fn routes() -> Vec<Route> {
     // Note that many routes have an `admin` variant; this seems to be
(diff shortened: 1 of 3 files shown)


dani-garcia:
Would it make more sense to do the `persist_to` operation first, and if that fails with a cross device move error, retry using `move_copy_to`? That way there won't be any need for an extra configuration option. What do you think?

williamdes:
> Would it make more sense to do the `persist_to` operation first, and if that fails with a cross device move error, retry using `move_copy_to`? That way there won't be any need for an extra configuration option. What do you think?

I also think it would be cleaner

ruifung:
For now the ErrorKind:CrossesDevices enum is still marked unstable though.

So I've just had it fall back on copying regardless of the error, should I perhaps have made use of the unstable Enum?

Apologies, but this is literally the first ever code I've ever written in rust.

williamdes (approved):
Congratulations for your first Rust code! 👏🏻

dani-garcia:
Thank you and good job!

PR #2611, opened 2022-07-10
Add more clippy checks for better code/readability

BlackDex:
A bit inspired by @paolobarbolini from this commit at lettre https://github.com/lettre/lettre/pull/784 .
I added a few more clippy lints here, and fixed the resulted issues.

Overall i think this could help in preventing future issues, and maybe
even peformance problems. It also makes some code a bit more clear.

We could always add more if we want to, i left a few out which i think
arn't that huge of an issue. Some like the `unused_async` are nice,
which resulted in a few `async` removals.

Some others are maybe a bit more estatic, like `string_to_string`, but i
think it looks better to use `clone` in those cases instead of `to_string` while they already are a string.

### DIFF
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -427,7 +427,7 @@ async fn update_user_org_type(data: Json<UserOrgTypeData>, _token: AdminToken, c
         }
     }
 
-    user_to_edit.atype = new_type as i32;
+    user_to_edit.atype = new_type;
     user_to_edit.save(&conn).await
 }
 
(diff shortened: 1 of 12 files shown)


BlackDex:
Let me know what you think of this, i have left out `string_add` for example, since i do not mind the `x + "some-string"` my self. But that is more a personal preference. We can always add/remove others if we want to make it more/less strict.

dani-garcia:
So far I agree 100% with those changes, yeah!

PR #2631, opened 2022-07-19
Fix persistent folder check within containers

### LINKED ISSUE #2622: Docker container v1.25.1 persistent check with volumes fails to start
### Subject of the issue
On starting of the latest (1.25.1) version, it will error stating there is no .env and no persistent storage although configured and working in 1.25.0

### Deployment environment

* vaultwarden version: 1.25.1

* Install method: Docker

* Clients used: Application doesn't start at all so presume all clients

* Reverse proxy and version: N/A

* MySQL/MariaDB or PostgreSQL version: N/A

* Other relevant details:

### Steps to reproduce

Run the below docker compose; (I have every variable populated but removed for this issue raise)
```yaml
services:
  vaultwarden:
    image: vaultwarden/server:latest
    container_name: vaultwarden
    restart: unless-stopped
    ports:
      - 50003:80
    networks:
      - proxy
    volumes:
      - vaultwarden_data:/data
    environment:
      - ADMIN_TOKEN=
      - WEBSOCKET_ENABLED=true
      - SMTP_HOST=
      - SMTP_FROM=
      - SMTP_FROM_NAME=VaultWarden
      - SMTP_SECURITY=starttls
      - SMTP_PORT=587
      - SMTP_USERNAME=
      - SMTP_PASSWORD=
      - SMTP_TIMEOUT=30
      - SMTP_AUTH_MECHANISM="Plain"
      - LOGIN_RATELIMIT_MAX_BURST=10
      - LOGIN_RATELIMIT_SECONDS=60
      - DOMAIN=
      - INVITATION_ORG_NAME=
      - INVITATIONS_ALLOWED=true
      - SIGNUPS_ALLOWED=false
      - SIGNUPS_DOMAINS_WHITELIST=
      - SIGNUPS_VERIFY=true
      - SIGNUPS_VERIFY_RESEND_TIME=3600
      - SIGNUPS_VERIFY_RESEND_LIMIT=6
      - EMERGENCY_ACCESS_ALLOWED=true
      - SENDS_ALLOWED=true
      - WEB_VAULT_ENABLED=true
```

### Expected behaviour

To launch the application and be accessible via it's port or the remapped port if provided i.e. 50003:80 in this case

### Actual behaviour

Application does not start and reviewing the logs, the below is returned;

```
[INFO] No .env file found.
[2022-07-17 04:21:50.752][vaultwarden][ERROR] No persistent volume!
########################################################################################
# It looks like you did not configure a persistent volume!                             #
# This will result in permanent data loss when the container is removed or updated!    #
# If you really want to use volatile storage set `I_REALLY_WANT_VOLATILE_STORAGE=true` #
########################################################################################
```

BlackDex:
The previous persistent folder check worked by checking if a file
exists. If you used a bind-mount, then this file is not there. But when
using a docker/podman volume those files are copied, and caused the
container to not start.

This change checks the `/proc/self/mountinfo` for a specific patern to
see if the data folder is persistent or not.

Fixes #2622

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -61,6 +61,11 @@ use std::{
     thread,
 };
 
+use tokio::{
+    fs::File,
+    io::{AsyncBufReadExt, BufReader},
+};
+
 #[macro_use]
 mod error;
 mod api;
@@ -89,7 +94,7 @@ async fn main() -> Result<(), Error> {
 
     let extra_debug = matches!(level, LF::Trace | LF::Debug);
 
-    check_data_folder();
+    check_data_folder().await;
     check_rsa_keys().unwrap_or_else(|_| {
         error!("Error creating keys, exiting...");
         exit(1);
@@ -286,7 +291,7 @@ fn create_dir(path: &str, description: &str) {
     create_dir_all(path).expect(&err_msg);
 }
 
-fn check_data_folder() {
+async fn check_data_folder() {
     let data_folder = &CONFIG.data_folder();
     let path = Path::new(data_folder);
     if !path.exists() {
@@ -299,9 +304,10 @@ fn check_data_folder() {
         exit(1);
     }
 
-    let persistent_volume_check_file = format!("{data_folder}/vaultwarden_docker_persistent_volume_check");
-    let check_file = Path::new(&persistent_volume_check_file);
-    if check_file.exists() && std::env::var("I_REALLY_WANT_VOLATILE_STORAGE").is_err() {
+    if is_running_in_docker()
+        && std::env::var("I_REALLY_WANT_VOLATILE_STORAGE").is_err()
+        && !docker_data_folder_is_persistent(data_folder).await
+    {
         error!(
             "No persistent volume!\n\
             ########################################################################################\n\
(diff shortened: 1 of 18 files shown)


williamdes (approved):

src/main.rs
```diff
@@ -314,6 +320,39 @@ fn check_data_folder() {
...
 }
 
+/// Detect when using Docker or Podman the DATA_FOLDER is either a bind-mount or a volume created manually.
+/// If not created manually, then the data will not be persistent.
+/// A none persistent volume in either Docker or Podman is represented by a 64 alphanumerical string.
+/// If we detect this string, we will alert about not having a persistent self defined volume.
+/// This probably means that someone forgot to add `-v /path/to/vaultwarden_data/:/data`
+async fn docker_data_folder_is_persistent(data_folder: &str) -> bool {
+    const MOUNTINFO_FILE: &str = "/proc/self/mountinfo";
+    if Path::new(MOUNTINFO_FILE).exists() {
```
RealOrangeOne, line 330: Rather than Checking it it exists and then opening it, perhaps it'd be better and easier to just try and open it instead?

src/main.rs
```diff
@@ -314,6 +320,39 @@ fn check_data_folder() {
     }
 }
 
+/// Detect when using Docker or Podman the DATA_FOLDER is either a bind-mount or a volume created manually.
+/// If not created manually, then the data will not be persistent.
+/// A none persistent volume in either Docker or Podman is represented by a 64 alphanumerical string.
+/// If we detect this string, we will alert about not having a persistent self defined volume.
+/// This probably means that someone forgot to add `-v /path/to/vaultwarden_data/:/data`
+async fn docker_data_folder_is_persistent(data_folder: &str) -> bool {
+    const MOUNTINFO_FILE: &str = "/proc/self/mountinfo";
```
RealOrangeOne, line 329: What happens if this is run on a non-linux system? It's not a valid mountpoint, so I doubt it'll behave.

src/main.rs
```diff
@@ -314,6 +320,39 @@ fn check_data_folder() {
...
+            let data_folder_match = if data_folder.starts_with('/') {
+                format!(" {data_folder} ")
+            } else {
+                format!(" /{data_folder} ")
+            };
+            let mut lines = BufReader::new(mountinfo).lines();
+            while let Some(line) = lines.next_line().await.unwrap_or_default() {
+                // Only execute a regex check if we find the base match
+                if line.contains(&data_folder_match) {
+                    let re = regex::Regex::new(r"/volumes/[a-z0-9]{64}/_data /").unwrap();
```
RealOrangeOne, line 341: Probably best to compile this regex once rather than with each loop. Ideally statically, but if not just moving it to the top level function scope.

src/main.rs
BlackDex: For one, this check is only done when `if is_running_in_docker()` is true, and that will only happen on Linux based environments.

Also, if the File does not exists, it will just continue.

src/main.rs
BlackDex: I think this is a nicer way of actually checking if the file exists, instead of matching a result on Ok or Err.

src/main.rs
BlackDex: That regex is only done once, since it will only be run if the `line.containts()` matches. Since there can only be one line matching that, the regex is only compiled once. Having that statically does not help compilation or startup speed.

src/main.rs
BlackDex: btw, i just cross-compiled vaultwarden for `x86_64-pc-windows-gnu`, it doesn't start because of some other errors (via wine). But it just skips that function.

src/main.rs
BlackDex: While i thought about it overnight, it is actually a redundant check wink

RealOrangeOne (approved):

PR #2632, opened 2022-07-19
Update build workflow for CI

BlackDex:
Because we want to support MSRV, we also need to run a CI for this.
This PR adds checks for the MSRV and rust-toolchain defined versions.

It will also run all cargo test, clippy and fmt checks no matter the outcome of the previous job.
This will help when there are multiple issues, like clippy errors and formatting.
Previously it would show only the first failed check and stopped.

It will also output a nice step summary with some details on which checks have failed.
Or it will output a success message.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,6 +20,7 @@ on:
 
 jobs:
   build:
+    runs-on: ubuntu-20.04
     # Make warnings errors, this is to prevent warnings slipping through.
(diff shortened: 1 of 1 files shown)


williamdes:
Can you provide a link to what is MSRV?

BlackDex:
> Can you provide a link to what is MSRV?

Sure: https://rust-lang.github.io/rfcs/2495-min-rust-version.html

williamdes (approved):

PR #2667, opened 2022-08-02
Group support

MFijak:
Hello all,
This is my first ever pull request and I also never used Rust before. Please tell me if I did anything wrong. 
I implemented the group support, which is related to issue #1623. 

To test my changes following sites can be used:
- Organizations->Manage->People
  - Groups can be assigned to people (hover over a member and click the gear symbol)
  - Delete a member
- Organizations->Manage->Collections
  - Collections can be assigned to groups (click on a collection)
  - Delete a collection
  - Add a collection
- Organizations->Manage->Groups
  - Create a group (top right “+ New Group”)
  - Modify a group (click on a group)
  - Delete a group (click on a group)
  - Users can be added to groups (hover over a group and click the gear symbol)
- Vaults
  - Should list any collections which are visible by the user

There was one code segment I was not so sure about if my implementation is the correct approach:

`pub async fn find_by_user_uuid(user_uuid: &str, conn: &DbConn)`
_Db/models/collection.rs_

This function basically returns all collections which the user has access to. I squeezed my group implementation into the query builder. Maybe a better approach would be to fire two independent queries. Each for user collections and group collections. I think this would be the better approach, because you can much clearer see what the query does. But I would assume there comes a small performance penalty.


BlackDex:
@MFijak Thanks for the contribution. I have quickly scanned the PR and added some remarks. There may be more items i need to check. But this is a nice start for someone's first Rust code.

I'm not sure when i have some time to check the rest of the code, and how it works etc.. But if you have any questions please leave a comment here, or come and join on our Matrix channel.

MFijak:
I am glad to hear that! No worries take your time. 

Just a small question. You wrote that you added some remarks. Are these visible to me? I clicked through the tabs but somehow I can not see any remarks. I probably misunderstood, because you wrote them down locally on your computer?

BlackDex (changes requested):

BlackDex:
> I am glad to hear that! No worries take your time.
> 
> Just a small question. You wrote that you added some remarks. Are these visible to me? I clicked through the tabs but somehow I can not see any remarks. I probably misunderstood, because you wrote them down locally on your computer?

No, i wrote them on Github. But i everytime forget that i need to `Start a review` hehe.
They should be visible now.

src/db/models/cipher.rs
```diff
@@ -354,6 +356,19 @@ impl Cipher {
...
 
+    /// Returns whether this cipher is owned by an group in which the user has full access.
+    async fn is_in_full_access_group(
+        &self,
+        user_uuid: &str,
+        cipher_sync_data: Option<&CipherSyncData>,
+        conn: &DbConn,
+    ) -> bool {
+        match cipher_sync_data {
+            Some(cipher_sync_data) => cipher_sync_data.user_groups.iter().any(|group| group.access_all),
```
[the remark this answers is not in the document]
MFijak: I changed it to a HashMap. But I still use iter().any(). Is this OK? I would assume that, because the HashMap will take care of the efficiency?

GeekCornerGH:
Looks neat, the thing you can do however is a rebase I assume

migrations/postgresql/2022-07-27-110000_add_group_support/up.sql
```diff
@@ -0,0 +1,23 @@
...
+  access_all                        BOOLEAN NOT NULL,
+  external_id                       VARCHAR(300) NULL,
+  creation_date                     TIMESTAMP NOT NULL,
+  revision_date                     TIMESTAMP NOT NULL
+);
+
+CREATE TABLE groups_users (
+  groups_uuid                       CHAR(36) NOT NULL REFERENCES groups (uuid),
+  users_organizations_uuid          VARCHAR(36) NOT NULL REFERENCES users_organizations (uuid),  
+  PRIMARY KEY (group_uuid, users_organizations_uuid)
```
dani-garcia, line 14:
```suggestion
  PRIMARY KEY (groups_uuid, users_organizations_uuid)
```

src/api/core/organizations.rs
```diff
@@ -1484,3 +1532,324 @@ async fn import(org_id: String, data: JsonUpcase<OrgImportData>, headers: Header
...
+        "ContinuationToken": null,
+    })))
+}
+
+#[derive(Deserialize)]
+#[allow(non_snake_case)]
+struct GroupRequest {
+    Name: String,
+    AccessAll: Option<bool>,
+    ExternalId: String,
```
dani-garcia, line 1552: If you don't set an externalId when creating a group in the web vault (testing on v2022.9.0), this field isn't sent and so `post_groups` returns a 422 Unprocessable entity.

If you write something in the text field and then delete it, at least it sends an empty string "". But if you don't touch the field at all it's not included in the request.

We should probably make this field an optional field, or at least return a more descriptive error message.

src/api/core/organizations.rs
```diff
@@ -1484,3 +1532,324 @@ async fn import(org_id: String, data: JsonUpcase<OrgImportData>, headers: Header
...
+    Collections: Vec<SelectionReadOnly>,
+}
+
+impl GroupRequest {
+    pub fn to_group(&self, organizations_uuid: &str) -> Result<Group, String> {
+        let access_all_value = match self.AccessAll {
+            Some(value) => value,
+            _ => return Err(String::from("Could not convert GroupRequest to Group, because AccessAll has no value!")),
```
dani-garcia, line 1560: Prefer returning an `ApiResult<Group>` here:

```suggestion
            _ => err!("Could not convert GroupRequest to Group, because AccessAll has no value!"),
```

src/api/core/organizations.rs
```diff
@@ -1484,3 +1532,324 @@ async fn import(org_id: String, data: JsonUpcase<OrgImportData>, headers: Header
...
+
+#[post("/organizations/<org_id>/groups", data = "<data>")]
+async fn post_groups(
+    org_id: String,
+    _headers: AdminHeaders,
+    data: JsonUpcase<GroupRequest>,
+    conn: DbConn,
+) -> JsonResult {
+    let group_request = data.into_inner().data;
+    let group = match group_request.to_group(&org_id) {
```
dani-garcia, line 1633: Returning `ApiResult` in `to_group`, you could do this here:

```
let group = group_request.to_group(&org_id)?;
```

src/api/core/organizations.rs
```diff
@@ -1484,3 +1532,324 @@ async fn import(org_id: String, data: JsonUpcase<OrgImportData>, headers: Header
...
+    _headers: AdminHeaders,
+    conn: DbConn,
+) -> JsonResult {
+    let group = match Group::find_by_uuid(&group_id, &conn).await {
+        Some(group) => group,
+        None => err!("Group not found"),
+    };
+
+    let group_request = data.into_inner().data;
+    let updated_group = match group_request.update_group(group) {
```
dani-garcia, line 1655: Same thing as with `to_group`

dani-garcia (commented):
So far it looks pretty good to me! I've noticed a couple of small things, but for a rust newbie this is solid work.

I haven't had time to fully review the query logic, but in a few small tests it has worked correctly so far.

src/api/core/organizations.rs
MFijak: Whoops I did not notice that. I changed it to an optional field.

src/db/models/cipher.rs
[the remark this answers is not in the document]
BlackDex: @MFijak , I just checked this part by building it and adding some users to groups etc...
And i think it can be a bit more optimized. It should be a bit more like the attachments cache.
```rust
// String == organization_uuid
HashMap<String, Vec<Group>>
```

If you put the identifier to set it to the organization_uuid, then you can minimize the calls.
Also, i think this is needed for a security reason, since, if i have full access in one org, i will get that in all org's.
But, i haven't checked that thoroughly yet.

And within the `is_in_full_access_group`, you can do a get on that HashMap on the org_uuid, and then you do need to iter over it. with the current info you have.

The other extra option would be, which may be a better solution in this case, is to create a new function within `Group`, something like `Group::find_by_user_all_access`, which will only return `all_access` groups when they exists. That would save iterations which are not needed anyway, and also would save memory and query processing.

You still need to HashMap group them by Org UUID, since that is probably needed per Org to prevent all access on orgs someone shouldn't have.


Here something i didn't tested my self, but it could be a good solution on how to address this.
```rust
  if let Some(ref org_uuid) = self.organization_uuid {
      if let Some(cipher_sync_data) = cipher_sync_data {
          if let Some(cached_user_groups) = cipher_sync_data.user_groups.get(org_uuid) {
              return cached_user_groups.len() >= 1;
          }
      // Note: This needs to be checked by org_uuid also!!!
      } else {
          return Group::is_in_full_access_group(user_uuid, org_uuid, conn).await;
      }
  }
  false
```

src/db/models/cipher.rs
```diff
@@ -368,22 +385,39 @@ impl Cipher {
...
-            self.get_collections_access_flags(user_uuid, conn).await
+            let mut user_collections_access_flags = self.get_user_collections_access_flags(user_uuid, conn).await;
+            let mut group_collections_access_flags = self.get_group_collections_access_flags(user_uuid, conn).await;
+
+            let mut result = Vec::new();
+
+            result.append(&mut user_collections_access_flags);
+            result.append(&mut group_collections_access_flags);
+
+            result
```
BlackDex, line 420: This might be a bit nicer:

```rust
let mut access_flags = self.get_user_collections_access_flags(user_uuid, conn).await;
access_flags.append(&mut self.get_group_collections_access_flags(user_uuid, conn).await);
access_flags
```

BlackDex (changes requested):

BlackDex:
@MFijak i did some further checking, and indeed the `Group::is_in_full_access_group()` function is returning true for all ciphers even if they are not part the of the organization to which they belong. So that function needs a `org_uuid` filter also.

And using the check where the `user_groups` is `HashMap<String, Vec<Group>>` works great with a function `Group::find_by_user_all_access()`. That will save queries, and keeps the speed into parsing each ciphers access for the user.

src/db/models/cipher.rs
```diff
@@ -354,6 +356,21 @@ impl Cipher {
         false
     }
 
+    /// Returns whether this cipher is owned by an group in which the user has full access.
+    async fn is_in_full_access_group(
+        &self,
+        user_uuid: &str,
+        cipher_sync_data: Option<&CipherSyncData>,
+        conn: &DbConn,
+    ) -> bool {
```
BlackDex, line 365: This needs a `org_uuid` argument also, and needs to be filtered during the query.

src/db/models/cipher.rs
```diff
@@ -354,6 +356,21 @@ impl Cipher {
...
+        match cipher_sync_data {
+            Some(cipher_sync_data) => {
+                cipher_sync_data.user_groups.iter().any(|hash_map_entry| hash_map_entry.1.access_all)
+            }
+            None => Group::is_in_full_access_group(user_uuid, conn).await,
```
BlackDex, line 370: This should provide the `org_uuid` to only query the current org_uuid, else it will get any all_access from any organization.

So with the changed `CipherSyncData` it could look like this:
```rust
if let Some(ref org_uuid) = self.organization_uuid {
    if let Some(cipher_sync_data) = cipher_sync_data {
        if let Some(cached_user_groups) = cipher_sync_data.user_groups.get(org_uuid) {
            return cached_user_groups.len() >= 1;
        }
    } else {
        return Group::is_in_full_access_group(user_uuid, org_uuid, conn).await;
    }
}
```

This also prevents a lookup when the cipher is not part of an org at all.

BlackDex (changes requested):

BlackDex:
Also, in general, I would like this whole PR to be squashed into one commit instead of the 23 commits currently, that would also make the merge/commit nicer to check in the future 😄. But that is maybe something for right before we are going to merge it.

src/db/models/cipher.rs
MFijak: Oh yes much better :+1:

src/db/models/cipher.rs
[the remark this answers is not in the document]
MFijak: I changed it. But I omitted the group model entirely, since we just need the org_uuid. If you don't like it, I cant change it back to the group model.

src/db/models/cipher.rs
MFijak: I changed the function, however I use the internal field from `&self.` Do you want me to add it as an argument? I would then pass `self.organization_uuid` in function `get_access_restrictions` to this function

src/db/models/cipher.rs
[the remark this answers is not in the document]
BlackDex: This is a nice solution also :), maybe even better, since this is less loops and data.

src/db/models/cipher.rs
BlackDex: Ow, wait, I marked the wrong `is_full_access_group`, i mend the one in `Group::is_in_full_access_group`, not this one, but you changed that already accordingly, that is good now i think.

dani-garcia:
This is ready for merging for me, if you can rebase it against the latest main and squash it into 1-5 commits, that would be perfect, otherwise I'll see if I can do it myself.

We've had a release last week which means if we merge this soon it's gonna have some time for user testing before the next release.

Nice job 👍

dani-garcia (approved):

PR #2693, opened 2022-08-17
External Links | Optimize behavior

Fvbor:
- Changed one link to open in a new Tab
- Prevent tabnabbing on three links

### DIFF
--- a/src/static/templates/admin/diagnostics.hbs
+++ b/src/static/templates/admin/diagnostics.hbs
@@ -168,8 +168,8 @@
                 <dl class="row">
                     <dd class="col-sm-12">
                         If you need support please check the following links first before you create a new issue:
-                         <a href="https://vaultwarden.discourse.group/" target="_blank" rel="noreferrer">Vaultwarden Forum</a>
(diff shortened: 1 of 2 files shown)


GeekCornerGH (approved):

BlackDex:
While `_blank` targets by default have `noopener` already in most modern browsers i think adding it doesn't do any harm either.
Also see: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#browser_compatibility

BlackDex:
Also, nice catch for the backup link. That should have been a `_blank` also from the start :).

BlackDex (approved):

Fvbor:
> While `_blank` targets by default have `noopener` already in most modern browsers i think adding it doesn't do any harm either. Also see: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#browser_compatibility

Ah! Good to know. Didn`t saw this in my source.

> Also, nice catch for the backup link. That should have been a `_blank` also from the start :).

I clicked on the link and my tab was gone... :D

PR #2713, opened 2022-08-28
fix: Allow admin page working on subpaths

GeekCornerGH:
Hi everyone,
I made this patch to fix this bug on the admin page when using subpath:
image

Have a great day


BlackDex:
Thanks for the PR, but this is really not needed, since if you configure the `DOMAIN` variable correctly it will just work fine.

Your changes do not take into account multiple sub-paths for example, while it will work the way it already was.

I'm not going to approve this one.

GeekCornerGH:
> Thanks for the PR, but this is really not needed, since if you configure the `DOMAIN` variable correctly it will just work fine.
> 
> Your changes do not take into account multiple sub-paths for example, while it will work the way it already was.
> 
> I'm not going to approve this one.

I had DOMAIN env variable actually

williamdes (approved):
That should work

BlackDex:
> > Thanks for the PR, but this is really not needed, since if you configure the `DOMAIN` variable correctly it will just work fine.
> > Your changes do not take into account multiple sub-paths for example, while it will work the way it already was.
> > I'm not going to approve this one.
> 
> I had DOMAIN env variable actually

Then something else would be wrong, since it works just fine for me.
Also make sure there is no config.json, amd if there is, then adjust the domain there too.

That whole `{{urlpath}}` is the special part to support these sub-paths, so that would be strange to remove and replace.

BlackDex:
> That should work

It will only work with one sub-path level, not two like `/secret/vaultwarden`, then it will break again probably.
Also, it's just bad to always use `../` i think.

GeekCornerGH:
> > That should work
> 
> It will only work with one sub-path level, not two like `/secret/vaultwarden`, then it will break again probably. Also, it's just bad to always use `../` i think.

Ah yes didn't thought about that.

GeekCornerGH:
Sent a message on Matrix, maybe you will be able to help me

BlackDex:
Since it's fixed now, let's close this

Issue #2712, opened 2022-08-27
Admin API rejects invalid tokens with 303 instead of 401 status code


linkvt:
Hi,

first of all thanks for this nice project, I really appreciate it!

### Subject of the issue
I was setting up fail2ban for my Vaultwarden instance and noticed, that on invalid tokens the Vaultwarden Admin API responds with a 303 HTTP status code instead of a 401 I usually see.

### Deployment environment
I removed most of the fields as they are IMO irrelevant.

* vaultwarden version: 1.25.2  

### Steps to reproduce

Try to log into the admin UI with an incorrect token.

### Expected behaviour
HTTP response with [401 Unauthorized](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401) status code.

### Actual behaviour
HTTP response with 303 See Other redirect status code.

### Comment
Is this really intended from a behaviour perspective, see the implementation at 
https://github.com/dani-garcia/vaultwarden/blob/60ed5ff99d15dec0b82c85987f9a3e244b8bde91/src/api/admin.rs#L181 

If I understand the comment https://github.com/dani-garcia/vaultwarden/discussions/2448#discussioncomment-2666185 correctly it might be intended by you as the authors/contributors, but this is as far as I know not the correct behaviour for bad requests due to user issues (-> 4xx status code) due to failing authentication (narrowing to 401 Unauthorized).

While I understand the reasoning that a GET should follow the POST in this case, it does not make sense to me that a 303 is used.

Thanks for looking into this already!

Best, Vincent

P.S.: If you agree that this can be changed to a 401 I would be happy to open the PR myself.

BlackDex:
I would need to check this. While a 401 would normally be a valid answer to an invalid login. There should be anything wrong with this 303 either. There are other sites which just redirect you to a different page when failed to login instead of providing a 401.

What is the main reason for not keeping this a 303?

linkvt:
Thanks for the quick reply!

My main reason is that I cannot distinguish successful from invalid logins while looking e.g. at access logs as both appear like `"POST /admin HTTP/2.0" 303`

As I also have never ever seen a site returning the 303 status code until today (while not being a junior wrt web dev) I googled and checked what others said and tbh I don't think that 303 is used correctly in more than this place.
When doing a logout I also see a 303 even when the logout is already a GET request - according to the HTTP Spec this behaviour is to my understanding not correct.

Examples results that indicate the 303 being used incorrectly are
[...]

PR #2715, opened 2022-08-29
Use a temporary redirect

RealOrangeOne:
Fixes #2448

### DIFF
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -172,13 +172,13 @@ fn post_admin_login(
     let data = data.into_inner();
 
     if crate::ratelimit::check_limit_admin(&ip.ip).is_err() {
-        return Err(Flash::error(Redirect::to(admin_url(referer)), "Too many requests, try again later."));
+        return Err(Flash::error(Redirect::temporary(admin_url(referer)), "Too many requests, try again later."));
     }
 
     // If the token is invalid, redirect to login page
     if !_validate_token(&data.token) {
         error!("Invalid admin token. IP: {}", ip.ip);
-        Err(Flash::error(Redirect::to(admin_url(referer)), "Invalid admin token, please try again."))
+        Err(Flash::error(Redirect::temporary(admin_url(referer)), "Invalid admin token, please try again."))
     } else {
         // If the token received is valid, generate JWT and save it as a cookie
         let claims = generate_admin_claims();
@@ -192,7 +192,7 @@ fn post_admin_login(
             .finish();
 
         cookies.add(cookie);
-        Ok(Redirect::to(admin_url(referer)))
+        Ok(Redirect::temporary(admin_url(referer)))
     }
 }
 
@@ -303,7 +303,7 @@ async fn test_smtp(data: Json<InviteData>, _token: AdminToken) -> EmptyResult {
 #[get("/logout")]
 fn logout(cookies: &CookieJar<'_>, referer: Referer) -> Redirect {
     cookies.remove(Cookie::build(COOKIE_NAME, "").path(admin_path()).finish());
-    Redirect::to(admin_url(referer))
+    Redirect::temporary(admin_url(referer))
 }
 
 #[get("/users")]


williamdes (approved):

BlackDex:
This could cause issues, as i already explained in the post you linked too.
A 301 could still redirect using the same POST/PUT method instead of using the required GET we need to happen (This could cause a recursion, which most browser will tend to stop after x amount of time).
Also. there is an other issue reported here: #2712 which states that it actually should be a 401, which i think would be better.
And the `too many requests` should also be different, it should be a `429` for example.

So, as this PR is currently I would not tend to approve it.
If this can be changed to have the correct error codes like suggested in #2712, that would be great :).

src/api/admin.rs
```diff
@@ -172,13 +172,13 @@ fn post_admin_login(
     let data = data.into_inner();
 
     if crate::ratelimit::check_limit_admin(&ip.ip).is_err() {
-        return Err(Flash::error(Redirect::to(admin_url(referer)), "Too many requests, try again later."));
+        return Err(Flash::error(Redirect::temporary(admin_url(referer)), "Too many requests, try again later."));
```
BlackDex, line 175: This should actually be a `429`
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429

src/api/admin.rs
```diff
@@ -172,13 +172,13 @@ fn post_admin_login(
...
     }
 
     // If the token is invalid, redirect to login page
     if !_validate_token(&data.token) {
         error!("Invalid admin token. IP: {}", ip.ip);
-        Err(Flash::error(Redirect::to(admin_url(referer)), "Invalid admin token, please try again."))
+        Err(Flash::error(Redirect::temporary(admin_url(referer)), "Invalid admin token, please try again."))
```
BlackDex, line 181: This should be a `401`:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401

src/api/admin.rs
```diff
@@ -192,7 +192,7 @@ fn post_admin_login(
             .finish();
 
         cookies.add(cookie);
-        Ok(Redirect::to(admin_url(referer)))
+        Ok(Redirect::temporary(admin_url(referer)))
```
BlackDex, line 195: This should still be a `303`, since method function only reacts to a `POST` and we need it to turn into a `GET`.

It is also stated to only use `302` for `GET` and `HEAD`, see: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302

And also see: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/303

src/api/admin.rs
```diff
@@ -303,7 +303,7 @@ async fn test_smtp(data: Json<InviteData>, _token: AdminToken) -> EmptyResult {
 #[get("/logout")]
 fn logout(cookies: &CookieJar<'_>, referer: Referer) -> Redirect {
     cookies.remove(Cookie::build(COOKIE_NAME, "").path(admin_path()).finish());
-    Redirect::to(admin_url(referer))
+    Redirect::temporary(admin_url(referer))
```
BlackDex, line 306: This would be fine, since here we respond to a `GET` request, and a `302` here would be fine.

BlackDex (changes requested):

BlackDex:
There probably need to be some changes in how the response will be displayed to the user, because we currently use Flash Messages for this, which seem to only support redirects and no other status codes.

src/api/admin.rs
RealOrangeOne: I think these codes all still need to be redirects of some sorts. These error-specific status codes will only work when it's an API call. If this were changed to a 429, then the user would never see the flash message, as they'd never end up back on the admin page for it to show up.

BlackDex: That is why i mentioned there are more code changes needed for this to fully work.
Instead of a Flash Message, it needs to return the HTML it self.

Since the template already has it there with the `{{error}}` value. It shouldn't be that hard to fix though i think.

BlackDex:
@RealOrangeOne I have created a PR which should solve this, I added you as a Co-Author since you initiated this change.
I marked it to have this one closed when that PR has been excepted and merged.

PR #2729, opened 2022-09-06
Change the handling of login errors.

### LINKED ISSUE #2712: Admin API rejects invalid tokens with 303 instead of 401 status code

BlackDex:
Previously FlashMessage was used to provide an error message during login.
This PR changes that flow to not use redirect for this, but renders the HTML and responds using the correct status code where needed. This should solve some issues which were reported in the past.

Thanks to @RealOrangeOne, for initiating this with a PR.

Fixes #2448
Fixes #2712
Closes #2715

### DIFF
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -7,8 +7,8 @@ use rocket::serde::json::Json;
 use rocket::{
     form::Form,
     http::{Cookie, CookieJar, SameSite, Status},
-    request::{self, FlashMessage, FromRequest, Outcome, Request},
-    response::{content::RawHtml as Html, Flash, Redirect},
+    request::{self, FromRequest, Outcome, Request},
+    response::{content::RawHtml as Html, Redirect},
     Route,
 };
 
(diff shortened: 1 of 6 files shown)


RealOrangeOne (approved):
Looks good to me, much nicer!

Notice you've removed boostrap - is that intentional? CI isn't failing, so presume it just wasn't ever referenced?

BlackDex:
@RealOrangeOne i didn't removed it, It's just updated, but since the diff is very large Github doesn't show it by default

PR #2732, opened 2022-09-07
Added support for web-vault v2022.9

BlackDex:
- The new web-vault version supports fastmail.com anon email, add the correct api host to support it.
- Removed Firefox Relay, this seems only to be supported on SaaS.
- Added a function to the two-factor api to prevent 404 errors.

### DIFF
--- a/src/api/core/two_factor/mod.rs
+++ b/src/api/core/two_factor/mod.rs
@@ -19,7 +19,14 @@ pub mod webauthn;
 pub mod yubikey;
 
 pub fn routes() -> Vec<Route> {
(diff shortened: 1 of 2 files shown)


williamdes (approved):

RealOrangeOne (approved):

MyDrift-user:
Hey @BlackDex, sorry if that sounds like a stupid question, but I can't find a setting to configure this.
I looked around in my organisation settings, then in my user settings and also in the admin panel.
Is this Feature still available in the newer Versions of Vaultwarden? Since this PR is almost 3 years old the feature may have been removed.

Issue #2745, opened 2022-09-15
Minor - Typo in Admin panel within Diagnostics view


djbrownbear:
I noticed this minor typo while troubleshooting a certificate update for my instance of vaultwarden. 

suggestion:  `<span class="badge bg-danger d-none" id="time-warning" title="Time offsets are too much at drift.">Error</span> `
https://github.com/dani-garcia/vaultwarden/blob/9c891baad1838eb3446a90b2d6d71ebace9b347c/src/static/templates/admin/diagnostics.hbs#L144

suggestion: `<span class="badge bg-success d-none" id="https-success" title="Configured to use HTTPS">HTTPS</span>`
https://github.com/dani-garcia/vaultwarden/blob/9c891baad1838eb3446a90b2d6d71ebace9b347c/src/static/templates/admin/diagnostics.hbs#L154

djbrownbear:
First time contributing, created pull request for this issue instead.

PR #2746, opened 2022-09-15
fix: tooltip typo

djbrownbear:
I opened an issue earlier (#2745) regarding this typo and closed that issue now.

### DIFF
--- a/src/static/templates/admin/diagnostics.hbs
+++ b/src/static/templates/admin/diagnostics.hbs
@@ -140,8 +140,8 @@
                         <span><b>Server:</b> {{page_data.server_time_local}}</span>
                     </dd>
                     <dt class="col-sm-5">Date & Time (UTC)
-                        <span class="badge bg-success d-none" id="time-success" title="Time offsets seem to be correct.">Ok</span>
-                        <span class="badge bg-danger d-none" id="time-warning" title="Time offsets are too much at drift.">Error</span>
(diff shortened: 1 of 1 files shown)


williamdes (approved):

src/static/templates/admin/diagnostics.hbs
```diff
@@ -141,7 +141,7 @@
                     </dd>
                     <dt class="col-sm-5">Date & Time (UTC)
                         <span class="badge bg-success d-none" id="time-success" title="Time offsets seem to be correct.">Ok</span>
-                        <span class="badge bg-danger d-none" id="time-warning" title="Time offsets are too mouch at drift.">Error</span>
+                        <span class="badge bg-danger d-none" id="time-warning" title="Time offsets are too much at drift.">Error</span>
```
jjlin, line 144: "too much at drift" reads awkwardly in English, so you might as well fix the verbiage while you're at it.

I'd suggest something more informative/actionable like:
* `Server and browser times are more than 30 seconds apart` for the warning case
* `Server and browser times are within 30 seconds of each other` for the success case

jjlin (changes requested):

src/static/templates/admin/diagnostics.hbs
djbrownbear: Good point, thank you for the suggestion! I completed the commits to update the verbiage for the warning and success case.

PR #2747, opened 2022-09-16
Lock accounts after X login failures, configurable

Willena:
I saw this feature request in #246 

The implementation reuse the user enabling feature to lock the account. It simply automatically disable the user when the number of retries is reach. When locked, only an admin can re-enable the user account.

A new column has been added to the user schema to store the number of failed attempts.
It is configurable with the `login_max_retry` configuration key. The default value is 0 which disable automatic account locking

*Any remarks tips and comments are welcomed as I'm not yet familiar with rust and currently trying to learn it*

### DIFF
--- a/migrations/mysql/2022-09-15-002500_add_login_attempts/up.sql
+++ b/migrations/mysql/2022-09-15-002500_add_login_attempts/up.sql
@@ -1,0 +1,1 @@
+ALTER TABLE users ADD COLUMN invalid_login_count INTEGER NOT NULL DEFAULT 0;
--- a/src/api/identity.rs
+++ b/src/api/identity.rs
@@ -105,15 +105,36 @@ async fn _password_login(data: ConnectData, conn: DbConn, ip: &ClientIp) -> Json
         None => err!("Username or password is incorrect. Try again", format!("IP: {}. Username: {}.", ip.ip, username)),
     };
 
-    // Check password
-    let password = data.password.as_ref().unwrap();
-    if !user.check_valid_password(password) {
-        err!("Username or password is incorrect. Try again", format!("IP: {}. Username: {}.", ip.ip, username))
-    }
-
     // Check if the user is disabled
     if !user.enabled {
         err!("This user has been disabled", format!("IP: {}. Username: {}.", ip.ip, username))
+    }
+
+    // Check password
+    let password = data.password.as_ref().unwrap();
+    if !user.check_valid_password(password) {
+        // When the configuration limits the number of attempts
+        if CONFIG.login_max_retry() > 0 {
+            let mut user = user;
+            let invalid_count = user.invalid_login_count;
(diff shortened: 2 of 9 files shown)


migrations/mysql/2022-09-15-002500_add_login_attempts/up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE users ADD COLUMN invalid_login_count INTEGER NOT NULL DEFAULT 0;
```
williamdes, line 1: I would say it's an unsigned integer
Not sure how this project handles integers

src/api/identity.rs
```diff
@@ -184,6 +205,13 @@ async fn _password_login(data: ConnectData, conn: DbConn, ip: &ClientIp) -> Json
         result["TwoFactorToken"] = Value::String(token);
     }
 
+    // Reset the number of attempts on success logins
+    let mut user = user;
+    user.invalid_login_count = 0;
+    if let Err(e) = user.save(&conn).await {
+        error!("Error updating user: {:#?}", e);
+    }
```
BlackDex, line 213: It's probably best if this is only saved when the count isn't 0, else it would trigger an unnecessary save action.

migrations/mysql/2022-09-15-002500_add_login_attempts/up.sql
BlackDex: Not sure from the top of my head if that will work nicely with diesel<>multiple db backends.

BlackDex (changes requested):
While overall i think it looks ok. I'm not sure if i would approve this. I haven't searched for it yet, but I remember that I have declined something similar like this because this could be abused. If i know your mail address i can lock you out from your own account, which i think is a very bad option.

Also, there already is a rate limit for logins active by default which prevents brute force attempts.

I did add some remarks, but that's more for learning, because I tend more to not having this merged my self.

Willena:
> While overall i think it looks ok. I'm not sure if i would approve this. I haven't searched for it yet, but I remember that I have declined something similar like this because this could be abused. If i know your mail address i can lock you out from your own account, which i think is a very bad option.
> 
> Also, there already is a rate limit for logins active by default which prevents brute force attempts.
> 
> I did add some remarks, but that's more for learning, because I tend more to not having this merged my self.

I totally understand your point of view. Maybe the feature should be removed from the "[META] Feature request" issue. 
It was quite a simple feature and as I'm learning Rust I thought it was a good candidate. 

Just to add details, the implementation keeps the ratelimiter check before checking and update the number of tries. 

I'm not against closing the PR if you think it will never be integrated.

BlackDex:
I'm going to close this as I think this could be abused in a wrong way.
There already is a rate-limiter available for the login attempts which I think should be enough.

PR #2810, opened 2022-10-11
Add SELinux instructions for persistent data

deajan:
Container stops working after a couple of seconds when SELinux is enabled with error message `[vaultwarden][ERROR] Error creating keys, exiting...` which isn't explicit. These optional instructions for SElinux make sure we won't fail to run vaultwarden as first experience ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,8 +31,16 @@ Basically full implementation of Bitwarden API is provided including:
  * YubiKey and Duo support
 
 ## Installation
+Create a local directory where to map container persistent storage
+```sh
+mkdir /vw-data
+```
+Optional SELinux configuration
+```sh
+semanage fcontext -a -t svirt_sandbox_file_t '/vw-data(/.*)?'
+restorecon -Rv /vw-data
+```
 Pull the docker image and mount a volume from the host for persistent storage:
-
 ```sh
 docker pull vaultwarden/server:latest
 docker run -d --name vaultwarden -v /vw-data/:/data/ -p 80:80 vaultwarden/server:latest


williamdes (approved):

BlackDex:
I would rather have this somewhere in the wiki instead of the README.
Somewhere like: https://github.com/dani-garcia/vaultwarden/wiki/Starting-a-Container
Maybe?

deajan:
My initial idea was to have a nice "out of the box" experience, which won't happen on Fedora/RHEL/AlmaLinux/CentOS/Rocky because SELinux is enabled by default.
Of course the instructions could go into a subpage of a wiki, but the first instructions on the README would just fail for all those who use one of these OSes.

beckerj:
I also think the wiki would be a better place for it, the readme shouldn't contain information specific to just one of many distribution families (Red Hat & Derivatives).
It only makes sense to keep SELinux enabled if the user knows how to configure it properly anyways.
For all other distributions, it would just complicate the "out of the box" experience.

deajan:
SELinux is enabled by default for all those linux families so yes, IMO it makes sense to give 2 lines of intructions that ensure a nice out of the box experience.
Anyway, I'll redo the PR.

stefan0xC:
I tried to reproduce the issue with a freshly installed Fedora 36 Workstation but `docker run` from the README.md command just works. Could it be that the default SELinux policies on the Fedora 36 Workstation are not as strict as RHEL/CentOS? It's been a while since I have used an enterprise linux.  :confused:

stefan0xC:
Ah, seems like if I just `dnf install docker` (which installs the moby-engine) instead of following the [Docker installation guide](https://docs.docker.com/engine/install/fedora) I'll have this problem.

stefan0xC:
> Container stops working after a couple of seconds when SELinux is enabled with error message `[vaultwarden][ERROR] Error creating keys, exiting...` which isn't explicit. 

I've added a PR to make the error message more explicit. #2811

deajan:
@stefan0xC Thanks for making the error message explicit ;)

BlackDex:
I'm fine with adding this to the wiki, but not in the README.
The README is not the place to describe specific OS or Hosts specific configurations.

deajan:
Done in https://github.com/dani-garcia/vaultwarden/wiki/Starting-a-Container/a37a424aee0e995ea2eea0afb4a02afa73438e5b

PR #2813, opened 2022-10-12
Add `CreationDate` to cipher response JSON

jjlin:
Upstream PR: https://github.com/bitwarden/server/pull/2142

I don't think this field is used for anything currently, but probably better to add it now than be surprised later.


BlackDex (approved):

BlackDex:
They are adding it to the cipher edit/view screens.
It also looks like they check if it exists, if not then it's `null` and not used/shown.

PR #2834, opened 2022-10-17
Fix master password hint update not working.

### LINKED ISSUE #2833: password hint not changed
After the password hint field was moved from `My Account` (2022.06.2) to `Security > Master Password` (2022.10.0), there are 2 issues:

- it is not possible to update the password hint
- even when changing the password, the password hint is not saved

Personally I don't use password hints, but I could imagine this is important to others.

I am not sure, if this is just a web UI issue, a server component issue, or both.

BlackDex:
- The Master Password Hint input has changed it's location to the password update form. This PR updates the code to process this.

- Also changed the `ProfileData` struct to exclude `Culture` and `MasterPasswordHint`, since both are not used at all, and when not defined they will also not be allocated.

Fixes #2833

### DIFF
--- a/src/api/core/accounts.rs
+++ b/src/api/core/accounts.rs
@@ -193,9 +193,8 @@ async fn profile(headers: Headers, conn: DbConn) -> Json<Value> {
 #[derive(Deserialize, Debug)]
 #[allow(non_snake_case)]
 struct ProfileData {
-    #[serde(rename = "Culture")]
-    _Culture: String, // Ignored, always use en-US
-    MasterPasswordHint: Option<String>,
+    // Culture: String, // Ignored, always use en-US
+    // MasterPasswordHint: Option<String>, // Ignored, has been moved to ChangePassData
     Name: String,
 }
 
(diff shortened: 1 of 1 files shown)


tessus:
Just out of curosity, this means that people have to change their password to change the password hint, correct? If yes, it's a bit of a strange workflow. ;-)

williamdes (approved):

BlackDex:
Well, the flow is kinda ok i think.
You probably only set the password hint when you want to change the password, because if you change the password, the hint needs to be changed to.

Also,  you can still provide the old-password as the new-password and that would work just fine.

stefan0xC (approved):
Looks good to me. But you can set your password as the hint when changing this in this form (not while signing up). Is this something we can/should prevent or something the client is missing?

BlackDex:
> Looks good to me. But you can set your password as the hint when changing this in this form (not while signing up). Is this something we can/should prevent or something the client is missing?

If there is a check right now during registering and not during change, then you need to report this upstream. We can't check if the password and hint match on the server side.

stefan0xC:
Done: bitwarden/clients#3821

Issue #2835, opened 2022-10-17
Vaultwarden silently falls back to SQLite if DATABASE_URL is invalid


netthier:
### Subject of the issue

When an invalid `DATABASE_URL` is specified, e.g. with a typo such as `postgreql://user:pwd@host` or with quotes like `"postgresql://user:pwd@host"`, Vaultwarden silently falls back to using the SQLite database without logging any errors or warnings.
### Deployment environment
Used image is `vaultwarden/server:1.26.0`

### Steps to reproduce

I started vaultwarden via a docker-compose file with the `DATABASE_URL` variable set

### Expected behaviour

If an invalid `DATABASE_URL` is specified, Vaultwarden should either log an error or a warning along the lines of "Failed to parse DATABASE_URL, falling back to SQLite"
### Actual behaviour

Vaultwarden starts regularly and runs migrations against an SQLite database.

tessus:
I've noticed the same beavior and IMO it should not fall back to sqlite, but abend with an error message.

BlackDex:
Where did you configured this? Docker compose file? Custom export? Systemd file. And could you give an example? Without the actual password etc if course.

netthier:
> Where did you configured this? Docker compose file? Custom export? Systemd file. And could you give an example? Without the actual password etc if course.

As mentioned in the issue, I specified this in my docker-compose file. The issue also contains examples without actual passwords. 
I noticed this issue after accidentally specifying the URL like this:
```yaml
environment:
  - DATABASE_URL="postgresql://user:pwd@host/vaultwarden"
 ```
Specifying it as `"DATABASE_URL=postgresql://user:pwd@host/vaultwarden"` instead worked.
Before opening this issue I further investigated by intentionally introducing a typo in the working variant, and removing a letter from the protocol, ending up with `postgreql://etc`, which led to the same issue.

BlackDex:
Thanks for the extra info.
The thing is that the quotes are an issue with compose, but not with export or a .env loaded by Vaultwarden it self.

But, it indeed should not continue if there is a faulty setting.
The only setting which could be seen as faulty but still should fall back is an empty string, since that indicates a default.

Thanks for the report.

BlackDex:
Should be resolved via #2873 already.

PR #2872, opened 2022-10-27
error if DATABASE_URL can't be parsed:

vilhelmbergsoe:
DbConnType.from_url() returns error if it fails to parse DATABASE_URL

This addresses issue #2835


stefan0xC (changes requested):
I think you have to change the default as well as currently if the parsing will fail if `DATABASE_URL` is unset:
https://github.com/dani-garcia/vaultwarden/blob/f60a6929a9a289bd39c38eeb793ebf54c4d88199/src/config.rs#L331

vilhelmbergsoe:
> I think you have to change the default as well otherwise the parsing will fail if `DATABASE_URL` is unset:
> 
> https://github.com/dani-garcia/vaultwarden/blob/f60a6929a9a289bd39c38eeb793ebf54c4d88199/src/config.rs#L331

Would prefixing the format! with "sqlite:" work, do you think?

stefan0xC:
I've looked it up and `diesel` does not know about a `sqlite:` protocol instead the [documentation](https://docs.diesel.rs/2.0.x/diesel/sqlite/struct.SqliteConnection.html) lists 3 options:
1. File paths (`test.db`)
2. [URIs](https://sqlite.org/uri.html) (`file://test.db`)
3. Special identifiers (`:memory:`)

The code currently assumes the first option and will create a `file:data` directory if you prepend it with `file:`. There are also probably other side effects

vilhelmbergsoe:
> I've looked it up and `diesel` does not know about a `sqlite:` protocol instead the [documentation](https://docs.diesel.rs/2.0.x/diesel/sqlite/struct.SqliteConnection.html) lists 3 options:
> 
>     1. File paths (`test.db`)
> 
>     2. [URIs](https://sqlite.org/uri.html) (`file://test.db`)
> 
>     3. Special identifiers (`:memory:`)
> 
> 
> The code currently assumes the first option and will create a `file:data` directory if you prepend it with `file:`. There are also probably other side effects

It seems it would be hard to interpret whether user specified DATABASE_URL is referring to a SQLite database or if it's just an invalid path, so do you think going back to SQLite as the default but also warning the user that the URL couldn't be parsed correctly is the right way to go here?

stefan0xC:
Maybe it would be easier to just not create directories here but check if the parent directory exists:
https://github.com/dani-garcia/vaultwarden/blob/f60a6929a9a289bd39c38eeb793ebf54c4d88199/src/db/mod.rs#L431-L436

Then a misconfigured `DATABASE_URL` as in the typo example `postgreql://user:pwd@host` will not fall back to using sqlite. (The other problem where the docker-compose gets rid of the environment variable I am not sure we can do anything about.)

BlackDex:
I haven't looked very well at it, but this will break a lot of instances using sqlite.
If this would be merged, and people upgrade which have a custom database location, they all would get a message that there database is wrong. Probably not the best idea.

There probably is a better option to solve the issue mentioned.

vilhelmbergsoe:
Like mentioned here: [https://github.com/dani-garcia/vaultwarden/pull/2873#issuecomment-1294312021](https://github.com/dani-garcia/vaultwarden/pull/2873#issuecomment-1294312021), it would make sense that if someone's intention is to provide a URI like "postgres://..." or "mysql://...", they would have a hard time forgetting the ":".

BlackDex:
I still have some issues with this.
If someone is using this on Windows they could use `C:/Vaultwarden/db.sqlite3` for example as a location. That would trigger this code and report that it is faulty, which it isn't of course.

So, this is still not a very good solution unfortunately.

jjlin:
You could check for that specific pattern, as `:` is not otherwise a valid char for Windows filenames.

BlackDex:
> You could check for that specific pattern, as `:` is not otherwise a valid char for Windows filenames.

It then should be `[A-Z]:/` which should match that specific exclude case then.

jjlin:
> > You could check for that specific pattern, as `:` is not otherwise a valid char for Windows filenames.
> 
> It then should be `[A-Z]:/` which should match that specific exclude case then.

I think just checking for `[a-zA-Z]:` would be sufficient. AFAIK Windows accepts both forward and back slash as path separator nowadays.

BlackDex:
> 

I just tested it, and a back slash isn't working unless you single-quote the value.
So, it indeed could work with a back slash also.

vilhelmbergsoe:
Nice catch!
The nice way of checking for the pattern would be using regex, but wouldn't something like this work?:
`if url.contains(':') && !url.chars().next().unwrap().is_ascii_alphabetic() && !url.chars().next().unwrap().is_ascii_uppercase()`
If there is a nicer way of doing this, please do tell, haha.

jjlin:
Might be simpler just to check the second character for `:` if the OS is Windows (e.g., https://doc.rust-lang.org/std/env/consts/constant.OS.html)

BlackDex:
@vilhelmbergsoe I think this should do, but test it first.

```rust
if url.contains(':') && !(url.len() > 2 && url.chars().nth(1).unwrap() == ':' && url.chars().nth(0).unwrap().is_ascii_alphabetic()) {
```

vilhelmbergsoe:
Okay so I did a quick test you can find [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3f288f9b7258e74ab0b9fea0d8c6c191), and it seems as though it correctly parses the windows path as an SQLite path and a mistyped URL as invalid.

src/db/mod.rs
```diff
@@ -190,23 +190,33 @@ generate_connections! {
...
+                #[cfg(postgresql)]
+                return Ok(DbConnType::postgresql);
+
+                #[cfg(not(postgresql))]
+                err!("`DATABASE_URL` is a PostgreSQL URL, but the 'postgresql' feature is not enabled")
+
+            // Failed to parse
+            } else {
+                err!("`DATABASE_URL` looks like a URI but it could not be parsed, make sure it is valid")
+            }
```
williamdes, line 217: this could be a match statement

vilhelmbergsoe: I just updated the commit

williamdes (approved):

src/db/mod.rs
BlackDex: I'm not sure a match statement makes it more readable.
And it doesn't matter in the end that much in asm/compiled binary.

williamdes: I was thinking to match on what was found before `:` but after that realized how difficult it was in the end last time I tried to do that :/

BlackDex:
I'm not sure what other people think, but I would prefer to remove the `match` and just use a simple `if...else` to have it less complex and more readable. Also since it doesn't do much on the compiled end result.

vilhelmbergsoe:
I like if .. else more personally, but maybe I'll wait for some more people to share what they like the best.

src/db/mod.rs
```diff
@@ -190,23 +190,35 @@ generate_connections! {
...
+                url if url.starts_with("postgresql:") | url.starts_with("postgres:") => {
+                    #[cfg(postgresql)]
+                    return Ok(DbConnType::postgresql);
+
+                    #[cfg(not(postgresql))]
+                    err!("`DATABASE_URL` is a PostgreSQL URL, but the 'postgresql' feature is not enabled")
+                }
+                // Failed to parse
+                _ => {
+                    err!("`DATABASE_URL` looks like a URI but it could not be parsed, make sure it is valid")
```
stefan0xC, line 217: I think it would be better to be explicit here and say that Vaultwarden disallows `:` in filenames and if it's not meant as filename that we only allow valid DATABASE_URLs that start with `mysql://`, `postgres://` or `postgresql://` to make sure this is configured correctly.

stefan0xC (commented):
What if someone used the `DATABASE_URL=file:data/db.sqlite3` for SQLite? Or if someone named their sqlite dabase file `really:important:db.sqlite3`? This would be a _valid_ filename and not an invalid URI. So I'm not sure I am not quite yet on board with this change.

Could we not simply check if the `url.contains("://")`? I think both mysql as well as postgres require the URI to contain the authority section.
If the url does not it's probably safe to assume it was meant as (SQLite) filename and only then check if `url` starts with `mysql://` or `postgresql://` or `postgres://`? And only fail if it's neither? I mean given enough users this will surely break some setup for someone but I think it's better than just checking for a `:` as that can be part of a filename.

I mean, I think there's no way we can solve this problem without introducing some restrictions since as pointed out before almost everything can be a valid filename. I mean even my suggestion above with looking for `://` would be unsound if we keep allowing it as a filepath. (I.d. `DATABASE_URL=bla://db.sqlite3` could be a valid way to specify a `db.sqlite3` file in the folder `bla:`.)

src/db/mod.rs
```diff
@@ -190,23 +190,35 @@ generate_connections! {
...
+                // MySQL
+                url if url.starts_with("mysql:") => {
+                    #[cfg(mysql)]
+                    return Ok(DbConnType::mysql);
+
+                    #[cfg(not(mysql))]
+                    err!("`DATABASE_URL` is a MySQL URL, but the 'mysql' feature is not enabled")
+                }
+                // Postgres
+                url if url.starts_with("postgresql:") | url.starts_with("postgres:") => {
```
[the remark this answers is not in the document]
vilhelmbergsoe: Yeah, this was a silly mistake, my bad

src/db/mod.rs
```diff
@@ -190,23 +190,30 @@ generate_connections! {
...
+                // Postgres
+            } else if url.starts_with("postgresql:") || url.starts_with("postgres:") {
+                #[cfg(postgresql)]
+                return Ok(DbConnType::postgresql);
+
+                #[cfg(not(postgresql))]
+                err!("`DATABASE_URL` is a PostgreSQL URL, but the 'postgresql' feature is not enabled")
+                // Failed to parse
+            } else {
+                err!("`DATABASE_URL` looks like a URI but it could not be parsed, make sure it is valid. Note that ':' in filepaths are disallowed and DATABASE_URL only accepts 'postgresql://', 'postgres://' and 'mysql://' as valid URIs")
```
BlackDex, line 214: I Would remove the `Note that ':' in filepaths are disallowed and` part and leave the rest, and add something like `or a valid path to a file for SQLite.`

BlackDex (changes requested):

src/db/mod.rs
```diff
@@ -190,23 +190,35 @@ generate_connections! {
...
-        } else if url.starts_with("postgresql:") || url.starts_with("postgres:") {
-            #[cfg(postgresql)]
-            return Ok(DbConnType::postgresql);
-
-            #[cfg(not(postgresql))]
-            err!("`DATABASE_URL` is a PostgreSQL URL, but the 'postgresql' feature is not enabled")
+        if url.contains(':')
+            && !(url.len() >= 2
+                && url.chars().next().unwrap().is_ascii_alphabetic()
+                && url.chars().nth(1).unwrap() == ':')
```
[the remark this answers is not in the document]
stefan0xC: @BlackDex I've tested this and diesel allows the `file:` URI syntax that SQLite supports. The [documentation](https://docs.diesel.rs/2.0.x/diesel/sqlite/struct.SqliteConnection.html) is wrong by saying it can be `file://`. And like I said the issue is that `:` can be a valid filename, so this would be an issue.

src/db/mod.rs
vilhelmbergsoe: So something like "`DATABASE_URL` looks like a URI but it could not be parsed, make sure it is valid or a valid path for SQLite. DATABASE_URL only accepts 'postgresql://', 'postgres://' and 'mysql://' as valid URIs"

BlackDex:
I think this PR isn't really needed anymore because of #2873 which does kinda the same, but in a different way.
Ill close this PR, if you think this could still help, please re-base and re-open this PR :).

PR #2877, opened 2022-10-31
make support string more secure

tessus:

- length of smtp username

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -229,19 +229,6 @@ macro_rules! make_config {
                     inner.config.clone()
                 };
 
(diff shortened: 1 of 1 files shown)


tessus:
Oops, I missed something.

BlackDex:
Actually, that is on purpose. It masks stuff, but also helps a bit in determining what people configured.

tessus:
I opened a new one, before you commented. I still don't think that this is ok, because people paste this support string publicly. e.g. when I opened a bug report I manually changed all these entries.

PR #2878, opened 2022-10-31
make support string more secure

tessus:

- length of smtp username

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -229,17 +229,18 @@ macro_rules! make_config {
                     inner.config.clone()
                 };
 
-                /// We map over the string and remove all alphanumeric, _ and - characters.
-                /// This is the fastest way (within micro-seconds) instead of using a regex (which takes mili-seconds)
                 fn _privacy_mask(value: &str) -> String {
-                    value.chars().map(|c|
-                        match c {
-                            c if c.is_alphanumeric() => '*',
-                            '_' => '*',
-                            '-' => '*',
-                            _ => c
+                    let mut masked: String = String::new();
+                    let mut n = 0;
+                    for c in value.chars() {
+                        if n <= 12 && [':', '/'].contains(&c) {
+                            masked += &c.to_string();
+                        } else {
+                            masked += &'*'.to_string();
                         }
-                    ).collect::<String>()
+                        n += 1;
+                    }
+                    masked
                 }
 
                 serde_json::Value::Object({
--- a/src/config.rs
+++ b/src/config.rs
@@ -233,6 +233,9 @@ macro_rules! make_config {
                     let mut masked: String = String::new();
                     let mut n = 0;
                     for c in value.chars() {
+                        if c == ',' {
+                            masked += &c.to_string();
+                        }
                         if n <= 12 && [':', '/'].contains(&c) {
                             masked += &c.to_string();
                         } else {


BlackDex:
Tis actually is done on porpoise. I understand that sometimes lengths could reveal something, but it also helps us determine misconfigurations on some parts.

People are still encouraged to double check there data they post.
I think it is a good tradeoff, and if people want they can obfuscate it more them selfs.

Passwords already are just `***` so that isn't guessable.
Maybe something could be done for the password within the db url though, but otherwise i think it's fine as is though.

tessus:
Hmm, vaultwarden is a security product, yet the support string is without modification not properly obfuscated.

By default, a publicly accessible support string should be more privacy oriented.

I understand your argument as well, but this should rather be the exception and not the default. If I have to manually change the support string to not reveal unwanted information, the mask algorithm has a problem.

tessus:
When it comes to misconfiguration, it is better to parse the config file during startup, which is already done afaik.

BlackDex:
> When it comes to misconfiguration, it is better to parse the config file during startup, which is already done afaik.

That isn't always possible though. Someone can still type `http://` instead of `https://`, have some other strange configs which we can't actually check if that is valid or not, but could have an idea it could be wrong.

Maybe we could leave some useful info like `http://` or `https://` visible in the `DOMAIN` string for example. And do the some for other stuff maybe, like `mysql://` and `postgres://` or `postgresql://` for the `DATABASE_URL`.

But for other items like `org_creation_users` or `allowed_iframe_ancestors`, those i would really like to be more verbose for example. Those are items where why sometimes things go wrong, which are hard to troubleshoot. Though maybe that is still up-to the user to verify, but i have encountered items where they were wrong and it was visible even though masked, which otherwise was not possible for me to see (No i do not have example, i can't find them that quickly).

So, I'm not against make it a bit more secure, but it still needs to be useful for us to troubleshoot issues. And if other people like just double check what they post and remove what don't want in there, that would be even better without making changes here.

tessus:
I am not quite sure how to respond to that. If the DB URL string is wrong, it will tell you when the service starts, or at least it's in the works #2872

If somebody uses http instead of https, despite the fact that http means unencrypted traffic and your passwords are transferred in clear text, I am speechless. Maybe in that case, it's better to hardcode https and let people only set the domain instead of the scheme as well.
Even for testing purposes http might not work, since certain browsers do not allow secure API calls via an unencrypted channel. But hardcoding https is maybe too extreme.

But I can change this PR to include the following info:

- DB scheme: `scheme://***` (scheme in clear text, since masking it is the same as using clear text)
- domain scheme: `scheme://***` (scheme in clear text)
- use your original masking algorithm for `allowed_iframe_ancestors` and `org_creation_users`

Is that a better approach?

BlackDex:
> 
> If somebody uses http instead of https, despite the fact that http means unencrypted traffic and your passwords are transferred in clear text, I am speechless. Maybe in that case, it's better to hardcode https and let people only set the domain instead of the scheme as well. Even for testing purposes http might not work, since certain browsers do not allow secure API calls via an unencrypted channel. But hardcoding https is maybe too extreme.
> 
Just a quick response on this part. Passwords are never ever transferred in clear text. They are always encrypted. Also, the master password is never sent to the server in clear text, or even recoverable for the server side.

http works just fine locally, good during testing.

tessus:
> Passwords are never ever transferred in clear text. 

So the client en/decrypts the passwords. Nice, good to know.

tessus:
Come to think of it, you never answered the following:

I can change this PR to include the following info:

- DB scheme: `scheme://***` (scheme in clear text, since masking it is the same as using clear text)
- domain scheme: `scheme://***` (scheme in clear text)
- use your original masking algorithm for `allowed_iframe_ancestors` and `org_creation_users`

Shall I do that?

Disclaimer: I have never done anything in Rust, so I will have to read up on a few things, but this is also not something that is super urgent, thus I don't have to code it in the next few minutes.

BlackDex:
Ill have to look at this a bit better. What would be a better choice, and if there are other ways to have the same info.

tessus:
Yes, this sounds like a good idea.

> We could even add an index into the loop, and only not convert the : and / when the index position is lower then 12 (starting from 0) which is the length of postgresql://, or some other way to detect this position maybe.

It's too bad that the `_privacy_mask` function does not include the PRIVACY_CONFIG string as an argument. That would make it a lot easier to mask certain strings differently.

But such an index would help, especially with the rest of the DB URL and the DOMAIN.

BlackDex:
The DB is not a problem, that is already there, so in the end ot doesn't really matter. For http, and the others format could be an issue, but also checked.

Adding options per item is a bit to much i think, that will slow down the filtering, though not that much, but i think that is overkill.

tessus:
@BlackDex please let me know what you think.

src/config.rs
```diff
@@ -229,17 +229,21 @@ macro_rules! make_config {
...
                         }
-                    ).collect::<String>()
+                        if n <= 12 && [':', '/'].contains(&c) {
+                            masked += &c.to_string();
+                        } else {
+                            masked += &'*'.to_string();
+                        }
+                        n += 1;
+                    }
+                    masked
```
BlackDex, line 246: That seems verify inefficient.
I suggest the following change, which is much quicker.

```suggestion
    let mut n = 0;
    value.chars().map(|c| {
        n += 1;
        match c {
            ':' if n <= 11  => c,
            '/' if n <= 13  => c,
            ',' => c,
            _ => '*'
        }
    }
    ).collect::<String>()
```

tessus: Ah, this is great. Thank you. Yep, I am new to Rust so the Rust way is still - well new.

However, I do have a few comments:

- I never had a `collect` statement in my code, yet in your suggestion it seems that way. how is this possible?
- I was wondering how I could use `map` with an index. now I know. Thanks!
- while your code looks a lot cleaner (and I love it), from a performance perspective the code is not so different. Both use a loop and both use if statements. It is true that I am using one more if, however this and the rest (var assignments) is something the compiler should be able to optimize.

Btw, my code also had a bug, so all of the code was contributed by you and I should not be credited for anything in this PR. I suggest to close this PR and you create a new PR for your code.

tessus: Thanks for the `godbolt.org` links. This is a great tool. Yep, the memory allocations have more impact than I thought.

> But, you did trigger thinking about it 😄 .

Unfortunately, an idea alone does not make one rich.

tessus:
Thanks @BlackDex for this learning experience.

PR #2888, opened 2022-11-04
Update Rust version, deps and workflow

### LINKED ISSUE #2847: Deprecated usages on Github Actions
I've noticed some deprecation messages on our Github Actions, both here and in `bw_web_builds`, caused by two things:

The use of the `save-state` and `set-output` commands:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Some actions using Node 12 which is EOL:
https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

We have around six moths until they become errors, but I'm opening this issue to track them, as the build logs aren't very visible.

BlackDex:
- Update Rust to v1.65.0
- Update dependencies
- Updated workflow files - Fixes #2847
- Added some extra clippy checks
- Fixed some clippy checks


williamdes (approved):

docker/Dockerfile.j2
```diff
@@ -251,4 +251,4 @@ COPY docker/start.sh /start.sh
 
 HEALTHCHECK --interval=60s --timeout=10s CMD ["/healthcheck.sh"]
 
-CMD ["/start.sh"]
+CMD ["sh", "/start.sh"]
```
jjlin, line 254: It makes no sense why it wouldn't work as is, but if you have to put this here, specifying `/bin/sh` would be better.

src/main.rs
```diff
@@ -12,9 +12,13 @@
     clippy::equatable_if_let,
     clippy::float_cmp_const,
     clippy::inefficient_to_string,
+    clippy::iter_on_empty_collections,
+    clippy::iter_on_single_items,
     clippy::linkedlist,
     clippy::macro_use_imports,
     clippy::manual_assert,
+    clippy::manual_instant_elapsed,
+    clippy::manual_string_new,
```
jjlin, line 21: I'm not in favor of this. Requiring `String::new()` for consistency is well-intentioned I suppose, but there is something to be said for the clarity of seeing `""` when there's an empty string. This or similar syntax is commonly used in most popular languages, so it's easy to read and understand for everyone, whereas constructor forms are relatively uncommon and will throw off many people I think. I would prefer standardizing on some form that contains `""`; `"".to_string()` might be clearest IMO, but I don't feel too strongly about the exact form.

BlackDex: I checked the source behind this, and using `"".to_string()` also uses different ways including trying to check for utf8 strings and unsafe functions. Using `String::new()` doesn't. So i think it this is a better way. Also, there are some different locations which were already using it, and then i vote for consistency in this case.

docker/Dockerfile.j2
BlackDex: I agree that it makes no sense. But the theory doesn't match the reality.

Underwater the `entry.sh` script uses `command -v` so `sh` would work fine. I'm not sure if that also works fine using `/bin/sh` in these same situations.

src/main.rs
jjlin: If you're saying that `String::new()` is the fastest, I wouldn't argue with that, though there's no particular reason that an optimizing compiler couldn't generate identical code. But I think clarity is more important than a negligible speed difference.

docker/Dockerfile.j2
jjlin: It's faster to use `/bin/sh`, and not all images use the `entry.sh` script. Personally, I wouldn't prioritize adding weird workarounds for systems that don't seem to work properly.

BlackDex: I'm not sure where you get it from that `/bin/sh` would be faster, but please enlighten me.
I understand that not all images (just two in our case) do not use `entry.sh`, but they still are able to work with just `sh`.

Even though, i have removed it from this PR, since this discussion, and thinking about it, it probably doesn't belong in this specific update PR, but rather in a separate issue solving PR.

src/main.rs
BlackDex: It's not that it's faster or what ever. Doesn't matter that the end-result is the same. I do like the consistency it brings. I understand that the eye-candy of `"".to_string()` when there are a few other `str`'s which need to be converted also have a better appeal. But i find that a bad argument in this case. We had `String::from("")`, `"".to_string()` and `String::new()` in several places, which maybe all result in the exact same ASM, but not during compile-time, and, i my self would then rather choose consistency.

jjlin: As I mentioned, "I don't feel too strongly about the exact form", so `String::from("")` is okay with me, though I'm not clear on why you think that makes it more consistent.

BlackDex: I mean, replace all `"something".to_string()` and `my_var.to_string()` to `String::from(my_var)` and `String::from("something")`, which matches more the `String::new()` and is visually the same, since that is what you mentioned about having it the same.

jjlin: What I had in mind was standardizing (for example) on `String::from("")`, `String::from("foo")`, and `String::from(foo_var)`, rather than having a mix of that and `"foo".into()`, `"foo".to_string()`, etc.

My main point was I dislike being forced to use `String::new()` for empty strings because it doesn't visually contain `""`, but whatever.

BlackDex: I just checked with godbolt.org, and `String::new()` is definitely less asm code, no utf8 checks etc.. 
So, I am still in favor of using `String::new()`.

And, maybe its a nice chore to change all other String variants to `String::from(T)` for consistency 😄

PR #2907, opened 2022-11-10
Impl the "org API key" and "Public API" needed to use "Bitwarden Directory Connector"

Kurnihil:
Hello, 
I've started with the org-api-key branch of @BlackDex  [https://github.com/BlackDex/vaultwarden/tree/org-api-key](url)
And added what was needed to use the "Bitwarden Directory Connector".
That means only the /public/organization/import route

I'm willing to continue with the implementation of the public API,
but before doing that I would like that somebody else looks at the code.

I've tested it locally but for sure some more testing is needed.

### DIFF
--- a/src/api/core/public.rs
+++ b/src/api/core/public.rs
@@ -1,0 +1,224 @@
+use rocket::{
+    Route,
+    Request,
+    request::{self, FromRequest, Outcome},
+};
+use chrono::Utc;
+
+use crate::{
+    auth,
+    db::{models::*, DbConn},
+    api::{JsonUpcase, EmptyResult},
+    mail,
+    CONFIG
+};
+
+pub fn routes() -> Vec<Route> {
+    routes![
+        ldap_import,
+    ]
+}
+
+
+#[derive(Deserialize, Debug)]
+#[allow(non_snake_case)]
+struct OrgImportGroupData {
+    Name: String,       // "GroupName"
+    ExternalId: String, // "cn=GroupName,ou=Groups,dc=example,dc=com"
+    MemberExternalIds: Vec<String>, // ["uid=user,ou=People,dc=example,dc=com"]
+}
+
+#[derive(Deserialize, Debug)]
+#[allow(non_snake_case)]
+struct OrgImportUserData {
+    Email: String, // "user@maildomain.net"
+    ExternalId: String, // "uid=user,ou=People,dc=example,dc=com"
+    Deleted: bool,
+}
+
+#[derive(Deserialize, Debug)]
+#[allow(non_snake_case)]
+struct OrgImportData {
+    Groups: Vec<OrgImportGroupData>,
+    Members: Vec<OrgImportUserData>,
+    OverwriteExisting: bool,
+    #[allow(dead_code)]
+    LargeImport: bool,
+}
+
+#[post("/public/organization/import", data = "<data>")]
+async fn ldap_import(data: JsonUpcase<OrgImportData>, token: PublicToken, mut conn: DbConn) -> EmptyResult {
+    let _ = &conn;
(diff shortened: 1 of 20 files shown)


williamdes (approved):

tessus:
BlackDex marked his branch and the commit as WIP. What you usually do, if you want to add or change code, is create a PR in BlackDex's repository.

BlackDex:
@tessus is missing some context, since i had contact with @Kurnihil regarding this, and i only created a starting point to help them further.

I haven't had any time yet to take a good look at this. And it does need to at least get a rebase especially since the last merge spree.

tessus:
Ah, I see. In this case I apologize for my comment.

hudsonventura:
Will this pull request be accepted? I hope so =D

BlackDex:
> Will this pull request be accepted? I hope so =D

Not in the current state, as it needs to be rebased, and we still need to test it of course.

Kurnihil:
I will try to rebase in the next few days, work permitting.
Sorry to have dropped it since now

BlackDex:
I did a quick test today, and using the latest available version of the Directory-Connector and a local LDAP deployment seems to work great as far as i can tell, it will create the users, add them as invited to the org.

So, apart from rebasing this PR, i think it at least works. Ill have to do some code-checking maybe later on.

src/db/models/user.rs
```diff
@@ -137,6 +139,21 @@ impl User {
...
+        match external_id {
+            Some(external_id) => {
+                if external_id.is_empty() {
+                    self.external_id = None;
+                } else {
+                    self.external_id = Some(external_id)
+                }
+            }
+            None => self.external_id = None,
+        }
```
BlackDex, line 154: This could be a bit less complex i think. And it will also look a bit more nice.

```suggestion
    let mut ext_id: Option<String> = None;
    if let Some(external_id) = external_id {
        if !external_id.is_empty() {
           ext_id = Some(external_id);
        }
    }
    self.external_id = ext_id;
```

src/db/models/collection.rs
```diff
@@ -276,7 +276,27 @@ impl Collection {
...
+                                users_organizations::uuid.eq(groups_users::users_organizations_uuid)
+                            ))
+                            .filter(users_organizations::user_uuid.eq(user_uuid))
+                            .select(groups_users::groups_uuid)
+                        ))
+                        .count()
+                        .first::<i64>(conn)
+                        .ok()
+                        .unwrap_or(0) != 0
+                }};
```
BlackDex, line 297: Is this something similar like #3108?

src/db/models/collection.rs
```diff
@@ -298,7 +318,27 @@ impl Collection {
...
+
+                direct || indirect
```
BlackDex, line 341: Is this something similar like #3108?

src/api/core/public.rs
```diff
@@ -0,0 +1,231 @@
...
+}
+
+#[derive(Deserialize, Debug)]
+#[allow(non_snake_case)]
+struct OrgImportData {
+    Groups: Vec<OrgImportGroupData>,
+    Members: Vec<OrgImportUserData>,
+    OverwriteExisting: bool,
+    #[allow(dead_code)]
+    LargeImport: bool,
```
BlackDex, line 41: Is there a reason this is not used? Of so, maybe add that as a comment?
If I'm correct, Bitwarden uses this for 2000+ users in a single run.

BlackDex (changes requested):

mkrauser:
I've tested this successfully with Google Workspace

Thanks you for this, I hope it gets merged soon!

BlackDex:
I have an updated version of this PR in #3568 .

PR #2910, opened 2022-11-11
Use constant size generic parameter for random bytes generation

samueltardieu:
All uses of `get_random()` were in the form of:

  `&get_random(vec![0u8; SIZE])`

with `SIZE` being a constant.

Building a `Vec` is unnecessary for two reasons. First, it uses a very short-lived dynamic memory allocation. Second, a `Vec` is a resizable object, which is useless in those context when random data have a fixed size and will only be read.

`get_random_bytes()` takes a constant as a generic parameter and returns an array with the requested number of random bytes.

Stack safety analysis: the random bytes will be allocated on the caller stack for a very short time (until the encoding function has been called on the data). In some cases, the random bytes take less room than the `Vec` did (a `Vec` is 24 bytes on a 64 bit computer). The maximum used size is 180 bytes, which makes it for 0.008% of the default stack size for a Rust thread (2MiB), so this is a non-issue.

Also, most of the uses of those random bytes are to encode them using an `Encoding`. The function `crypto::encode_random_bytes()` generates random bytes and encode them with the provided `Encoding`, leading to code deduplication.

`generate_id()` has also been converted to use a constant generic parameter as well since the length of the requested String is always a constant.

### DIFF
--- a/src/crypto.rs
+++ b/src/crypto.rs
@@ -37,10 +37,6 @@ pub fn hmac_sign(key: &str, data: &str) -> String {
 // Random values
 //
 
-pub fn get_random_64() -> Vec<u8> {
-    get_random_bytes::<64>().to_vec()
-}
-
 /// Return an array holding `N` random bytes.
 pub fn get_random_bytes<const N: usize>() -> [u8; N] {
     use ring::rand::{SecureRandom, SystemRandom};
--- a/src/crypto.rs
+++ b/src/crypto.rs
@@ -3,7 +3,7 @@
 //
 use std::num::NonZeroU32;
 
-use data_encoding::HEXLOWER;
+use data_encoding::{Encoding, HEXLOWER};
 use ring::{digest, hmac, pbkdf2};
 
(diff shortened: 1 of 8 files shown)


BlackDex (approved):

src/crypto.rs
```diff
@@ -38,17 +38,24 @@ pub fn hmac_sign(key: &str, data: &str) -> String {
 //
 
 pub fn get_random_64() -> Vec<u8> {
-    get_random(vec![0u8; 64])
+    get_random_bytes::<64>().to_vec()
```
jjlin, line 41: It's a bit inconsistent to have `get_random_bytes()` return an array, but `get_random_64()` return a vector. There are only a few callers to this function, and it looks like it would be straightforward to directly use `get_random_bytes()` or `encode_random_bytes()` in all of those cases, so I'd suggest getting rid of this function entirely.

jjlin (changes requested):

src/crypto.rs
samueltardieu: Makes sense. I've kept the `to_vec()` in two places in order not to change the database interface.

PR #2911, opened 2022-11-12
Update config comment to reflect rfc8314.

skid9000:
TL;DR : Port 465 with implicit TLS for SMTP isn't outdated since RFC 8314

As stated in [RFC 8314](https://tools.ietf.org/html/rfc8314) : 
image

Also as stated [here](https://github.com/iredmail/iRedMail/issues/22#issuecomment-731779992) 

The IANA registry has been modified to register port 465 for native tls smtp connections : 

image

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -352,7 +352,7 @@
 # SMTP_FROM=vaultwarden@domain.tld
 # SMTP_FROM_NAME=Vaultwarden
 # SMTP_SECURITY=starttls # ("starttls", "force_tls", "off") Enable a secure connection. Default is "starttls" (Explicit - ports 587 or 25), "force_tls" (Implicit - port 465) or "off", no encryption (port 25)
-# SMTP_PORT=587          # Ports 587 (submission) and 25 (smtp) are standard without encryption and with encryption via STARTTLS (Explicit TLS). Port 465 is outdated and used with Implicit TLS.
+# SMTP_PORT=587          # Ports 587 (submission), 465 (submissions) and 25 (smtp) are standard without encryption and with encryption ports, 587 use STARTTLS (Explicit encryption) and 465 use native TLS (Implicit encryption)
 # SMTP_USERNAME=username
 # SMTP_PASSWORD=password
 # SMTP_TIMEOUT=15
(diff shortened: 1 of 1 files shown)


.env.template
```diff
@@ -352,7 +352,7 @@
 # SMTP_FROM=vaultwarden@domain.tld
 # SMTP_FROM_NAME=Vaultwarden
 # SMTP_SECURITY=starttls # ("starttls", "force_tls", "off") Enable a secure connection. Default is "starttls" (Explicit - ports 587 or 25), "force_tls" (Implicit - port 465) or "off", no encryption (port 25)
-# SMTP_PORT=587          # Ports 587 (submission) and 25 (smtp) are standard without encryption and with encryption via STARTTLS (Explicit TLS). Port 465 is outdated and used with Implicit TLS.
+# SMTP_PORT=587          # Ports 587 (submission), 465 (submissions) and 25 (smtp) are standard without encryption and with encryption ports, 587 use STARTTLS (Explicit encryption) and 465 use native TLS (Implicit encryption)
```
tessus, line 355: No, this change is wrong.

The sentence `Ports 587 (submission), 465 (submissions) and 25 (smtp) are standard without encryption` is wrong.

465 is NEVER without encryption. The TLS handshake happens upon connection establishment. There is never unencrypted data flow.

This is correct for 587 and 25, because traffic is unencrypted until the STARTTLS command is sent.

```suggestion
# SMTP_PORT=587          # Ports 587 (submission) and 25 (smtp) are standard without encryption and with encryption via STARTTLS (Explicit TLS). Port 465 (submissions) is used for encrypted submission (Implicit TLS).
```

skid9000: Ah right, i typed that a little too fast.

Thanks for noticing it !

tessus: You're welcome.

BlackDex (changes requested):
The change looks ok. But this also needs to be change inside the `config.rs` file.
https://github.com/dani-garcia/vaultwarden/blob/7a7673103fb180098f18abe77b75ba085710b559/src/config.rs#L591

BlackDex (approved):

PR #2919, opened 2022-11-14
Fully remove DuckDuckGo email service.

### LINKED ISSUE #2828: When using the duckduckgo service, there is a cors error in the request
### Subject of the issue

When using the duckduckgo service, there is a cors error in the request.
截屏2022-10-15 20 06 12

### Deployment environment

* vaultwarden version: 1.26.0

* Clients used: web vault in chrome 

* Reverse proxy and version:  Nginx 1.22.0 

* MySQL/MariaDB or PostgreSQL version: sqlite

* Other relevant details:

### Actual behaviour

cors error in request

BlackDex:
The DuckDuckGo email service is not supported for self-hosted servers. This option is already hidden via the latest web-vault.

This PR also removes some server side headers.

Fixes #2828

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -83,7 +83,6 @@ impl Fairing for AppHeaders {
                   https://app.simplelogin.io/api/ \
                   https://app.anonaddy.com/api/ \
                   https://api.fastmail.com/ \
-                  https://quack.duckduckgo.com/api/email/ \
                   ;\
                 ",
                 icon_service_csp = CONFIG._icon_service_csp(),


GeekCornerGH (approved):

t3dium:
this'll still work fine in the local clients right? since i haven't experienced any issues using duckduckgo's email forwarding within the clients, so im hesitant in updating vaultwarden

BlackDex:
It's all client side. But not supported by the web-vault though.
Nothing server side.

ItsDrike:
Just curious here, why was duckduckgo removed from the web vault here? The linked issue this PR is supposed to be fixing seems to have been deleted, and I've only noticed this message in the release logs, would you mind explaining why this was done?

stefan0xC:
Seems like the user who reported the issue initially has deleted their GitHub profile taking the reported issue with them.

On self-hosted servers it does not work because it is [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) protected. @BlackDex reported the issue upstream https://github.com/bitwarden/clients/issues/3798

PR #2928, opened 2022-11-21
Update settings description

karbobc:
Update description from `login` to `admin` in admin page.

`Read-Only Config`
image

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -543,9 +543,9 @@ make_config! {
         /// Max burst size for login requests |> Allow a burst of requests of up to this size, while maintaining the average indicated by `login_ratelimit_seconds`. Note that this applies to both the login and the 2FA, so it's recommended to allow a burst size of at least 2
         login_ratelimit_max_burst:     u32, false, def, 10;
 
-        /// Seconds between admin requests |> Number of seconds, on average, between admin requests from the same IP address before rate limiting kicks in
+        /// Seconds between admin login requests |> Number of seconds, on average, between admin requests from the same IP address before rate limiting kicks in
         admin_ratelimit_seconds:       u64, false, def, 300;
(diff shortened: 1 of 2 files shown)


src/config.rs
```diff
@@ -545,7 +545,7 @@ make_config! {
 
         /// Seconds between admin requests |> Number of seconds, on average, between admin requests from the same IP address before rate limiting kicks in
         admin_ratelimit_seconds:       u64, false, def, 300;
-        /// Max burst size for login requests |> Allow a burst of requests of up to this size, while maintaining the average indicated by `admin_ratelimit_seconds`
+        /// Max burst size for admin requests |> Allow a burst of requests of up to this size, while maintaining the average indicated by `admin_ratelimit_seconds`
```
BlackDex, line 548: Then, it actually should be `admin login` instead. Because now it implies on all admin requests, which is not the case.

BlackDex (changes requested):

src/config.rs
BlackDex: the latter

BlackDex: Also, this probably needs to be changed in the template: https://github.com/dani-garcia/vaultwarden/blob/7a7673103fb180098f18abe77b75ba085710b559/.env.template#L305

karbobc: I'll update it at the same time.

karbobc: > Also, this probably needs to be changed in the template:
> 
> https://github.com/dani-garcia/vaultwarden/blob/7a7673103fb180098f18abe77b75ba085710b559/.env.template#L305

This doesn't seem to need an update.

BlackDex: Why not? It says `admin`, and not `admin login`. Would be best to keep both items consistent not?

karbobc: You're right, I will update it later.

PR #2940, opened 2022-12-01
Update Vaultwarden Logo's

BlackDex:
Updated the logo's so the `V` is better visible.
Also the cog it self is better now, the previous version wasn't fully round. These versions also are used with the PR to update the web-vault and use these logo's.

Also updated the images in the static folder.


stefan0xC:
@dani-garcia can you maybe also update the Logo of [Vaultwarden Discourse](https://vaultwarden.discourse.group/)? 


vaultwarden-logo

And I've read somewhere that you can also use a different logo for dark mode which is currently almost invisible.
vaultwarden-discourse-darkmode

So it would be nice if you could change it to the white variant:
vaultwarden-logo-white

dani-garcia:
@stefan0xC Should be fixed now!

stefan0xC:
@dani-garcia Thanks. I think the smaller white logo is missing the holes:
discourse-white-logo

Could be an artifact of svg to png conversion, so maybe if you upload it as png directly it will work better?
vaultwarden-icon-white
vaultwarden-icon

stefan0xC:
For [#vaultwarden:matrix.org](https://matrix.to/#/#vaultwarden:matrix.org) there would also be a white background and a border:
vaultwarden-icon-with-bg-and-border


I've created it with 
```
convert -density 1200 -resize 412x412 vaultwarden-icon{.svg,-with-bg.png}
convert -bordercolor white -border 50 vaultwarden-icon-with-bg{,-and-border}.png
```

BlackDex:
I changed the avatar on matrix just now.

PR #2963, opened 2022-12-08
Increase privacy of masked config

### LINKED ISSUE #2929: make support string more secure
Opening this just as a reference.

- length of smtp username

/ref #2878

BlackDex:
This changes the masking function to hide a bit more information from the generated support string. It will still keep showing the `://` for example, and `,`, but other characters will be hidden.

Also did some small changes on some key's which all showed up as `Internal` on the Settings page.

Fixes #2929

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -232,14 +232,23 @@ macro_rules! make_config {
                 /// We map over the string and remove all alphanumeric, _ and - characters.
                 /// This is the fastest way (within micro-seconds) instead of using a regex (which takes mili-seconds)
                 fn _privacy_mask(value: &str) -> String {
-                    value.chars().map(|c|
-                        match c {
-                            c if c.is_alphanumeric() => '*',
-                            '_' => '*',
-                            '-' => '*',
-                            _ => c
-                        }
-                    ).collect::<String>()
+                    let mut n: u16 = 0;
+                    let mut colon_match = false;
+                    value
+                        .chars()
+                        .map(|c| {
+                            n += 1;
+                            match c {
+                                ':' if n <= 11 => {
+                                    colon_match = true;
+                                    c
+                                }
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -232,13 +232,17 @@ macro_rules! make_config {
...
+                    let mut collon_match = false;
+                    value.chars().map(|c| {
+                        n += 1;
                         match c {
-                            c if c.is_alphanumeric() => '*',
-                            '_' => '*',
-                            '-' => '*',
-                            _ => c
+                            ':' if n <= 11  => { collon_match = true ; c },
+                            '/' if n <= 13 && collon_match  => c,
```
dani-garcia, line 241: collon -> colon :)

BlackDex: I have to digest this a bit 🤣

BlackDex:
@dani-garcia Done :)

src/config.rs
```diff
@@ -232,13 +232,17 @@ macro_rules! make_config {
...
+                    let mut colon_match = false;
+                    value.chars().map(|c| {
+                        n += 1;
                         match c {
-                            c if c.is_alphanumeric() => '*',
-                            '_' => '*',
-                            '-' => '*',
-                            _ => c
+                            ':' if n <= 11  => { colon_match = true ; c },
+                            '/' if n <= 13 && colon_match  => c,
```
tessus, line 241:
```suggestion
                            '/' if n <= 13 && colon_match => c,
```

removed extra space, unless this was intentional

BlackDex: It looks like `cargo fmt` doesn't work inside macro's, not even for _simple_ code.
But i agree, that should be different. I have run `fmt` on that part manually, and replaced it with the right formatted code.

PR #2968, opened 2022-12-09
Remove patched multer-rs

manofthepeace:
Seems like the fix was merged into 2022.10.0: https://github.com/bitwarden/mobile/releases/tag/v2022.10.0. Which is few months old now, so I think it is safe to remvove the workarounds

cc @BlackDex

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1516,7 +1516,8 @@ dependencies = [
 [[package]]
 name = "multer"
 version = "2.0.4"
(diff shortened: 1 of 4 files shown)


BlackDex:
We want to wait with this until next year, and at least after a newer release with this patched version included.
The reason is that we want to give late adapters to newer versions a bit more time.

But thanks for the PR.
We can leave it open until we want to remove it, but that will probably mean you need to perform some re-bases in the meantime.

BlackDex:
@manofthepeace from my point of view this PR is allowed to be merged now.
We have a new release with this patch. And i think all is fine right now. So merging this into the `testing` images would be good.

So if you can re-base, @dani-garcia can merge this :).

manofthepeace:
Rebase done.

BlackDex (approved):

PR #2970, opened 2022-12-09
use black favicon for /admin

tessus:
as mentioned in https://github.com/dani-garcia/bw_web_builds/pull/101#issuecomment-1343289313

image


williamdes (approved):

BlackDex:
Not to be rude here. But @stefan0xC already created some good looking favicon's. This version deviates from those. I would rather have them all the same.

tessus:
Hmm, the white favicon had a 64x64 dimension, which is why I thought I should be using it for the black one as well.

But I can use the black 32x32 from the other PR. Shall I do that?

tessus:
Yep, it is probably better for consistency. Thanks for pointing this out.

PR #3033, opened 2022-12-23
update favicon

wutoz:
remove background color


BlackDex:
We actually added it just a few weeks ago. So, i don't see a point in removing it again actually.
The main reason is, some people have a dark background, others a light one, this way it will always be visible.

wutoz:
get！😄

PR #3058, opened 2022-12-28
Removed unsafe-inline JS from CSP and other fixes

BlackDex:
- Removed `unsafe-inline` for javascript from CSP. The admin interface now uses files instead of inline javascript.
- Modified javascript to work not being inline.
- Run eslint over javascript and fixed some items.
- Added a `to_json` Handlebars helper. Used at the diagnostics page.
- Changed `AdminTemplateData` struct to be smaller. The `config` was always added, but only used at one page. Same goes for `can_backup` and `version`.
- Also inlined CSS.
  We can't remove the `unsafe-inline` from css, because that seems to
  break the web-vault currently. That might need some further checks.
  But for now the 404 page and all the admin pages are clear of inline scripts and styles.

### DIFF
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -144,7 +144,6 @@ fn render_admin_login(msg: Option<&str>, redirect: Option<String>) -> ApiResult<
     let msg = msg.map(|msg| format!("Error: {msg}"));
     let json = json!({
         "page_content": "admin/login",
-        "version": VERSION,
         "error": msg,
         "redirect": redirect,
         "urlpath": CONFIG.domain_path()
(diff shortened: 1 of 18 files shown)


src/api/admin.rs
```diff
@@ -617,13 +602,14 @@ async fn diagnostics(_token: AdminToken, ip_header: IpHeader, mut conn: DbConn)
...
         "dns_resolved": dns_resolved,
+        "current_release": VERSION,
         "latest_release": latest_release,
         "latest_commit": latest_commit,
         "web_vault_enabled": &CONFIG.web_vault_enabled(),
         "web_vault_version": web_vault_version.version,
         "latest_web_build": latest_web_build,
         "running_within_docker": running_within_docker,
-        "docker_base_image": docker_base_image(),
+        "docker_base_image": if running_within_docker { docker_base_image() } else { "Not checked" },
```
tessus, line 612: Wouldn't be `Not available` better?

Not checked means that it was not determined for an unspecified reason. Not available means that this value makes no sense for the current setup.

tessus, line 612:
```suggestion
        "docker_base_image": if running_within_docker { docker_base_image() } else { "Not available" },
```

BlackDex: Then it would be **not applicable**, because technically it is available somewhere. Also, in this case we do not check this, so it is correct i think.

tessus: Ah, yes. You are correct. `Not applicable` is the correct term.

"Not checked" leaves the reader with a question: Why? Why was it not checked? "Not applicable" makes it perfectly clear.

GeekCornerGH (approved):
Haven't looked at rust code, but the rest semms good

PR #3059, opened 2022-12-28
Validate YUBICO_SERVER string (#3003)

### LINKED ISSUE #3003: Not posible to store a YubiKey OTP in the user Account Settings
### Subject of the issue

After following the steps in the Vaultwarden Wiki how to obtain a Client ID and Secret Key at Yubico and setting those values as enviromental variables for the docker container (https://github.com/dani-garcia/vaultwarden/wiki/Enabling-Yubikey-OTP-authentication) I still can't register any YubKey in Vaultwarden. When trying to do so, after touching the gold contact of the YubiKey at one of the YubiKey fields in user account settings, an red error notification is given:

Invalid YubiKey OTP provided.

Tested it with multiple YubiKey Keys (2x YubiKey 5 NFC and 2x YubiKey 5C NFC)

The error is also logged in the logfile:
```
[2022-12-17 12:46:21.330][request][INFO] PUT /api/two-factor/yubikey
[2022-12-17 12:46:21.402][error][ERROR] Invalid Yubikey OTP provided.
[CAUSE] Network(
    reqwest::Error {
        kind: Builder,
        source: RelativeUrlWithoutBase,
    },
)
[2022-12-17 12:46:21.403][response][INFO] (activate_yubikey_put) PUT /api/two-factor/yubikey => 400 Bad Request
```
The YubiKey OTP DEMO site validates all my keys successful:
```
{
  "otp": "xxx",
  "nonce": "xxx",
  "serial": xxx,
  "sl": "100",
  "status": "OK",
  "t": "2022-12-17T12:51:00Z0068"
}
```

I've found the given error in the source code of yubikey.rs at an if/then statement at line 150. The statement looks to accept a string of which the total length is only 12 characters long, while in fact a YubiKey is a 44 character long string where the first 12 characters remain constant (https://developers.yubico.com/OTP/OTPs_Explained.html).

If I for testing purposes only enter the first 12 characters of my YubiKey in the field, the value is stored at the settings. Only this will cause YubiKey 2FA login for Vaultwarden to always fail.

* Vaultwarden version: v1.26.0
* Web-vault version: v2022.10.0
* Running within Docker: true (Base: Alpine)
* Uses a reverse proxy: true
* Internet access: true
* DNS Check: true
* Time Check: true
* HTTPS Check: true
* Database type: SQLite
* Database version: 3.35.4
* Clients used: 
* Reverse proxy and version: 
* Other relevant information: 

**Environment settings which are overridden:** DOMAIN, SIGNUPS_ALLOWED, SHOW_PASSWORD_HINT, ADMIN_TOKEN, IP_HEADER, YUBICO_CLIENT_ID, YUBICO_SECRET_KEY, SMTP_HOST, SMTP_SECURITY, SMTP_PORT, SMTP_FROM, SMTP_FROM_NAME, SMTP_USERNAME, SMTP_PASSWORD

```json
{
  "_duo_akey": "***",
  "_enable_duo": true,
  "_enable_email_2fa": true,
  "_enable_smtp": true,
  "_enable_yubico": true,
  "_icon_service_csp": "",
  "_icon_service_url": "",
  "_ip_header_enabled": true,
  "admin_token": "***",
  "data_folder": "data",
  "database_conn_init": "",
  "database_max_conns": 10,
  "database_timeout": 30,
  "database_url": "****/**.*******",
  "db_connection_retries": 15,
  "disable_admin_token": false,
  "domain": "*****://*********.*****.**",
  "domain_origin": "*****://*********.*****.**",
  "domain_path": "",
  "domain_set": true,
  "duo_host": "***",
  "duo_ikey": "***",
  "duo_skey": "***",
  "email_attempts_limit": 3,
  "email_expiration_time": 600,
  "email_token_size": 6,
  "emergency_notification_reminder_schedule": "0 5 * * * *",
  "emergency_request_timeout_schedule": "0 5 * * * *",
  "enable_db_wal": true,
  "extended_logging": true,
  "helo_name": null,
  "hibp_api_key": null,
  "icon_blacklist_regex": null,
  "icon_cache_negttl": 259200,
  "icon_cache_ttl": 2592000,
  "icon_download_timeout": 10,
  "icon_redirect_code": 302,
  "icon_service": "internal",
  "invitation_org_name": "Bitwarden",
  "invitations_allowed": true,
  "log_file": "/data/bitwarden.log",
  "log_level": "Info",
  "org_attachment_limit": null,
  "org_creation_users": "",
  "reload_templates": false,
  "sends_allowed": true,
  "sends_folder": "data/sends",
  "show_password_hint": false,
  "signups_allowed": false,
  "signups_verify": true,
  "smtp_auth_mechanism": null,
  "smtp_debug": false,
  "smtp_explicit_tls": null,
  "smtp_from": "*******@*****.***",
  "smtp_from_name": "***",
  "smtp_host": "****.*****.***",
  "smtp_password": "***",
  "smtp_port": 587,
  "smtp_security": "starttls",
  "smtp_ssl": null,
  "smtp_timeout": 15,
  "smtp_username": "*******",
  "tmp_folder": "data/tmp",
  "use_syslog": false,
  "web_vault_enabled": true,
  "websocket_enabled": true,
  "websocket_port": 3012,
  "yubico_client_id": "***",
  "yubico_secret_key": "***",
  "yubico_server": ""
}
```
</details>

### Steps to reproduce

1. Follow steps "Enabling Yubikey OTP authentication" in Vaultwarden Wiki.
2. After login in at the Bitwarden Web Vault go to Account Settings.
3. Select Security > Two-step login > YubiKey OTP Security Key > Manage.
4. Enter Master Password and click Continue.
5. Hilight the first YukiKey 1 field, insert the YubiKey and press the gold contact of the YubiKey.

### Expected behaviour
Posible to store the YubiKey's as 2FA for Vaultwarden at the user account settings by just touching the gold contact.

### Actual behaviour
Not posible to use YubiKey as MFA in Vaultwarden.

BlackDex:
If the `YUBICO_SERVER` is defined to an empty string, the whole yubikey implementation doesn't work anymore.

This PR adds a check for this variable that it at least starts with `https://`.

Resolves #3003

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -698,8 +698,17 @@ fn validate_config(cfg: &ConfigItems) -> Result<(), Error> {
         err!("All Duo options need to be set for global Duo support")
     }
 
-    if cfg._enable_yubico && cfg.yubico_client_id.is_some() != cfg.yubico_secret_key.is_some() {
-        err!("Both `YUBICO_CLIENT_ID` and `YUBICO_SECRET_KEY` need to be set for Yubikey OTP support")
+    if cfg._enable_yubico {
+        if cfg.yubico_client_id.is_some() != cfg.yubico_secret_key.is_some() {
+            err!("Both `YUBICO_CLIENT_ID` and `YUBICO_SECRET_KEY` must be set for Yubikey OTP support")
+        }
+
+        if let Some(yubico_server) = &cfg.yubico_server {
+            let yubico_server = yubico_server.to_lowercase();
+            if !yubico_server.starts_with("https://") {
+                err!("`YUBICO_SERVER` must be a valid URL and start with 'https://'. Either unset this variable or provide a valid URL.")
+            }
+        }
     }
 
     if cfg._enable_smtp {


src/config.rs
```diff
@@ -698,8 +698,17 @@ fn validate_config(cfg: &ConfigItems) -> Result<(), Error> {
...
-        err!("Both `YUBICO_CLIENT_ID` and `YUBICO_SECRET_KEY` need to be set for Yubikey OTP support")
+    if cfg._enable_yubico {
+        if cfg.yubico_client_id.is_some() != cfg.yubico_secret_key.is_some() {
+            err!("Both `YUBICO_CLIENT_ID` and `YUBICO_SECRET_KEY` need to be set for Yubikey OTP support")
+        }
+
+        if let Some(yubico_server) = &cfg.yubico_server {
+            let yubico_server = yubico_server.to_lowercase();
+            if !yubico_server.starts_with("https://") {
+                err!("`YUBICO_SERVER` needs to be a valid URL and start with 'https://'. Either unset this variable or provide a valid URL.")
```
tessus, line 709:
```suggestion
                err!("`YUBICO_SERVER` must be a valid URL and start with 'https://'. Either unset this variable or provide a valid URL.")
```

tessus:
IMO an empty string should be handled as an unset variable.

I know that an empty string is not the same, however, in case of most config parameters, this is the best and most user-friendly approach.

BlackDex:
> IMO an empty string should be handled as an unset variable.
> 
> I know that an empty string is not the same, however, in case of most config parameters, this is the best and most user-friendly approach.
> 
> What do you think?

No, i do not think that we should do that. The main reason is that if someone wants to configure this, but for some strange reason the value isn't passed correctly it's going to be an empty string. By allowing this to continue as if it was not set, it could disclose there data to yubico. It also makes it harder for people to troubleshoot if there think it is set correctly. So, better to have it not set at all then to allow an empty string in this case.

PR #3065, opened 2022-12-29
Resolve uninlined_format_args clippy warnings

BlackDex:
The upcomming release of Rust 1.67.0 will warn on `uninlined_format_args`.
This PR resolves that by inlining all these items. It also looks nicer.

Also added a `clippy.toml` which indicates the **MSRV** to prevent clippy suggests which could break the **MSRV** version compile.


tessus (commented):
You missed the following 2, or was this on purpose?

https://github.com/dani-garcia/vaultwarden/blob/f34d7e37c359ea46f5023916ec4d977db5c17c0b/src/main.rs#L310

https://github.com/dani-garcia/vaultwarden/blob/f34d7e37c359ea46f5023916ec4d977db5c17c0b/src/main.rs#L319

P.S.: I didn't check the entire code base, but these 2 caught my eye.

BlackDex:
Well, those are not triggered by that clippy check.

While there are more then just those two which can also be converted, they aren't causing a clippy warning.

That doesn't mean we shouldn't change them though.

tessus:
I am new to Rust, so why don't they trigger a warning? I did not see a compiler directive to not check that part of the code, or maybe I missed it?

BlackDex:
Those macros aren't part of the rust eco system it self, but provide by custom macros. `format!` and `println!` for example are.

tessus:
I see. Thanks for the info.

PR #3076, opened 2022-12-30
Update WebSocket Notifications

### LINKED ISSUE #2591: Unable to update cipher after editing attachments using Desktop client
### Subject of the issue

* Unexpected behavior when deleting an attachment
* Problem of editing and deleting an attachment at the same time

### Deployment environment

* Vaultwarden version: v1.25.0-af50eae6
* Web-vault version: v2022.6.0 (compiled w/ PR72)
* Running within Docker: false (Base: Debian)
* Uses a reverse proxy: true
* Internet access: true
* DNS Check: true
* Time Check: true
* HTTPS Check: true
* Database type: SQLite
* Database version: 3.35.4
* Clients used: Web, Android, Windows, Chrome, Firefox
* Reverse proxy and version: Nginx 1.21.6
* Other relevant information: 

### Steps to reproduce

1. Create an item with an attachment
2. Edit the item
3. Delete the attachment, confirm deletion
4. Before save & close, change any other field
5. Click on save to finish editing the item, error occurs : https://i.imgur.com/63SzWOq.png

### Expected behaviour / Actual behaviour

If the deletion of the attachment has already been taken into account on the server side, there should be no error here.
It is currently impossible to edit a field and delete an attachment at the same time.

### Troubleshooting data

* Image 01 : https://i.imgur.com/63SzWOq.png
* Sorry, I couldn't test on the official version because I don't have my premium subscription anymore, so I can't add attachments.

BlackDex:
Previously the websocket notifications were using the string `app_id` as the
`ContextId`. This was incorrect and should have been the device_uuid
from the client device executing the request. The clients will ignore
the websocket request if the uuid matches. This also fixes some issues
with the Desktop client which is able to modify attachments within the
same screen and causes an issue when saving the attachment afterwards.

Also changed the way to handle removed attachments, since that causes an
error saving the vault cipher afterwards, complaining about a missing
attachment. Bitwarden ignores this, and continues with the remaining
attachments (if any). This also fixes #2591 .

Further some more websocket notifications have been added to some other
functions which enhance the user experience.

- Logout users when deauthed, changed password, rotated keys
- Trigger OrgSyncKeys on user confirm and removal
- Added some extra to the send feature

Also renamed UpdateTypes to match Bitwarden naming.

### DIFF
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -13,7 +13,7 @@ use rocket::{
 };
 
 use crate::{
-    api::{core::log_event, ApiResult, EmptyResult, JsonResult, NumberOrString},
+    api::{core::log_event, ApiResult, EmptyResult, JsonResult, Notify, NumberOrString, UpdateType},
     auth::{decode_admin, encode_jwt, generate_admin_claims, ClientIp},
     config::ConfigBuilder,
     db::{backup_database, get_sql_server_version, models::*, DbConn, DbConnType},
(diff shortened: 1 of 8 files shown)


tessus:
> The clients will ignore the websocket request if the uuid matches.

Hmm, aren't the websocket requests supposed to work for a specific device? Thus the device (client) should ignore the websocket request, if the uuid DOES NOT match.

Or am I missing something here?

BlackDex:
> > The clients will ignore the websocket request if the uuid matches.
> 
> Hmm, aren't the websocket requests supposed to work for a specific device? Thus the device (client) should ignore the websocket request, if the uuid DOES NOT match.
> 
> Or am I missing something here?

All applicable clients will receive the exact same websocket notification. That includes the client which made the change. Because that client already has the correct changes it doesn't need to pull them in again. So the device id if the client which triggered the notification is sent with it, and then that specific client will ignore the request, all others will act upon the message and fetch the latest cipher data, or sync etc..

tessus:
Ah, ok. For some reason I had the impression that some event had to be sent to a signle device, but yep, it's basically a broadcast. Thus I had it backwards. Makes sense.

PR #3108, opened 2023-01-05
allow editing/unhiding by group

### LINKED ISSUE #2989: Group members cannot add new entries to collection on testing-branch
### Subject of the issue
When using the groups-feature on the testing-branch, a "regular" group-member (non-manager) cannot create any entries in a collection, that is given access to via groups.

* Vaultwarden version: v1.26.0-d0b53a6a
* Web-vault version: v2022.11.2
* Uses a reverse proxy: true
* Internet access: true
* DNS Check: true
* Time Check: true
* HTTPS Check: true
* Database type: PostgreSQL
* Database version: PostgreSQL 13.3 on x86_64-pc-linux-gnu, compiled by gcc (Debian 8.3.0-6) 8.3.0, 64-bit
* Clients used: 
* Reverse proxy and version: 
* Other relevant information: 

### Steps to reproduce
1. Create an organisation (or use existing)
2. Add another user "U" (any name is fine) to the organisation (as a regular user)
3. Create(or use existing) collection "C" (any name is fine)
4. Create(or use existing) group "G" (any name is fine)
4.1. Give permission of that collection "C" to the group "G"
4.2. Assign the user "U" to the group "G"
4.3. Make sure that the user "U" does **not** have direct permission on the collection "C". In other words: the permission should be configured via the group only!
5. Login as user "U"
5.1. Add a new entry to the collection "C"

### Expected behavior
The entry should be added to the collection.

### Actual behavior
An error toast appears in the top right: "You lack the necessary permissions to perform this action."

farodin91:
Fixes #2989

### DIFF
--- a/src/db/models/collection.rs
+++ b/src/db/models/collection.rs
@@ -287,47 +287,95 @@ impl Collection {
     }
 
     pub async fn is_writable_by_user(&self, user_uuid: &str, conn: &mut DbConn) -> bool {
-        match UserOrganization::find_by_user_and_org(user_uuid, &self.org_uuid, conn).await {
-            None => false, // Not in Org
-            Some(user_org) => {
-                if user_org.has_full_access() {
-                    return true;
-                }
-
-                db_run! { conn: {
-                    users_collections::table
-                        .filter(users_collections::collection_uuid.eq(&self.uuid))
-                        .filter(users_collections::user_uuid.eq(user_uuid))
-                        .filter(users_collections::read_only.eq(false))
-                        .count()
-                        .first::<i64>(conn)
-                        .ok()
-                        .unwrap_or(0) != 0
-                }}
-            }
-        }
+        let user_uuid = user_uuid.to_string();
+        db_run! { conn: {
+            collections::table
+            .left_join(users_collections::table.on(
+                users_collections::collection_uuid.eq(collections::uuid).and(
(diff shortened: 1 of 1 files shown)


farodin91:
I wanted to fix and saw some other query's that are huge. I find it hard to debug query's that have multiple joins. This way you can reuse some code. The location is probably wrong, I will fix this it by moving directly into collections.

farodin91:
@BlackDex  I changed it to one query by using find_by_uuid_and_user.

BlackDex (approved):
While I'm still in favor of changing the DB Schema around the whole group handling to make it less complex, probably with a bit less of normalization, this _quick fix_ does provide a good way of allowing this feature to be tested a bit further by more people and give us time to think about a proper and better solution.

So, ill approve this PR, with the side note that i think we do need to rework this feature to be less complex.

BlackDex:
@farodin91 Could you try to rebase 😄 .

farodin91:
@BlackDex  Rebased.

Do i have write in the discussion for re license thread?

BlackDex:
> @BlackDex Rebased.
> 
> Do i have write in the discussion for re license thread?

That would be nice :)

dani-garcia (approved):

bendem:
Awesome, and thank you for favoring performance. It is indeed a bit harder to read, but users and database admins thank you for it (re #1402).

PR #3116, opened 2023-01-06
Admin password reset

sirux88:
Since i'd like to use the admin reset password functionality within my instance I implemented the missing "Reset Master Password" policy and the needed functions

There are two points that (maybe) need improvment:

- Auto-enrollment won't work for instances without email enabled
- Bitwarden's reference implementation logs out the user to reset after resetting the password. I haven't found a api/method for this and also don't know how to implement

As i'm not used to programming in rust please feel free to comment and throw any improvements concering syntax.
All other improvements are welcome aswell

### DIFF
--- a/migrations/mysql/2023-01-06-151600_add_reset_password_support/up.sql
+++ b/migrations/mysql/2023-01-06-151600_add_reset_password_support/up.sql
@@ -1,0 +1,2 @@
+ALTER TABLE users_organizations
+ADD COLUMN reset_password_key VARCHAR(255);
--- a/migrations/mysql/2023-01-06-151600_add_reset_password_support/up.sql
+++ b/migrations/mysql/2023-01-06-151600_add_reset_password_support/up.sql
@@ -1,2 +1,2 @@
 ALTER TABLE users_organizations
-ADD COLUMN reset_password_key VARCHAR(255);
+ADD COLUMN reset_password_key TEXT;
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -2583,7 +2583,7 @@ async fn put_reset_password(
 
     let reset_request = data.into_inner().data;
 
-    user.set_password_and_key(reset_request.NewMasterPasswordHash.as_str(), reset_request.Key.as_str(), None);
+    user.set_password(reset_request.NewMasterPasswordHash.as_str(), Some(reset_request.Key), true, None);
     user.save(&mut conn).await?;
 
     nt.send_user_update(UpdateType::LogOut, &user).await;
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -2542,7 +2542,7 @@ async fn put_reset_password(
 
     // Sending email before resetting password to ensure working email configuration and the resulting
     // user notification. Also this might add some protection against security flaws and misuse
-    if let Err(e) = mail::send_admin_reset_password(&user.email.to_lowercase(), &user.name, &org.name).await {
+    if let Err(e) = mail::send_admin_reset_password(&user.email, &user.name, &org.name).await {
         error!("Error sending user reset password email: {:#?}", e);
     }
 
@@ -2615,19 +2615,11 @@ async fn check_reset_password_applicable_and_permissions(
     };
 
     // Resetting user must be higher/equal to user to reset
-    let mut reset_allowed = false;
-    if headers.org_user_type == UserOrgType::Owner {
-        reset_allowed = true;
+    match headers.org_user_type {
+        UserOrgType::Owner => Ok(()),
+        UserOrgType::Admin if target_user.atype <= UserOrgType::Admin => Ok(()),
+        _ => err!("No permission to reset this user's password"),
     }
-    if headers.org_user_type == UserOrgType::Admin {
-        reset_allowed = target_user.atype != (UserOrgType::Owner as i32);
-    }
-
-    if !reset_allowed {
-        err!("No permission to reset this user's password");
-    }
-
-    Ok(())
 }
 
 async fn check_reset_password_applicable(org_id: &str, conn: &mut DbConn) -> EmptyResult {
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -711,10 +711,6 @@ async fn send_invite(
         err!("Only Owners can invite Managers, Admins or Owners")
     }
 
-    if !CONFIG.mail_enabled() && OrgPolicy::org_is_reset_password_auto_enroll(&org_id, &mut conn).await {
-        err!("With mailing disabled and auto-enrollment-feature of reset-password-policy enabled it's not possible to invite users");
-    }
-
     for email in data.Emails.iter() {
         let email = email.to_lowercase();
         let mut user_org_status = UserOrgStatus::Invited as i32;
@@ -729,10 +725,6 @@ async fn send_invite(
                 }
 
                 if !CONFIG.mail_enabled() {
-                    if OrgPolicy::org_is_reset_password_auto_enroll(&org_id, &mut conn).await {
-                        err!("With disabled mailing and enabled auto-enrollment-feature of reset-password-policy it's not possible to invite existing users");
-                    }
-
                     let invitation = Invitation::new(&email);
                     invitation.save(&mut conn).await?;
                 }
@@ -748,10 +740,6 @@ async fn send_invite(
                     // automatically accept existing users if mail is disabled
                     if !CONFIG.mail_enabled() && !user.password_hash.is_empty() {
                         user_org_status = UserOrgStatus::Accepted as i32;
-
-                        if OrgPolicy::org_is_reset_password_auto_enroll(&org_id, &mut conn).await {
-                            err!("With disabled mailing and enabled auto-enrollment-feature of reset-password-policy it's not possible to invite existing users");
-                        }
                     }
                     user
                 }
@@ -1597,17 +1585,8 @@ async fn put_policy(
         }
     }
 
-    // This check is required since invited users automatically get accepted if mailing is not enabled (this seems like a vaultwarden specific feature)
-    // As a result of this the necessary "/accepted"-endpoint doesn't get hit.
-    // But this endpoint is required for autoenrollment while invitation.
-    // Nevertheless reset password is fully fuctiontional in settings without mailing by manual enrollment
-
     if pol_type_enum == OrgPolicyType::ResetPassword && data.enabled && !CONFIG.mail_enabled() {
-        if let Some(policy_data) = &data.data {
-            if policy_data["autoEnrollEnabled"].as_bool().unwrap_or(false) {
-                err!("Autoenroll can't be used since it requires enabled emailing")
-            }
-        }
+        err!("Due to potential security flaws and/or misuse reset password policy is disabled on mail disabled instances")
     }
 
     let mut policy = match OrgPolicy::find_by_org_and_type(&org_id, pol_type_enum, &mut conn).await {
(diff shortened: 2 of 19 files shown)


BlackDex (changes requested):
Thanks for the contribution.
In general, since i haven't looked at the code that good yet.

You mentioned that it will not work without email being enabled.
In that case, please add restrictions and check if mail is enabled with `if CONFIG.mail_enabled() {`
and prevent this policy from being activated at all.
And, also check if the policy is enabled when an admin tries to use this feature, mail is still enabled, and not turned of afterwards. Maybe even prevent this policy from being active in that case?
That way we can prevent issue reports, and you can return an error message stating mail has to be enabled.

In regards to logging the user out. I have a PR open which helps with that for at least the web-socket enabled clients (#3076). But other then that, depending on if they need a hard logout/refresh from the user, you can use the `user.reset_security_stamp()` that will invalidate there security stamp, and will prevent any further valid communication from any client.

sirux88:
Concerning logging out:

> In regards to logging the user out. I have a PR open which helps with that for at least the web-socket enabled clients (#3076). But other then that, depending on if they need a hard logout/refresh from the user, you can use the `user.reset_security_stamp()` that will invalidate there security stamp, and will prevent any further valid communication from any client.

1. Since resetting the password relies on `user.set_password` `user.reset_security_stamp()` is already been executed. Therefore this should be fine for non-websocket-enabled clients
2. I added your mentioned web-socket logout within my latest commit. 
Seems to me like the default behavior of at least web-vault

BlackDex:
>
> 2. I added your mentioned web-socket logout within my latest commit.
>    Seems to me like the default behavior of at least web-vault

That is true for the one you are currently using. But desktop client and browser extensions will now also receive the logout request, else they would in the end act upon the reset of the security stamp, but this is a bit nicer i think.

sirux88:
> You mentioned that it will not work without email being enabled.
In that case, please add restrictions and check if mail is enabled with if CONFIG.mail_enabled() {
and prevent this policy from being activated at all.

Let me clarify a bit what I ment with "auto-enrollment":
The restriction I mentioned is only about auto-enrollment-feature. Auto-enrollment means newly invited users get enrolled automatically while accepting the invation. This is implemented within org-endpoint `/organizations/<org_id>/users/<_org_user_id>/accept`
On email-disabled-instances this endpoint is never gone to be hit. As a consequence reset-password can't be enrolled

Manual enrollment can always be done. Also in email-disabled instances. 
So there's no need to disable the policy at all for email-disabled instances.

src/api/core/organizations.rs
```diff
@@ -1546,6 +1562,19 @@ async fn put_policy(
...
+    //Nevertheless reset password is fully fuctiontional in settings without mailing by manual enrollment
+    if pol_type_enum == OrgPolicyType::ResetPassword && data.enabled {
+        //Maybe this check could get improved to get away from string checking
+        //Deserializing from "serde_json::from_value" won't work with current scope
+        let pol_add = serde_json::to_string(&data.data)?;
+        if pol_add.contains("\"autoEnrollEnabled\":true") && !CONFIG.mail_enabled() {
+            err!("Autoenroll can't be used since it requires enabled emailing")
+        }
+    }
+
```
sirux88, line 1577: I already added a check on email-disabled instances. While setting the policy, an error is thrown if auto-enrollment is enabled on those instances preventing its usage

src/api/core/organizations.rs
```diff
@@ -740,6 +740,10 @@ async fn send_invite(
                     // automatically accept existing users if mail is disabled
                     if !CONFIG.mail_enabled() && !user.password_hash.is_empty() {
                         user_org_status = UserOrgStatus::Accepted as i32;
+
+                        if OrgPolicy::org_is_reset_password_auto_enroll(&org_id, &mut conn).await {
+                            err!("With disabled mailing and enabled auto-enrollment-feature of reset-password-policy it's not possible to invite existing users");
+                        }
```
sirux88, line 746: > And, also check if the policy is enabled when an admin tries to use this feature, mail is still enabled, and not turned of afterwards. Maybe even prevent this policy from being active in that case?
That way we can prevent issue reports, and you can return an error message stating mail has to be enabled.

I added a check that is preventing existing user invitation to current org if auto-enrollment is enabled but mailing is disabled.
This state can be reached if mailing was enabled while setting up the policy and disabling mailing afterwards.
I think this was what you ment.

sirux88: I put the check aswell for non existing users if they get invited. They habe the same restrictions.

sirux88: I refactored the stuff I did in my latest commit

BlackDex: While I still have not looked at this yet, but could you fix the clippy warnings `cargo clippy --features sqlite,mysql,postgresql` ? 
You can also install pre-commit and enable it for this repo so all checks are done before you commit :).

sirux88: If I should do a rebase to current main, please tell me

GeekCornerGH (changes requested):
You should rebase your PR

sirux88:
Rebase done @GeekCornerGH

src/api/core/organizations.rs
BlackDex: This is probably a nicer option. Which also does some type-checking.
It also prevents from checking the `policy_data` if mail is enabled. This saves some checking and converting cycles.

```suggestion
    if pol_type_enum == OrgPolicyType::ResetPassword && data.enabled && !CONFIG.mail_enabled() {
        if let Some(policy_data) = &data.data {
            if policy_data["autoEnrollEnabled"].as_bool().unwrap_or(false) {
                err!("Autoenroll can't be used since it requires enabled emailing")
            }
        }
    }
```

src/api/core/organizations.rs
```diff
@@ -1570,6 +1597,19 @@ async fn put_policy(
         }
     }
 
+    //This check is required since invited users automatically get accepted if mailing is not enabled (this seems like a vaultwarden specific feature)
+    //As a result of this the necessary "/accepted"-endpoint doesn't get hit.
+    //But this endpoint is required for autoenrollment while invitation.
+    //Nevertheless reset password is fully fuctiontional in settings without mailing by manual enrollment
```
BlackDex, line 1603: Please add a space after the `//` 😄

src/api/core/organizations.rs
```diff
@@ -1570,6 +1597,19 @@ async fn put_policy(
...
+    if pol_type_enum == OrgPolicyType::ResetPassword && data.enabled {
+        //Maybe this check could get improved to get away from string checking
+        //Deserializing from "serde_json::from_value" won't work with current scope
```
BlackDex, line 1606: Same here regarding the `//`

BlackDex (changes requested):

src/api/core/organizations.rs
sirux88: I had something similar in the first place but I struggeld a bit with Rust.
Thanks for the improvement. I took it as you suggested

migrations/mysql/2023-01-06-151600_add_reset_password_support/up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE users_organizations
+ADD COLUMN reset_password_key VARCHAR(255);
```
BlackDex, line 2: Bitwarden uses `longtext` for this, we use `text` for all the other key columns.
I suggest to change this to `TEXT` to prevent issues with long keys.

src/api/core/organizations.rs
```diff
@@ -2460,6 +2500,198 @@ async fn delete_group_user(
...
+        Some(user) => user,
+        None => err!("User not found"),
+    };
+
+    let reset_request = data.into_inner().data;
+
+    user.set_password(reset_request.NewMasterPasswordHash.as_str(), Some(reset_request.Key), true, None);
+    user.save(&mut conn).await?;
+
+    nt.send_user_update(UpdateType::LogOut, &user).await;
```
BlackDex, line 2589: Please update this to `nt.send_logout()`;

```suggestion
    nt.send_logout(&user, None).await;
```

BlackDex (changes requested):
Some small changes on items i see right now.
I haven't tested the code yet.

src/api/core/organizations.rs
```diff
@@ -2460,6 +2500,198 @@ async fn delete_group_user(
...
+#[put("/organizations/<org_id>/users/<org_user_id>/reset-password", data = "<data>")]
+async fn put_reset_password(
+    org_id: String,
+    org_user_id: String,
+    headers: AdminHeaders,
+    data: JsonUpcase<OrganizationUserResetPasswordRequest>,
+    mut conn: DbConn,
+    ip: ClientIp,
+    nt: Notify<'_>,
+) -> EmptyResult {
```
BlackDex, line 2539: This function needs to check if `CONFIG.mail_enabled()` at the start of the function, and deny a password reset if not enabled. This could be a security risk when an admin in some way is able to disable this.

I might even say that we should send the mail first, if that is successful then reset the password, else don't even reset the password at all, since that also could cause a security issue.

It might even be better to do this during the `get_reset_password_details` function. That way it should prevent the UI from continuing i think.

Maybe you need to put some other `CONFIG.mail_enabled()` checks at some functions to prevent security issues regarding the password reset?

BlackDex (changes requested):

migrations/mysql/2023-01-06-151600_add_reset_password_support/up.sql
BlackDex: I just tested the code, and my keys are bigger then 255.

sirux88: I changed this as suggested. Don't know why I did this in the first place and only for mysql.
Sorry for that

src/api/core/organizations.rs
sirux88: Changed as suggested

src/api/core/organizations.rs
sirux88: I changed a couple of think to get along with what you suggested.
I refactored two check routines and added those to the relevant methods for password resetting. 

Since the additional checks now exclude using reset password policy in email-disabled instances at all I ended up in blocking reset password policy (on email disabled instances) while setting it (an error is thrown).
This may disappoint users but in my eyes security goes first for that functionality

sirux88: Is there a chance to hide a policy conditionally @BlackDex? 
I don't see any endpoint that queries available policies (`"/organizations/<org_id>/policies")` only fetches the once that exist in database context not in general) or any other way to do this

src/api/core/organizations.rs
```diff
@@ -2647,16 +2601,41 @@ async fn get_reset_password_details(
...
+    }
+    if headers.org_user_type == UserOrgType::Admin {
+        reset_allowed = target_user.atype != (UserOrgType::Owner as i32);
+    }
+
+    if !reset_allowed {
+        err!("No permission to reset this user's password");
+    }
+
+    Ok(())
```
BlackDex, line 2630: A `match` would be better suited here. Now all if statements are always processed while only one should be needed.

```suggestion
    match headers.org_user_type {
        UserOrgType::Owner => Ok(()),
        UserOrgType::Admin if target_user.atype <= UserOrgType::Admin => Ok(()),
        _ => err!("No permission to reset this user's password"),
    }
```

BlackDex (changes requested):
Some quick items i saw just now. Will try to further test and check later today.

src/api/core/organizations.rs
BlackDex: You can hide that policy by setting https://github.com/dani-garcia/vaultwarden/pull/3116/files#diff-1fbd2918c7f409ef277132370231b3fb9646fe76f7661af3eeead167540ec671R162 to `false` if mail is disabled.

That should solve at least the checking during the enabling of the policy. I also think that it will prevent the interface from even showing the option to reset a users password if that is set to false.

Then the only thing still needed to be done is make sure the password isn't reset if the mail was unable to be sent, to prevent the users from not knowing it was reset.

sirux88: > You can hide that policy by setting https://github.com/dani-garcia/vaultwarden/pull/3116/files#diff-1fbd2918c7f409ef277132370231b3fb9646fe76f7661af3eeead167540ec671R162 to `false` if mail is disabled.
> 
> That should solve at least the checking during the enabling of the policy. I also think that it will prevent the interface from even showing the option to reset a users password if that is set to false.

Yeah you are right. Setting 'UseResetPassword' to false (conditionally by 'CONFIG.mail_enabled()') hides the policy.
Therefore I removed the check within setting the policy (as you mentioned aswell)

src/api/core/organizations.rs
sirux88: much better than mine

BlackDex (approved):
Great! Looks nice, and seems to work fine during my tests.
Thanks for the contribution!

src/api/core/organizations.rs
sirux88: > Then the only thing still needed to be done is make sure the password isn't reset if the mail was unable to be sent, to prevent the users from not knowing it was reset. Because all other items are already hidden.

Is https://github.com/dani-garcia/vaultwarden/pull/3116/files?diff=unified&w=0#diff-4a9695da74664125793a066394e03e68892b306081e00dc7861f2da1145b0c48R2539-R2543 not enough?

I put sending the email before resetting the password itself. The error thrown should interrupt processing if the email could not be sent.

One single point would remain, if the email won't get delivered by the email server. 
But I think we can't add any protection against that since it is out of vaultwarden's scope

BlackDex: I think it's fine right now. We can't ensure the mail is delivered indeed. But we at least tried to be to be as certain as possible.

dani-garcia (approved):

dani-garcia:
Thanks!

PR #3124, opened 2023-01-07
Change `text/plain` API responses to `application/json`

jjlin:
Recent versions of the Bitwarden clients (see bitwarden/clients#3574) won't parse non-JSON responses. The most noticeable consequence is that `/api/accounts/revision-date` responses won't be parsed, leading to `/api/sync` always being called, even when it's not necessary.


tessus:
Hmm, this is interesting. The bw clients do not check whether the response is a valid JSON document, but whether the content-type is JSON. e.g. I can send any data with any type of content-type....

BlackDex (approved):

PR #3126, opened 2023-01-09
Remove `shrink-to-fit=no` from viewport-meta-tag

redwerkz:
This was a workaroud needed for iOS versions before 9.3 and is not part of the recommended viewport meta tag anymore.
https://www.scottohara.me/blog/2018/12/11/shrink-to-fit.html


BlackDex (approved):
While it doesn't do any harm sitting there, it probably can be removed indeed.

PR #3134, opened 2023-01-11
Add avatar color support

BlackDex:
The new web-vault v2023.1.0 supports a custom color for the avatar. https://github.com/bitwarden/server/pull/2330

This PR adds this feature.

### DIFF
--- a/src/api/core/accounts.rs
+++ b/src/api/core/accounts.rs
@@ -39,6 +39,7 @@ pub fn routes() -> Vec<rocket::Route> {
         api_key,
         rotate_api_key,
         get_known_device,
+        put_avatar,
     ]
 }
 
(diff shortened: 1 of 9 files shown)


src/api/core/accounts.rs
```diff
@@ -228,6 +229,32 @@ async fn post_profile(data: JsonUpcase<ProfileData>, headers: Headers, mut conn:
...
+#[put("/accounts/avatar", data = "<data>")]
+async fn put_avatar(data: JsonUpcase<AvatarData>, headers: Headers, mut conn: DbConn) -> JsonResult {
+    let data: AvatarData = data.into_inner().data;
+
+    // It looks like it only supports the 6 hex color format.
+    // If you try to add the short value it will not show that color.
+    // Check and force 7 chars, including the #.
+    if let Some(color) = &data.AvatarColor {
+        if color.len() != 7 {
+            err!("The field AvatarColor must be a string with a length of 7 characts")
```
tessus, line 247: There's a typo:

```suggestion
            err!("The field AvatarColor must be a string with a length of 7 characters")
```

However, maybe you should just mention the proper format:

```suggestion
            err!("The field AvatarColor must be a string containing the HEX color in the format #rrggbb")
```

BlackDex: Well, it doesn't really matter what we put there. It is never shown to the user.
But ill change the spelling issue at least :)

tessus: That is true, but it also helps to present the proper info to admins who have access to the log without having them to read the source code. ;-)

BlackDex: I have changed it to a better sentence :)

PR #3139, opened 2023-01-13
Minor refactoring concering user.setpassword

sirux88:
Setting a new users password hash always requires setting a new key value aswell
Therefore I refactored the specific method user.set_password and all corresponding usages


BlackDex (approved):
LGTM

PR #3147, opened 2023-01-16
add support for system mta though sendmail

soruh:
This PR enables using `lettre`'s `SendmailTransport` in Vaultwarden.

It adds two configuration options:
    - `USE_SENDMAIL` wether to use sendmail instead of directly connecting to a SMTP server
    -  `SENDMAIL_COMMAND` what sendmail command to use. By default the one from the `$PATH` is used, but I think there are cases where it makes sense to specify a different one

Supporting the system MTA allows for more flexible mail configuration without too much effort and does not require packaging any sendmail client as assumed in #2198.

While there are other usecases like sending through a local mailserver without having to create a dedicated SMTP Account as mentioned in #2198, my specific usecase for this is the following:

I have msmtp configured with an account that all applications use without me having to paste the SMTP configuration and credentials into each of their configs and allows me to only have to maintain the one SMTP configuration instead of potentially forgetting one (not fun).

Testing can be done as follows:

```bash
> sudo pacman -S msmtp-mta
> sudo tee /etc/msmtprc << EOF
defaults

account default
auth on
from <from>
host <hostname>
password <passowrd>
tls on
user <username>
EOF
> tee .env << EOF
USE_SENDMAIL = true
SMTP_FROM = "bitwarden@tld"
EOF
> cargo run [...] # run vaultwarden
```

log into vaultwarden, click
User icon in top right > account settings > security > Two-step login > Email > Manage > Send Email

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -373,7 +373,7 @@
 # ROCKET_WORKERS=10
 # ROCKET_TLS={certs="/path/to/certs.pem",key="/path/to/key.pem"}
 
-## Mail specific settings, set SMTP_HOST and SMTP_FROM to enable the mail service.
+## Mail specific settings, set SMTP_FROM and either SMTP_HOST or USE_SENDMAIL to enable the mail service.
 ## To make sure the email links are pointing to the correct host, set the DOMAIN variable.
 ## Note: if SMTP_USERNAME is specified, SMTP_PASSWORD is mandatory
 # SMTP_HOST=smtp.domain.tld
@@ -384,6 +384,11 @@
 # SMTP_USERNAME=username
 # SMTP_PASSWORD=password
 # SMTP_TIMEOUT=15
+
+# Whether to send mail via the `sendmail` command
+# USE_SENDMAIL=false
+# Which sendmail command to use. The one found in the $PATH is used if not specified.
+# SENDMAIL_COMMAND="/path/to/sendmail"
 
 ## Defaults for SSL is "Plain" and "Login" and nothing for Non-SSL connections.
 ## Possible values: ["Plain", "Login", "Xoauth2"].
--- a/src/config.rs
+++ b/src/config.rs
@@ -749,31 +749,34 @@ fn validate_config(cfg: &ConfigItems) -> Result<(), Error> {
         }
 
         if cfg.use_sendmail {
-            if let Some(ref command) = cfg.sendmail_command {
-                let path = std::path::Path::new(&command);
+            let command = cfg.sendmail_command.as_deref().unwrap_or("sendmail");
 
-                if !path.is_absolute() {
-                    err!(format!("path to sendmail command `{path:?}` is not absolute"));
+            let mut path = std::path::PathBuf::from(command);
+
+            if !path.is_absolute() {
+                match which::which(command) {
+                    Ok(result) => path = result,
+                    Err(_) => err!(format!("sendmail command {command:?} not found in $PATH")),
                 }
+            }
 
-                match path.metadata() {
-                    Err(err) if err.kind() == std::io::ErrorKind::NotFound => {
-                        err!(format!("sendmail command not found at `{path:?}`"))
+            match path.metadata() {
+                Err(err) if err.kind() == std::io::ErrorKind::NotFound => {
+                    err!(format!("sendmail command not found at `{path:?}`"))
+                }
+                Err(err) => {
+                    err!(format!("failed to access sendmail command at `{path:?}`: {err}"))
+                }
+                Ok(metadata) => {
+                    if metadata.is_dir() {
+                        err!(format!("sendmail command at `{path:?}` isn't a directory"));
                     }
-                    Err(err) => {
-                        err!(format!("failed to access sendmail command at `{path:?}`: {err}"))
-                    }
-                    Ok(metadata) => {
-                        if metadata.is_dir() {
-                            err!(format!("sendmail command at `{path:?}` isn't a directory"));
-                        }
 
-                        #[cfg(unix)]
-                        {
-                            use std::os::unix::fs::PermissionsExt;
-                            if !metadata.permissions().mode() & 0o111 != 0 {
-                                err!(format!("sendmail command at `{path:?}` isn't executable"));
-                            }
+                    #[cfg(unix)]
+                    {
+                        use std::os::unix::fs::PermissionsExt;
+                        if !metadata.permissions().mode() & 0o111 != 0 {
+                            err!(format!("sendmail command at `{path:?}` isn't executable"));
                         }
                     }
                 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,3 +1,4 @@
+use std::env::consts::EXE_SUFFIX;
 use std::process::exit;
 use std::sync::RwLock;
 
@@ -749,12 +750,12 @@ fn validate_config(cfg: &ConfigItems) -> Result<(), Error> {
         }
 
         if cfg.use_sendmail {
-            let command = cfg.sendmail_command.as_deref().unwrap_or("sendmail");
+            let command = cfg.sendmail_command.clone().unwrap_or_else(|| format!("sendmail{EXE_SUFFIX}"));
 
-            let mut path = std::path::PathBuf::from(command);
+            let mut path = std::path::PathBuf::from(&command);
 
             if !path.is_absolute() {
-                match which::which(command) {
+                match which::which(&command) {
                     Ok(result) => path = result,
                     Err(_) => err!(format!("sendmail command {command:?} not found in $PATH")),
                 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -614,6 +614,10 @@ make_config! {
     smtp: _enable_smtp {
         /// Enabled
         _enable_smtp:                  bool,   true,   def,     true;
+        /// Use Sendmail |> Whether to send mail via the `sendmail` command
+        use_sendmail:                  bool,   true,   def,     false;
+        /// Sendmail Command |> Which sendmail command to use. The one found in the $PATH is used if not specified.
+        sendmail_command:              String, true,   option;
         /// Host
         smtp_host:                     String, true,   option;
         /// DEPRECATED smtp_ssl |> DEPRECATED - Please use SMTP_SECURITY
@@ -653,7 +657,7 @@ make_config! {
     /// Email 2FA Settings
     email_2fa: _enable_email_2fa {
         /// Enabled |> Disabling will prevent users from setting up new email 2FA and using existing email 2FA configured
-        _enable_email_2fa:      bool,   true,   auto,    |c| c._enable_smtp && c.smtp_host.is_some();
+        _enable_email_2fa:      bool,   true,   auto,    |c| c._enable_smtp && (c.smtp_host.is_some() || c.use_sendmail);
         /// Email token size |> Number of digits in an email 2FA token (min: 6, max: 255). Note that the Bitwarden clients are hardcoded to mention 6 digit codes regardless of this setting.
         email_token_size:       u8,     true,   def,      6;
         /// Token expiration time |> Maximum time in seconds a token is valid. The time the user has to open email client and copy token.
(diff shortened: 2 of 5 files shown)


src/config.rs
```diff
@@ -738,27 +742,29 @@ fn validate_config(cfg: &ConfigItems) -> Result<(), Error> {
...
-            err!("SMTP_FROM does not contain a mandatory @ sign")
-        }
+        if !cfg.use_sendmail {
+            if cfg.smtp_host.is_some() ^ !cfg.smtp_from.is_empty() {
+                err!("Both `SMTP_HOST` and `SMTP_FROM` need to be set for email support without `USE_SENDMAIL`")
+            }
 
-        if cfg.smtp_username.is_some() != cfg.smtp_password.is_some() {
-            err!("Both `SMTP_USERNAME` and `SMTP_PASSWORD` need to be set to enable email authentication")
+            if cfg.smtp_username.is_some() ^ cfg.smtp_password.is_some() {
```
BlackDex, line 750: Is there a reason why you changed the expression here from `!=` to `^`?

src/config.rs
```diff
@@ -738,27 +742,29 @@ fn validate_config(cfg: &ConfigItems) -> Result<(), Error> {
...
 
-        if cfg.smtp_host.is_some() == cfg.smtp_from.is_empty() {
-            err!("Both `SMTP_HOST` and `SMTP_FROM` need to be set for email support")
-        }
-
-        if cfg.smtp_host.is_some() && !cfg.smtp_from.contains('@') {
-            err!("SMTP_FROM does not contain a mandatory @ sign")
-        }
+        if !cfg.use_sendmail {
+            if cfg.smtp_host.is_some() ^ !cfg.smtp_from.is_empty() {
```
BlackDex, line 746: Is there a reason why you changed the expression here from `==` to `^`?

src/config.rs
```diff
@@ -612,6 +612,10 @@ make_config! {
     smtp: _enable_smtp {
         /// Enabled
         _enable_smtp:                  bool,   true,   def,     true;
+        /// Wether or not to send mails through sendmail
+        use_sendmail:                  bool,   true,   def,     false;
+        /// Which sendmail command to use. The one found in the $PATH is used if not specified.
+        sendmail_command:              String, true,   option;
```
BlackDex, line 618: Please adjust the comments so that it will look better in the admin interface.
It needs to be a short identifier separated by `|>` followed by a description.
Is is done for `smtp_ssl` for example.

BlackDex (changes requested):
I think there are some adjustments missing here: https://github.com/dani-garcia/vaultwarden/blob/9991563573ea43b997da47a62bbb83669c667462/src/config.rs#L1046-L1049

src/config.rs
soruh: Fixed in [41faff5](https://github.com/dani-garcia/vaultwarden/pull/3147/commits/41faff5ec789df635d6af9b65ffd0eea73617757)

soruh:
> I think there are some adjustments missing here:
> 
> https://github.com/dani-garcia/vaultwarden/blob/9991563573ea43b997da47a62bbb83669c667462/src/config.rs#L1046-L1049

Fixed in [41faff5](https://github.com/dani-garcia/vaultwarden/pull/3147/commits/41faff5ec789df635d6af9b65ffd0eea73617757)

src/config.rs
soruh: I had originally had changed a lot more here only to notice that it wasn't necessary so this is what is left of that. 
Both are equivalent, but in my mind XOR much easier to reason about than bool == bool (which is why I changed it to understand the code better). I'm happy to revert this if you want.

src/config.rs
soruh: same as above

BlackDex (changes requested):
Could you btw also add these variables to the `.env.template` with the same comments :)

soruh:
Oh, I missed that. Might be because it was hidden :)

src/config.rs
BlackDex: I think it is just what someone perverse on using. I actually find the `!=` and `==` a bit more clear. Though, in the end for the compiler it doesn't matter in this case 😄 .

BlackDex (changes requested):
Ok, i just tested this locally, and it seems to work fine.

Though it does add extra config items, i do think this could be useful for some people.
I know it has been requested in the past. And since there are more and more standalone builds made of Vaultwarden, like for Arch Linux, Alpine, Nixos, FreeBSD, Void etc... this could be useful.

There is just one item i think we may need to address. If someone provides `SENDMAIL_COMMAND` i think it should be the full path to the executable, and therefor, i think we should validate if that file exists or not during the config validation. This will prevent users from configuring this variable, but a non working environment.

Other then that, LGTM.

soruh:
Thank you, that's a good point. I'll add that when I get home.

soruh:
It looks like the `SENDMAIL_COMMAND` is untimately passed into tokio::process::Command so if it's an absolute path it will just be executed but if it's not it will be looked up in the `$PATH`. I see two options to check validity:
- force the command to be an absolute path and check if the file exists and is executable
- allow commands in the `$PATH` and figure out a way to check if the command is valid (like spawning a tokio::process::Command with the command and checking for ENOENT)

I think the latter is definitely more involved, but there could be use cases where it's needed. Do you have any preference, or a better way to do this?

soruh:
I think I would prefer forcing an absolute path

BlackDex:
I would say the first. Make a path mandatory to be absolute.
For one, it feels safer to me and if someone wants to override it, they probably know the full path.

Checking if it is executable is going to be difficult. That isn't supported on Windows. And probably to much i think.

soruh:
Great, I agree. I'll implement that when I'm home then.

soruh:
[07f754b](https://github.com/dani-garcia/vaultwarden/pull/3147/commits/07f754b8b10bdb1105e921523b7917d2fe34c01b) adds the check. If you think the check for executability is too complicated I can remove it again.

.env.template
```diff
@@ -385,6 +385,11 @@
 # SMTP_PASSWORD=password
 # SMTP_TIMEOUT=15
 
+# Wether or not to send mails through sendmail
```
jjlin, line 388:
```suggestion
# Whether to send mail via the `sendmail` command
```

src/config.rs
```diff
@@ -612,6 +612,10 @@ make_config! {
     smtp: _enable_smtp {
         /// Enabled
         _enable_smtp:                  bool,   true,   def,     true;
+        /// Use Sendmail |> Wether or not to send mails through sendmail
```
jjlin, line 615: Same comment as in `.env.template`

src/config.rs
```diff
@@ -738,27 +742,59 @@ fn validate_config(cfg: &ConfigItems) -> Result<(), Error> {
...
-            err!("Both `SMTP_HOST` and `SMTP_FROM` need to be set for email support")
-        }
+        if cfg.use_sendmail {
+            if let Some(ref command) = cfg.sendmail_command {
+                let path = std::path::Path::new(&command);
 
-        if cfg.smtp_host.is_some() && !cfg.smtp_from.contains('@') {
-            err!("SMTP_FROM does not contain a mandatory @ sign")
-        }
+                if !path.is_absolute() {
```
jjlin, line 749: Just FYI, it doesn't seem like lettre requires this to be an absolute path. If it's just a command (like the default value of `sendmail`), it uses the one found via `PATH`.

src/config.rs
```diff
@@ -612,6 +612,10 @@ make_config! {
     smtp: _enable_smtp {
         /// Enabled
         _enable_smtp:                  bool,   true,   def,     true;
+        /// Use Sendmail |> Wether or not to send mails through sendmail
+        use_sendmail:                  bool,   true,   def,     false;
```
jjlin, line 616: This variable could potentially be removed by inferring whether `sendmail` is desired based on whether `sendmail_command` is set. In that case, setting `smtp_host` and `sendmail_command` should probably be made mutually exclusive.

src/mail.rs
```diff
@@ -496,6 +504,58 @@ pub async fn send_test(address: &str) -> EmptyResult {
...
 }
 
+async fn send_with_selected_transport(email: Message) -> EmptyResult {
+    if CONFIG.use_sendmail() {
+        match sendmail_transport().send(email).await {
+            Ok(_) => Ok(()),
+            // Match some common errors and make them more user friendly
+            Err(e) => {
+                if e.is_client() {
+                    debug!("Sendmail Client error: {:#?}", e);
```
jjlin, line 514: I know it was already like this before, but it's not clear to me why `Client` is arbitrarily capitalized here (same with `Response` and `Encryption`, but not `timeout`). It might be good to normalize this while you're changing this code anyway.

src/config.rs
soruh: Yes, unfortunately it is very hard to validate that a command exists in the `$PATH` (see comment above). If you know of a good way to do this, that'd be great. Especially as the default will very often be the `sendmail` from the `$PATH` which may not exist if `sendmail` is not installed.

src/config.rs
soruh: I feel like actually setting `SENDMAIL_COMMAND` is the exception rather than the rule and enabling sending mails via sendmail by specifying explicitely a `sendmail command` which is just `sendmail` seems very counterintuitive to me...

src/mail.rs
soruh: Fixed in [2f247d8](https://github.com/dani-garcia/vaultwarden/pull/3147/commits/2f247d84a9dc4e401e1eecd24f5b25b63c5b191f)

.env.template
soruh: Fixed in [7cc583c](https://github.com/dani-garcia/vaultwarden/pull/3147/commits/7cc583c37e3e58435f6e1250fac8e83c6d97a956)

src/config.rs
soruh: Fixed in [7cc583c](https://github.com/dani-garcia/vaultwarden/pull/3147/commits/7cc583c37e3e58435f6e1250fac8e83c6d97a956)

src/config.rs
jjlin: I'm not sure validation at startup is really required (the command could disappear after startup anyway, so one would hope there would still be a useful error message if that were to happen). But if you really want to do this check on a non-absolute path, https://crates.io/crates/which would probably do the job.

BlackDex: That is why we determined we want it to be an absolute path.
And i think that validation at startup is wanted, we check a lot to prevent issues later on. That someone can remove or uninstall the binary is in the there responsibility. But if we can prevent issued at startup i would prefer that, it's just a simple check during validation, thats all.

src/config.rs
BlackDex: I agree with @soruh. The option to configure the sendmail binary isn't mandatory. We could also say, we only support the sendmail command and thats it. And people need to make sure it's located somewhere within the $PATH and be done with it, so not checking for a binary would be my preferred option instead of the other way around.

Also, stating to use this and not validating if the command exist as you mentioned before sounds a bit strange to me. That would be counter productive in my opinion.

src/config.rs
soruh: I mean it _is_ somewhat unfortunate that we can't check if a `sendmail` client is actually installed when the user sets `USE_SENDMAIL=true` but no `SENDMAIL_COMMAND` which will be the most common use case... The [which](https://crates.io/crates/which) crate does seem pretty lightweight, but I really can't make the decision if an extra dependency is worth it here...

The alternative of just trying to run the command and check if we get `ENOENT` seems mostly reasonable but since we don't know the implementation specific detail of what the installed sendmail client will do when called without arguments it feels pretty janky to me...

soruh:
I have just pushed [82452b3](https://github.com/dani-garcia/vaultwarden/pull/3147/commits/82452b3b2f8b49275f99e0048ce50217088be484) which adds the check using the `which` crate. If there is a consensus on what to do we can either revert it for the old behavior or keep it.

soruh:
I feel like it would be unfortunate for this to get stuck on something that feels pretty minor to me. I would really like this to be in the next version when it is released as I need it myself :). Is there a way a consensus can be found for this? Maybe a compromise of not using the which crate for now, but having the configuration description be more adamant about having `sendmail` installed would be acceptable?

BlackDex:
I might have some time to check this this evening (UTC). But you do need to rebase this PR it looks like.

soruh:
Do you know if there is a way for github to automatically squash PR commits or should I clean up the commit history a bit?

BlackDex:
Not that i know. I do stuff like that via the cli with a `reset --soft`.

BlackDex (changes requested):

src/config.rs
```diff
@@ -742,27 +746,62 @@ fn validate_config(cfg: &ConfigItems) -> Result<(), Error> {
...
         }
 
         if cfg._enable_email_2fa && cfg.email_token_size < 6 {
             err!("`EMAIL_TOKEN_SIZE` has a minimum size of 6")
         }
     }
 
+    if cfg._enable_email_2fa && !(cfg.smtp_host.is_none() || cfg.use_sendmail) {
+        err!("To enable email 2FA, a mail transport must be configured")
+    }
```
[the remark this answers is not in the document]
soruh: You're absolutely correct, should be fixed now

src/config.rs
```diff
@@ -742,27 +746,62 @@ fn validate_config(cfg: &ConfigItems) -> Result<(), Error> {
...
-            err!("SMTP_FROM does not contain a mandatory @ sign")
-        }
+            let mut path = std::path::PathBuf::from(command);
+
+            if !path.is_absolute() {
+                match which::which(command) {
+                    Ok(result) => path = result,
+                    Err(_) => err!(format!("sendmail command {command:?} not found in $PATH")),
+                }
+            }
```
BlackDex, line 759: While i do not think Windows builds will be there much (if at all), and also want to use sendmail, then you need to fix the executable here, since Windows has `.exe` as a suffix. I suggest to use https://doc.rust-lang.org/std/env/consts/constant.EXE_SUFFIX.html which would be set to `.exe` when build for Windows, and should be empty for Linux.

`which` seems to work on a Windows build, so that should be fine.

BlackDex (approved):

src/config.rs
BlackDex, line 759: While i do not think Windows builds will be there much (if at all), and also want to use sendmail, then you need to fix the executable here, since Windows has `.exe` as a suffix. I suggest to use https://doc.rust-lang.org/std/env/consts/constant.EXE_SUFFIX.html which would be set to `.exe` when build for Windows, and should be empty for Linux.

`which` seems to work on a Windows build, so that should be fine.

BlackDex (changes requested):

src/config.rs
soruh: [50bb24e](https://github.com/dani-garcia/vaultwarden/pull/3147/commits/50bb24e34c534a2b8693bdc8ddf77d8f57a0e566) does a bit more copying than strictly necessary, but since it's only run once. I think the readability outweighs using Cow etc. here.

BlackDex: And it's run during config changes, but that is also not done a lot.

BlackDex (approved):

soruh:
Thank's alot for your help on this @BlackDex!

dani-garcia (approved):

soruh:
Curious that git thinks there is a conflict here. I think it assumes my new function is the same as the function added on main. I've rebased it now...

dani-garcia:
Yeah, I had trouble with git conflicts while trying to batch merge all the PR's manually, not sure why. Ended up merging them one by one.

We've just relicensed the project as AGPL but this PR was started before the license change was merged, so i'd appreciate if you can add your comment to the relicensing agreement, even if it's just a small formality by now:
https://github.com/dani-garcia/vaultwarden/discussions/2450

Thanks for your contribution!

soruh:
[done](https://github.com/dani-garcia/vaultwarden/discussions/2450#discussioncomment-4950651)

P.S.: It should probably say:
```patch
-And i hereby grant the Vaultwarden project team to re-license Vaultwarden
+And I hereby grant the Vaultwarden project team permission to re-license Vaultwarden
```

PR #3163, opened 2023-01-24
Update KDF Configuration and processing

BlackDex:
- Change default Password Hash KDF Storage from 100_000 to 350_000 iterations
- Update Password Hash when the default iteration value is different
- Validate password_iterations
- Validate client-side KDF to prevent it from being set lower than 100_000


src/config.rs
```diff
@@ -465,7 +465,7 @@ make_config! {
         emergency_access_allowed:    bool,   true,   def,    true;
         /// Password iterations |> Number of server-side passwords hashing iterations.
         /// The changes only apply when a user changes their password. Not recommended to lower the value
-        password_iterations:    i32,    true,   def,    100_000;
+        password_iterations:    i32,    true,   def,    350_000;
```
drizuid, line 468: OWASP updated the recommendation from 350,000 to 600,000 for PBKDF2-HMAC-SHA256 at the end of December, 2022.

BlackDex: I understand that. But if you want that  you can change it your self.
We also need to keep in mind that they changed to because of the current hardware specs.
Since Vaultwarden is also able to run on very low spec hardware, we need to at least have the default lower to provide a better experience out-of-the-box.

But, that doesn't prevent you from increasing this value for your own server instance.

src/config.rs
```diff
@@ -465,7 +465,7 @@ make_config! {
         emergency_access_allowed:    bool,   true,   def,    true;
         /// Password iterations |> Number of server-side passwords hashing iterations.
         /// The changes only apply when a user changes their password. Not recommended to lower the value
```
tessus, line 467: This comment is not valid anymore
```suggestion
        /// The changes apply when a user logs in. The minimum allowed value is 100000.
```

src/config.rs
mweinelt: I suggest we follow the outcome of https://github.com/bitwarden/mobile/pull/2305 wrt usability.

src/config.rs
BlackDex: Actually, that comment was never valid.

BlackDex: Thanks, i have changed this at the `config.rs` and `.env.template` file.

.env.template
```diff
@@ -298,9 +298,9 @@
 ## This setting applies globally to all users.
 # INCOMPLETE_2FA_TIME_LIMIT=3
 
-## Controls the PBBKDF password iterations to apply on the server
-## The change only applies when the password is changed
-# PASSWORD_ITERATIONS=100000
+## Number of server-side passwords hashing iterations for the password hash.
+## The default for new users, and it will update during login for existing users if changed.
```
tessus, line 302:
```suggestion
## The default for new users. It will be updated during login for existing users, if changed.
```

Sorry for the nitpicking.

src/config.rs
```diff
@@ -463,9 +463,9 @@ make_config! {
         invitation_expiration_hours: u32, false, def, 120;
         /// Allow emergency access |> Controls whether users can enable emergency access to their accounts. This setting applies globally to all users.
         emergency_access_allowed:    bool,   true,   def,    true;
-        /// Password iterations |> Number of server-side passwords hashing iterations.
-        /// The changes only apply when a user changes their password. Not recommended to lower the value
-        password_iterations:    i32,    true,   def,    100_000;
+        /// Password iterations |> Number of server-side passwords hashing iterations for the password hash.
+        /// The default for new users, and it will update during login for existing users if changed.
```
tessus, line 467:
```suggestion
        /// The default for new users. It will be updated during login for existing users, if changed.
```

src/config.rs
BlackDex: Well, it looks they also changed it on the client side to 600_000 already via https://github.com/bitwarden/clients/pull/4546.
Ill adjust this PR to match.

src/config.rs
BlackDex: Thanks for the grammar check!

PR #3164, opened 2023-01-24
Remove `arm32v6`-specific tag

jjlin:
This section of code seems to be breaking the Docker release workflow as of a few days ago, though it's unclear why. This tag only existed to work around an issue with Docker pulling the wrong image for ARMv6 platforms; that issue was resolved in Docker 20.10.0, which has been out for a few years now, so it seems like a reasonable time to drop this tag.


dani-garcia:
Thanks!

PR #3165, opened 2023-01-24
the default is PASSWORD_ITERATIONS=600000

tessus:
The 350000 in the `.env` file was an oversight.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -300,7 +300,7 @@
 
 ## Number of server-side passwords hashing iterations for the password hash.
 ## The default for new users. If changed, it will be updated during login for existing users.
-# PASSWORD_ITERATIONS=350000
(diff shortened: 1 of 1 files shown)


tessus:
I know, but this file shows the default values. But in this case it shows the wrong one.

It's not a bug, but an inconsistency.

BlackDex:
> I know, but this file shows the default values. But in this case it shows the wrong one.
> 
> It's not a bug, but an inconsistency.

No, that file doesn't show the defaults

tessus:
Ok. my bad. I thought it shows the default values for the settings.

BlackDex:
> Ok. my bad. I thought it shows the default values for the settings.

It currently doesn't. It might be ok to do that though, but then i suggest to check all values and create a different PR 🙂

Issue #2937, opened 2022-11-28
Very large notes cause problems


przemub:
### Subject of the issue

Thank you so much for this re-implementation, I love it!

When I was bored, I created a secure note, of approximately 15 MB size, consisting of repeating 4-character string. I used the Firefox extension.

I observed the following problems:
* Searching for anything hanged the clients for ~10 seconds, no matter the client
* It becomes impossible to use the official iOS client (you get a Syncing failed. message with no further info)
* The vault can be exported to JSON, but not imported to the official Bitwarden server (exceeded maximum note length).

Actually, the last problem was how I discovered what caused this :)

Probably, some limit is needed, or maybe not searching inside the whole note?

### Your environment (Generated via diagnostics page)
* Vaultwarden version: v1.26.0
* Web-vault version: v2022.10.0
* Running within Docker: true (Base: Debian)
* Environment settings overridden: false
* Uses a reverse proxy: true
* IP Header check: true (X-Real-IP)
* Internet access: true
* Internet access via a proxy: false
* DNS Check: true
* Time Check: true
* Domain Configuration Check: false
* HTTPS Check: false
* Database type: SQLite
* Database version: 3.35.4
* Clients used: iOS, Firefox, macOS
* Reverse proxy and version: nginx 1.18
* Other relevant information: 

### Config (Generated via diagnostics page)
<details><summary>Show Running Config</summary>

**Environment settings which are overridden:** 

```json
{
  "_duo_akey": null,
  "_enable_duo": false,
  "_enable_email_2fa": false,
  "_enable_smtp": true,
  "_enable_yubico": true,
  "_icon_service_csp": "",
  "_icon_service_url": "",
  "_ip_header_enabled": true,
  "admin_ratelimit_max_burst": 3,
  "admin_ratelimit_seconds": 300,
  "admin_token": "***",
  "allowed_iframe_ancestors": "",
  "attachments_folder": "data/attachments",
  "authenticator_disable_time_drift": false,
  "data_folder": "data",
  "database_conn_init": "",
  "database_max_conns": 10,
  "database_timeout": 30,
  "database_url": "****/**.*******",
  "db_connection_retries": 15,
  "disable_2fa_remember": false,
  "disable_admin_token": false,
  "disable_icon_download": false,
  "domain": "****://*********",
  "domain_origin": "****://*********",
  "domain_path": "",
  "domain_set": false,
  "duo_host": null,
  "duo_ikey": null,
  "duo_skey": null,
[...]

PR #3168, opened 2023-01-25
Increase secure notes length to match LastPass.

STRML:
As users are migrating from LastPass, I am seeing more and more people hitting the max length of 10,000 on the "Notes" field, which applies to both CipherType.Login and CipherType.SecureNote.

This default is too low, and by being significantly lower than competing products, needlessly complicates the transition.

This is not helped by how unhelpful the import process is on web (showing errors in console only, silently dropping items), which is a PR for another day. By merging this, we can ensure that one of the most common drop-offs is fixed.

45,000 matches LastPass. As we know that the encrypted length can be longer than the raw text, I have increased this to 50,000 so there is a healthy buffer to account for this inflation.

Mirrors PR at https://github.com/bitwarden/server/pull/2625

### DIFF
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -380,8 +380,8 @@ pub async fn update_cipher_from_data(
     }
 
     if let Some(note) = &data.Notes {
-        if note.len() > 10_000 {
-            err!("The field Notes exceeds the maximum encrypted value length of 10000 characters.")
+        if note.len() > 50_000 {
+            err!("The field Notes exceeds the maximum encrypted value length of 50000 characters.")
         }
     }
 
--- a/src/db/models/cipher.rs
+++ b/src/db/models/cipher.rs
@@ -82,11 +82,11 @@ impl Cipher {
         let mut validation_errors = serde_json::Map::new();
         for (index, cipher) in cipher_data.iter().enumerate() {
             if let Some(note) = &cipher.Notes {
-                if note.len() > 10_000 {
+                if note.len() > 50_000 {
                     validation_errors.insert(
                         format!("Ciphers[{index}].Notes"),
                         serde_json::to_value([
-                            "The field Notes exceeds the maximum encrypted value length of 10000 characters.",
+                            "The field Notes exceeds the maximum encrypted value length of 50000 characters.",
                         ])
                         .unwrap(),
                     );


BlackDex:
While I'm all for increasing this limit, I'm a bit hesitant to do so because of #2937. That's why we added a limit in the first place actually.

Also, on this part i would like to keep compatibiliteit with Bitwarden in regards to export from Vaultwarden and import to Bitwarden.

So, if your PR gets accepted at Bitwarden, i see no reason not to increase it here either.

STRML:
Fair, thanks for the input. I would suspect #2937 is not going to be such an issue with 50k notes versus 15M, given the 2+ order of magnitude size difference. But let's wait and see what happens at Bitwarden, as I don't want to create a compatibility issue.

STRML:
Honestly it's pretty frustrating, because it doesn't look like that solution is coming any time soon, but migrations into bw/vw happen every day. 

It's such a common mistake to let the perfect be the enemy of the good.

tessus:
Send me an email and I'll let you know what I think. I will not post this publicly.

stefan0xC:
Bitwarden server has increased the limit to `35000` in https://github.com/bitwarden/server/pull/3193

STRML:
Great, just matching the official server here will make migration much easier. 

Although, arguably, the ship of mass migration has already sailed.

BlackDex:
> Bitwarden has apparently increased the encrypted string length to `35000` in [bitwarden/server#3193](https://github.com/bitwarden/server/pull/3193) (which should be relevant to this issue)

I Would suggest we update to the same value.

jjlin:
> Bitwarden has apparently increased the encrypted string length to `35000` in [bitwarden/server#3193](https://github.com/bitwarden/server/pull/3193) (which should be relevant to this issue)

That's for their secrets manager, not for the password manager secure notes.

stefan0xC:
The client seems to restrict the note length to 25,000 characters: https://github.com/bitwarden/clients/pull/6030
Not sure if we implement this differently but 35k should be safe, shouldn't it?

stefan0xC:
> That's for their secrets manager, not for the password manager secure notes.

Ah, thanks. I'm apparently tired. I was wondering why they said 25000 ;)

Jayd603:
So... I ran into this issue.   We wanted to store notes that others can edit and they happen to be larger than 10k characters.  We can do an attachment for now but it creates extra steps for people, ie, download file, edit, upload again.   Could this possibly be user configurable to an extent?   What is the max we can do without it becoming an issue?    I may experiment with it myself and will report back if I find anything interesting.

Jayd603:
Bump - looks like this is the same size still.  10k seems too restrictive for a secure note on modern hardware.  The Bitwarden request was rejected but it sounds like the issue is with their back-end storage/architecture due to the large number of users they have on their hosted platform.  I guess i don't care as long as it remains a simple patch for us, just one more step during deployment, but other users may want larger notes without having to patch/build a docker image etc.

mwa85:
I can't migrate from LastPass do to this limit so I'll like this limit to be lifted and be the same as LastPass to make migration (importing the LastPass-export) possible

Thanks.

tessus:
I don't think this PR will ever be merged, because the devs want vw to behave the same way as bw, even if bw's design is idiotic and detrimental to a successful migration from Lastpass.

However, you can always compile your own version that includes this PR.

STRML:
It is infuriating - exactly the kind of stuff that prevents repeat contribution.

tessus:
@STRML in most cases it makes sense not to divert from bw's design. But the reasoning for not adding this escapes me.

Currently people can neither migrate fro Lastpass to bw or vw. (if they are using longer notes.) With this change they could migrate to vw. After that migrating from vw to bw would fail the same as it would from Lastpass to bw, so nothing new here.

The problem could be that the bw clients might not support longer notes in the future. But that's neither confirmed nor on the roadmap.

STRML:
I don't have a business degree, but refusing to provide a hassle-free path for your competitors' users to migrate to your product feels like a miss.

If I were at BW this would have been my top priority as soon as the LP drama started.

BlackDex:
We have been talking about this PR.
And increasing this limit to `50_000` might still be to low for some users maybe.
Making it variable is also not something we would like to do. Maybe a skip length check is an option to add.

But our main concern is you will lose the ability to export from Vaultwarden, and import to Bitwarden (Or other password managers even).
Also, it could be that the clients in the future might break or fail to sync if the notes are too large, or unable to process updates for these items.

encrypting and decrypting those notes when they become larger and larger also mean a client needs more processing power. And, when increasing your KDF settings, that might also cause extra load of course.

An other option of course is using attachments, but i can understand that attachments are not easy to import/export.

What would you prefer to have done?
I my self am for just leaving it as-is and keep in-sync with Bitwarden.

Jayd603:
> What would you prefer to have done? I my self am for just leaving it as-is and keep in-sync with Bitwarden.

Good question, perhaps some type of option to allow larger note sizes with a warning that it may break export/import into bitwarden etc.    Maybe down the road vaultwarden can stray away from bitwarden in some cases with a BITWARDEN strict compatibility option and when strict mode is on, none of the vaultwarden specific changes will work/build.  This is a question for the core devs here and what their goals/vision are, if down the road they want to break away from bitwarden more or just keep it as a near mirror of bitwarden. My increase to 100k didn't seem to break the clients we use and it performed great for us , the main issue we had with using attachments is we are allowing developers to edit environment files for our devops platform, environment files are stored as secure notes and pulled by jenkins etc.   Editing a secure note is much much easier for them rather than needing to download a file, edit it, then upload it etc.   there is much more room for error that way too.

STRML:
My only goal here was to provide a much more reasonable max length (so that, say, RSA keys and certificate chains can easily be put into a Note) and to ease a transition from LP.

Rather than overthink it, I'd say just choose the bigger number (100,000) as there is little chance this will cause any processing power issues and is big enough to fit most reasonable data, and will fix the LP import issue.

PR #3170, opened 2023-01-25
Allow listening on privileged ports (below 1024) as non-root

jjlin:
This is done by running `setcap cap_net_bind_service=+ep` on the executable in the build stage (doing it in the runtime stage creates an extra copy of the executable that bloats the image). This only works when using the BuildKit-based builder, since the `COPY` instruction doesn't copy capabilities on the legacy builder.


jjlin:
Note that executables running as non-root actually can already bind to "privileged" ports even without setting `cap_net_bind_service` in [Docker 20.10.0](https://docs.docker.com/engine/release-notes/#security-2) and later (see moby/moby#41030).

Nevertheless, setting `cap_net_bind_service` should still be useful for people running earlier versions of Docker, or using another container runtime that doesn't have this change.

docker/Dockerfile.j2
```diff
@@ -93,32 +91,33 @@ ENV DEBIAN_FRONTEND=noninteractive \
...
     USER="root"
 
-
 # Create CARGO_HOME folder and don't download rust docs
 RUN {{ mount_rust_cache -}} mkdir -pv "${CARGO_HOME}" \
     && rustup set profile minimal
 
 {% if "alpine" in target_file %}
+# Install build dependencies
+RUN apt-get update && apt-get install -y --no-install-recommends libcap2-bin
```
jjlin, line 100: This could be avoided if `libcap2-bin` were included in the `blackdex/rust-musl` images.

jjlin:
https://hub.docker.com/r/jjlin/vaultwarden/tags shows the results of this PR

docker/Dockerfile.j2
BlackDex: Sure happy to add.

BlackDex: @jjlin It's currently building the images: https://github.com/BlackDex/rust-musl/actions/runs/4004773326

jjlin: Removed in 1f26241.

BlackDex (approved):

dani-garcia (approved):

PR #3210, opened 2023-02-01
add argon2 kdf fields

tessus:
Changes:

- added columns to db schema with DEFAULT NULL
- added a pub enum `UserKdfType`
- if UserKdfType is Argon2id, server responses will include 2 more KVPs for memory and parallelism

Test plan:

- used quexten's webvault (which includes the UI changes for Argon2) to switch KDF to argon2 and back to pbkdf2

<details>
<summary>Previous text (for reference)</summary>

This is still a work in progess. The following items have to be clarified first:

- default db values or NULL?
- default iterations might have to be changed depending on the type (PBKDF2, ARGON2)
- enum for kdf type?

During compilation I get an error:

```
error: recursion limit reached while expanding `__static_cond!`

consider increasing the recursion limit by adding a `#![recursion_limit = "194"]` attribute to your crate (`vaultwarden`)
```

However in the code it says:

```
// The recursion_limit is mainly triggered by the json!() macro.
// The more key/value pairs there are the more recursion occurs.
// We want to keep this as low as possible, but not higher then 128.
// If you go above 128 it will cause rust-analyzer to fail,
```

So, what now?
</details>

### DIFF
--- a/src/api/core/accounts.rs
+++ b/src/api/core/accounts.rs
@@ -49,6 +49,8 @@ pub struct RegisterData {
     Email: String,
     Kdf: Option<i32>,
     KdfIterations: Option<i32>,
+    KdfMemory: Option<i32>,
+    KdfParallelism: Option<i32>,
     Key: String,
     Keys: Option<KeysData>,
     MasterPasswordHash: String,
@@ -153,13 +155,16 @@ pub async fn _register(data: JsonUpcase<RegisterData>, mut conn: DbConn) -> Json
     // Make sure we don't leave a lingering invitation.
     Invitation::take(&email, &mut conn).await;
 
+    if let Some(client_kdf_type) = data.Kdf {
+        user.client_kdf_type = client_kdf_type;
+    }
+
     if let Some(client_kdf_iter) = data.KdfIterations {
         user.client_kdf_iter = client_kdf_iter;
     }
 
-    if let Some(client_kdf_type) = data.Kdf {
-        user.client_kdf_type = client_kdf_type;
-    }
+    user.client_kdf_parallelism = data.KdfMemory;
+    user.client_kdf_memory = data.KdfParallelism;
 
     user.set_password(&data.MasterPasswordHash, Some(data.Key), true, None);
     user.password_hint = password_hint;
@@ -337,6 +342,8 @@ async fn post_password(
 struct ChangeKdfData {
     Kdf: i32,
     KdfIterations: i32,
+    KdfMemory: Option<i32>,
+    KdfParallelism: Option<i32>,
 
     MasterPasswordHash: String,
     NewMasterPasswordHash: String,
@@ -352,10 +359,31 @@ async fn post_kdf(data: JsonUpcase<ChangeKdfData>, headers: Headers, mut conn: D
         err!("Invalid password")
     }
 
-    if data.KdfIterations < 100_000 {
-        err!("KDF iterations lower then 100000 are not allowed.")
+    if data.Kdf == UserKdfType::Pbkdf2 as i32 && data.KdfIterations < 100_000 {
+        err!("PBKDF2 KDF iterations must be at least 100000.")
     }
 
+    if data.Kdf == UserKdfType::Argon2id as i32 {
+        if data.KdfIterations < 1 {
+            err!("Argon2 KDF iterations must be at least 1.")
+        }
+        if let Some(m) = data.KdfMemory {
+            if !(15..=1024).contains(&m) {
+                err!("Argon2 memory must be between 15 MB and 1024 MB.")
+            }
+            user.client_kdf_memory = data.KdfMemory;
+        } else {
+            err!("Argon2 memory parameter is required.")
+        }
+        if let Some(p) = data.KdfParallelism {
+            if !(1..=16).contains(&p) {
+                err!("Argon2 parallelism must be between 1 and 16.")
+            }
+            user.client_kdf_parallelism = data.KdfParallelism;
+        } else {
+            err!("Argon2 parallelism parameter is required.")
+        }
+    }
     user.client_kdf_iter = data.KdfIterations;
     user.client_kdf_type = data.Kdf;
     user.set_password(&data.NewMasterPasswordHash, Some(data.Key), true, None);
@@ -770,15 +798,22 @@ async fn prelogin(data: JsonUpcase<PreloginData>, conn: DbConn) -> Json<Value> {
 pub async fn _prelogin(data: JsonUpcase<PreloginData>, mut conn: DbConn) -> Json<Value> {
     let data: PreloginData = data.into_inner().data;
 
-    let (kdf_type, kdf_iter) = match User::find_by_mail(&data.Email, &mut conn).await {
-        Some(user) => (user.client_kdf_type, user.client_kdf_iter),
-        None => (User::CLIENT_KDF_TYPE_DEFAULT, User::CLIENT_KDF_ITER_DEFAULT),
+    let (kdf_type, kdf_iter, kdf_mem, kdf_para) = match User::find_by_mail(&data.Email, &mut conn).await {
+        Some(user) => (user.client_kdf_type, user.client_kdf_iter, user.client_kdf_memory, user.client_kdf_parallelism),
+        None => (User::CLIENT_KDF_TYPE_DEFAULT, User::CLIENT_KDF_ITER_DEFAULT, None, None),
     };
 
-    Json(json!({
+    let mut result = json!({
         "Kdf": kdf_type,
-        "KdfIterations": kdf_iter
-    }))
+        "KdfIterations": kdf_iter,
+    });
+
+    if kdf_type == UserKdfType::Argon2id as i32 {
+        result["KdfMemory"] = Value::Number(kdf_mem.expect("Argon2 memory parameter is required.").into());
+        result["KdfParallelism"] = Value::Number(kdf_para.expect("Argon2 parallelism parameter is required.").into());
+    }
+
+    Json(result)
 }
 
 // https://github.com/bitwarden/server/blob/master/src/Api/Models/Request/Accounts/SecretVerificationRequestModel.cs
--- a/src/db/models/user.rs
+++ b/src/db/models/user.rs
@@ -44,6 +44,8 @@ db_object! {
 
         pub client_kdf_type: i32,
         pub client_kdf_iter: i32,
+        pub client_kdf_memory: Option<i32>,
+        pub client_kdf_parallelism: Option<i32>,
 
         pub api_key: Option<String>,
 
@@ -56,6 +58,11 @@ db_object! {
     pub struct Invitation {
         pub email: String,
     }
+}
+
+pub enum UserKdfType {
+    Pbkdf2 = 0,
+    Argon2id = 1,
 }
 
 enum UserStatus {
@@ -73,7 +80,7 @@ pub struct UserStampException {
 
 /// Local methods
 impl User {
-    pub const CLIENT_KDF_TYPE_DEFAULT: i32 = 0; // PBKDF2: 0
+    pub const CLIENT_KDF_TYPE_DEFAULT: i32 = UserKdfType::Pbkdf2 as i32;
     pub const CLIENT_KDF_ITER_DEFAULT: i32 = 600_000;
 
     pub fn new(email: String) -> Self {
@@ -113,6 +120,8 @@ impl User {
 
             client_kdf_type: Self::CLIENT_KDF_TYPE_DEFAULT,
             client_kdf_iter: Self::CLIENT_KDF_ITER_DEFAULT,
+            client_kdf_memory: None,
+            client_kdf_parallelism: None,
 
             api_key: None,
 
--- a/src/db/schemas/mysql/schema.rs
+++ b/src/db/schemas/mysql/schema.rs
@@ -199,6 +199,8 @@ table! {
         excluded_globals -> Text,
         client_kdf_type -> Integer,
         client_kdf_iter -> Integer,
+        client_kdf_memory -> Nullable<Integer>,
+        client_kdf_parallelism -> Nullable<Integer>,
         api_key -> Nullable<Text>,
         avatar_color -> Nullable<Text>,
     }
--- a/src/db/schemas/mysql/schema.rs
+++ b/src/db/schemas/mysql/schema.rs
@@ -224,6 +224,7 @@ table! {
         akey -> Text,
         status -> Integer,
         atype -> Integer,
+        reset_password_key -> Nullable<Text>,
     }
 }
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,7 +48,10 @@ jobs:
         ports:
           - 5000:5000
     env:
-      DOCKER_BUILDKIT: 1 # Disabled for now, but we should look at this because it will speedup building!
+      # Use BuildKit (https://docs.docker.com/build/buildkit/) for better
+      # build performance and the ability to copy extended file attributes
+      # (e.g., for executable capabilities) across build phases.
+      DOCKER_BUILDKIT: 1
       # DOCKER_REPO/secrets.DOCKERHUB_REPO needs to be 'index.docker.io/<user>/<repo>'
       DOCKER_REPO: ${{ secrets.DOCKERHUB_REPO }}
       SOURCE_COMMIT: ${{ github.sha }}
--- a/.hadolint.yaml
+++ b/.hadolint.yaml
@@ -3,5 +3,7 @@ ignored:
   - DL3008
   # disable explicit version for apk install
   - DL3018
+  # disable check for consecutive `RUN` instructions
+  - DL3059
 trustedRegistries:
   - docker.io
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ resolver = "2"
 
 repository = "https://github.com/dani-garcia/vaultwarden"
 readme = "README.md"
-license = "GPL-3.0-only"
+license = "AGPL-3.0-only"
 publish = false
 build = "build.rs"
 
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,5 +1,5 @@
-                    GNU GENERAL PUBLIC LICENSE
-                       Version 3, 29 June 2007
+                    GNU AFFERO GENERAL PUBLIC LICENSE
+                       Version 3, 19 November 2007
 
  Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
  Everyone is permitted to copy and distribute verbatim copies
@@ -7,17 +7,15 @@
 
                             Preamble
 
-  The GNU General Public License is a free, copyleft license for
-software and other kinds of works.
+  The GNU Affero General Public License is a free, copyleft license for
+software and other kinds of works, specifically designed to ensure
+cooperation with the community in the case of network server software.
 
   The licenses for most software and other practical works are designed
 to take away your freedom to share and change the works.  By contrast,
-the GNU General Public License is intended to guarantee your freedom to
+our General Public Licenses are intended to guarantee your freedom to
 share and change all versions of a program--to make sure it remains free
-software for all its users.  We, the Free Software Foundation, use the
-GNU General Public License for most of our software; it applies also to
-any other work released this way by its authors.  You can apply it to
-your programs, too.
+software for all its users.
 
   When we speak of free software, we are referring to freedom, not
 price.  Our General Public Licenses are designed to make sure that you
@@ -26,44 +24,34 @@ them if you wish), that you receive source code or can get it if you
 want it, that you can change the software or use pieces of it in new
 free programs, and that you know you can do these things.
 
-  To protect your rights, we need to prevent others from denying you
-these rights or asking you to surrender the rights.  Therefore, you have
-certain responsibilities if you distribute copies of the software, or if
-you modify it: responsibilities to respect the freedom of others.
+  Developers that use our General Public Licenses protect your rights
+with two steps: (1) assert copyright on the software, and (2) offer
+you this License which gives you legal permission to copy, distribute
+and/or modify the software.
 
-  For example, if you distribute copies of such a program, whether
-gratis or for a fee, you must pass on to the recipients the same
-freedoms that you received.  You must make sure that they, too, receive
-or can get the source code.  And you must show them these terms so they
-know their rights.
+  A secondary benefit of defending all users' freedom is that
+improvements made in alternate versions of the program, if they
+receive widespread use, become available for other developers to
+incorporate.  Many developers of free software are heartened and
+encouraged by the resulting cooperation.  However, in the case of
+software used on network servers, this result may fail to come about.
+The GNU General Public License permits making a modified version and
+letting the public access it on a server without ever releasing its
+source code to the public.
 
-  Developers that use the GNU GPL protect your rights with two steps:
-(1) assert copyright on the software, and (2) offer you this License
-giving you legal permission to copy, distribute and/or modify it.
+  The GNU Affero General Public License is designed specifically to
+ensure that, in such cases, the modified source code becomes available
+to the community.  It requires the operator of a network server to
+provide the source code of the modified version running there to the
+users of that server.  Therefore, public use of a modified version, on
+a publicly accessible server, gives the public access to the source
+code of the modified version.
 
-  For the developers' and authors' protection, the GPL clearly explains
-that there is no warranty for this free software.  For both users' and
-authors' sake, the GPL requires that modified versions be marked as
-changed, so that their problems will not be attributed erroneously to
-authors of previous versions.
-
-  Some devices are designed to deny users access to install or run
-modified versions of the software inside them, although the manufacturer
-can do so.  This is fundamentally incompatible with the aim of
-protecting users' freedom to change the software.  The systematic
-pattern of such abuse occurs in the area of products for individuals to
-use, which is precisely where it is most unacceptable.  Therefore, we
-have designed this version of the GPL to prohibit the practice for those
-products.  If such problems arise substantially in other domains, we
-stand ready to extend this provision to those domains in future versions
-of the GPL, as needed to protect the freedom of users.
-
-  Finally, every program is threatened constantly by software patents.
-States should not allow patents to restrict development and use of
-software on general-purpose computers, but in those that do, we wish to
-avoid the special danger that patents applied to a free program could
-make it effectively proprietary.  To prevent this, the GPL assures that
-patents cannot be used to render the program non-free.
+  An older license, called the Affero General Public License and
+published by Affero, was designed to accomplish similar goals.  This is
+a different license, not a version of the Affero GPL, but Affero has
+released a new version of the Affero GPL which permits relicensing under
+this license.
 
   The precise terms and conditions for copying, distribution and
 modification follow.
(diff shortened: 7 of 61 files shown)


quexten:
`default db values or NULL?`

In the official server, kdfmemory and kdfparallelism are null if pbkdf2 is selected, and not null when argon2 is selected. If they are null, the fields can (should) be omitted from the API responses.

`default iterations might have to be changed depending on the type (PBKDF2, ARGON2)`
Currently, PBKDF2 is default anyways for the clients. The only way they can get argon2 is to create an account, log in, and then change to argon2.

BlackDex:
I'd rather not have any defaults set into the database. Also, as @quexten mentioned the `NULL` is probably better.
Regarding the recursions, just up it +1 at a time until that message disappears, Rust just mentions a number mostly twice as high.

quexten:
Some checks such as :
https://github.com/dani-garcia/vaultwarden/blob/9366e3145231034040d7d2e0216070163d6d889f/src/config.rs#L676-L679

Need adjusting to distinguish between pbkdf2 and argon2.

`Right, but when someone switches to argon2, I rather not have 600,000 iterations... I haven't seen the UI yet, thus I don't know what the UI will do. e.g. will it automatically change the iterations to another value or will the user have to set the iterations manually?`

The default iterations are set in the client, not on the server. I think they should only matter on the server in case the client omits the values:

https://github.com/dani-garcia/vaultwarden/blob/9366e3145231034040d7d2e0216070163d6d889f/src/api/core/accounts.rs#L773-L778

quexten:
By the way, the limits should be:
0 < iterations
14 < memory < 1025
0 < parallelism < 17

The defaults are:
iterations: 3
memory: 64 (mb)
parallelism: 4

https://github.com/bitwarden/server/blob/cb1ba50ce26ce33b2a5acf30536a2075e4fadebd/src/Core/Utilities/KdfSettingsValidator.cs#L10-L36

quexten:
Oh, one more note: The CSP needs to be changed for the webvault. Due to webassembly not supporting hashes for the CSP yet, it requires the "wasm-unsafe-eval" policy in the CSP.

https://github.com/bitwarden/server/commit/522df6e91a937e02fba018a0276bb1561fd40e3d

Since the current CSP does not allow it:

quexten:
> @quexten sorry, I read your edits too late, since I read the email notifications instead. I can change the err messages later.
> 
> I only add memory/parallelism when argon2 is used to the JSON in `_password_login`.
> 
> Let me know, if I should do the same for:
> 
> ```
> _refresh_login
> _api_key_login
> _prelogin
> takeover_emergency_access
> ```
> 
> I can do the rest this evening.



> _refresh_login

Yep, everwhere where KdfType, KdfIterations are sent, KdfMemory and KdfParallelism should now be sent too.

tessus:
@quexten 

> Yep, everwhere where KdfType, KdfIterations are sent, KdfMemory and KdfParallelism should now be sent too.

This is already happening. What I was asking was whether I should also exclude the argon2 values from the JSON response as you have stated here https://github.com/dani-garcia/vaultwarden/pull/3210#issuecomment-1411656951 for the 4 functions I have listed.

I have already done this for one function: https://github.com/dani-garcia/vaultwarden/pull/3210/files#diff-1c5b0c87dec2154a167b89110c637a7a4bc04f59af0b83e8ddba39eb2134518cR254

I can look into this again in 5-6 hours.

quexten:
Oh, sorry misunderstood. I don't think it would cause issues to respond values, since for the pbkdf2 case the wouldn't be used, but since the should be null in that case, it's better to just exclude them from the response (the same way as on the official server). For argon2 of course do send them.

src/api/core/accounts.rs
```diff
@@ -352,12 +364,26 @@ async fn post_kdf(data: JsonUpcase<ChangeKdfData>, headers: Headers, mut conn: D
         err!("Invalid password")
     }
 
-    if data.KdfIterations < 100_000 {
-        err!("KDF iterations lower then 100000 are not allowed.")
+    if data.Kdf == 0 && data.KdfIterations < 100_000 {
+        err!("KDF iterations lower than 100000 are not allowed.")
+    }
+
+    if data.Kdf == 1 {
```
BlackDex, line 371: I would suggest to create an enum for the `Kdf_Type`.
```rust
pub enum KdfType {
  Pbkdf = 0,
  Argon2 = 1,
}
```

This makes it easier to see the difference between the two options, and could also become handy during `match` checks and could prevent missing KDF Types from being handled.

BlackDex (changes requested):
Looks good for your (as far as i know) first shot at Rust :).
Some small items to address.

I haven't validated/run the code it self though.

src/api/core/accounts.rs
quexten: Small note: Probably Pbkdf2 instead of Pbkdf?
In the official server it's:

```Csharp

namespace Bit.Core.Enums;

public enum KdfType : byte
{
    PBKDF2_SHA256 = 0,
    Argon2id = 1
}
```

BlackDex: I updated my example to be a bit more precise. Lets try to match upstream where possible and not getting inconvenient.

tessus:
I am sorry, I had to take care of somehing last evening. I'll try to finish it in the morning.

tessus:
Unfortunstely I have issues with `enum`. Even though it is `pub` it's not recognized in the other files, even though it seems to be imported with 

```
use crate::{
    db::{models::*, DbConn},
}
```

Sorry, I just don't know Rust very well. 

But other than that the code _should_ be fine. Can someone please help me with this enum?

BlackDex:
> Unfortunately I have issues with `enum`. It is a `pub`, but it's not recognized in the other files, even though it seems to be imported with
> 
> ```
> use crate::{
>     db::{models::*, DbConn},
> }
> ```
> 
> Sorry, I just don't know Rust very well.
> 
> But other than that the code _should_ be fine. Can someone please help me with this enum?

I suggest to place that enum in the `src/db/models/user.rs` either before or after the `enum UserStatus`
You also need to update the `src/db/models/mod.rs` and add that enum to the list of the `pub use self::user{...}` list.

That should make it globally available when either `db::{models::*, ..}` is used or `db::models::{UserKdfType, ...}`

tessus:
> I suggest to place that enum in the src/db/models/user.rs either before or after the enum UserStatus

Yes, I have done that already in my last commit.

> You also need to update the src/db/models/mod.rs and add that enum to the list of the pub use self::user{...} list.

This was the missing part. Thank you! I'll push another commit in a sec. At the end, after the review and when it is good to go, I'll squash my commits and force-push.

tessus:
@quexten I think it's ready for a test spin.

tessus:
Seriously, the pipeline failed because of the list not in alphabetical order?
This is ridiculous. 

```
-pub use self::user::{Invitation, User, UserStampException, KdfType};
+pub use self::user::{Invitation, KdfType, User, UserStampException};
```

Anyway, it would make sense to have the format check in the beginning and abend BEFORE building the code....

quexten:
> @quexten I think it's ready for a test spin.

Very nice. I'll test it in a few hours.

quexten:
> > @quexten I think it's ready for a test spin.
> 
> Very nice. I'll test it in a few hours.

@tessus 
Initial test works correctly. For pbkdf2 the fields are omitted. I can switch to argon2 and configure different fields, and the customized fields get sent correctly when logging in.

However, switching back to pbkdf2 does not work. I get "422 Unprocessable Entity"
screenshot


Did not tet database migrations (yet).

quexten:
> > > @quexten I think it's ready for a test spin.
> > 
> > 
> > Very nice. I'll test it in a few hours.
> 
> @tessus Initial test works correctly. For pbkdf2 the fields are omitted. I can switch to argon2 and configure different fields, and the customized fields get sent correctly when logging in.
> 
> However, switching back to pbkdf2 does not work. I get "422 Unprocessable Entity" screenshot
> 
> Did not tet database migrations (yet).

From the log:

`[2023-02-02 14:34:31.440][_][WARN] Data guard `JsonUpcase < ChangeKdfData >` failed: Parse("{\"kdf\":0,\"kdfIterations\":600000,\"masterPasswordHash\":\"CrB8NCVWzgH3A9rXcV0Ks4e1g9wuu7n/aLGImbPJOkM=\",\"newMasterPasswordHash\":\"W6YUqnaEd2ns3TR42KPCrKVZ3CIVvevrTBjXUcV+wKI=\",\"key\":\"2.bK/vVVrmGEQ3UCa78xymkg==|5eZFTrmR2J8A+FoRckzr8d8JlMAB0SSve5ENGSbIJdLkAXi6692MC22fobnJmDpkwYpNaG8YgOeRaaXc0ALuRTtJ4rY/CDY2eKzUVVz8Omw=|TpAPAbtZIC078vr6rsengMlenrqiXTcKB/H19q6CRIg=\"}", Error("missing field `KdfMemory`", line: 1, column: 360)).
`

tessus:
Hmm, not sure why it needs a KdfMemory field when switching back to pbkdf2. The field is in the database, but it won't be in a response when the type is not argon2.

Any chance you can send me the web-vault? http://g0to.ca/drop should do it.

quexten:
> Hmm, not sure why it needs a KdfMemory field when switching back to pbkdf2. The field is in the database, but it won't be in a response when the type is not argon2.
> 
> Any chance you can send me the web-vault? http://g0to.ca/drop should do it.

I think some structure or function in the server code might be expecting the field when parsing the json request. Since it is not sent when using pbkdf2, the server throws an error.
I have uploaded the web vault to the site you linked.

BlackDex:
> Thanks a bunch for the zip. I will look at this error this evening (hopefully). I think I might know what the problem could be (from thinking about the code I changed - I don't have it in front of me right now). But maybe BlackDex will find the issue during the review, who knows. ;-)

Yea, that was very easy :).
https://github.com/dani-garcia/vaultwarden/pull/3210/files#diff-e6ed38cdf0d7a02240d1d2a5d7ad72682905ea9bfb7e4cc2f0edef05f9de2d88R350-R351

Those probably need to be converted to an `Option<>`.

tessus:
Nice, I thought along the same lines. I have a few minutes. I'll fix it now.

quexten:
> Nice, I thought along the same lines. I have a few minutes. I'll fix it now.

Switching back to Pbkdf2 works now.

src/db/models/user.rs
```diff
@@ -58,6 +60,11 @@ db_object! {
     }
 }
 
+pub enum KdfType {
```
BlackDex, line 63: Maybe rename this to `UserKdfType` which makes a bit more more clear throughout the rest of the code.
It also matches the other enums/structs naming.

BlackDex (changes requested):

src/db/models/user.rs
tessus: done

tessus:
Thanks for your help quexten and BlackDex. I am gonna squash and force-push. 

It was an interesting experience, because I am not sure why the compiler doesn't handle certain things and why they have to be specified explicitly:

```
if data.Kdf == UserKdfType::Argon2id as i32 {
        if data.KdfMemory < Some(15) || data.KdfMemory > Some(1024) {
            err!("Argon2 memory must be between 15mb and 1024mb.")
        } 
```

IMO there is no reason to cast the enum to i32. And if such a thing was necessary, I'd rather specify the type in the enum itself, instead of having it to type it in a myriad of places throughout the code.
The same goes for the requirement to use `Some()`. The compiler knows that `KdfMemory` is `Option<>`, it even tells me to use `Some`. So why do I have to specify it? I know, there's also `None`, but as a default the compiler could just pick `Some`.

But I learned a bunch of things, so thanks for letting me dabble with Rust.

BlackDex:
> Thanks for your help quexten and BlackDex. I am gonna squash and force-push.
> 
> It was an interesting experience, because I am not sure why the compiler doesn't handle certain things and why they have to be specified explicitly:
> 
> ```
> if data.Kdf == UserKdfType::Argon2id as i32 {
>         if data.KdfMemory < Some(15) || data.KdfMemory > Some(1024) {
>             err!("Argon2 memory must be between 15mb and 1024mb.")
>         } 
> ```
> 
> IMO there is no reason to cast the enum to `i32`. And if such a thing was necessary, I'd rather specify the type in the enum itself, instead of having to type it in a myriad of places throughout the code. The same goes for the requirement to use `Some()`. The compiler knows that `KdfMemory` is `Option<>`, it even tells me to use `Some`. So why do I have to specify it? I know, there's also `None`, but as a default the compiler could just pick `Some`.
> 
> But I learned a bunch of things, so thanks for letting me dabble with Rust.

`enums` can support more then just a specific type. Setting a type on an enum is possible for exporting as a c-repr for example.
There are also some ways to use derive to make it somewhat compatible without using that, but that makes the code more bloated. Casting it your self is more specific and maybe also safer.

Regarding the some/none. The compiler doesn't know what you want to do exactly. It tries to guess this and provide you with a possible solution, if that is correct or what you want is something else. That part of the code could also be written like this for example.

```rust
        if let Some(kdf_memory) = data.KdfMemory {
            if !(15..1024).contains(kdf_memory) {
            err!("Argon2 memory must be between 15mb and 1024mb.")
        } 
```

What the nice thing of this is that it skips the range check if `KdfMemory` is `None`.
Though, I'm not sure which is better. That would be something to check on https://godbolt.org/ and look at the asm for example.

src/api/core/accounts.rs
tessus: I think this one is resolved as well. 

How is this handled in this project?

I usually always let the person who started the thread/review/conversation resolve the conversation. But in some projects I have seen that the person who implements the changes resolves the conversation.

src/db/models/user.rs
```diff
@@ -44,6 +44,8 @@ db_object! {
 
         pub client_kdf_type: i32,
         pub client_kdf_iter: i32,
+        pub client_kdf_memory: i32,
+        pub client_kdf_parallelism: i32,
```
jjlin, line 48: These should be `Option` to reflect the actual database schema. Some other parts of the code will probably need to be adjusted accordingly

src/db/models/user.rs
```diff
@@ -73,8 +80,10 @@ pub struct UserStampException {
 
 /// Local methods
 impl User {
-    pub const CLIENT_KDF_TYPE_DEFAULT: i32 = 0; // PBKDF2: 0
+    pub const CLIENT_KDF_TYPE_DEFAULT: i32 = UserKdfType::Pbkdf2 as i32;
     pub const CLIENT_KDF_ITER_DEFAULT: i32 = 600_000;
+    pub const CLIENT_KDF_MEMO_DEFAULT: i32 = 19;
+    pub const CLIENT_KDF_PARA_DEFAULT: i32 = 1;
```
jjlin, line 86: Remove; these parameters don't need to be set by default if they're made `Option`

src/db/models/user.rs
```diff
@@ -113,6 +122,8 @@ impl User {
 
             client_kdf_type: Self::CLIENT_KDF_TYPE_DEFAULT,
             client_kdf_iter: Self::CLIENT_KDF_ITER_DEFAULT,
+            client_kdf_memory: Self::CLIENT_KDF_MEMO_DEFAULT,
+            client_kdf_parallelism: Self::CLIENT_KDF_PARA_DEFAULT,
```
jjlin, line 126: Set to `None`

src/db/schemas/mysql/schema.rs
```diff
@@ -199,6 +199,8 @@ table! {
         excluded_globals -> Text,
         client_kdf_type -> Integer,
         client_kdf_iter -> Integer,
+        client_kdf_memory -> Integer,
+        client_kdf_parallelism -> Integer,
```
jjlin, line 203: `Nullable<Integer>`

src/api/core/accounts.rs
```diff
@@ -352,10 +364,28 @@ async fn post_kdf(data: JsonUpcase<ChangeKdfData>, headers: Headers, mut conn: D
...
+            err!("Argon2 parallelism must be between 1 and 16.")
+        }
     }
 
+    if let Some(client_kdf_parallelism) = data.KdfParallelism {
+        user.client_kdf_parallelism = client_kdf_parallelism;
+    }
+    if let Some(client_kdf_memory) = data.KdfMemory {
+        user.client_kdf_memory = client_kdf_memory;
+    }
```
jjlin, line 388:
```suggestion
if data.Kdf == UserKdfType::Argon2id as i32 {
    if data.KdfIterations < 1 {
        err!("Argon2 KDF iterations must be at least 1.")
    }
    if let Some(m) = data.KdfMemory {
        if m < 15 || m > 1024 {
            err!("Argon2 memory must be between 15 MB and 1024 MB.")
        }
        user.client_kdf_memory = data.KdfMemory;
    } else {
        err!("Argon2 memory parameter is required.")
    }
    if let Some(p) = data.KdfParallelism {
        if p < 1 || p > 16 {
            err!("Argon2 parallelism must be between 1 and 16.")
        }
        user.client_kdf_parallelism = data.KdfParallelism;
    } else {
        err!("Argon2 parallelism parameter is required.")
    }
}
```

src/api/core/accounts.rs
```diff
@@ -352,10 +364,28 @@ async fn post_kdf(data: JsonUpcase<ChangeKdfData>, headers: Headers, mut conn: D
         err!("Invalid password")
     }
 
-    if data.KdfIterations < 100_000 {
-        err!("KDF iterations lower then 100000 are not allowed.")
+    if data.Kdf == UserKdfType::Pbkdf2 as i32 && data.KdfIterations < 100_000 {
+        err!("PBKDF2 KDF iterations lower than 100000 are not allowed.")
```
jjlin, line 368:
```suggestion
        err!("PBKDF2 KDF iterations must be at least 100000.")
```
(to make the wording more consistent with the Argon2 case)

src/api/core/accounts.rs
tessus: Ah, nice. Thanks, this looks much better.

src/db/schemas/mysql/schema.rs
tessus: yes, thanks. I did miss changing it.

tessus: although now the `Build / Build and Test msrv (pull_request)` pipeline fails and the error messge tells me nothing, or let me put it that way. I don't understand what the issue is.

BlackDex: You forgot to charge the user db object to match the nullable to be an Option<>

tessus: ah, I only did that later locally. thanks. but I ran into another issue. I'll put that in the other thread though.

src/db/models/user.rs
tessus: I don't know how to fix the resulting compile error:

```
error[E0277]: the trait bound `serde_json::Number: std::convert::From<std::option::Option<i32>>` is not satisfied
   --> src/api/identity.rs:115:45
    |
115 |         result["KdfMemory"] = Value::Number(user.client_kdf_memory.into());
    |                                             ^^^^^^^^^^^^^^^^^^^^^^ ---- required by a bound introduced by this call
    |                                             |
    |                                             the trait `std::convert::From<std::option::Option<i32>>` is not implemented for `serde_json::Number`
```

I tried a bunch of different things, like:

- `result["KdfMemory"] = Value::Number(Some(user.client_kdf_memory).into());`
- `result["KdfMemory"] = Value::Number((user.client_kdf_memory as i32).into());`

but they didn't work.

tessus: I also tried a few other `let Some` constructs, but I just can't fix the errors. For a Rust person this must be ridiculously easy to solve, but I need one example how this is done properly.

tessus:
Back to WIP. No idea how to fix those compile errors yet. Will have to do some research, but any hint might help.

src/api/core/accounts.rs
```diff
@@ -153,6 +155,14 @@ pub async fn _register(data: JsonUpcase<RegisterData>, mut conn: DbConn) -> Json
     // Make sure we don't leave a lingering invitation.
     Invitation::take(&email, &mut conn).await;
 
+    if let Some(client_kdf_parallelism) = data.KdfParallelism {
+        user.client_kdf_parallelism = Some(client_kdf_parallelism);
+    }
+
+    if let Some(client_kdf_memory) = data.KdfMemory {
+        user.client_kdf_memory = Some(client_kdf_memory);
+    }
```
jjlin, line 164:
```suggestion
    user.client_kdf_parallelism = client_kdf_parallelism;
    user.client_kdf_memory = client_kdf_memory;
```

If they aren't `Some`, then they must be `None` anyway, so this is simpler. I assume the other KDF fields have a check to avoid overwriting the default value in the user model.

Although I would probably move these down in reorder as `type`, `iter`, `memory`, `parallelism`.

src/db/models/user.rs
jjlin: You can probably just do `user.client_kdf_memory.unwrap().into()`. `expect()` is another option but it might be too verbose adding the same error messages in multiple places

jjlin: You can probably just do `user.client_kdf_memory.unwrap().into()`. `expect()` is another option but it might be too verbose adding the same error messages in multiple places

tessus: isn't unwrap dangerous and can lead to a panic? What's the pattern in Rust to prevent a panic? Check if not none and then unwrap? unwrap_or? but unwrap_or would be also wrong in this case. Hmm, I still have to learn the ins and outs of Rust. Some concepts are rather new to me.

src/api/core/accounts.rs
tessus: I will reorder them but keep them as written. registerData uses `Option<i32>` for all 4, so to keep it consistent I should use the same pattern as the other 2. Or we should change the other 2 to the same "simple" assignment as you suggested.

tessus:
Sqashed and force-pushed again. I dislike a messy git history, although it probably wouldn't matter, since merge commits are done when PRs are merged to master.

tessus:
Sorry, it seems that I can't re-request a review from more than one person.

src/db/models/user.rs
jjlin: `unwrap()` and `expect()` are used throughout the codebase; `expect()` is just `unwrap()` with a custom panic message.

It's okay to panic when there's no reasonable way to recover. If your KDF type is supposed to be Argon2, but there's no memory or parallelism value, then there's either a bug in the validation code (the part that's supposed to reject requests to set Argon2 without those values), or the database has been corrupted somehow. Anyway, in this framework, the panic will be caught and show a traceback instead of exiting the whole program.

src/api/core/accounts.rs
jjlin: As I said, the `type` and `iter` fields are different because they have default values in the user model, so you want to avoid overwriting them. See the analogous code in upstream:

https://github.com/bitwarden/server/blob/cf669286eda971e060a52b1852ba910df7e34a87/src/Core/Models/Api/Request/Accounts/RegisterRequestModel.cs#L40-L43

src/db/models/user.rs
tessus: ok. that makes sense. thanks.

src/api/core/accounts.rs
tessus: yep, once again, when I do this, it doesn't compile. But it compiles and works as it is. I have re-tested it.

jjlin: I meant to write
```rust
user.client_kdf_memory = data.KdfMemory;
user.client_kdf_parallelism = data.KdfParallelism;
```
I verified that this builds.

tessus: It's official. I am an idiot. I actually did that since it didn't compile and was the logical choice. Hmm.... Well... It helps to save the freakin file. hahaha. ;-)

src/api/core/accounts.rs
```diff
@@ -770,15 +798,22 @@ async fn prelogin(data: JsonUpcase<PreloginData>, conn: DbConn) -> Json<Value> {
...
         "Kdf": kdf_type,
-        "KdfIterations": kdf_iter
-    }))
+        "KdfIterations": kdf_iter,
+    });
+
+    if kdf_type == UserKdfType::Argon2id as i32 {
+        result["KdfMemory"] = Value::Number(kdf_mem.unwrap().into());
+        result["KdfParallelism"] = Value::Number(kdf_para.unwrap().into());
+    }
```
BlockListed, line 814: I think using if let Some(x) = Option<> syntax would be a lot better than unwrapping.
```suggestion
    if let Some(m) = kdf_mem {
        result["KdfMemory"] = Value::Number(m.into())
    }
    if let Some(p) = kdf_para {
        result["KdfParallelism"] = Value::Number(p.into())
    }
```

BlackDex: Then, instead of an `unwrap()` i suggest to use a `expect()` instead, and provide a correct message in case it fails.
Now it's probably guessing.

tessus: This will make the code look like crap and cluttered. See the comment from djinn in another thread.

If there are 3 people who are reviewing the PR and they have different opinions, this PR will never be finished. I am not against suggestions and help, especially since I definitely need help with Rust patterns, but can you plese agree on an approach.

I have fixed all the outstanding issues and now it seems (at least to me) there is a contradictory suggestion than to what I got before.

tessus: There are 4 options right now:

1. leave it as it is
2. replace `unwrap` with `expect`
3. use BlockListed's suggestion
4. change it to the following (IMO the best option):

```rust
    if kdf_type == UserKdfType::Argon2id as i32 {
        match kdf_mem {
            Some(m) => {
                result["KdfMemory"] = Value::Number(m.into());
            }
            None => panic!("KdfMemory must not be empty, when Argon2 is used."),
        }
        match kdf_para {
            Some(p) => {
                result["KdfParallelism"] = Value::Number(p.into());
            }
            None => panic!("KdfParallelism must not be empty, when Argon2 is used."),
        }
    }
```

tessus: please let me know which one to pick and I make the changes. I favor option number 4.

BlockListed: Panicking seems like the best solution and `unwrap`ping is not really intended for production code in rust, so the most rusty thing to choose would be `expect` in my opinion.

BlackDex: If you are only going to use `match` to send an `panic` which i would actually not want to be used at all, then an `except` would be the better choice. So, i agree with @BlockListed

tessus: ok, `expect` it is.

tessus:
Is there still something missing for this change?

BlackDex:
> Is there still something missing for this change?

In the sense of an approval?
I did not had time yet to check it. Also a web-vault needs to be released with supports for this of course. It could be that Bitwarden will make some changes to this part while not yet released by them of course, wouldn't be the first time.

tessus:
> In the sense of an approval?

No, I know that reviewing a PR can take time. But rather whether I have addressed all the issues that came up in previous conversations. I think I addressed everything, but I could have missed something. But the people who have left comments would most likely know right away and can say "hold on, dude, what about xyz? you missed that one."

> Also a web-vault needs to be released with supports for this of course.

Well, this change does not require a new web-vault. It's the other way around. The new clients that have the argon2 code will require this code change.
I am not saying it must be merged now, I am just saying that merging this PR will not mess up the current or older clients.

dani-garcia (approved):
Tested with the main branch of the web vault, works correctly for me

pjs128:
In what version should we start seeing Argon2id in the Security>Keys>KDF_algorithm dropdown?

BlackDex:
Either currently in testing, or wait until next release.

PR #3231, opened 2023-02-07
Generate distinct log messages for regex vs. IP blacklisting.

kpfleming:
When an icon will not be downloaded due to matching a configured blacklist, ensure that the log message indicates the type of blacklist that was matched.

### DIFF
--- a/src/api/icons.rs
+++ b/src/api/icons.rs
@@ -265,7 +265,7 @@ enum DomainBlacklistReason {
 }
 
 use cached::proc_macro::cached;
-#[cached(key = "String", convert = r#"{ domain.to_string() }"#, size = 16, time = 60, option = true)]
+#[cached(key = "String", convert = r#"{ domain.to_string() }"#, size = 16, time = 60)]
 async fn check_domain_blacklist_reason(domain: &str) -> Option<DomainBlacklistReason> {
     // First check the blacklist regex if there is a match.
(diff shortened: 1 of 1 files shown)


BlackDex (approved):

BlackDex:
Hmm, wait, it looks like it will only cache `Some()` results. That kinda defeats the whole caching for non-blocked items.

src/api/icons.rs
```diff
@@ -258,9 +258,15 @@ mod tests {
...
+#[derive(Debug, Clone)]
+enum DomainBlacklistReason {
+    Regex,
+    IP,
+}
+
 use cached::proc_macro::cached;
-#[cached(key = "String", convert = r#"{ domain.to_string() }"#, size = 16, time = 60)]
-async fn is_domain_blacklisted(domain: &str) -> bool {
+#[cached(key = "String", convert = r#"{ domain.to_string() }"#, size = 16, time = 60, option = true)]
```
BlackDex, line 268: Please remove the `option = true` for `cached` here.
That causes None responses to be ignored, which is not what we want.

BlackDex (changes requested):

src/api/icons.rs
kpfleming: Will do; I had added it at point to work around a different issue which is no longer present.

BlackDex (approved):

dani-garcia (approved):

PR #3234, opened 2023-02-08
Updated Rust and crates

BlackDex:
- Updated Rust to v1.67.0
- Updated all crates except for `cookies` and `webauthn`


Cargo.toml
```diff
@@ -114,7 +114,7 @@ webauthn-rs = "0.3.2"
 url = "2.3.1"
 
 # Email librariese-Base, Update crates and small change.
```
tessus, line 116: May I ask what `librariese-Base` is supposed to mean? And what small change? Is there a create for small changes?

I came across this when looking at the changes for this PR and it was right above a line that had changes. I did not go through the file trying to nitpick on stuff. ;-)

Anyway the plural of `library` is `libraries`, but maybe there is something that is called librariese-Base or was it meant to mean "base libraries"?

BlackDex: I have no clue actually haha. But maybe in a next update ill change it. Unless there are some crates updated after what already is updated.

tessus: No worries, I was just curious. For me it was like having a comment that says "Apple, oranges, and bananas`, or whatever that doesn't have anything to do with the text it was supposed to describe. Haha.

BlackDex: Changed it :)

dani-garcia (approved):

tessus:
Can this be merged as well? I am just curious. I am still trying to figure out when certain things are merged. 

e.g. Would it make sense to add auto-merge, so that when the binding approvals are reached, the PR is automatically merged to main?

Or are the PRs picked for what should be in the testing doker images?

dani-garcia:
Usually things are just merged manually right after they are approved, in this case it wasn't done because there are some conflicts in the dockerfiles that need to be resolved. Was planning to do that myself after I reviewed everything else but didn't have enough time in the end.

tessus:
Ah, I see. So in this case, auto-merge might be an idea, since it will only merge after approval and when there are no conflicts. But I also understand that a project lead wants to do this mnually. Just wanted to get a better feel on how this project is managed.

Thanks a bunch for the answer.

BlackDex:
Auto-merge will not work in this case.
The issue here is that there was a PR which renamed the Dockerfiles and made some other changes. And also a Cargo.toml conflict. Those are difficult to fix sometimes.

I wasn't home my self, and no other way to resolve those conflicts a bit easy, else i would have done it my self. Which i might still do tomorrow.

tessus:
I thought that gh's auto-merge will only merge to main, if 2 conditions are met:

1. all approval that are necessary have been given
2. PR does not have any conflicts

I'm just saying that in general, since the PRs are merged pretty much after approcal there's no need to do it manually.
In this specific instance, you are correct, the auto-merge would not have happened.

BlackDex:
So that wouldn't have solved this case then, since there were conflicts.

tessus:
Yep

PR #3240, opened 2023-02-11
Version string verbosity

strump:
Build script doesn't produce version string if no git history is available.

**Steps to reproduce:**

1. Clone repo with **no history** (Mogenius.com does this): `git clone --depth 1 https://github.com/dani-garcia/vaultwarden.git`
2. Build docker image: `docker build --build-arg DB=sqlite -f docker/amd64/Dockerfile -t vaultwarden:local .`
3. Run it: `docker run -it --rm vaultwarden:local`

**Actual result**: no version information in output
```
/--------------------------------------------------------------------\
|                        Starting Vaultwarden                        |
|--------------------------------------------------------------------|
| This is an *unofficial* Bitwarden implementation, DO NOT use the   |
| official channels to report bugs/features, regardless of client.   |
| Send usage/configuration questions or feature requests to:         |
|   https://vaultwarden.discourse.group/                             |
| Report suspected bugs/issues in the software itself at:            |
|   https://github.com/dani-garcia/vaultwarden/issues/new            |
\--------------------------------------------------------------------/
```

**After this PR:**
```
/--------------------------------------------------------------------\
|                        Starting Vaultwarden                        |
|             Version main-7d7da921 2023-02-11 17:27:10              |
|--------------------------------------------------------------------|
| This is an *unofficial* Bitwarden implementation, DO NOT use the   |
| official channels to report bugs/features, regardless of client.   |
| Send usage/configuration questions or feature requests to:         |
|   https://vaultwarden.discourse.group/                             |
| Report suspected bugs/issues in the software itself at:            |
|   https://github.com/dani-garcia/vaultwarden/issues/new            |
\--------------------------------------------------------------------/
```

**Description:**
* Last tag is now optional.
* Timestamp is added to version string.

Four types of version string are generated in build.rs:
1. Main branch + last tag → `{last_tag}-{rev_short} {commit_time}`
2. Main branch  → `main-{rev_short} {commit_time}`
3. Non-main branch + last tag  → `{last_tag}-{rev_short} ({branch}) {commit_time}`
4. Non-main branch  → `{rev_short} ({branch}) {commit_time}`

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -57,8 +57,10 @@ fn version_from_git_info() -> Result<String, std::io::Error> {
 
     // The last available tag, equal to exact_tag when
     // the current commit is tagged
-    let last_tag = run(&["git", "describe", "--abbrev=0", "--tags"])?;
-    println!("cargo:rustc-env=GIT_LAST_TAG={last_tag}");
+    let last_tag_maybe = run(&["git", "describe", "--abbrev=0", "--tags"]).ok();
+    if let Some(ref last_tag) = last_tag_maybe {
+        println!("cargo:rustc-env=GIT_LAST_TAG={last_tag}");
+    }
 
     // The current branch name
     let branch = run(&["git", "rev-parse", "--abbrev-ref", "HEAD"])?;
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -57,8 +57,10 @@ fn version_from_git_info() -> Result<String, std::io::Error> {
 
     // The last available tag, equal to exact_tag when
     // the current commit is tagged
-    let last_tag = run(&["git", "describe", "--abbrev=0", "--tags"])?;
-    println!("cargo:rustc-env=GIT_LAST_TAG={last_tag}");
+    let last_tag_maybe = run(&["git", "describe", "--abbrev=0", "--tags"]).ok();
```
williamdes, line 60: You should use https://crates.io/crates/vergen for that

BlackDex: That is maybe a nice one to use instead of our custom calls, and it could provide some more info.

BlackDex:
Besides the tests that are failing. There is an env variable that can be set if it is known to not having a full git clone.
You can use `VW_VERSION` to set a custom version string instead of this modification.

build.rs
williamdes: yean and its very easy to use, I use it on https://github.com/wdes/mail-autodiscover-autoconfig/issues/3#issuecomment-1230181511
if examples are needed

build.rs
```diff
@@ -69,12 +71,25 @@ fn version_from_git_info() -> Result<String, std::io::Error> {
     let rev_short = rev.get(..8).unwrap_or_default();
     println!("cargo:rustc-env=GIT_REV={rev_short}");
 
+    // The current git commit time
+    let commit_time = run(&["git", "log", "-1", "--format=%cd", "--date=format:%Y-%m-%d %H:%M:%S"])?;
```
jjlin, line 75: Adding the commit time to the version string seems overly verbose, and isn't something that's usually included in a version string itself. While I don't feel that including the commit time is all that valuable, if you really want to include it, I think it should go on its own line of output, and it should also include timezone info.

BTW, `hooks/build` already includes the build (not commit) time in the Docker image. To me it seems a bit odd for those times not to match, so that's another reason I'd favor not including the commit time.

BlackDex:
I have looked into `vergen`, and it looks nice, but it would change our current version string and not the same info we would want.
While I can see a good point into using the commit date as build date in case no other info is present, it isn't very useful in the end I think.

If some maintainer or builder is using a shallow pull they should export `VW_VERSION` with the correct version them self, since they probably also need to give there package or image a version number, and they can set the same.

Going to close this as we are not going to implement this.
I might see if i can add some more features to `vergen` so we can use it with our may of versioning, and i think it could be useful for other projects as well. But that is something for later.

PR #3251, opened 2023-02-16
Validate all needed fields for client API login

BlackDex:
During the client API login we need to have a `device_identifier`, `device_name` and `device_type`.
When these were not provided Vaultwarden would panic.

This PR add checks for these fields and makes sure it returns a better error message instead of causing a panic.

### DIFF
--- a/src/api/identity.rs
+++ b/src/api/identity.rs
@@ -52,6 +52,10 @@ async fn login(data: Form<ConnectData>, client_header: ClientHeaders, mut conn: 
             _check_is_some(&data.client_secret, "client_secret cannot be blank")?;
             _check_is_some(&data.scope, "scope cannot be blank")?;
 
+            _check_is_some(&data.device_identifier, "device_identifier cannot be blank")?;
+            _check_is_some(&data.device_name, "device_name cannot be blank")?;
+            _check_is_some(&data.device_type, "device_type cannot be blank")?;
+
             _api_key_login(data, &mut user_uuid, &mut conn, &ip).await
         }
         t => err!("Invalid type", t),


src/api/identity.rs
```diff
@@ -52,6 +52,9 @@ async fn login(data: Form<ConnectData>, client_header: ClientHeaders, mut conn:
             _check_is_some(&data.client_secret, "client_secret cannot be blank")?;
             _check_is_some(&data.scope, "scope cannot be blank")?;
 
+            _check_is_some(&data.device_identifier, "device_identifier cannot be blank")?;
+            _check_is_some(&data.device_name, "device_name cannot be blank")?;
```
jjlin, line 56: It seems like `device_type` should be checked too then? At least that's what the client code seems to be doing:

1. https://github.com/bitwarden/clients/blob/65b9ab66d0b9d51a467997abbec484cf7b95b0cf/libs/common/src/auth/login-strategies/user-api-login.strategy.ts#L59
2. https://github.com/bitwarden/clients/blob/65b9ab66d0b9d51a467997abbec484cf7b95b0cf/libs/common/src/auth/login-strategies/login.strategy.ts#L81-L82
3. https://github.com/bitwarden/clients/blob/65b9ab66d0b9d51a467997abbec484cf7b95b0cf/libs/common/src/auth/models/request/identity-token/device.request.ts#L11-L13

BlackDex: Hmm, ill have to check, but that isn't really used during session/refresh token generation and validation. But, if the cli does send that it would be best to add it also. Just having those two fields checked was enough to be able to login without a panic.

Thx for the pointers.

BlackDex: Thanks @jjlin , it should be there indeed. While we catch a non-processable value to the default `14` aka `unknown`, checking this would be better indeed.

PR #3262, opened 2023-02-20
Make the admin session lifetime adjustable

mittler-works:
In my opinion, it may be useful if this value is adjustable to personal needs. For example, I want to access the Admin API programmatically and accordingly I want to set this value to as short as possible. Others may want to set the value higher if they often spend a long time in the Admin UI or something.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -335,6 +335,9 @@
 ## Allow a burst of requests of up to this size, while maintaining the average indicated by `ADMIN_RATELIMIT_SECONDS`.
 # ADMIN_RATELIMIT_MAX_BURST=3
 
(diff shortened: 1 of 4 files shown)


BlackDex:
This will not work. If we would want this, and you want this to actually work you need to adjust the JWT lifetime.
https://github.com/dani-garcia/vaultwarden/blob/1ba8275dcb53bf8e6f41ad2fa2e0a09f3674ff74/src/auth.rs#L240-L248

mittler-works:
Thank you for pointing out! I guess it must then be set in both places, so that the cookie containing the JWT does not expire before the JWT becomes invalid.

PR #3267, opened 2023-02-21
change WEBSOCKET_ENABLED default value to true

JohnnyRacket:
Currently `WEBSOCKET_ENABLED` defaults to `false`, which causes the inability to log into BitWarden apps (iOS & Android). The only thing I saw in the wiki about this ENV_VAR was related to getting notifications [wiki](https://github.com/dani-garcia/vaultwarden/wiki/Enabling-WebSocket-notifications). In fact the wiki specifically mentions how it doesn't affect mobile because of push notifications. 

Proposing the default value be changed to `true` because neither iOS nor Android apps can log in without `WEBSOCKET_ENABLED=true` env var and this seems like a reasonable baseline setting.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -66,7 +66,7 @@
 # WEB_VAULT_ENABLED=true
 
 ## Enables websocket notifications
-# WEBSOCKET_ENABLED=false
+# WEBSOCKET_ENABLED=true
 
 ## Controls the WebSocket server address and port
 # WEBSOCKET_ADDRESS=0.0.0.0
--- a/src/config.rs
+++ b/src/config.rs
@@ -371,7 +371,7 @@ make_config! {
     },
     ws {
         /// Enable websocket notifications
-        websocket_enabled:      bool,   false,  def,    false;
+        websocket_enabled:      bool,   false,  def,    true;
         /// Websocket address
         websocket_address:      String, false,  def,    "0.0.0.0".to_string();
         /// Websocket port


jjlin:
The mobile apps don't even use WebSockets, so you have some other issue going on

BlackDex:
I'm going to decline this PR, since login and web-sockets have nothing to do with the mobile clients.
The mobile clients can't even use web-sockets, so this will not have any effect.

Also, we want this to be kept on `false` as default, since you need to deploy a reverse proxy when you want to use the web-sockets feature at all.

And, if you want to change a default, just adjust the ENV as mentioned via docker-compose or `-e WEBSOCKET_ENABLED=false`

JohnnyRacket:
That may be true, but it fixed it for me as well as others:
[discourse link](https://vaultwarden.discourse.group/t/problems-with-apps-on-mac-and-ios/2153/7)
I can confirm that is the only setting changed between deployments on my cluster.

BlackDex:
It sounds more like the container needed a restart instead of changing that env.
Again, the mobile clients do not use web-sockets at all, they do not have support for that feature built-in.

And, web-sockets have no connection to the login process, except that once logged in, it uses the cookie/token to authenticate the user over the web-socket.

JohnnyRacket:
It is not the login process per say, it is the first step of contacting the server, only the self hosted server and username have been entered when this error  occurs, no password to kick off a login flow.
I will take off the env var and see if it re-breaks, and also test how it affects an already logged in mobile device.

BlackDex:
Then you were either running an older version of Vaultwarden which did not had the correct new endpoints yet available which are needed for the newer clients. And your re-creation of the container caused it to use the newly pulled image to be used.

Both me and @jjlin are maintainers of Vaultwarden, just trust us that web-sockets have nothing to do with the whole login process.

JohnnyRacket:
I confirmed via exec'ing in to be on `1.27.0` this whole time, I troubleshot this for about 20 minutes before finding my solution in that discourse thread.

I have re-confirmed this `WEBSOCKET_ENABLED=false` breaks my mobile apps login. I am unable to get to the password screen with the error "We were unable to process your request. Please try again or contact us."

I do not doubt either of your knowledge, perhaps mine and those two people in the discourse thread have an exceptionally odd setup along with me causing this to be the case, I am running a k8s cluster behind cloudflare.

For me iOS and Android only work (get past the username screen without error) with this flag enabled, so I will not bother to argue with either of you, I fully respect your understanding, but it is perhaps worth looking into since 3 people have found the same solution to an issue in the past month or two.

BlackDex:
In that case i would suggest to double check your reverse proxy config instead.
You probably have an issue there.

JohnnyRacket:
I appreciate the advice, upon more testing you are both correct that this flag does not seem to be the issue. I have sat here spam spinning up and down pods and it seems like there is a 50% the error happens for the life of the pod and a 50% chance everything works fine (same healthy logs either way) for the life of the pod. 

I think those posters probably encountered something similar and the env_var was a reason to restart their container like it was for me, perhaps they are both on a similar setup with k8s, postgres, and cloudflare.

It is very odd behaviour, l'll chalk it up to my setup, appreciate your patience and explanations.

BlackDex:
Are you running multiple pods for Vaultwarden with a single ingress?

JohnnyRacket:
single pod behind a single ingress for it, I can post a sanitized version on my deploy if you are interested. buts its just a single replica vault warden with a single replica postgres. My working theory is that it is silently failing to connect to postgres on some boots.

JohnnyRacket:
server deployment:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    io.kompose.service: vaultwarden
  name: vaultwarden
spec:
  replicas: 1
  selector:
    matchLabels:
      io.kompose.service: vaultwarden
  strategy:
    type: Recreate
  template:
    metadata:
      labels:
        io.kompose.service: vaultwarden
    spec:
      containers:
        - env:
            - name: DATABASE_URL
              value: postgresql://user:pass@vaultwarden-postgres:5432/vaultwarden
            - name: SIGNUPS_ALLOWED
              value: "false"
            - name: WEBSOCKET_ENABLED
              value: "true"
          image: vaultwarden/server:latest
          name: vaultwarden
          ports:
            - containerPort: 80
          volumeMounts:
            - mountPath: /data
              name: vaultwarden-claim
      restartPolicy: Always
      volumes:
        - name: vaultwarden-claim
          persistentVolumeClaim:
            claimName: vaultwarden-claim
```
postgres deployment:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    io.kompose.service: vaultwarden-postgres
  name: vaultwarden-postgres
spec:
  replicas: 1
  selector:
    matchLabels:
      io.kompose.service: vaultwarden-postgres
  template:
    metadata:
      labels:
        io.kompose.service: vaultwarden-postgres
    spec:
      containers:
        - env:
            - name: POSTGRES_DB
              value: vaultwarden
            - name: POSTGRES_PASSWORD
              value: pass
            - name: POSTGRES_USER
              value: user
          image: postgres
          name: vaultwarden-postgres
          ports:
            - containerPort: 5432
      restartPolicy: Always
```

Ill drop these here if youre interested.

PR #3282, opened 2023-02-24
Add confirmation for removing 2FA and deauthing sessions in admin panel

JCBird1012:
As someone who has inadvertently pressed "Remove all 2FA" instead of "Deauthorize sessions" one too many times (and had to re-add all my YubiKeys) - there should be a confirmation prompt before taking either of those actions.

Regardless of my accidental misclicks - any action that modifies user state should be protected by a confirmation prompt.

### DIFF
--- a/src/static/scripts/admin_users.js
+++ b/src/static/scripts/admin_users.js
@@ -32,10 +32,13 @@ function remove2fa(event) {
         alert("Required parameters not found!");
         return false;
     }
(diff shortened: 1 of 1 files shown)


BlackDex (approved):

dani-garcia (approved):

dpinse:
Currently the buttons don't work after this merge. I noticed `const email = event.target.parentNode.dataset.vwUserEmail;` is missing from the functions as well as `!email` is missing for the if statements.

PR #3281, opened 2023-02-24
Fix the web-vault v2023.2.0 API calls

BlackDex:
- Supports the new Collection/Group/User editing UI's
- Support `/partial` endpoint for cipher updating to allow folder and favorite update for read-only ciphers.
- Prevent `Favorite`, `Folder`, `read-only` and `hide-passwords` from being added to the organizational sync.
- Added and corrected some `Object` key's to the output json.

Fixes #3279

PR #3285, opened 2023-02-26
Save members when updating groups

C0rby:
The web client uses this API endpoint to add users to groups.


BlackDex:
Duplicate of #3281

BlackDex:
Also #3281 had more fixes which are needed.
Also, I'm currently in progress to check all the other endpoints for changes which might be needed.

C0rby:
Ah thanks, I must have missed that when looking through the open PRs.

PR #3288, opened 2023-02-28
Some Admin Interface updates

BlackDex:
- Updated datatables
- Added NTP Time check
- Added Collections, Groups and Events count for orgs
- Renamed `Items` to `Ciphers`
- Some small style updates

### DIFF
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -300,8 +300,9 @@ fn logout(cookies: &CookieJar<'_>) -> Redirect {
 
 #[get("/users")]
 async fn get_users_json(_token: AdminToken, mut conn: DbConn) -> Json<Value> {
-    let mut users_json = Vec::new();
-    for u in User::get_all(&mut conn).await {
+    let users = User::get_all(&mut conn).await;
+    let mut users_json = Vec::with_capacity(users.len());
+    for u in users {
         let mut usr = u.to_json(&mut conn).await;
         usr["UserEnabled"] = json!(u.enabled);
         usr["CreatedAt"] = json!(format_naive_datetime_local(&u.created_at, DT_FMT));
@@ -313,8 +314,9 @@ async fn get_users_json(_token: AdminToken, mut conn: DbConn) -> Json<Value> {
 
 #[get("/users/overview")]
 async fn users_overview(_token: AdminToken, mut conn: DbConn) -> ApiResult<Html<String>> {
-    let mut users_json = Vec::new();
-    for u in User::get_all(&mut conn).await {
+    let users = User::get_all(&mut conn).await;
+    let mut users_json = Vec::with_capacity(users.len());
+    for u in users {
         let mut usr = u.to_json(&mut conn).await;
         usr["cipher_count"] = json!(Cipher::count_owned_by_user(&u.uuid, &mut conn).await);
         usr["attachment_count"] = json!(Attachment::count_by_user(&u.uuid, &mut conn).await);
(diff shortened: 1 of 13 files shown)


dani-garcia (approved):

jjlin:
I'm not sure renaming `Items` to `Ciphers` in the UI is an improvement. IMO, this is an overly technical term that may cause confusion, and Bitwarden itself uses the "items" term in their UI and user-facing docs.

Also, you might consider using the Cloudflare [`/cdn-cgi/trace`](https://cloudflare.com/cdn-cgi/trace) API instead of timeapi.io, which may not be freely accessible over the long term.

BlackDex:
> I'm not sure renaming `Items` to `Ciphers` in the UI is an improvement. IMO, this is an overly technical term that may cause confusion, and Bitwarden itself uses the "items" term in their UI and user-facing docs.
> 

Maybe we should call it `Vault Items` then? That is more descriptive.

> Also, you might consider using the Cloudflare [`/cdn-cgi/trace`](https://cloudflare.com/cdn-cgi/trace) API instead of timeapi.io, which may not be freely accessible over the long term.

Thanks for the tip. While i think timeapi.io should be enough. It's not that people keep bashing the diagnostics page. And on there site they claim to keep a free plan available, which I think would suffice. But in case it's not, we can switch easily.

jjlin:
> > I'm not sure renaming `Items` to `Ciphers` in the UI is an improvement. IMO, this is an overly technical term that may cause confusion, and Bitwarden itself uses the "items" term in their UI and user-facing docs.
> 
> Maybe we should call it `Vault Items` then? That is more descriptive.

Sounds fine to me.

> > Also, you might consider using the Cloudflare [`/cdn-cgi/trace`](https://cloudflare.com/cdn-cgi/trace) API instead of timeapi.io, which may not be freely accessible over the long term.
> 
> Thanks for the tip. While i think timeapi.io should be enough. It's not that people keep bashing the diagnostics page. And on there site they claim to keep a free plan available, which I think would suffice. But in case it's not, we can switch easily.

I assume a free plan means that an API key would need to be provided, though.

BlackDex:
> > > I'm not sure renaming `Items` to `Ciphers` in the UI is an improvement. IMO, this is an overly technical term that may cause confusion, and Bitwarden itself uses the "items" term in their UI and user-facing docs.
> > 
> > 
> > Maybe we should call it `Vault Items` then? That is more descriptive.
> 
> Sounds fine to me.
> 
> > > Also, you might consider using the Cloudflare [`/cdn-cgi/trace`](https://cloudflare.com/cdn-cgi/trace) API instead of timeapi.io, which may not be freely accessible over the long term.
> > 
> > 
> > Thanks for the tip. While i think timeapi.io should be enough. It's not that people keep bashing the diagnostics page. And on there site they claim to keep a free plan available, which I think would suffice. But in case it's not, we can switch easily.
> 
> I assume a free plan means that an API key would need to be provided, though.

That might be a possibility indeed. But they have a 180 days of grace period for changing the API and depreciate it, that should be more then enough time.

tessus:
I never understood why they would call notes, logins, identity, and cards: ciphers. A cipher is an algorithm for performing encryption or decryption, not a password or any of the other item types.

I am not sure "Vault Items" is an improvement, because where else would the items be?

So there a few terms that fit the profile:

- items
- entries
- records

Anything is better than cipher. ;-)

tessus:
@BlackDex Btw, is it ok to comment on a merged PR or shall I rather create a new discussion?

BlackDex:
This is fine. Ill take these comments into a new pr with some other changes.

PR #3290, opened 2023-03-01
Fix confirmation for removing 2FA and deauthing sessions in admin panel

dpinse:
Fixes #3282

Adds the email variable as this is required for the buttons to function.


JCBird1012:
You beat e9ba0ea by minutes  - either way, big oversight on my part. 
thanks for catching this.

BlackDex:
Also an oversight on my party.

BlackDex (approved):

PR #3304, opened 2023-03-04
feat: Implement Push Notifications sync

GeekCornerGH:
# Implements the Push Notification feature
#### Resolves #2469

[Push-demo.webm](https://user-images.githubusercontent.com/45696571/222894552-109e3c34-e9b2-44a0-9474-304123f78b0a.webm)

Special thanks to @samb-devel for your help

## How to test:
Get your keys from [the Bitwarden website](https://bitwarden.com/host)
Set environment variables to the required values:
```env
PUSH_ENABLED=true
PUSH_INSTALLATION_ID=CHANGEME
PUSH_INSTALLATION_KEY=CHANGEME
```
PUSH_RELAY_URI is already configured to use default Bitwarden server.
Start the server, then test.

This was my first shot at Rust.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -72,6 +72,13 @@
 # WEBSOCKET_ADDRESS=0.0.0.0
 # WEBSOCKET_PORT=3012
 
+## Enables push notifications (requires key and id from https://bitwarden.com/host)
+# PUSH_ENABLED=true
+# PUSH_INSTALLATION_ID=CHANGEME
+# PUSH_INSTALLATION_KEY=CHANGEME
+## Don't change this unless you know what you're doing.
+# PUSH_RELAY_BASE_URI=https://push.bitwarden.com
+
 ## Controls whether users are allowed to create Bitwarden Sends.
 ## This setting applies globally to all users.
 ## To control this on a per-org basis instead, use the "Disable Send" org policy.
--- a/migrations/mysql/2023-02-18-125735_push_uuid_table/up.sql
+++ b/migrations/mysql/2023-02-18-125735_push_uuid_table/up.sql
@@ -1,0 +1,1 @@
+ALTER TABLE devices ADD COLUMN push_uuid TEXT;
--- a/migrations/postgresql/2023-02-18-125735_push_uuid_table/up.sql
+++ b/migrations/postgresql/2023-02-18-125735_push_uuid_table/up.sql
@@ -1,0 +1,1 @@
+ALTER TABLE devices ADD COLUMN push_uuid TEXT;
--- a/migrations/sqlite/2023-02-18-125735_push_uuid_table/up.sql
+++ b/migrations/sqlite/2023-02-18-125735_push_uuid_table/up.sql
@@ -1,0 +1,1 @@
+ALTER TABLE devices ADD COLUMN push_uuid TEXT;
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -13,7 +13,7 @@ use rocket::{
 };
 
 use crate::{
-    api::{core::log_event, ApiResult, EmptyResult, JsonResult, Notify, NumberOrString},
+    api::{core::log_event, unregister_push_device, ApiResult, EmptyResult, JsonResult, Notify, NumberOrString},
     auth::{decode_admin, encode_jwt, generate_admin_claims, ClientIp},
     config::ConfigBuilder,
     db::{backup_database, get_sql_server_version, models::*, DbConn, DbConnType},
@@ -402,14 +402,22 @@ async fn delete_user(uuid: &str, token: AdminToken, mut conn: DbConn) -> EmptyRe
 #[post("/users/<uuid>/deauth")]
 async fn deauth_user(uuid: &str, _token: AdminToken, mut conn: DbConn, nt: Notify<'_>) -> EmptyResult {
     let mut user = get_user_or_404(uuid, &mut conn).await?;
+
+    nt.send_logout(&user, None, &mut conn).await;
+
+    if CONFIG.push_enabled() {
+        for device in Device::find_push_device_by_user(&user.uuid, &mut conn).await {
+            match unregister_push_device(device.uuid).await {
+                Ok(r) => r,
+                Err(e) => error!("Unable to unregister devices from Bitwarden server: {}", e),
+            };
+        }
+    }
+
     Device::delete_all_by_user(&user.uuid, &mut conn).await?;
     user.reset_security_stamp();
 
-    let save_result = user.save(&mut conn).await;
-
-    nt.send_logout(&user, None).await;
-
-    save_result
+    user.save(&mut conn).await
 }
 
 #[post("/users/<uuid>/disable")]
@@ -421,7 +429,7 @@ async fn disable_user(uuid: &str, _token: AdminToken, mut conn: DbConn, nt: Noti
 
     let save_result = user.save(&mut conn).await;
 
-    nt.send_logout(&user, None).await;
+    nt.send_logout(&user, None, &mut conn).await;
 
     save_result
 }
--- a/src/api/core/accounts.rs
+++ b/src/api/core/accounts.rs
@@ -4,7 +4,8 @@ use serde_json::Value;
 
 use crate::{
     api::{
-        core::log_user_event, EmptyResult, JsonResult, JsonUpcase, Notify, NumberOrString, PasswordData, UpdateType,
+        core::log_user_event, register_push_device, unregister_push_device, EmptyResult, JsonResult, JsonUpcase,
+        Notify, NumberOrString, PasswordData, UpdateType,
     },
     auth::{decode_delete, decode_invite, decode_verify_email, Headers},
     crypto,
@@ -35,6 +36,7 @@ pub fn routes() -> Vec<rocket::Route> {
         post_verify_email_token,
         post_delete_recover,
         post_delete_recover_token,
+        post_device_token,
         delete_account,
         post_delete_account,
         revision_date,
@@ -46,6 +48,9 @@ pub fn routes() -> Vec<rocket::Route> {
         get_known_device,
         get_known_device_from_path,
         put_avatar,
+        put_device_token,
+        put_clear_device_token,
+        post_clear_device_token,
     ]
 }
 
@@ -338,7 +343,7 @@ async fn post_password(
     // Prevent loging out the client where the user requested this endpoint from.
     // If you do logout the user it will causes issues at the client side.
     // Adding the device uuid will prevent this.
-    nt.send_logout(&user, Some(headers.device.uuid)).await;
+    nt.send_logout(&user, Some(headers.device.uuid), &mut conn).await;
 
     save_result
 }
@@ -398,7 +403,7 @@ async fn post_kdf(data: JsonUpcase<ChangeKdfData>, headers: Headers, mut conn: D
     user.set_password(&data.NewMasterPasswordHash, Some(data.Key), true, None);
     let save_result = user.save(&mut conn).await;
 
-    nt.send_logout(&user, Some(headers.device.uuid)).await;
+    nt.send_logout(&user, Some(headers.device.uuid), &mut conn).await;
 
     save_result
 }
@@ -485,7 +490,7 @@ async fn post_rotatekey(data: JsonUpcase<KeyData>, headers: Headers, mut conn: D
     // Prevent loging out the client where the user requested this endpoint from.
     // If you do logout the user it will causes issues at the client side.
     // Adding the device uuid will prevent this.
-    nt.send_logout(&user, Some(headers.device.uuid)).await;
+    nt.send_logout(&user, Some(headers.device.uuid), &mut conn).await;
 
     save_result
 }
@@ -508,7 +513,7 @@ async fn post_sstamp(
     user.reset_security_stamp();
     let save_result = user.save(&mut conn).await;
 
-    nt.send_logout(&user, None).await;
+    nt.send_logout(&user, None, &mut conn).await;
 
     save_result
 }
@@ -611,7 +616,7 @@ async fn post_email(
 
     let save_result = user.save(&mut conn).await;
 
-    nt.send_logout(&user, None).await;
+    nt.send_logout(&user, None, &mut conn).await;
 
     save_result
 }
@@ -929,4 +934,65 @@ impl<'r> FromRequest<'r> for KnownDevice {
             uuid,
         })
     }
+}
+
+#[derive(Deserialize)]
+#[allow(non_snake_case)]
+struct PushToken {
+    PushToken: String,
+}
+
+#[post("/devices/identifier/<uuid>/token", data = "<data>")]
+async fn post_device_token(uuid: &str, data: JsonUpcase<PushToken>, headers: Headers, conn: DbConn) -> EmptyResult {
+    put_device_token(uuid, data, headers, conn).await
+}
+
+#[put("/devices/identifier/<uuid>/token", data = "<data>")]
+async fn put_device_token(uuid: &str, data: JsonUpcase<PushToken>, headers: Headers, mut conn: DbConn) -> EmptyResult {
+    if !CONFIG.push_enabled() {
+        return Ok(());
+    }
+
+    let data = data.into_inner().data;
+    let token = data.PushToken;
+    let mut device = match Device::find_by_uuid_and_user(&headers.device.uuid, &headers.user.uuid, &mut conn).await {
+        Some(device) => device,
+        None => err!(format!("Error: device {uuid} should be present before a token can be assigned")),
+    };
+    device.push_token = Some(token);
+    if device.push_uuid.is_none() {
+        device.push_uuid = Some(uuid::Uuid::new_v4().to_string());
+    }
+    if let Err(e) = device.save(&mut conn).await {
+        err!(format!("An error occured while trying to save the device push token: {e}"));
+    }
+    if let Err(e) = register_push_device(headers.user.uuid, device).await {
+        err!(format!("An error occured while proceeding registration of a device: {e}"));
+    }
+
+    Ok(())
+}
+
+#[put("/devices/identifier/<uuid>/clear-token")]
+async fn put_clear_device_token(uuid: &str, mut conn: DbConn) -> EmptyResult {
+    // This only clears push token
+    // https://github.com/bitwarden/core/blob/master/src/Api/Controllers/DevicesController.cs#L109
+    // https://github.com/bitwarden/core/blob/master/src/Core/Services/Implementations/DeviceService.cs#L37
+    // This is somehow not implemented in any app, added it in case it is required
+    if !CONFIG.push_enabled() {
+        return Ok(());
+    }
+
+    if let Some(device) = Device::find_by_uuid(uuid, &mut conn).await {
+        Device::clear_push_token_by_uuid(uuid, &mut conn).await?;
+        unregister_push_device(device.uuid).await?;
+    }
+
+    Ok(())
+}
+
+// On upstream server, both PUT and POST are declared. Implementing the POST method in case it would be useful somewhere
+#[post("/devices/identifier/<uuid>/clear-token")]
+async fn post_clear_device_token(uuid: &str, conn: DbConn) -> EmptyResult {
+    put_clear_device_token(uuid, conn).await
 }
--- a/src/api/notifications.rs
+++ b/src/api/notifications.rs
@@ -21,7 +21,10 @@ use tokio_tungstenite::{
 
 use crate::{
     auth::ClientIp,
-    db::models::{Cipher, Folder, Send as DbSend, User},
+    db::{
+        models::{Cipher, Folder, Send as DbSend, User},
+        DbConn,
+    },
     Error, CONFIG,
 };
 
@@ -32,6 +35,8 @@ static WS_USERS: Lazy<Arc<WebSocketUsers>> = Lazy::new(|| {
         map: Arc::new(dashmap::DashMap::new()),
     })
 });
+
+use super::{push_cipher_update, push_folder_update, push_logout, push_send_update, push_user_update};
 
 pub fn routes() -> Vec<Route> {
     routes![websockets_hub]
@@ -233,19 +238,33 @@ impl WebSocketUsers {
         );
 
         self.send_update(&user.uuid, &data).await;
+
+        if CONFIG.push_enabled() {
+            push_user_update(ut, user).await;
+        }
     }
 
-    pub async fn send_logout(&self, user: &User, acting_device_uuid: Option<String>) {
+    pub async fn send_logout(&self, user: &User, acting_device_uuid: Option<String>, conn: &mut DbConn) {
         let data = create_update(
             vec![("UserId".into(), user.uuid.clone().into()), ("Date".into(), serialize_date(user.updated_at))],
             UpdateType::LogOut,
-            acting_device_uuid,
+            acting_device_uuid.clone(),
         );
 
         self.send_update(&user.uuid, &data).await;
+
+        if CONFIG.push_enabled() {
+            push_logout(user, acting_device_uuid, conn).await;
+        }
     }
 
-    pub async fn send_folder_update(&self, ut: UpdateType, folder: &Folder, acting_device_uuid: &String) {
+    pub async fn send_folder_update(
+        &self,
+        ut: UpdateType,
+        folder: &Folder,
+        acting_device_uuid: &String,
+        conn: &mut DbConn,
+    ) {
         let data = create_update(
             vec![
                 ("Id".into(), folder.uuid.clone().into()),
@@ -257,6 +276,10 @@ impl WebSocketUsers {
         );
 
         self.send_update(&folder.user_uuid, &data).await;
+
+        if CONFIG.push_enabled() {
+            push_folder_update(ut, folder, acting_device_uuid, conn).await;
+        }
     }
 
     pub async fn send_cipher_update(
@@ -266,6 +289,7 @@ impl WebSocketUsers {
         user_uuids: &[String],
         acting_device_uuid: &String,
         collection_uuids: Option<Vec<String>>,
+        conn: &mut DbConn,
     ) {
         let org_uuid = convert_option(cipher.organization_uuid.clone());
         // Depending if there are collections provided or not, we need to have different values for the following variables.
@@ -295,9 +319,13 @@ impl WebSocketUsers {
         for uuid in user_uuids {
             self.send_update(uuid, &data).await;
         }
+
+        if CONFIG.push_enabled() && user_uuids.len() == 1 {
+            push_cipher_update(ut, cipher, acting_device_uuid, conn).await;
+        }
     }
 
-    pub async fn send_send_update(&self, ut: UpdateType, send: &DbSend, user_uuids: &[String]) {
+    pub async fn send_send_update(&self, ut: UpdateType, send: &DbSend, user_uuids: &[String], conn: &mut DbConn) {
         let user_uuid = convert_option(send.user_uuid.clone());
 
         let data = create_update(
@@ -312,6 +340,9 @@ impl WebSocketUsers {
 
         for uuid in user_uuids {
             self.send_update(uuid, &data).await;
+        }
+        if CONFIG.push_enabled() && user_uuids.len() == 1 {
+            push_send_update(ut, send, conn).await;
         }
     }
 }
@@ -354,7 +385,7 @@ fn create_ping() -> Vec<u8> {
 }
 
 #[allow(dead_code)]
-#[derive(Eq, PartialEq)]
+#[derive(Copy, Clone, Eq, PartialEq)]
 pub enum UpdateType {
     SyncCipherUpdate = 0,
     SyncCipherCreate = 1,
--- a/src/api/push.rs
+++ b/src/api/push.rs
@@ -1,0 +1,280 @@
+use reqwest::header::{ACCEPT, AUTHORIZATION, CONTENT_TYPE};
+use serde_json::Value;
+use tokio::sync::RwLock;
+
+use crate::{
+    api::{ApiResult, EmptyResult, UpdateType},
+    db::models::{Cipher, Device, Folder, Send, User},
+    util::get_reqwest_client,
+    CONFIG,
+};
+
+use once_cell::sync::Lazy;
+use std::time::{Duration, Instant};
+
+#[derive(Deserialize)]
+struct AuthPushToken {
+    access_token: String,
+    expires_in: i32,
+}
+
+#[derive(Debug)]
+struct LocalAuthPushToken {
+    access_token: String,
+    valid_until: Instant,
+}
+
+async fn get_auth_push_token() -> ApiResult<String> {
+    static PUSH_TOKEN: Lazy<RwLock<LocalAuthPushToken>> = Lazy::new(|| {
+        RwLock::new(LocalAuthPushToken {
+            access_token: String::new(),
+            valid_until: Instant::now(),
+        })
+    });
+    let push_token = PUSH_TOKEN.read().await;
+
+    if push_token.valid_until.saturating_duration_since(Instant::now()).as_secs() > 0 {
+        debug!("Auth Push token still valid, no need for a new one");
+        return Ok(push_token.access_token.clone());
+    }
+    drop(push_token); // Drop the read lock now
+
+    let installation_id = CONFIG.push_installation_id();
+    let client_id = format!("installation.{installation_id}");
+    let client_secret = CONFIG.push_installation_key();
+
+    let params = [
+        ("grant_type", "client_credentials"),
+        ("scope", "api.push"),
+        ("client_id", &client_id),
+        ("client_secret", &client_secret),
+    ];
+
+    let res = match get_reqwest_client().post("https://identity.bitwarden.com/connect/token").form(&params).send().await
+    {
+        Ok(r) => r,
+        Err(e) => err!(format!("Error getting push token from bitwarden server: {e}")),
+    };
+
+    let json_pushtoken = match res.json::<AuthPushToken>().await {
+        Ok(r) => r,
+        Err(e) => err!(format!("Unexpected push token received from bitwarden server: {e}")),
+    };
+
+    let mut push_token = PUSH_TOKEN.write().await;
+    push_token.valid_until = Instant::now()
+        .checked_add(Duration::new((json_pushtoken.expires_in / 2) as u64, 0)) // Token valid for half the specified time
+        .unwrap();
+
+    push_token.access_token = json_pushtoken.access_token;
+
+    debug!("Token still valid for {}", push_token.valid_until.saturating_duration_since(Instant::now()).as_secs());
+    Ok(push_token.access_token.clone())
+}
+
+pub async fn register_push_device(user_uuid: String, device: Device) -> EmptyResult {
+    if !CONFIG.push_enabled() {
+        return Ok(());
+    }
+    let auth_push_token = get_auth_push_token().await?;
+
+    //Needed to register a device for push to bitwarden :
+    let data = json!({
+        "userId": user_uuid,
+        "deviceId": device.push_uuid,
+        "identifier": device.uuid,
+        "type": device.atype,
+        "pushToken": device.push_token
+    });
+
+    let auth_header = format!("Bearer {}", &auth_push_token);
+
+    get_reqwest_client()
+        .post(CONFIG.push_relay_uri() + "/push/register")
+        .header(CONTENT_TYPE, "application/json")
+        .header(ACCEPT, "application/json")
+        .header(AUTHORIZATION, auth_header)
+        .json(&data)
+        .send()
+        .await?
+        .error_for_status()?;
+    Ok(())
+}
+
+pub async fn unregister_push_device(uuid: String) -> EmptyResult {
+    if !CONFIG.push_enabled() {
+        return Ok(());
+    }
+    let auth_push_token = get_auth_push_token().await?;
+
+    let auth_header = format!("Bearer {}", &auth_push_token);
+
+    match get_reqwest_client()
+        .delete(CONFIG.push_relay_uri() + "/push/" + &uuid)
+        .header(AUTHORIZATION, auth_header)
+        .send()
+        .await
+    {
+        Ok(r) => r,
+        Err(e) => err!(format!("An error occured during device unregistration: {e}")),
+    };
+    Ok(())
+}
+
+pub async fn push_cipher_update(
+    ut: UpdateType,
+    cipher: &Cipher,
+    acting_device_uuid: &String,
+    conn: &mut crate::db::DbConn,
+) {
+    // We shouldn't send a push notification on cipher update if the cipher belongs to an organization, this isn't implemented in the upstream server too.
+    if cipher.organization_uuid.is_some() {
+        return;
+    };
+    let user_uuid = match &cipher.user_uuid {
+        Some(c) => c,
+        None => {
+            debug!("Cipher has no uuid");
+            return;
+        }
+    };
+
+    for device in Device::find_by_user(user_uuid, conn).await {
+        let data = json!({
+            "userId": user_uuid,
+            "organizationId": (),
+            "deviceId": device.push_uuid,
+            "identifier": acting_device_uuid,
+            "type": ut as i32,
+            "payload": {
+                "Id": cipher.uuid,
+                "UserId": cipher.user_uuid,
+                "OrganizationId": (),
+                "RevisionDate": cipher.updated_at
+            }
+        });
+
+        send_to_push_relay(data).await;
+    }
+}
+
+pub async fn push_logout(user: &User, acting_device_uuid: Option<String>, conn: &mut crate::db::DbConn) {
+    if let Some(d) = acting_device_uuid {
+        for device in Device::find_by_user(&user.uuid, conn).await {
+            let data = json!({
+                "userId": user.uuid,
+                "organizationId": (),
+                "deviceId": device.push_uuid,
+                "identifier": d,
+                "type": UpdateType::LogOut as i32,
+                "payload": {
+                    "UserId": user.uuid,
+                    "Date": user.updated_at
+                }
+            });
+            send_to_push_relay(data).await;
+        }
+    } else {
+        let data = json!({
+            "userId": user.uuid,
+            "organizationId": (),
+            "deviceId": (),
+            "identifier": (),
+            "type": UpdateType::LogOut as i32,
+            "payload": {
+                "UserId": user.uuid,
+                "Date": user.updated_at
+            }
+        });
+        send_to_push_relay(data).await;
+    }
+}
+
+pub async fn push_user_update(ut: UpdateType, user: &User) {
+    let data = json!({
+        "userId": user.uuid,
+        "organizationId": (),
+        "deviceId": (),
+        "identifier": (),
+        "type": ut as i32,
+        "payload": {
+            "UserId": user.uuid,
+            "Date": user.updated_at
+        }
+    });
+
+    send_to_push_relay(data).await;
+}
+
+pub async fn push_folder_update(
+    ut: UpdateType,
+    folder: &Folder,
+    acting_device_uuid: &String,
+    conn: &mut crate::db::DbConn,
+) {
+    for device in Device::find_by_user(&folder.user_uuid, conn).await {
+        let data = json!({
+            "userId": folder.user_uuid,
+            "organizationId": (),
+            "deviceId": device.push_uuid,
+            "identifier": acting_device_uuid,
+            "type": ut as i32,
+            "payload": {
+                "Id": folder.uuid,
+                "UserId": folder.user_uuid,
+                "RevisionDate": folder.updated_at
+            }
+        });
+
+        send_to_push_relay(data).await;
+    }
+}
+
+pub async fn push_send_update(ut: UpdateType, send: &Send, conn: &mut crate::db::DbConn) {
+    if let Some(s) = &send.user_uuid {
+        for device in Device::find_by_user(s, conn).await {
+            let data = json!({
+                "userId": send.user_uuid,
+                "organizationId": (),
+                "deviceId": device.push_uuid,
+                "identifier": (),
+                "type": ut as i32,
+                "payload": {
+                    "Id": send.uuid,
+                    "UserId": send.user_uuid,
+                    "RevisionDate": send.revision_date
+                }
+            });
+
+            send_to_push_relay(data).await;
+        }
+    }
+}
+
+async fn send_to_push_relay(data: Value) {
+    if !CONFIG.push_enabled() {
+        return;
+    }
+
+    let auth_push_token = match get_auth_push_token().await {
+        Ok(s) => s,
+        Err(e) => {
+            debug!("Could not get the auth push token: {}", e);
+            return;
+        }
+    };
+
+    let auth_header = format!("Bearer {}", &auth_push_token);
+
+    if let Err(e) = get_reqwest_client()
+        .post(CONFIG.push_relay_uri() + "/push/send")
+        .header(ACCEPT, "application/json")
+        .header(CONTENT_TYPE, "application/json")
+        .header(AUTHORIZATION, auth_header)
+        .json(&data)
+        .send()
+        .await
+    {
+        error!("An error occured while sending a send update to the push relay: {}", e);
+    };
+}
--- a/src/config.rs
+++ b/src/config.rs
@@ -377,6 +377,16 @@ make_config! {
         /// Websocket port
         websocket_port:         u16,    false,  def,    3012;
     },
+    push {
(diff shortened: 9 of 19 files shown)


stefan0xC:
Thanks for the contribution. I have not looked at the changes yet but can you run `cargo fmt` and fix the errors reported by `cargo clippy` so the automatic checks won't fail?

GeekCornerGH:
> Thanks for the contribution. I have not looked at the changes yet but can you run `cargo fmt` and fix the errors reported by `cargo clippy` so the automatic checks won't fail? 

Sure thing, will do it later

GeekCornerGH:
Removed pretty much every error.
If someone can suggest changes to remove the last 2 ones, feel free to do so, not sure how to resolve them (removing return throw an error about response type for instance)

src/push.rs
```diff
@@ -0,0 +1,297 @@
...
+            "Id": cipher.uuid,
+            "UserId": cipher.user_uuid,
+            "OrganizationId": (),
+            "RevisionDate": cipher.updated_at
+        }
+    });
+
+    send_to_push_relay(data).await;
+}
+pub async fn push_logout(user: &User, acting_device_identifier: Option<String>, conn: &mut crate::db::DbConn) {
```
VarChar42, line 168: Some functions don't check for the 'push_enabled' config value at the beginning. Is this intentionally?

GeekCornerGH: Fair question, I guess I forgot to add back the function checking for enabled configuration :(

BlackDex (changes requested):
I'm looking at the code right now.
At a first glance the code looks ok, but have not looked detailed at it.

First things that i notice are the following.
There are two migrations done on the database, while that could be done in one go, which would make a `down.sql` a bit easier I think, and probably wanted.

Also, i have not checked what the effects are yet of this change, since there is a primary key on the old `UUID` and `USER_UUID` combined. I don't know if that will break stuff with this rename.
I still tend do keep `uuid` as is, and use `push_token` as a new column. Even though that deviates from upstream Bitwarden. But we can put that into the comments at the object.

I also see a lot of extra calls being done in regards to pushing these notifications.
Now every function which has a `nt.send...` also has a `push_...` in there.
Can't we merge this in some way and just have one entry point for sending out changes?
Like, and the `push::` to the `notifications.rs` file?
That would solve future notification updates or additions to just add one, instead of two calls every time.

migrations/mysql/2023-02-18-133907_create-new-uuid-table/up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE devices ADD uuid TEXT NOT NULL DEFAULT "00000000-0000-0000-0000-000000000000";
```
BlackDex, line 1: Does this really need to have a default? I'd rather have no default for this.
If we really want this to be a valid uuid, we need to solve this a different way. And that is why i think we need to rename this to `push_uuid` instead, and have it optional.
That way we can check if it is set or not.

src/api/core/accounts.rs
```diff
@@ -872,12 +882,77 @@ async fn rotate_api_key(data: JsonUpcase<SecretVerificationRequest>, headers: He
...
+) -> EmptyResult {
+    let data: PushTokenData = data.into_inner();
+    let token = data.pushToken.clone();
+    let mut device =
+        match Device::find_by_identifier_and_user(&headers.device.identifier, &headers.user.uuid, &mut conn).await {
+            Some(device) => device,
+            None => Device::new(identifier, headers.user.uuid.clone(), headers.device.name, headers.device.atype),
+        };
+    device.push_token = Some(token);
+    if device.uuid == *"00000000-0000-0000-0000-000000000000".to_string() {
```
BlackDex, line 916: If we change this to an optional db field, there is no need to do a string checking here, which is kinda expensive compared to checking for an `Option`

src/db/models/user.rs
```diff
@@ -9,6 +9,7 @@ db_object! {
     #[diesel(table_name = users)]
     #[diesel(treat_none_as_null = true)]
     #[diesel(primary_key(uuid))]
+    #[derive(Clone)]
```
BlackDex, line 12: Is this really needed? I think it works just fine without having this derive.

src/api/notifications.rs
```diff
@@ -278,7 +278,7 @@ fn create_ping() -> Vec<u8> {
 }
 
 #[allow(dead_code)]
-#[derive(Eq, PartialEq)]
+#[derive(Clone, Copy, Eq, PartialEq)]
```
BlackDex, line 281: This can probably also be solved if we merge the push and notify calls into one.
Then you could either change `ut` in `push.rs` to be an `i32` or we need to make sure we can cast/convert from an `&UpdateType` to an `i32`. Kinda the same as we do for the event logging.

BlackDex (changes requested):

src/api/notifications.rs
BlackDex: Or, now that i think if, that might be an issue too. Ill have to check a bit better.

BlackDex: I Think what a small modification we could re-define `ut` as an `i32` and pass it by reference for one, and let it consume by the other.

src/db/models/user.rs
GeekCornerGH: Good question, might forgot to remove it from debug tests

GeekCornerGH: Putting in resolved as this is fixed in my local branch

src/config.rs
```diff
@@ -377,6 +377,16 @@ make_config! {
         /// Websocket port
         websocket_port:         u16,    false,  def,    3012;
     },
+    push {
+        /// Enable push notifications
+        push_enabled:           bool,   false,  def,    false;
+        /// Push relay base uri
+        push_relay_uri:         String, false,  def,    "https://push.bitwarden.com".to_string();
```
BlackDex, line 384: Is this actually configurable, what are the alternatives?
Else we are probably better of having this not configurable.

src/push.rs
```diff
@@ -0,0 +1,302 @@
...
+    auth_header.push_str(&PUSH_TOKEN.lock().unwrap().access_token);
+    match get_reqwest_client()
+        .delete(CONFIG.push_relay_uri() + "/push/" + &uuid)
+        .header(AUTHORIZATION, auth_header)
+        .send()
+        .await
+    {
+        Ok(r) => r,
+        Err(e) => err!(format!("An error occured during device unregistration: {e}")),
+    };
```
BlackDex, line 127: Maybe `if let Err(e) = get_reqwest_client()` is better suited here then a match which doesn't do anything with the `Ok` result here.

src/push.rs
```diff
@@ -0,0 +1,302 @@
...
+}
+
+pub async fn unregister_push_device(uuid: String) -> EmptyResult {
+    if let Err(_e) = Ok::<(),()>(()) {
+        ();
+    };
+    get_auth_push_token().await?;
+
+    let mut auth_header = "Bearer ".to_string();
+    auth_header.push_str(&PUSH_TOKEN.lock().unwrap().access_token);
```
BlackDex, line 118:
```suggestion
let auth_header = format!("Bearer {}", &PUSH_TOKEN.lock().unwrap().access_token);
```

src/push.rs
```diff
@@ -0,0 +1,302 @@
...
+    access_token: String,
+    valid_until: Instant,
+}
+
+static PUSH_TOKEN: Lazy<Mutex<LocalAuthPushToken>> = Lazy::new(|| {
+    Mutex::new(LocalAuthPushToken {
+        access_token: String::new(),
+        valid_until: Instant::now(),
+    })
+});
```
BlackDex, line 35: I think an `RwLock` would be better here. A `Mutex` will lock reads and writes, while an `RwLock` will only lock during writes, which will only occur when the push relay token is expired.

See https://docs.rs/tokio/latest/tokio/sync/struct.RwLock.html (Though i do not know if the tokio RwLock can be used here, else use std).

BlackDex (changes requested):
Some more reviews from my side.
I haven't fully checked everything detailed though.

src/push.rs
```diff
@@ -0,0 +1,302 @@
...
+        .send()
+        .await?
+        .error_for_status()?;
+    Ok(())
+}
+
+pub async fn unregister_push_device(uuid: String) -> EmptyResult {
+    if let Err(_e) = Ok::<(),()>(()) {
+        ();
+    };
```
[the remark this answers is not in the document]
GeekCornerGH: Just ignores if there is an error, was meant to pass clippy tests

src/push.rs
```diff
@@ -0,0 +1,302 @@
...
+        return;
+    }
+
+    match get_auth_push_token().await {
+        Ok(_) => (),
+        Err(e) => {
+            debug!("Cannot get the auth push token: {}", e);
+            return;
+        }
+    };
```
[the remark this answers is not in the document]
GeekCornerGH: Yep, my bad

src/push.rs
```diff
@@ -0,0 +1,302 @@
...
+    let mut client_id = "installation.".to_string();
+    let client_secret = CONFIG.push_installation_key();
+
+    client_id.push_str(&installationid);
+
+    let mut params = HashMap::new();
+    params.insert("grant_type", "client_credentials");
+    params.insert("scope", "api.push");
+    params.insert("client_id", &client_id);
+    params.insert("client_secret", &client_secret);
```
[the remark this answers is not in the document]
GeekCornerGH: This pr contains my first lines of rust, so I used stackoverflow and other websites :/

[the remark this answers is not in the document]
BlackDex: Ah, no prob. Just wondered if there was a specific reason. But, an array would be enough i think.
```rust
let params = [
    ("grant_type", "client_credentials"),
    ("scope", "api.push"),
    ("client_id", &client_id),
    ("client_secret", &client_secret)
];
```

src/push.rs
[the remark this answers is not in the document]
BlackDex: An error from what/where? You can just use `return;` here i think instead of this.
Because you now check if there is an Err from an Ok. Which will never happen.

src/push.rs
dani-garcia: As far as I know, as long as you aren't holding the lock guard between await points, using the std locks is fine, I remember that being mentioned somewhere in the tokio docs. I think it should be there still somewhere.

We're only ever swapping values into the mutex so I don't think it'll matter much whether we use sync/async or mutex/rwlock.

That said, we should avoid unnecessary locks and unlocks, like there is on here:
https://github.com/dani-garcia/vaultwarden/blob/59da014e86f181531d3ff9ec5b3830c3bc6e19e8/src/push.rs#L66-L74

And replace it with something like this  (Haven't actually tested the code, just wrote it by hand now):

```rust
let valid_until = {
    let lock = PUSH_TOKEN.lock().unwrap();

    *lock.valid_until = Instant::now()
    .checked_add(Duration::new((json_pushtoken.expires_in / 2) as u64, 0)) // Token valid for half the specified time
    .unwrap();

    *lock.access_token = json_pushtoken.access_token;

    lock.valid_until.saturating_duration_since(Instant::now()).as_secs()
};
debug!("Token still valid for {}", valid_until);
```
Note how i've stored the lock guard inside an inner {} block, to make sure it gets dropped in the end. That way we are saving some lock/unlock cycles.

BlackDex: Doesn't the Mutex locks also on reads? In case of multiple push events they could be waiting on the other.

dani-garcia: Yes, but the lock is very short lived (we are only replacing a variable and that's it), so the chance of two events trying to lock the mutex at the same time is very low, and even when it happens, they would only lock it for a very short time.

At our scale I don't think the distinction between mutex and rwlock is going to cause any performance difference at all.

dani-garcia: For what it's worth, here's how i'd do it with a rwlock, and also returning the access token to avoid having to do another lock/unlock cycle (not tested):
```rust
pub async fn get_auth_push_token() -> ApiResult<String> {
    static PUSH_TOKEN: Lazy<RwLock<LocalAuthPushToken>> = Lazy::new(|| {
        RwLock::new(LocalAuthPushToken {
            access_token: String::new(),
            valid_until: Instant::now(),
        })
    });

    let push_token = PUSH_TOKEN.read().unwrap();
    if push_token.valid_until.saturating_duration_since(Instant::now()).as_secs() > 0 {
        debug!("Auth Push token still valid, no need for a new one");
        return Ok(push_token.access_token.clone());
    }
    drop(push_token); // Drop the read lock now

    let installationid = CONFIG.push_installation_id();
    let mut client_id = "installation.".to_string();
    let client_secret = CONFIG.push_installation_key();

    client_id.push_str(&installationid);

    let mut params = HashMap::new();
    params.insert("grant_type", "client_credentials");
    params.insert("scope", "api.push");
    params.insert("client_id", &client_id);
    params.insert("client_secret", &client_secret);

    let res = match get_reqwest_client().post("https://identity.bitwarden.com/connect/token").form(&params).send().await {
        Ok(r) => r,
        Err(e) => err!(format!("Error getting push token from bitwarden server: {e}")),
    };

    let json_pushtoken = match res.json::<AuthPushToken>().await {
        Ok(r) => r,
        Err(e) => err!(format!("Unexpected push token received from bitwarden server: {e}")),
    };

    let mut push_token = PUSH_TOKEN.write().unwrap();
    push_token.valid_until = Instant::now()
        .checked_add(Duration::new((json_pushtoken.expires_in / 2) as u64, 0)) // Token valid for half the specified time
        .unwrap();

    push_token.access_token = json_pushtoken.access_token;

    debug!("Token still valid for {}", push_token.valid_until.saturating_duration_since(Instant::now()).as_secs());
    Ok(push_token.access_token.clone())
}
```
Then you can do:

```rust
 let auth_push_token = get_auth_push_token().await?;

    //Needed to register a device for push to bitwarden :
    let data = json!({
        "userId": user_uuid,
        "deviceId": device.uuid,
        "identifier": device.identifier,
        "type": device.atype,
        "pushToken": device.push_token
    });

    let mut auth_header = "Bearer ".to_string();
    auth_header.push_str(&auth_push_token);

    get_reqwest_client()
        .... the rest
```

src/push.rs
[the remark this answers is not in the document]
GeekCornerGH: idk, clippy complained about that :(

src/push.rs
BlackDex: With v1.65.0 as MSRV, you can even get rid of `Lazy` now i think 😉

src/api/core/accounts.rs
```diff
@@ -872,12 +882,77 @@ async fn rotate_api_key(data: JsonUpcase<SecretVerificationRequest>, headers: He
     _api_key(data, true, headers, conn).await
 }
 
-#[get("/devices/knowndevice/<email>/<uuid>")]
-async fn get_known_device(email: String, uuid: String, mut conn: DbConn) -> JsonResult {
+#[get("/devices/knowndevice/<email>/<identifier>")]
+async fn get_known_device(email: String, identifier: String, mut conn: DbConn) -> JsonResult {
```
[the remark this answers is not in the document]
BlackDex: Thanks @justindbaur for letting us know. This probably need to be fixed outside of this PR I think. And we probably need to support both for a while.

src/push.rs
samb-devel: I'm trying to implement this and would need your support here. I have the following error message attached to each line implementing a route, for instance in `admin.rs` on the line:
`#[post("/users/<uuid>/deauth")]`

```
future cannot be sent between threads safely
within `[async block@src/api/admin.rs:382:1: 382:32]`, the trait `std::marker::Send` is not implemented for `std::sync::RwLockReadGuard<'_, LocalAuthPushToken>`
required for the cast from `[async block@src/api/admin.rs:382:1: 382:32]` to the object type `dyn futures::Future<Output = Outcome<rocket::Response<'__r>, rocket::http::Status, rocket::Data<'__r>>> + std::marker::Send`
```

Could you please help ? Thanks in advance

BlackDex: Try to use RwLock from tokio.

src/api/notifications.rs
samb-devel: We worked on other parts and just started that one. What would make more sense? To have both `push.rs` and `notifications.rs` merged into the same file? If you could briefly describe the best architecture on how to manage those calls and in which file they should be, we'll give a look. Thanks

src/config.rs
GeekCornerGH: You can using env variables

BlackDex: I bet you can. But I actually mean is this something you can change to something else, and have it all still work?
What are the alternatives if there are any. Because, if there arn't any alternatives, its better to NOT have this as a config option.

GeekCornerGH: Oh you're talking about the push_relay_uri option? I thought you were talking about the whole config block. Anyways this was on the official server so I added it there :)

BlackDex: I understand, but is there any other server you can configure and have it all still working?? I don't think it is of use to have this configuarble.

GeekCornerGH: Fair question, and to be honnest, I don't know haha, I can remove it if you want

stefan0xC: As far as I've looked into it, it won't work as because self-hosted servers have ["to relay their messages through the Bitwarden cloud-hosted instance. [Because] only Bitwarden is allowed to send notifications to the store-distributed Android and iOS mobile applications"](https://contributing.bitwarden.com/architecture/push-notifications/mobile/). 

Not sure how difficult it would be to setup your own push relay (and get the mobile clients to work). I imagine it would be rather tedious.

Would it make sense to have this as an internal configuration option so it's theoretically possible to configure this to another URL? (Only asking because it's also a [configuration option in Bitwarden server](https://bitwarden.com/help/configure-push-relay/))

migrations/mysql/2023-02-18-125735_push_uuid_table/up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE devices CREATE COLUMN push_uuid TEXT;
```
dpinse, line 1: MariaDB errors out with a migration syntax error should be `ALTER TABLE devices ADD COLUMN push_uuid TEXT;`

dpinse (changes requested):
MariaDB migration fails with synax error, should be `ALTER TABLE devices ADD COLUMN push_uuid TEXT;`

migrations/mysql/2023-02-18-125735_push_uuid_table/up.sql
BlackDex: I think this should still be changed to be one file, which is not done yet.
Do you have a more detailed message maybe?

dpinse: [2023-03-14 12:11:43.582][panic][ERROR] thread 'main' panicked at 'Error running migrations: QueryError(DieselMigrationName { name: "2023-02-18-125735_push_uuid_table", version: MigrationVersion("20230218125735") }, DatabaseError(Unknown, "You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'CREATE COLUMN push_uuid TEXT' at line 1"))': src/db/mod.rs:471
   0: std::backtrace::Backtrace::create
   1: vaultwarden::init_logging::{{closure}}
   2: std::panicking::rust_panic_with_hook
   3: std::panicking::begin_panic_handler::{{closure}}
   4: std::sys_common::backtrace::__rust_end_short_backtrace
   5: rust_begin_unwind
   6: core::panicking::panic_fmt
   7: core::result::unwrap_failed
   8: vaultwarden::db::mysql_migrations::run_migrations
   9: vaultwarden::db::DbPool::from_config
  10: vaultwarden::main::{{closure}}
  11: tokio::runtime::park::CachedParkThread::block_on
  12: tokio::runtime::scheduler::multi_thread::MultiThread::block_on
  13: tokio::runtime::runtime::Runtime::block_on
  14: rocket::async_main
  15: vaultwarden::main
  16: std::sys_common::backtrace::__rust_begin_short_backtrace
  17: std::rt::lang_start::{{closure}}
  18: std::rt::lang_start_internal
  19: main

GeekCornerGH: Good to know, could you test on Postgres too?

tessus: `CREATE COLUMN` is not a valid SQL statement. @dpinse is correct, it must be `ADD COLUMN`.

src/config.rs
tessus: I think it makes sense adding it as an internal config option. This makes it more future-proof, in case bw allows to specify a relay at one point.
But I don't have any say, so my opinion is probabaly moot. ;-)

BlackDex:
@GeekCornerGH , my apologies already for my changes, hope it won't be to hard to merge hehe.

migrations/postgresql/2023-02-18-125735_push_uuid_table/up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE devices CREATE COLUMN push_uuid TEXT;
```
tessus, line 1:
```suggestion
ALTER TABLE devices ADD COLUMN push_uuid TEXT;
```

migrations/sqlite/2023-02-18-125735_push_uuid_table/up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE devices CREATE COLUMN push_uuid TEXT;
```
tessus, line 1:
```suggestion
ALTER TABLE devices ADD COLUMN push_uuid TEXT;
```

migrations/mysql/2023-02-18-125735_push_uuid_table/up.sql
tessus, line 1:
```suggestion
ALTER TABLE devices ADD COLUMN push_uuid TEXT;
```

tessus (commented):
The SQL statements are wrong. I am not sure how they ever ran successfully during a migration.

The proper statement is `ALTER TABLE bla ADD COLUMN ...`

GeekCornerGH:
Wooops, I guess git didn't caught the changes 😂 I'll rebase anyways

src/main.rs
```diff
@@ -417,6 +418,22 @@ async fn check_data_folder() {
...
 }
 
+fn check_push_config() {
+    if CONFIG.push_enabled()
+        && (CONFIG.push_installation_id() == String::new() || CONFIG.push_installation_key() == String::new())
+    {
+        error!(
+            "Misconfigured Push Notification service\n\
+            ########################################################################################\n\
+            # It looks like you enabled Push Notification feature, but you didn't configured the   #\n\
```
tessus, line 428:
```suggestion
            # It looks like you enabled Push Notification feature, but you didn't configure the    #\n\
```

Apart from the fact that this should go in `config.rs`, there's also a typo.

src/main.rs
```diff
@@ -417,6 +418,22 @@ async fn check_data_folder() {
...
+            # value. Make sure the installation id and key from https://bitwarden.com/host are     #\n\
```
tessus, line 429: Apart from the fact that this should go in `config.rs`, there's also a typo and I have improved the wording.

```suggestion
            # It looks like you enabled Push Notification feature, but didn't configure it         #\n\
            # properly. Make sure the installation id and key from https://bitwarden.com/host are  #\n\
```

I hope the border (the `#`s) is aligned, because github splits the lines, even in unified view.

GeekCornerGH:
@tessus since GitHub won't allow me to request 2 reviewers at one time, I just tag you haha, is everything fine now?

tessus (commented):
I think there are still 2 points outstanding, from other reviewers:

1. don't add an option in config.rs for `PUSH_RELAY_URI` (or at least it should not be public for now)
2. merge the function calls

ad 1) currently there is only one push server and setting it to anything other than the bw address won't work. It is too error prone to allow users to change this value. However, in the code the variable should still be used, since bw might introduce other servers or allow customers to setup their own push server (which would also require a change in he client code).

ad 2) currently there are always 2 fuction calls, as an example:

```
send_folder_update(...)
push_folder_update(...)
```

If I understood BlackDex's comment correctly, the second call should be added to the first one:

```
fn send_folder_update(...) {
    original code;
    
    if (PUSH_ENABLED) {
        push_folder_update(...);
    }
}
```

But I am not the right person to review the entire PR. Someone who actually knows Rust (I only dabble with it) and the code base will hvae to do a proper review.

samb-devel:
Thank you @tessus for your remarks.

@BlackDex : could you please confirm that the code suggested by @tessus is what you inteded? Can we simply call `push_folder_update(..)` from within `send_folder_update` which is inside the `impl WebSocketUsers` block without any problem?

Thank you in advance for your feedback

BlackDex:
> Thank you @tessus for your remarks.
> 
> @BlackDex : could you please confirm that the code suggested by @tessus is what you inteded? Can we simply call `push_folder_update(..)` from within `send_folder_update` which is inside the `impl WebSocketUsers` block without any problem?
> 
> Thank you in advance for your feedback

My main point is i don't like to have two functions being called at all those functions while using the same arguments.
I actually have not yet looked into the best way to do this. Maybe we need to create an overlay on both, so that in the future it could be easy to add a new way.

But i do think the `nt` can be used for both to simplify and consolidate this.

samb-devel:
@BlackDex thanks for the answer. I get your point. My main concern is that we're not very deep into rust development, We're happy to learn and do, we just need a bit of guidance on the best practices and on how Vaultwarden may develop.

Does it make sense to keep push calls and websocket notifications in two separate files, or should we merge them `push.rs` and `notifications.rs`?
Do you see any possible issue in calling the `push_folder_update` from within the `impl WebSocketUsers` block, or are there issues of any kind and that should be avoided?

tessus:
> Does it make sense to keep push calls and websocket notifications in two separate files, or should we merge them `push.rs` and `notifications.rs`?

I think this is up to you. Everyone has their preference. I don't see an issue to have 2 files. I suspect other reviewers would have already objected otherwise.

> Do you see any possible issue in calling the `push_folder_update` from within the `impl WebSocketUsers` block, or are there issues of any kind and that should be avoided?

Personally I'd just go for it. It's just a bit of refactoring. But BlackDex also mentioned that a wrapper might be beneficial. However, as a first step, I'd just try adding it to the impl block and see what happens. ;-)

BlackDex:
> @BlackDex thanks for the answer. I get your point. My main concern is that we're not very deep into rust development, We're happy to learn and do, we just need a bit of guidance on the best practices and on how Vaultwarden may develop.
> 
> Does it make sense to keep push calls and websocket notifications in two separate files, or should we merge them `push.rs` and `notifications.rs`? Do you see any possible issue in calling the `push_folder_update` from within the `impl WebSocketUsers` block, or are there issues of any kind and that should be avoided?

While I would like to consolidate this to make it a bit more simple. It isn't something i would trip over for an approval on my side. It can be optimized later when tested by others.

GeekCornerGH:
Talking about testing, while implemented, I couldn't get Send Notifications to work, while I had the same issue on the official server fyi

BlackDex (changes requested):

src/api/core/accounts.rs
```diff
@@ -930,3 +941,62 @@ impl<'r> FromRequest<'r> for KnownDevice {
...
+        error!("An error occured while trying to save the device push token: {}", e);
+        return Err(e);
+    }
+    if CONFIG.push_enabled() {
+        if let Err(e) = register_push_device(headers.user.uuid, device).await {
+            error!("An error occured while proceeding registration of a device: {}", e);
+        };
+    }
+
+    Ok(())
```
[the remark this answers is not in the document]
GeekCornerGH: I guess it is needed. Else, it won't compile at all.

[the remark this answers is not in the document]
BlackDex: I mean, you removed function from an other file, and there this function returned a JSON.
Now it doesn't return a JSON anymore, just an EmptyResult (aka 200).

I'm not sure if this is correct. It could be. But i just wondered because it did before.

BlackDex (changes requested):

GeekCornerGH:
Marking this as ready for review, since I can't reproduce issues I got on my windows pc on a Linux laptop

tessus (commented):
The PR hasn't changed since last time.

The function calls are still separate.

stefan0xC (changes requested):
While looking at the code I noticed a few smaller and some bigger issues and thus have some suggestions and also a few questions. Not that since I don't have a hosting id and key from Bitwarden I could not actually test if the push notifications work. But running it **without** the feature enabled I noticed that there are some checks missing:

```
[2023-04-03 22:23:16.325][request][INFO] PUT /api/ciphers/e9d45f5f-f619-43c9-9f36-f70f573f303e
[2023-04-03 22:23:16.333][reqwest::connect][DEBUG] starting new connection: https://identity.bitwarden.com/
[2023-04-03 22:23:16.339][trust_dns_proto::xfer::dns_handle][DEBUG] querying: identity.bitwarden.com A
[2023-04-03 22:23:16.339][trust_dns_resolver::name_server::name_server_pool][DEBUG] sending request: [Query { name: Name("identity.bitwarden.com"), query_type: A, query_class: IN }]
[2023-04-03 22:23:16.339][trust_dns_resolver::name_server::name_server][DEBUG] reconnecting: NameServerConfig { socket_addr: 127.0.0.53:53, protocol: Udp, tls_dns_name: None, trust_nx_responses: false, bind_addr: None }
[2023-04-03 22:23:16.339][trust_dns_proto::xfer][DEBUG] enqueueing message:QUERY:[Query { name: Name("identity.bitwarden.com"), query_type: A, query_class: IN }]
[2023-04-03 22:23:16.339][trust_dns_proto::udp::udp_client_stream][DEBUG] final message: ; header 49878:QUERY:RD:NoError:QUERY:0/0/0
; query
;; identity.bitwarden.com. IN A

[2023-04-03 22:23:16.339][trust_dns_proto::udp::udp_stream][DEBUG] created socket successfully
[2023-04-03 22:23:16.356][trust_dns_proto::udp::udp_client_stream][DEBUG] received message id: 49878
[2023-04-03 22:23:16.357][trust_dns_resolver::error][DEBUG] Response:; header 49878:RESPONSE:RD,RA:NoError:QUERY:2/0/0
; query
;; identity.bitwarden.com. IN A
; answers 2
identity.bitwarden.com. 177 IN A 104.18.13.33
identity.bitwarden.com. 177 IN A 104.18.12.33
; nameservers 0
; additionals 0

[2023-04-03 22:23:16.357][trust_dns_resolver::error][DEBUG] Response:; header 49878:RESPONSE:RD,RA:NoError:QUERY:2/0/0
; query
;; identity.bitwarden.com. IN A
; answers 2
identity.bitwarden.com. 177 IN A 104.18.13.33
identity.bitwarden.com. 177 IN A 104.18.12.33
; nameservers 0
; additionals 0

[2023-04-03 22:23:16.681][trust_dns_proto::xfer::dns_exchange][DEBUG] io_stream is done, shutting down
[2023-04-03 22:23:16.681][vaultwarden::api::push][ERROR] Unexpected push token received from bitwarden server: error decoding response body: missing field `access_token` at line 1 column 26
[2023-04-03 22:23:16.681][vaultwarden::api::push][DEBUG] Could not get the auth push token: {"ErrorModel":{"Message":"Unexpected push token received from bitwarden server: error decoding response body: missing field `access_token` at line 1 column 26","Object":"error"},"ExceptionMessage":null,"ExceptionStackTrace":null,"InnerExceptionMessage":null,"Message":"Unexpected push token received from bitwarden server: error decoding response body: missing field `access_token` at line 1 column 26","Object":"error","ValidationErrors":{"":["Unexpected push token received from bitwarden server: error decoding response body: missing field `access_token` at line 1 column 26"]},"error":"","error_description":""}
```

So before this gets merged I think we should check if there are no unintended side effects like this when this feature is **not** enabled.

tessus:
> So before this gets merged I think we should check if there are no unintended side effects like this when this feature is not enabled.

I don't want to be a stickler, but I did mention the following change:

```
fn send_folder_update(...) {
    original code;
    
    if (PUSH_ENABLED) {
        push_folder_update(...);
    }
}
```

This would have solved this issue already, since the `push_*` functions don't test for the service being enabled....

src/api/admin.rs
```diff
@@ -396,14 +399,21 @@ async fn delete_user(uuid: String, token: AdminToken, mut conn: DbConn) -> Empty
...
     let mut user = get_user_or_404(&uuid, &mut conn).await?;
-    Device::delete_all_by_user(&user.uuid, &mut conn).await?;
-    user.reset_security_stamp();
-
-    let save_result = user.save(&mut conn).await;
-
     nt.send_logout(&user, None).await;
+    push_logout(&user, None, &mut conn).await; // Send logout notifications before deleting devices
 
-    save_result
```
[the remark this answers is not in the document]
GeekCornerGH: Would make more sense to group both call in the first place, as database access is needed for push

src/api/core/accounts.rs
```diff
@@ -930,3 +939,64 @@ impl<'r> FromRequest<'r> for KnownDevice {
...
+        Err(e) => error!("{}", e),
+        Ok(_r) => (),
+    };
+    if let Some(device) = Device::find_by_uuid(&uuid, &mut conn).await {
+        if let Err(e) = unregister_push_device(device.uuid).await {
+            error!("{}", e);
+        };
+    };
+    ""
+}
```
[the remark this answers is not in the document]
GeekCornerGH: It was implemented in the server code. So I added it there too, in order to have it ready in case it get implemented somewhere

src/db/models/device.rs
```diff
@@ -175,4 +196,15 @@ impl Device {
                 .from_db()
         }}
     }
+    pub async fn find_push_device_by_user(user_uuid: &str, conn: &mut DbConn) -> Option<Self> {
```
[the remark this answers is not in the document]
GeekCornerGH: Figured out that's how Bitwarden official server works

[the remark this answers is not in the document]
stefan0xC: I don't think [this is how the official server works](https://contributing.bitwarden.com/architecture/push-notifications/mobile#using-the-push-token-to-send-notifications-to-the-device). I've checked where this is called and let me clarify my point: This function is only used used when clearing the token from the `/admin` page via `deauth_user`. However this function only unregisters the result of this function (i.e. the first one it finds) and then it deletes all Devices for a user. However, according to my understanding it should be possible to add multiple push devices.

src/api/admin.rs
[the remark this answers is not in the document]
stefan0xC: I don't understand your explanation. I might be wrong but I think the order is intended to do the operation first and only then notify the user. If the notification fails the operation will not be executed.

Zoruk:
@GeekCornerGH Do you want some help to clean up your MR and fix the requested changes ?

GeekCornerGH:
Most changes aren't pushed yet, they are on my local branch. But thanks for your help. If you want to make wrapper, feel free to make a pull request to my GitHub branch

src/api/admin.rs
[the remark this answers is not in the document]
stefan0xC: @GeekCornerGH After thinking about it some more (and talking about it in the matrix chatroom) I think I might have been wrong about this and your explanation about the order makes a lot more sense now. Sorry for the inconvenience. :pensive: 
(The unregistering of only 1 device still seems wrong though.)

src/api/core/accounts.rs
```diff
@@ -930,3 +935,68 @@ impl<'r> FromRequest<'r> for KnownDevice {
...
+    uuid: String,
+    data: JsonUpcase<PushToken>,
+    headers: Headers,
+    mut conn: DbConn,
+) -> EmptyResult {
+    if !CONFIG.push_enabled() {
+        return Ok(());
+    }
+
+    let token = data.0.data.PushToken;
```
BlackDex, line 961:
```suggestion
let data = data.into_inner().data;
let token = data.PushToken;
```

src/api/push.rs
```diff
@@ -0,0 +1,297 @@
...
+        }
+    });
+
+    send_to_push_relay(data).await;
+}
+
+pub async fn push_logout(user: &User, acting_device_uuid: Option<String>, conn: &mut crate::db::DbConn) {
+    let data: JsonValue;
+    if let Some(d) = acting_device_uuid {
+        if let Some(device) = Device::find_by_uuid_and_user(d.as_str(), &user.uuid, conn).await {
```
BlackDex, line 172:
```suggestion
        if let Some(device) = Device::find_by_uuid_and_user(&d, &user.uuid, conn).await {
```

src/api/push.rs
```diff
@@ -0,0 +1,297 @@
...
+    db::models::{Cipher, Device, Folder, Send, User},
+    util::get_reqwest_client,
+    CONFIG,
+};
+
+use once_cell::sync::Lazy;
+use std::time::{Duration, Instant};
+
+#[derive(Deserialize, Debug)]
+#[allow(non_snake_case)]
```
BlackDex, line 17: This can be removed i think

src/api/push.rs
```diff
@@ -0,0 +1,297 @@
...
+            "UserId": folder.user_uuid,
+            "RevisionDate": folder.updated_at
+        }
+    });
+
+    send_to_push_relay(data).await;
+}
+
+pub async fn push_send_update(ut: UpdateType, send: &Send) {
+    if send.user_uuid.is_none() {
```
BlackDex, line 251: Maybe a good idea too also add the check if push is enabled or not?

```suggestion
    if !CONFIG.push_enabled() || send.user_uuid.is_none() {
```
That prevents the code from continuing and in the end still being aborted.

src/api/push.rs
```diff
@@ -0,0 +1,297 @@
...
+    send_to_push_relay(data).await;
+}
+
+pub async fn push_folder_update(
+    ut: UpdateType,
+    folder: &Folder,
+    acting_device_uuid: &String,
+    conn: &mut crate::db::DbConn,
+) {
+    let device = match Device::find_by_uuid_and_user(acting_device_uuid, &folder.user_uuid, conn).await {
```
BlackDex, line 226: Add an if check here if push is enabled or not, will prevent database lookups if not enabled in the first place.

src/api/push.rs
```diff
@@ -0,0 +1,297 @@
...
+                "UserId": user.uuid,
+                "Date": user.updated_at
+            }
+        });
+        send_to_push_relay(data).await;
+    }
+}
+
+pub async fn push_user_update(ut: UpdateType, user: &User) {
+    let data = json!({
```
BlackDex, line 205: Add an if check here if push is enabled or not.

src/api/push.rs
```diff
@@ -0,0 +1,297 @@
...
+            "OrganizationId": (),
+            "RevisionDate": cipher.updated_at
+        }
+    });
+
+    send_to_push_relay(data).await;
+}
+
+pub async fn push_logout(user: &User, acting_device_uuid: Option<String>, conn: &mut crate::db::DbConn) {
+    let data: JsonValue;
```
BlackDex, line 170: Add an if check here if push is enabled or not, will prevent database lookups if not enabled in the first place.

src/api/push.rs
```diff
@@ -0,0 +1,297 @@
...
+}
+
+pub async fn push_cipher_update(
+    ut: UpdateType,
+    cipher: &Cipher,
+    acting_device_uuid: &String,
+    conn: &mut crate::db::DbConn,
+) {
+    // We shouldn't send a push notification on cipher update if the cipher belongs to an organization, this isn't implemented in the upstream server too.
+    if cipher.organization_uuid.is_some() {
```
BlackDex, line 133: Add an if check here if push is enabled or not, will prevent database lookups if not enabled in the first place.

BlackDex (changes requested):
In overall i think it looks nice (but i think i have mentioned that before).
Some small adjustments to be done I think. Though I did not yet builded this PR and gave it a test run my self, so I can't comment on that yet.

It also needs a rebase (as already mentioned on Matrix).

BlackDex (changes requested):
In overall i think it looks nice (but i think i have mentioned that before).
Some small adjustments to be done I think. Though I did not yet builded this PR and gave it a test run my self, so I can't comment on that yet.

It also needs a rebase (as already mentioned on Matrix).

BlackDex (changes requested):
In overall i think it looks nice (but i think i have mentioned that before).
Some small adjustments to be done I think. Though I did not yet builded this PR and gave it a test run my self, so I can't comment on that yet.

It also needs a rebase (as already mentioned on Matrix).

BlackDex (changes requested):
In overall i think it looks nice (but i think i have mentioned that before).
Some small adjustments to be done I think. Though I did not yet builded this PR and gave it a test run my self, so I can't comment on that yet.

It also needs a rebase (as already mentioned on Matrix).

BlackDex (changes requested):
In overall i think it looks nice (but i think i have mentioned that before).
Some small adjustments to be done I think. Though I did not yet builded this PR and gave it a test run my self, so I can't comment on that yet.

It also needs a rebase (as already mentioned on Matrix).

src/api/push.rs
tessus: @BlackDex I don't think this is necessary, since it is only called when push is enabled. or am I missing something here?

```
        if CONFIG.push_enabled() {
            push_user_update(ut, user).await;
        }
```

This is also true for the other function as far as I can tell...

tessus:
@GeekCornerGH It seems you are merging main into your PR. You need to rebase it. Currently your PR has so many changes that do not even belong to this PR that it makes reviewing it a real mess. e.g. you added 2 commits. None of them had anything to do with BlackDex's review. But I had to read all of it to find that out, and I might have missed somthing since BlackDex had quite a few comments.

GeekCornerGH:
> @GeekCornerGH It seems you are merging main into your PR. You need to rebase it. Currently your PR has so many changes that do not even belong to this PR that it makes reviewing it a real mess. e.g. you added 2 commits. None of them had anything to do with BlackDex's review. But I had to read all of it to find that out, and I might have missed somthing since BlackDex had quite a few comments.

I'm in fact using rebase, not merge. Also I missed BlackDex's comments, as my phone didn't showed the reviews themselves. Checking this rn

tessus:
> I'm in fact using rebase, not merge.

This is very strange then. Maybe I clicked on the wrong button. I think I clicked on `Compare` instead of the `Files changed` button at the top.

In that case my apologies. My bad.

src/api/push.rs
GeekCornerGH: Already done [here](https://github.com/dani-garcia/vaultwarden/blob/4ec5334a77d471dd8f6082f482b80b785e4f65e0/src/api/push.rs#L271)

src/api/push.rs
GeekCornerGH: Check is done [here](https://github.com/dani-garcia/vaultwarden/blob/4ec5334a77d471dd8f6082f482b80b785e4f65e0/src/api/push.rs#L271)

src/api/push.rs
GeekCornerGH: Deserialize is required fyi

src/db/models/device.rs
```diff
@@ -175,4 +196,15 @@ impl Device {
...
+        db_run! { conn: {
+            devices::table
+                .filter(devices::user_uuid.eq(user_uuid))
+                .filter(devices::push_token.is_not_null())
+                .order(devices::updated_at.desc())
+                .first::<DeviceDb>(conn)
```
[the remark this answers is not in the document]
tessus: I guess this is the last issue that has to be resolved, correct?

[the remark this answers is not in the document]
tessus: @GeekCornerGH I saw that you pushed a change a while back, but you are using `.first::` in the query.
This only returns one record. Shouldn't the query return all records (devices that belong to user X)?
Then you can iterate over those records and clear the token.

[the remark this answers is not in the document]
BlackDex: @GeekCornerGH I quickly tested it today, and this is not working. You need to extract all `push_tokens` and send it to each push_token. Else only one device will receive an update, and the other will not.

I tested this with two identical Virtual Android Devices, just one device receives this update.
So this definitely needs some work.

GeekCornerGH:
Latest changes were tested against 2 cloned AVD, running the latest apk 2023.05.0

src/api/admin.rs
```diff
@@ -402,14 +402,22 @@ async fn delete_user(uuid: &str, token: AdminToken, mut conn: DbConn) -> EmptyRe
 #[post("/users/<uuid>/deauth")]
 async fn deauth_user(uuid: &str, _token: AdminToken, mut conn: DbConn, nt: Notify<'_>) -> EmptyResult {
     let mut user = get_user_or_404(uuid, &mut conn).await?;
-    Device::delete_all_by_user(&user.uuid, &mut conn).await?;
-    user.reset_security_stamp();
 
-    let save_result = user.save(&mut conn).await;
+    nt.send_logout(&user, None, &mut conn).await;
```
BlackDex, line 406: Is there a reason for this to be at the top right now?
It was at the bottom before?
I'm not really seeing a good reason for this to be here actually.

src/config.rs
BlackDex: This currently isn't an internal config, and if we keep it here, i would also suggest to put it back into the `.env.template`, just to make it match everything.

So, I'm fine leaving it here, but then also add it to the `.env.template` file again,

.env.template
```diff
@@ -72,6 +72,11 @@
 # WEBSOCKET_ADDRESS=0.0.0.0
 # WEBSOCKET_PORT=3012
 
+## Enables push notifications (requires key and id from https://bitwarden.com/host)
+# PUSH_ENABLED=true
+# PUSH_INSTALLATION_ID=CHANGEME
+# PUSH_INSTALLATION_KEY=CHANGEME
+
```
BlackDex, line 79: Please add the `push_relay_uri` env here again, since we seem to keep it in the config anyways. Not mentioning it here will not help. Just add a good comment to mention that it is the default, and shouldn't be changed at all, unless there are special reason too.

src/api/core/accounts.rs
```diff
@@ -930,3 +935,69 @@ impl<'r> FromRequest<'r> for KnownDevice {
...
+
+    if let Some(device) = Device::find_by_uuid(&uuid, &mut conn).await {
+        Device::clear_push_token_by_uuid(&uuid, &mut conn).await?;
+        unregister_push_device(device.uuid).await?;
+    }
+
+    Ok(())
+}
+
+// On upstream server, both PUT and POST are declared.
```
BlackDex, line 999: If there also is a `PUT`, then i would suggest to add this also. Why leave it out? Might be that they use it somewhere for some reason?

Could be done the same as for `post_device_token` and `put_device_token`.

BlackDex (changes requested):

BlackDex:
Looks nice, some final tweaks i think.

src/api/admin.rs
GeekCornerGH: Indeed there is a reason. Calls between websockets and push got merged, so we have first to send logout, before unregistering push devices.

src/api/core/accounts.rs
GeekCornerGH: The post request was actually above.

src/api/core/accounts.rs
```diff
@@ -930,3 +935,64 @@ impl<'r> FromRequest<'r> for KnownDevice {
...
+        unregister_push_device(device.uuid).await?;
+    }
+
+    Ok(())
+}
+
+// On upstream server, both PUT and POST are declared. Implementing the POST method in case it would be useful somewhere
+#[post("/devices/identifier/<uuid>/clear-token")]
+async fn clear_device_token_post(uuid: &str, conn: DbConn) -> EmptyResult {
+    clear_device_token(uuid, conn).await
```
tessus, line 997: To make it consistent with previous function names, shouldn't this be named `clear_device_token_put`?

Btw, great job. I can't wait to see this merged!

BlackDex (approved):

src/api/core/accounts.rs
tessus: BlackDex already approved it, so please ignore my comment.

BlackDex: While that would be better, and it still can be changed, it isn't a major issue.

Also i would like @dani-garcia look on this whole PR first too.

BlackDex (approved):

dani-garcia (approved):

tetsumaki:
I think you should delete `EXPOSE 3012` from many Dockerfile here : https://github.com/dani-garcia/vaultwarden/tree/main/docker

BlackDex:
> I think you should delete `EXPOSE 3012` from many Dockerfile here : https://github.com/dani-garcia/vaultwarden/tree/main/docker

For now we don't, since we need a transition from the old to the new way of websockets. And we don't want to break current installations.

TryAnixx:
its not working for me: https://github.com/dani-garcia/vaultwarden/issues/6328 can somebody help?

PR #3307, opened 2023-03-05
Add HEAD routes to avoid spurious error messages

jjlin:
Rocket automatically implements a HEAD route when there's a matching GET route, but relying on this behavior also means a spurious error gets logged due to <https://github.com/SergioBenitez/Rocket/issues/1098>.

Add explicit HEAD routes for `/` and `/alive` to prevent uptime monitoring services from generating error messages like `No matching routes for HEAD /`. With these new routes, `HEAD /` only checks that the server can respond over the network, while `HEAD /alive` also checks that the database connection is alive, similar to `GET /alive`.

### DIFF
--- a/src/api/web.rs
+++ b/src/api/web.rs
@@ -4,7 +4,7 @@ use rocket::{fs::NamedFile, http::ContentType, response::content::RawHtml as Htm
 use serde_json::Value;
 
 use crate::{
-    api::{core::now, ApiResult},
+    api::{core::now, ApiResult, EmptyResult},
     error::Error,
     util::{Cached, SafeString},
     CONFIG,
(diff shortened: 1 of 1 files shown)


BlackDex (approved):
You beat me to it.

BlackDex:
@jjlin Why not use `EmptyResult` instead of allowing a specific clippy lint globally btw?
It may be an erroneous message for these two functions, but not for everything.

jjlin:
> @jjlin Why not use `EmptyResult` instead of allowing a specific clippy lint globally btw? It may be an erroneous message for these two functions, but not for everything. And, it can be easily solved by returning `Ok(())`

It didn't occur to me, thanks.

BlackDex (approved):

dani-garcia (approved):

PR #3310, opened 2023-03-05
Upd Crates, Rust, MSRV, GHA and remove Backtrace

BlackDex:
- Changed MSRV to v1.65. Discussed this with @dani-garcia, and we will support **N-2**. This is/will be the same as for the `time` crate we use. Also updated the wiki regarding this https://github.com/dani-garcia/vaultwarden/wiki/Building-binary
- Removed backtrace crate in favor of `std::backtrace` stable since v1.65
- Updated Rust to v1.67.1
- Updated all the crates
- Updated the GHA action versions
- Adjusted the GHA MSRV build to extract the MSRV from `Cargo.toml`


dani-garcia (approved):
Needs to resolve the conflicts caused by the argon pr

PR #3329, opened 2023-03-11
Add support for `/api/devices/knowndevice` with HTTP header params

jjlin:
Upstream PR: https://github.com/bitwarden/server/pull/2682

### DIFF
--- a/src/api/core/accounts.rs
+++ b/src/api/core/accounts.rs
@@ -12,6 +12,11 @@ use crate::{
     mail, CONFIG,
 };
 
+use rocket::{
+    http::Status,
+    request::{FromRequest, Outcome, Request},
+};
+
 pub fn routes() -> Vec<rocket::Route> {
     routes![
         register,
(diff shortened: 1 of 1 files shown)


src/api/core/accounts.rs
```diff
@@ -872,12 +878,47 @@ async fn rotate_api_key(data: JsonUpcase<SecretVerificationRequest>, headers: He
...
+    email: String,
+    uuid: String,
+}
+
+#[rocket::async_trait]
+impl<'r> FromRequest<'r> for KnownDevice {
+    type Error = &'static str;
+
+    async fn from_request(req: &'r Request<'_>) -> Outcome<Self, Self::Error> {
+        let email = if let Some(email) = req.headers().get_one("X-Request-Email") {
```
BlackDex, line 907: This value is base64 encoded. Shouldn't we decode that here?

BlackDex (changes requested):

src/api/core/accounts.rs
jjlin: Sigh, forgot about that. Thanks.

BlackDex (approved):

dani-garcia (approved):

PR #3332, opened 2023-03-11
Merge ClientIp with Headers.

BlackDex:
Since we now use the `ClientIp` Guard on a lot more places, it also increases the size of binary, and the macro generated code because of this extra Guard. By merging the `ClientIp` Guard with the several `Header` guards we have it reduces the amount of code generated (including LLVM IR), but also a small speedup in build time.

I also spotted some small `json!()` optimizations which also reduced the amount of code generated.

There was no decrease in the speed of the code as far as i could tell. It btw also lowered the amount of allocations done.

### DIFF
--- a/src/api/notifications.rs
+++ b/src/api/notifications.rs
@@ -10,8 +10,7 @@ use std::{
 use chrono::NaiveDateTime;
 use futures::{SinkExt, StreamExt};
 use rmpv::Value;
-use rocket::{serde::json::Json, Route};
-use serde_json::Value as JsonValue;
+use rocket::Route;
 use tokio::{
     net::{TcpListener, TcpStream},
     sync::mpsc::Sender,
@@ -23,13 +22,12 @@ use tokio_tungstenite::{
 
 use crate::{
     api::EmptyResult,
-    auth::Headers,
     db::models::{Cipher, Folder, Send, User},
     Error, CONFIG,
 };
 
 pub fn routes() -> Vec<Route> {
-    routes![negotiate, websockets_err]
+    routes![websockets_err]
 }
 
 #[get("/hub")]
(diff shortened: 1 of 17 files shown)


src/api/notifications.rs
```diff
@@ -51,29 +49,6 @@ fn websockets_err() -> EmptyResult {
     }
 }
 
-#[post("/hub/negotiate")]
-fn negotiate(_headers: Headers) -> Json<JsonValue> {
```
dani-garcia, line 55: Is this no longer used by the clients?

BlackDex: I can't find that anywhere in the client/mobile code at all.
I could be wrong.

dani-garcia: They are using SignalR so there is a chance that the library is using it internally, but the ws connection works without it, maybe it's not required anymore.

BlackDex: Yea, i would have expected it to turn up in the builded coded in the end

dani-garcia: Just checked the clients and they don't seem to be making the request so it's probably fine

dani-garcia (approved):

src/api/notifications.rs
jjlin: * https://github.com/bitwarden/clients/commit/31a257407be7f8f47624b0d021363aaf2cfda2d7
* https://learn.microsoft.com/en-us/aspnet/core/signalr/configuration?view=aspnetcore-7.0&tabs=javascript

BlackDex: Thanks! I couldn't find that part. But of course i should have looked for a different spelling.

PR #3365, opened 2023-03-24
add missing eol at eof (POSIX compliance)

tessus:
According to the POSIX standard a line is defined as follows:

```
3.206 Line

A sequence of zero or more non-<newline> characters plus a terminating <newline> character.
```

/ref https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206

----

I am very sorry for the nitpicking, but this really drives me nuts. I never understood why people started to omit the last EOL. At university (almost 30 years ago) we had a system's programming course and it was drilled into us that a line without an EOL (newline) is not a line. Additionally many compilers complained, others threw an error.

I do understand that there is this unwritten rule that for minimized js there shouldn't be an EOL. IMO this is also a mistake, but who knows - that last byte could throw the world into chaos, if it had to be transferred as well.

I hope you don't see me as a total nut case.


BlackDex:
While i personally thing this is too much, and i do know still some tools fail if there isn't a newline at the end. I also think that this might trigger Diesel to do some nasty stuff with the new files now, which we need to validate

And, we need to check what happens with those mail templates in plain/text, we need to prevent empty newlines between header, body and footer.

jjlin:
While I agree that text files should generally have a final newline, I don't know of any modern tools that will choke on files without one, so changing all of these files (some of which are machine-generated or copied verbatim from other places) adds a lot of noise just to satisfy your personal OCD.

I'm okay with the `.editorconfig` change (except the `*.min.js` case probably isn't necessary as no one should be editing those anyway), but it's a hard no from me on the other changes.

tessus:
> I also think that this might trigger Diesel to do some nasty stuff with the new files now, which we need to validate

I can run some tests in a bit.

> we need to prevent empty newlines between header, body and footer.

I did not add an empty line at the end of the file. I added a newline character at the supposedly last line. So there won't be any empty lines, unless you added them in the templates.

BlackDex:
> While I agree that text files should generally have a final newline, I don't know of any modern tools that will choke on files without one, so changing all of these files (some of which are machine-generated or copied verbatim from other places) adds a lot of noise just to satisfy your personal OCD.
> 
> I'm okay with the `.editorconfig` change (except the `*.min.js` case probably isn't necessary as no one should be editing those anyway), but it's a hard no from me on the other changes.

I agree on this. And, besides the `*.min.js`, there are `*.js` which we did not created our selfs, and used an online version, which would now be changed because the addition of a newline, while the original doesn't have that, which in turn causes the checksum of that file to be different from the original. That is also something i would like to prevent.

BlackDex:
> I did not add an empty line at the end of the file. I added a newline character at the supposedly last line. So there won't be any empty lines, unless you added them in the templates.

It's not that you added it, but I'm not sure what the merging of those files will do and reading the plain/text mail messages. It could be that the header, body and footer now get an extra newline in between them, so before and after the body text.

tessus:
> changing all of these files (some of which are machine-generated or copied verbatim from other places) adds a lot of noise just to satisfy your personal OCD.

Forget my OCD for just a second. So you are ok that the text files which are the source code for this project are in fact not compliant and in reality are missing a last line. Just because tools ignore the fact that the last line is not a line and still read it as a line doesn't make it right. Strictly speaking you are telling me that correcting wrong files isn't necessary because it seems to work anyway. I am sorry but one shouldn't be that complacent.

Anyway, I am going to close it. I guess ignorance is bliss. I just have to work on that.

BlackDex:
> > changing all of these files (some of which are machine-generated or copied verbatim from other places) adds a lot of noise just to satisfy your personal OCD.
> 
> Forget my OCD for just a second. So you are ok that the text files which are the source code for this project are in fact not compliant and in reality are missing a last line. Just because tools ignore the fact that the last line is not a line and still read it as a line doesn't make it right. Strictly speaking you are telling me that correcting wrong files isn't necessary because it seems to work anyway. I am sorry but one shouldn't be that complacent.
> 
> Anyway, I am going to close it. I guess ignorance is bliss. I just have to work on that.

To bad that you react to this feedback like that.
While I'm for having this on new files being created or added (except vendor `js` and `css` files), adding this noise to existing files (and currently not knowing what the effect would be) is a bit to much I think.

jjlin:
> Forget my OCD for just a second. So you are ok that the text files which are the source code for this project are in fact not compliant and in reality are missing a last line. Just because tools ignore the fact that the last line is not a line and still read it as a line doesn't make it right. Strictly speaking you are telling me that correcting wrong files isn't necessary because it seems to work anyway. I am sorry but one shouldn't be that complacent.

You've cited the definition of a line in POSIX, and that could influence say, whether `wc -l` counts the final "line" of a file.

But I'm skeptical that any mainstream language defines valid source files in terms of POSIX lines. For one thing, Windows is a very common platform, so most languages will support CRLF line terminators. If you can cite any mainstream language spec that says that a final newline is required to have a valid source file, I'd be curious to see it.

tessus:
> To bad that you react to this feedback like that.

What exactly did rub you the wrong way?

That I closed the PR? There ia no point in discussing something that you are clearly against. 

Or that I said 'ignorance is bliss'? This was not a malicious comment. It's a fact. You are willingly accepting that the files are wrong. Even when you take files from somewhere else, they are in your source tree. You are not downloading them from a CDN and do a checksum comparision. They live in your source tree, so the files are part of this project. But as I mentioned before, there is no point in trying to convince you. It's not my project. I know when to give up.

You have to forgive my bluntness. I don't mean to be abrasive or hostile, but I understand when it comes across as such. Please note that this is definitely not my intention. My words are without an undertone, nor do I say them to hurt or offend anyone. They are just candidly spoken. I dislike sugarcoating or being overly diplomatic - I rather put my opinions out there unfiltered (you will always know where I stand and don't have to be afraid that I stab you in the back). I apologize, if I offended anyone.

tessus:
> You've cited the definition of a line in POSIX, and that could influence say, whether wc -l counts the final "line" of a file.

Yes and it does:

```
$ echo -n "not a line" >text.txt
$ cat text.txt |wc -l
       0
```

> For one thing, Windows is a very common platform, so most languages will support CRLF line terminators. If you can cite any mainstream language spec that says that a final newline is required to have a valid source file, I'd be curious to see it.

Windows is not POSIX compliant, but the posix definition speaks of a newline character. This means LF on Unix and CRLF on Windows (and I think there was another newline definition for another OS). So CRLF is still an EOL terminator. So on Windows you can also either have a CRLF at the end of the last line or not. So I don't completely understand your argument.

I cannot cite any language spec that says that a final newline is required to have a valid source file. I would have to look into it. I just recall issues in the past (warnings and sometimes errors), but I couldn't tell you what compilers exactly. Maybe xlc on AIX, or older versions of gcc on Linux. It's been a while.

BlackDex:
> > To bad that you react to this feedback like that.
> 
> What exactly did rub you the wrong way?
> 
> That I closed the PR? There ia no point in discussing something that you are clearly against.

If you would have read my last comment fully, you would read that I'm not fully against it, but okay 😉 .

> 
> Or that I said 'ignorance is bliss'? This was not a malicious comment. It's a fact. 

To be fair Ignorance is a bliss, I would love to be more ignorant, that would make my life a bit easier i guess 🤣 .

> You are willingly accepting that the files are wrong.

I think that is a bit short sighted. And to be fair, if files were wrong stuff wouldn't have worked.
Also, a file is basically just a stream of bytes, and if it ends with an EOF or `\n` that doesn't matter for most modern applications.
I even know of application which fail because of a newline at the end, so it can go both ways.

> Even when you take files from somewhere else, they are in your source tree. You are not downloading them from a CDN and do a checksum comparision. They live in your source tree, so the files are part of this project.

That is not how i see it. We also do not use `min.js` files to make it clear what is in there. It could also be that people want to verify that these files are original. So we differ from opinion on this, which is fine.

> But as I mentioned before, there is no point in trying to convince you. It's not my project. I know when to give up.

It's also not my project in the POSIX sense of the word, but i think that items like this could be discussed with good arguments.
And again, there might be no reason to modify existing files unless they are edited for example. Which would not create this noise and prevent possible issues (which we have not checked)

> 
> You have to forgive my bluntness. I don't mean to be abrasive or hostile, but I understand when it comes across as such. Please note that this is definitely not my intention. My words are without an undertone, nor do I say them to hurt or offend anyone. They are just candidly spoken. I dislike sugarcoating or being overly diplomatic - I rather put my opinions out there unfiltered (you will always know where I stand and don't have to be afraid that I stab you in the back). I apologize, if I offended anyone.

Forgiven, and no issues with this.

PR #3390, opened 2023-03-29
Fix abort on pw reset mail error

### LINKED ISSUE #3388: Admin reset not working
### Subject of the issue
Enabled admin reset for a organization. Master-password reset.

i used it on a test user  to test it how how it works etc and if it works.

When i reset the password and try to login with that account it just fails the login says username / password wrong.
i tried to change the password multiple times but the same Error.

### Deployment environment

<!--
    =========================================================================================
    Preferably, use the `Generate Support String` button on the admin page's Diagnostics tab.
    That will auto-generate most of the info requested in this section.
    =========================================================================================
* Vaultwarden version: v1.28.0
* Web-vault version: v2023.3.0b
* OS/Arch: linux/x86_64
* Uses a reverse proxy: false
* Internet access: true
* DNS Check: true
* Server/NTP Time Check: true
* HTTPS Check: true
* Database type: SQLite
* Database version: 3.39.2
* Clients used: 
* Reverse proxy and version: 
* Other relevant information: 

```json
{
  "_duo_akey": null,
  "_enable_duo": false,
  "_enable_email_2fa": false,
  "_enable_smtp": true,
  "_enable_yubico": true,
  "_icon_service_csp": "",
  "_icon_service_url": "",
  "_ip_header_enabled": true,
  "_smtp_img_src": "cid:",
  "admin_session_lifetime": 20,
  "admin_token": "***",
  "data_folder": "data",
  "database_conn_init": "",
  "database_max_conns": 10,
  "database_timeout": 30,
  "db_connection_retries": 15,
  "disable_admin_token": false,
  "disable_icon_download": true,
  "domain": "*****://***************",
  "domain_origin": "*****://***************",
  "domain_path": "",
  "domain_set": true,
  "duo_host": null,
  "duo_ikey": null,
  "duo_skey": null,
  "email_attempts_limit": 3,
  "email_expiration_time": 600,
  "email_token_size": 6,
  "enable_db_wal": true,
  "events_days_retain": null,
  "extended_logging": true,
  "helo_name": null,
  "hibp_api_key": null,
  "icon_blacklist_regex": null,
  "icon_cache_negttl": 259200,
  "icon_cache_ttl": 2592000,
  "icon_download_timeout": 10,
  "icon_redirect_code": 302,
  "icon_service": "internal",
  "invitations_allowed": true,
  "ip_header": "X-Real-IP",
  "log_file": null,
  "log_level": "Info",
  "org_attachment_limit": null,
  "org_creation_users": "",
  "org_events_enabled": false,
  "org_groups_enabled": false,
  "password_iterations": 2000000,
  "reload_templates": false,
  "sendmail_command": null,
  "sends_allowed": true,
  "sends_folder": "data/sends",
  "show_password_hint": false,
  "signups_allowed": false,
  "signups_domains_whitelist": "*****************,************,********,******************",
  "signups_verify": false,
  "smtp_auth_mechanism": null,
  "smtp_debug": false,
  "smtp_embed_images": true,
  "smtp_explicit_tls": null,
  "smtp_from": "*****************************",
  "smtp_host": "***********",
  "smtp_password": null,
  "smtp_port": 25,
  "smtp_security": "off",
  "smtp_ssl": null,
  "smtp_timeout": 15,
  "smtp_username": null,
  "tmp_folder": "data/tmp",
  "use_sendmail": false,
  "use_syslog": false,
  "web_vault_enabled": true,
  "websocket_enabled": true,
  "websocket_port": 3012,
  "yubico_client_id": "00000",
  "yubico_secret_key": "***",
  "yubico_server": null
}
```
</details>

BlackDex:
There was used a wrong macro to produce an error message when mailing
the user his password was reset failed. It was using `error!()` which
does not return an `Err` and aborts the rest of the code.

This resulted in the users password still being resetted, but not being
notified. This PR fixes this by using `err!()`. Also, do not set the
user object as mutable until it really is needed.

Second, when a user was using the new Argon2id KDF with custom values
like memory and parallelism, that would have rendered the password
incorrect. The endpoint which should return all the data did not
returned all the new Argon2id values. (Thanks to @stefan0xC for spotting the specific culprit).

Fixes #3388

### DIFF
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -2694,7 +2694,7 @@ async fn put_reset_password(
         None => err!("User to reset isn't member of required organization"),
     };
 
-    let mut user = match User::find_by_uuid(&org_user.user_uuid, &mut conn).await {
+    let user = match User::find_by_uuid(&org_user.user_uuid, &mut conn).await {
         Some(user) => user,
         None => err!("User not found"),
     };
@@ -2711,11 +2711,12 @@ async fn put_reset_password(
     // Sending email before resetting password to ensure working email configuration and the resulting
     // user notification. Also this might add some protection against security flaws and misuse
     if let Err(e) = mail::send_admin_reset_password(&user.email, &user.name, &org.name).await {
-        error!("Error sending user reset password email: {:#?}", e);
+        err!(format!("Error sending user reset password email: {e:#?}"));
     }
 
     let reset_request = data.into_inner().data;
 
+    let mut user = user;
     user.set_password(reset_request.NewMasterPasswordHash.as_str(), Some(reset_request.Key), true, None);
     user.save(&mut conn).await?;
 
@@ -2759,12 +2760,15 @@ async fn get_reset_password_details(
 
     check_reset_password_applicable_and_permissions(&org_id, &org_user_id, &headers, &mut conn).await?;
 
+    // https://github.com/bitwarden/server/blob/3b50ccb9f804efaacdc46bed5b60e5b28eddefcf/src/Api/Models/Response/Organizations/OrganizationUserResponseModel.cs#L111
     Ok(Json(json!({
         "Object": "organizationUserResetPasswordDetails",
         "Kdf":user.client_kdf_type,
         "KdfIterations":user.client_kdf_iter,
+        "KdfMemory":user.client_kdf_memory,
+        "KdfParallelism":user.client_kdf_parallelism,
         "ResetPasswordKey":org_user.reset_password_key,
-        "EncryptedPrivateKey":org.private_key ,
+        "EncryptedPrivateKey":org.private_key,
 
     })))
 }


src/api/core/organizations.rs
```diff
@@ -2759,12 +2760,15 @@ async fn get_reset_password_details(
 
     check_reset_password_applicable_and_permissions(&org_id, &org_user_id, &headers, &mut conn).await?;
 
+    // https://github.com/bitwarden/server/blob/3b50ccb9f804efaacdc46bed5b60e5b28eddefcf/src/Api/Models/Response/Organizations/OrganizationUserResponseModel.cs#L111
     Ok(Json(json!({
         "Object": "organizationUserResetPasswordDetails",
         "Kdf":user.client_kdf_type,
         "KdfIterations":user.client_kdf_iter,
+        "KdfMemory":user.client_kdf_memory,
+        "KdfParallelism":user.client_kdf_parallelism,
```
tessus, line 2769: What happens, if they are not set?

I used the following method to only return these values when they are set:

https://github.com/dani-garcia/vaultwarden/blob/main/src/api/identity.rs#L110-L132

tessus: ok, but this means that the keys are still returned with the value `null`, correct?

I am just asking, because I was told for the argon2 kdf fields PR to only return the keys when they are set. In that case, I could have kept the code shorter. ;-)

stefan0xC: @tessus Yeah, but they have to be set if you use `argon2id` as KDF otherwise you would not be able to login (alternatively we could assume default some values). And if you use `PBKDF2` those values will be ignored (at least as far as I've tested it now in this reset password call, not sure what happens in the `_refresh_login` case).

tessus: No worries, I just noticed this inconsistency and thought that it might have been an oversight.

BlackDex: I added a link to the upstream source, where you see that they return all values, no filtering done.

Took be fair, i have not checked for the other endpoints what they do. But thinking about it, i would guess that Bitwarden will send them also, but that i would need to check to be sure.

stefan0xC: @tessus I mean you are right that in that case it still works for users using `PBKDF2` if the returned JSON contains a `null` in both `KdfMemory` and `KdfParallelism`. So this check might be superfluous here. If I set those parameters to `null` in the database for a user using `Argon2id` it  already panics earlier in the call to `identity/accounts/prelogin`:

https://github.com/dani-garcia/vaultwarden/blob/525e6bb65a6926e0f9de3fc5dafd5c5b63981f9f/src/api/core/accounts.rs#L811-L814

stefan0xC: Okay, a login with the default `argon2ìd` settings would work if those are set to `null` (which should not even happen because there is a check that prevents setting the KDF to `Argon2id` if those are unset). Otherwise login will fail like in the linked issue. (And with `PBKDF2` those settings are ignored even if they are not `null`.)

dani-garcia (approved):

PR #3403, opened 2023-03-31
Revert setcap, update rust and crates

BlackDex:
- Revert #3170 as discussed in #3387
  This appears to be a breaking change on some platforms.
  And in hindsight it's better to not have this feature as there might be (hypothetical) security issues with this.
- Update Dockerfile.j2 for easy version changes.
  Just change it in one place instead of multiple
- Updated to Rust to latest patched version
- Updated crates to latest available
- Pinned mimalloc to an older version, as it breaks on musl builds

Resolves #3387


jjlin:
For the record, I disagree with the characterization that it's "better" not to have #3170 just because a handful of users who are fixated with dropping all possible privileges (even though they haven't presented any practical downside to keeping `NET_BIND_SERVICE`) feels strongly about it. Having stronger feelings and being more vocal doesn't make them more right.

jjlin:
IMO, that argument is even more hypothetical than the "hypothetical" users that #3170 is supposed to benefit, but if that's your opinion, that's fine.

BlackDex:
Both arguments apply. Not just one or the other.
I don't think it's security theater.
Also arguing that Docker already has this feature built in is a bad argument, since docker isn't the only container platform.

Anyways, main reasons are breaking change and security.

williamdes (approved):

docker/Dockerfile.j2
```diff
@@ -206,7 +194,7 @@ RUN [ "cross-build-start" ]
 # Create data folder and Install needed libraries
 RUN mkdir /data \
 {% if "alpine" in runtime_stage_base_image %}
-    && apk add --no-cache \
+    && apk update && apk add --no-cache \
```
stefan0xC, line 197: Why do we need to update the package index? I thought `apk add --no-cache` is sufficient to fetch and automatically remove the package index? :thinking:

stefan0xC (changes requested):

docker/Dockerfile.j2
BlackDex: I could be wrong, but i think the `--no-cache` means not caching the packages locally. And the update ensures the latest version info is pulled. During my test i thought i noticed without the update, it downloaded older versions of the packages instead of the latest patched versions.

stefan0xC: @BlackDex as far as I've looked into it the alpine image ships with an empty `/var/cache/apk/` directory and running `apk add --no-cache`  apparently fetches the index too:

```bash
/ # apk add --no-cache tree
fetch https://dl-cdn.alpinelinux.org/alpine/v3.17/main/x86_64/APKINDEX.tar.gz
fetch https://dl-cdn.alpinelinux.org/alpine/v3.17/community/x86_64/APKINDEX.tar.gz
(1/1) Installing tree (2.0.4-r0)
Executing busybox-1.35.0-r29.trigger
OK: 7 MiB in 16 packages
/ # ls -l /var/cache/apk/
total 0
```

BlackDex: We have to check the images used for the other architectures of that is the same.

stefan0xC: It's the same in sofar that the `/var/cache/apk` directories of those images are also empty (I've just mounted the image but I did not run them). I mean if they were not I would probably consider it a bug of the image as almost all best practices regarding alpine images recommend this way of installing packages.

BlackDex: Thanks for checking!
I removed it :)

PR #3404, opened 2023-03-31
WebSockets via Rocket's Upgrade connection

### LINKED ISSUE #685: Run websocket server on same port as other HTTP serving
This is currently blocked on upstream Rocket feature which has seen some recent progress: https://github.com/SergioBenitez/Rocket/issues/90

BlackDex:
This PR implements a (not yet fully released) new feature of Rocket which allows WebSockets/Upgrade connections.
No more need for multiple ports to be opened for Vaultwarden.
No explicit need for a reverse proxy to get WebSockets to work (Although I still suggest to use a reverse proxy).

- Using a git revision for Rocket, since `rocket_ws` is not yet released.
- Updated other crates as well.
- Added a connection guard to clear the WS connection from the Users list.

Fixes #685
Fixes #2917
Fixes #1424

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -79,9 +79,9 @@ dependencies = [
 
 [[package]]
 name = "async-executor"
-version = "1.5.0"
+version = "1.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "17adb73da160dfb475c183343c8cccd80721ea5a605d3eb57125f0a7b7a92d0b"
+checksum = "6fa3dc5f2a8564f07759c008b9109dc0d39de92a88d5588b8a5036d286383afb"
 dependencies = [
  "async-lock",
  "async-task",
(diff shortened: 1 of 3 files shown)


BlackDex:
I think the implementation is working. It can use both port 3012 Direct Tungstenite and the Rocket/Tungstenite implementation.
It can be mixed and notification will be sent to both.

Best would be if Rocket 0.5 would be released as stable so that we do not have to use a patched version.

BlackDex:
@dani-garcia , i think for testing it's fine.
Not sure if we want to release a .2 soon with some fixes currently merged? I do have a fix pending on my laptop regarding some admin interface fixes which were caused by the recent updates.

dani-garcia:
True, I don't expect this change to break anything, and we still have the 3012 fallback anyway.

BlackDex:
Just as a note here for now.

If you want to use this, you still need to make sure that your reverse proxy pass-through the upgrade request correctly.

For nginx you can use the following.

Somewhere within the `http {}`
```nginx
map $http_upgrade $connection_upgrade {
    default upgrade;
    ''      "";
}
```

And then within the location.
```nginx
location / {
  proxy_http_version 1.1;
  proxy_set_header Upgrade $http_upgrade;
  proxy_set_header Connection $connection_upgrade;
```

qymab:
With this change should I remove this line from my docker compose :

```
- traefik.http.routers.bwws.middlewares=redirect-to-https
- traefik.http.routers.bwws.service=bw-ws
- traefik.http.services.bw-ws.loadbalancer.server.port=3012
```

thanks

PR #3434, opened 2023-04-08
Update deb to Bookworm, fix vulnerabilities linked to bullseye image

mysther:
Fix high vulnerabilities CVE-2021-3847, CVE-2023-23914, CVE-2022-27385, CVE-2019-8457, CVE-2022-42916 and 268 others vulnerabilities.

Reported by grype tool

### DIFF
--- a/docker/Dockerfile.j2
+++ b/docker/Dockerfile.j2
@@ -3,7 +3,7 @@
 # This file was generated using a Jinja2 template.
 # Please make your changes in `Dockerfile.j2` and then `make` the individual Dockerfiles.
 {% set rust_version = "1.68.2" %}
-{% set debian_version = "bullseye" %}
+{% set debian_version = "bookworm" %}
 {% set alpine_version = "3.17" %}
 {% set build_stage_base_image = "rust:%s-%s" % (rust_version, debian_version) %}
 {% if "alpine" in target_file %}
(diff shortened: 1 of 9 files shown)


BlackDex:
Bookworm is not yet released as stable by Debian.
I rather keep using the current stable release bullseye.

It probably won't take long for a final Bookworm release as they are on RC1 right now.

mysther:
Ok thanks for your reply, I'll see how I can help you on the topics you mentioned.

BlackDex:
> Ok thanks for your reply, I'll see how I can help you on the topics you mentioned.

Nothing much to do. My Alpine/Musl build images are already available with OpenSSL v3 and PostgreSQL v15 😉

williamdes:
I'd say we keep this PR open in draft state so the work is not lost and we can rebase it?

BlackDex (changes requested):
I'm fine with keeping this open. But that would also mean some more changes to be done.
1. All the `runs-on:` need to be changed to `22.04` in the workflow files at `.github/workflows`
2. The Alpine images need to use the `-openssl3` tagged images ( See: https://github.com/BlackDex/rust-musl/#openssl-v30 )
3. The Alpine images need to use the v15 PostgreSQL library instead of the default v11 ( See: https://github.com/BlackDex/rust-musl/#using-a-dockerfile )

This needs to be done, else OpenSSL and PostgreSQL will not match (or too big a difference) the versions in the Debian container images which could lead to unexplained issues between tests and the releases.

BlackDex:
I'm going to close this as a duplicate of #3573 and the other reasons given above.
Also, a new Rust version has been released for which #3573 also has changes to make that work.

Besides that, still thanks for the PR! We appreciate all the help.

PR #3439, opened 2023-04-11
Fixed missing footer_text and a few inconsistencies in email templates

kennymc-c:
Fixed some missing footer_text partials and a few inconsistencies between plain text and html email templates


vilgotf (commented):
The UNIX file permission changes from 644 to 755 should be reverted.

dani-garcia (approved):

PR #3475, opened 2023-04-23
inline static rsa keys

vilgotf:
From my understanding, rust never inlines static variables meaning that these variables are separately statically allocated and wasting space.

Even if there's not codegen diff, separately defining these static variables is (imo) code smell.


BlackDex:
I think you are right. This also looks a bit better i think.
Could you also please rebase on main, so it's up-to-date?

Thx.

dani-garcia (approved):
Yeah, not sure why we have them as separate variables, maybe the `*_VEC`s were used in more places before? No reason to have them now yeah.

BlackDex:
> Yeah, not sure why we have them as separate variables, maybe the `*_VEC`s were used in more places before? No reason to have them now yeah.

I think we did before indeed. Long long time ago.

PR #3502, opened 2023-05-02
Use Rocket `v0.5` branch to fix endpoints

### LINKED ISSUE #3494: latest git version is broken with nginx proxy
### Subject of the issue
Updated to the latest version in git (1ec049e2b5f2f1aa3b8d035e94915b131c5e9e25) and everything is broken.  Getting either a 502 bad gateway error or the following error in my nginx logs:

```
[error] 1467253#1467253: *225 upstream prematurely closed connection while reading upstream, client: 1.2.3.4, server: fqdn.com, request: "GET /vault/ HTTP/2.0", upstream: "http://127.0.0.1:8000/vault/", host: "fqdn.com"
```

### Deployment environment

* vaultwarden version: 1ec049e2b5f2f1aa3b8d035e94915b131c5e9e25

* Install method: built from source

* Clients used: web browser

* Reverse proxy and version: nginx 1.22.1-9 via Debian testing/unstable using the exact configuration described in the wiki under "Nginx with sub-path (by BlackDex)"

* MySQL/MariaDB or PostgreSQL version: neither/ SQLite

* Other relevant details:

### Steps to reproduce

Nothing special other than what is described above.  Rolling back to 393645617efa6866001d40cc7f6a9485947e9b23 and  bw_web v2023.3.0b fixes everything again.

### Expected behaviour

Vault page loads in my browser.

### Actual behaviour

502 Bad Gateway or even more cryptic client error due to the stream simply crapping out mid request.

### Troubleshooting data

Nothing much to provide sadly beyond the error above.  I never even see the requests making it through to Rocket as nothing is ever logged there unless I access Rocket directly.  That does at least give me the following:

```
% curl http://127.0.0.1:8000
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="utf-8">
    <title>404 Not Found</title>
</head>
<body align="center">
    <div role="main" align="center">
        <h1>404: Not Found</h1>
        <p>The requested resource could not be found.</p>
        <hr />
    </div>
    <div role="contentinfo" align="center">
        <small>Rocket</small>
    </div>
</body>
</html>

% curl http://127.0.0.1:8000/vault/
curl: (52) Empty reply from server
```

When I roll back to the previously working version, that second request actually returns the expected:

```
% curl http://127.0.0.1:8000/vault/
<!doctype html><html class="theme_light"><head><meta charset="utf-8"/><meta name="viewport" content="width=1010"/><meta name="theme-color" content="#175DDC"/><title page-title>Vaultwarden Web Vault</title><link rel="apple-touch-icon" sizes="180x180" href="images/apple-touch-icon.png"/><link rel="icon" type="image/png" sizes="32x32" href="images/favicon-32x32.png"/><link rel="icon" type="image/png" sizes="16x16" href="images/favicon-16x16.png"/><link rel="mask-icon" href="images/safari-pinned-tab.svg" color="#175DDC"/><link rel="manifest" href="ca8f66ed7fccfcd0809f.json"/><script defer="defer" src="theme_head.5f24ba8d7aa944e6f52b.js"></script><link href="app/main.c107f7b9354b6317446a.css" rel="stylesheet"></head><body class="layout_frontend"><app-root><div class="mt-5 d-flex justify-content-center"><div><img class="mb-4 logo logo-themed" alt="Bitwarden"/><p class="text-center"><i class="bwi bwi-spinner bwi-spin bwi-2x text-muted" title="Loading" aria-hidden="true"></i></p></div></div></app-root><script defer="defer" src="app/polyfills.428c25638840333a09ee.js"></script><script defer="defer" src="app/vendor.776201439bfdefee32dc.js"></script><script defer="defer" src="app/main.3a27378a743dd4ad9f70.js"></script></body></html>
```

Nothing else has changed, either between the working and broken versions aside from the vaultwarden code and web contents themselves.  The systemd unit file and nginx configurations are identical.  But I can provide whatever else might be helpful.

### LINKED ISSUE #3506: Latest testing version returns 404 when creating a new org
### Subject of the issue
Latest testing version returns 404 when creating a new org

### Deployment environment
Vaultwarden version: v1.28.1-1ec049e
Web-vault version: v2023.4.0
OS/Arch: linux/x86_64
Running within Docker: true (Base: Alpine)
Environment settings overridden: true
Uses a reverse proxy: true
IP Header check: true (X-Real-IP)
Internet access: true
Internet access via a proxy: false
DNS Check: true
Time Check: true
Domain Configuration Check: false
HTTPS Check: true
Database type: MySQL
Database version: 10.6.12-MariaDB-log
Clients used: Web vault
Reverse proxy and version: Nginx 1.22
Environment settings which are overridden: SIGNUPS_VERIFY_RESEND_TIME, ADMIN_TOKEN

### Steps to reproduce
Goto the organizations tab. Click new organization. It loads forever and the returns 404 for /api/plans

### Expected behaviour
/api/plans should return a valid response. 

### Troubleshooting data
image

Going to https://vault.placeholder.com/api/plans in a browser returns a response but I suspect some things were changed on the Bitwarden side in v2023.4.0.

BlackDex:
There now is a `v0.5` branch which will be the final release version
when the time is there. Switched to this instead of the `master` branch
which contains other fixes and enhancements as well (for `v0.6`).

This should solve all the endpoint issue we were having.

Fixes #3495
Fixes #3494
Fixes #3506

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1510,9 +1510,9 @@ dependencies = [
 
 [[package]]
 name = "libc"
-version = "0.2.142"
+version = "0.2.143"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6a987beff54b60ffa6d51982e1aa1146bc42f19bd26be28b0586f252fccf5317"
+checksum = "edc207893e85c5d6be840e969b496b53d94cec8be2d501b214f50daa97fa8024"
 
 [[package]]
 name = "libmimalloc-sys"
@@ -1552,9 +1552,9 @@ checksum = "0717cef1bc8b636c6e1c1bbdefc09e6322da8a9321966e8928ef80d20f7f770f"
 
 [[package]]
 name = "linux-raw-sys"
-version = "0.3.6"
+version = "0.3.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b64f40e5e03e0d54f03845c8197d0291253cdbedfb1cb46b13c2c117554a9f4c"
+checksum = "ece97ea872ece730aed82664c424eb4c8291e1ff2480247ccf7409044bc6479f"
 
 [[package]]
 name = "lock_api"
@@ -2100,9 +2100,9 @@ checksum = "8b870d8c151b6f2fb93e84a13146138f05d02ed11c7e7c54f8826aaaf7c9f184"
 
 [[package]]
 name = "pkg-config"
-version = "0.3.26"
+version = "0.3.27"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6ac9a59f73473f1b8d852421e59e64809f025994837ef743615c6d0c5b305160"
+checksum = "26072860ba924cbfa98ea39c8c19b4dd6a4a25423dbdf219c1eca91aa0cf6964"
 
 [[package]]
 name = "polling"
@@ -2447,7 +2447,7 @@ dependencies = [
 [[package]]
 name = "rocket"
 version = "0.5.0-rc.3"
-source = "git+https://github.com/SergioBenitez/Rocket?rev=9b0564ed27f90686b333337d9f6ed76484a84b27#9b0564ed27f90686b333337d9f6ed76484a84b27"
+source = "git+https://github.com/SergioBenitez/Rocket?rev=be496682c209e008215d81ca9dd396617c4f810a#be496682c209e008215d81ca9dd396617c4f810a"
 dependencies = [
  "async-stream",
  "async-trait",
@@ -2485,7 +2485,7 @@ dependencies = [
 [[package]]
 name = "rocket_codegen"
 version = "0.5.0-rc.3"
-source = "git+https://github.com/SergioBenitez/Rocket?rev=9b0564ed27f90686b333337d9f6ed76484a84b27#9b0564ed27f90686b333337d9f6ed76484a84b27"
+source = "git+https://github.com/SergioBenitez/Rocket?rev=be496682c209e008215d81ca9dd396617c4f810a#be496682c209e008215d81ca9dd396617c4f810a"
 dependencies = [
  "devise",
  "glob",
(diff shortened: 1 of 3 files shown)


src/api/admin.rs
```diff
@@ -256,6 +257,16 @@ fn admin_page(_token: AdminToken) -> ApiResult<Html<String>> {
     render_admin_page()
 }
 
+#[get("/<p..>", rank = 100)] // Only match this if the other routes don't match
+fn admin_page_path(p: PathBuf, token: AdminToken) -> ApiResult<Html<String>> {
+    // Check if there is no filename given, in that case, we need to redirect to web_index() function.
```
stefan0xC, line 262: This comment refers to the wrong function. Aside from that it seems to work. :+1:

BlackDex: copy/paste 😇

BlackDex:
Looks like this needs a bit more work, so, no merging yet.

BlackDex:
Switched to the `v0.5` branch of Rocket, which will be the the final release version when the time comes.
This should solve the endpoint issues 😄 .

tessus:
I am slightly puzzled by this. Rocket won't change the fact that the client is sending `/plans/`.

I didn't see any other changes, e.g. using a fairing to use the old behavior. So did Benito revert everything again so that a fairing is not necessary?

However, if the code always strips the last trailing slash, a ligitimate API endpoint `/path/` is not possible. As I mentioned, I am a bit confused.

BlackDex:
The `v0.5` branch doesn't contain the trailing slash changes. That is something he mentioned on Matrix. And later created this branch to keep master/future`v0.6` commits and the coming final `v0.5` separated.

tessus:
Thanks for the info. I wasn't using Matrix.

But since we already know that these changes are coming in 0.6, wouldn't it make sense to already make the code ready to use 0.6-rc (or whatever the name will be of what will be released for the 0.6 branch).

Versioning in Rocket seems a bit backwards. I reckon most people use the 0.5-rcX releases in production. As soon as 0.5 is out, people will use the 0.6-rc releases... ;-)

BlackDex:
We tend to use stable versions only. Rocket had some lifecycle challenges a while back. The next stable version is going to be `v0.5` which is the version we intend to u keep using. `v0.6` isn't even a think yet. Also, the trailing slash changes, according to Sergio need good testing and feedback from several users before it can be even final.

So, using the branch which is going to be the next stable is the best to do. Also, the only reason we are using commits right now instead of the `rc3` is because the new changes of being able to use web-sockets via Rocket instead of using a separate port and service.

dani-garcia (approved):

PR #3547, opened 2023-05-25
Prevent 401 on main admin page

### LINKED ISSUE #3540: Admin login page always responds with a 401
### Subject of the issue

Ever since https://github.com/dani-garcia/vaultwarden/commit/5a05139efe484b37edbc2db6eb618676d26f3723, the login page consistently returns a 401 status code. I believe this is the intended behavior, but it's causing some login issues for me so I thought I'd see if this warrants changing.

For more context, I protect my admin page behind SSO with [Authentik](https://goauthentik.io) and the nginx [auth request](http://nginx.org/en/docs/http/ngx_http_auth_request_module.html) module. This module sends each request to an SSO auth endpoint before connecting to the upstream server. If the the SSO endpoint returns a 401 response, the user is redirected to the login page. Otherwise, the user is allowed through. It also seems to handle a 401 response from the app by redirecting to the SSO login URL.

When this module is enabled, Vaultwarden's 401 response results in a redirect loop.

During my troubleshooting, I disabled the SSO auth request. Instead of seeing the Vaultwarden login page, I got a response from the [default backend](https://github.com/tarampampam/error-pages) I use since I have it set to replace 401s.

I'm sure I could disable the default 401 backend for Vaultwarden and keep SSO disabled, but all of this troubleshooting made me wonder if there would be a better way to handle this. Would it make more sense for Vaultwarden to return a 200 when the login page is first shown, then return a 401 if the entered code is wrong?

If this behavior shouldn't change, I don't mind closing the issue and finding a workaround.

### Deployment environment

* vaultwarden version: 1.28.1

* Install method: Docker image in Kubernetes

* Clients used:  Web vault, desktop, Android

* Reverse proxy and version:  ingress-nginx v1.7.1

* MySQL/MariaDB or PostgreSQL version: 

* Other relevant details: Running Authentik v2023.5.1

### Steps to reproduce

- Navigate to `/admin`
- Check the response status code

### Expected behaviour

- The initial login page returns a 200 status code
- An invalid token causes the form to be shown again with a 401 status code

### Actual behaviour

- The initial login page returns a 401 status code
- An invalid token causes the form to be shown again with a 401 status code

BlackDex:
When you are not loggedin, and have no cookie etc.. we always returned a 401. This was mainly to allow the login page on all the sub pages, and after login being redirected to the requested page, for these pages a 401 is a valid response, since, you do not have access.

But for the main `/admin` page, it should just respond with a `200` and show the login page.

This PR fixes this flow and response. It should prevent people using Fail2ban, or other tools being triggered by only accessing the login page.

Resolves #3540

### DIFF
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -36,6 +36,7 @@ pub fn routes() -> Vec<Route> {
         get_user_by_mail_json,
         post_admin_login,
         admin_page,
+        admin_page_login,
         invite_user,
         logout,
         delete_user,
@@ -254,6 +255,11 @@ fn render_admin_page() -> ApiResult<Html<String>> {
 #[get("/")]
 fn admin_page(_token: AdminToken) -> ApiResult<Html<String>> {
     render_admin_page()
+}
+
+#[get("/", rank = 2)]
+fn admin_page_login() -> ApiResult<Html<String>> {
+    render_admin_login(None, None)
 }
 
 #[derive(Deserialize, Debug)]
@@ -761,7 +767,17 @@ impl<'r> FromRequest<'r> for AdminToken {
 
             let access_token = match cookies.get(COOKIE_NAME) {
                 Some(cookie) => cookie.value(),
-                None => return Outcome::Failure((Status::Unauthorized, "Unauthorized")),
+                None => {
+                    let requested_page =
+                        request.segments::<std::path::PathBuf>(0..).unwrap_or_default().display().to_string();
+                    // When the requested page is empty, it is `/admin`, in that case, Forward, so it will render the login page
+                    // Else, return a 401 failure, which will be caught
+                    if requested_page.is_empty() {
+                        return Outcome::Forward(Status::Unauthorized);
+                    } else {
+                        return Outcome::Failure((Status::Unauthorized, "Unauthorized"));
+                    }
+                }
             };
 
             if decode_admin(access_token).is_err() {


src/api/admin.rs
```diff
@@ -761,7 +767,17 @@ impl<'r> FromRequest<'r> for AdminToken {
 
             let access_token = match cookies.get(COOKIE_NAME) {
                 Some(cookie) => cookie.value(),
-                None => return Outcome::Failure((Status::Unauthorized, "Unauthorized")),
+                None => {
+                    let requested_page =
+                        request.segments::<std::path::PathBuf>(0..).unwrap_or_default().display().to_string();
+                    // When the requested page is empty, it is `/admin`, in that case, Forward, so it will render the login page
+                    // Else, return a 401 failure, which will be catched
```
tessus, line 774:
```suggestion
                    // Else, return a 401 failure, which will be caught
```

tessus: sorry for the nitpicking

BlackDex: np, English is not my main language ;)

tessus:
Yep, this shows a 200 now. Cheers.

Btw, do you know, whether there is a standard after trying to login? Currently a 401 is returned, which I believe is fine. But I have also seen 200s. e.g. try to login to github with a wrong user/password. The browser still gets a 200.
I am just wondering, because I have seen these things all over the place. (IMO the worst is to have an API return 404, when the API does not find an element. The API itself should return a 200 status code, but the error code in the response should be a 404. Anyway, off topic.)
Just wondering what your take on this is.

BlackDex:
> Yep, this shows a 200 now. Cheers.
> 
> Btw, do you know, whether there is a standard after trying to login? Currently a 401 is returned, which I believe is fine. But I have also seen 200s. e.g. try to login to github with a wrong user/password. The browser still gets a 200. I am just wondering, because I have seen these things all over the place. (IMO the worst is to have an API return 404, when the API does not find an element. The API itself should return a 200 status code, but the error code in the response should be a 404. Anyway, off topic.) Just wondering what your take on this is.

Well, a 401 should be used if the login is invalid, either when trying to access a page which needs authentication, or when sending your credentials and they are invalid. A 200 is OK, and invalid credentials, or accessing a page which you do not have access to, is not OK in my opinion.

If you build some kind of API, and return a 200 for invalid login, that would mean you need to scrape the body to see if you are logged in or not? That is bad in my opinion.

dani-garcia (approved):

PR #3563, opened 2023-06-03
Update Rust and Crates

tessus:
- Updated Rust to v1.70.0
- Updated MSRV to v1.68.2
- Updated crates


Cargo.toml
```diff
@@ -40,15 +40,15 @@ syslog = "6.1.0"
...
+log = "0.4.18"
 fern = { version = "0.6.2", features = ["syslog-6"] }
 tracing = { version = "0.1.37", features = ["log"] } # Needed to have lettre and webauthn-rs trace logging to work
 
 # A `dotenv` implementation for Rust
 dotenvy = { version = "0.15.7", default-features = false }
 
 # Lazy initialization
-once_cell = "1.17.1"
+once_cell = "1.17.2"
```
BlackDex, line 51: Seems to be `1.18.0` already, just a few hours ago :)

BlackDex (changes requested):
First of all, thanks!

I just noticed some missing updates. Of which a few were updated after your PR.
`once_cell` and `cached` are both updated, to `1.18.0` and `0.44.0`

Further, you missed `diesel_logger`, which is currently at `0.3.0`, which fixes compatibility with diesel v2.1.x.

A nice tool for some extra checks is https://crates.io/crates/cargo-outdated, you can then run `cargo outdated -R` to see all outdated crates. Some can't be updated like `cookie` and `webauthn-rs` for example. But most of the time others can.

Also, since I released the new web-vault just a few minutes ago, would be cool if you could add the new web-vault version also.

Cargo.toml
tessus: Hmm, I can't travel to the future. ;-) Haha. When I ran the `cargo update` I only got that version.

BlackDex: I know

tessus:
> I just noticed some missing updates. Of which a few were updated after your PR.
once_cell and cached are both updated, to 1.18.0 and 0.44.0

I'll check it out.

> Further, you missed diesel_logger, which is currently at 0.3.0, which fixes compatibility with diesel v2.1.x.

Hmm, this was not shown in cargo update. Weird.

> A nice tool for some extra checks is https://crates.io/crates/cargo-outdated, you can then run cargo outdated -R to see all outdated crates.

This is a great advise. I'll try using this tool, but I am getting a compile error when doing a `cargo install --locked cargo-outdated` on my my macOS. I have to look into it, or maybe I'll just use the Linux build VM for it.
For this PR I actually just ran `cargo update` and looked at what was in the cargo.toml and updated them, but this tool seems much better.

> Also, since I released the new web-vault just a few minutes ago, would be cool if you could add the new web-vault version also.

Will do.

Cargo.toml
tessus: Which is why I put that winking smiley...

tessus:
> Some can't be updated like cookie and webauthn-rs for example. 

Ok, I have updated all but those 2. I've also updated the web-vault to v2023.5.0.

BlackDex:
> > Some can't be updated like cookie and webauthn-rs for example.
> 
> Ok, I have updated all but those 2. I've also updated the web-vault to v2023.5.0.

Great thx again!

BlackDex (approved):

PR #3568, opened 2023-06-09
Implement the Organization API Key support for the new Directory Connector v2022

BlackDex:
I previously started a branch to implement the API Key support for Organizations.
Later @Kurnihil used that branch to further implement this and created a PR #2907, which hasn't been touched for a while now.

I refreshed this by rebasing and merging and updateing the work already done by @Kurnihil and tested this using a very small set of users and groups via an OpenLDAP server. I do not have access to an Microsoft AD or Azure AD environment which I can use for testing, so that is not taken into account here.

Closes #2907

Co-authored-by: Daniele Andrei <daniele.andrei@geo-satis.com>
Co-authored-by: Kurnihil

### DIFF
--- a/migrations/mysql/2022-07-21-200424_create_organization_api_key/up.sql
+++ b/migrations/mysql/2022-07-21-200424_create_organization_api_key/up.sql
@@ -1,0 +1,8 @@
+CREATE TABLE organization_api_key (
+	uuid			CHAR(36) NOT NULL,
+	org_uuid		CHAR(36) NOT NULL REFERENCES organizations(uuid),
+	atype			INTEGER NOT NULL,
+	api_key			VARCHAR(255) NOT NULL,
+	revision_date	DATETIME NOT NULL,
+	PRIMARY KEY(uuid, org_uuid)
+);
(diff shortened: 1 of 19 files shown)


dani-garcia (approved):

deajan:
@BlackDex Does that mean we get to use official Directory Connector with all it's functions (ie disabling/deleting users) ?

BlackDex:
> @BlackDex Does that mean we get to use official Directory Connector with all it's functions (ie disabling/deleting users) ?

As mentioned, i do not have access to an AzureAD/Microsoft AD environment, so I'm not able to test disabled users, as that is not a feature which ldap supports the way it's implemented in the Directory Connector.

But it does remove users from the org which are not in the ldap environment anymore.

deajan:
Then I guess I'll have to honor my pledge at the next release ;)
Do you guys provide any "donation" papers ?

BlackDex:
> Then I guess I'll have to honor my pledge at the next release ;) Do you guys provide any "donation" papers ?

Only the ones on our personal pages i guess.
https://github.com/sponsors/BlackDex and https://github.com/sponsors/dani-garcia

There is no project wide donation.

PR #3573, opened 2023-06-11
Update images to Bookworm and PQ15 and Rust v1.71

BlackDex:
This PR updates the base images to use Debian Bookworm as base image. Also the MUSL/Alpine builds now use OpenSSLv3 and PostgreSQL v15.

The GHA Workflows are updated to use Ubuntu 22.04 to better match the versions of Debian Bookworm.

Also:
- Enabled spares crate registry
- Updated workflow actions
- Updated Rust to v1.71.1
- The rust-musl images now use musl v1.2.3 for the 32bit arch's if the Rust version is v1.71.0 or higher.
   The 64bit arch's already used musl v1.2.3.
- Updated crates.

Improves / Closes #3434


dani-garcia (approved):

tessus:
@BlackDex unfortunately gh does not parse the entire line, but only looks for the keywords at the beginning - unless `keyword PR` is/are separated by comma/s.

`keyword` can be one of: `close closes closed fix fixes fixed resolve resolves resolved`

```
Improves / Closes #3434
```

won't do anything. You have to use one of the following:

```
Fixed #3434
Closes #3434
resolve #3434
```

Or for multiple references on one line:

```
Closes ISSUE1, Fixes ISSUE2, Resolves ISSUE3
```

P.S.: This behavior also has been annoying me greatly, but I haven't been able to convince gh otherwise. ;-)

BlackDex:
@tessus as far as i know `improves` isn't a valid keyword, so that doesn't have any effect. It should however act upon the keyword `Closes` which is exactly what I intended.

tessus:
@BlackDex as you can see, it is not linked to anything in the gh web UI. I guess we'll see what happens when you merge it...

tessus:
@BlackDex ok, I've just tested it in my test repo. Yes, it seems to work now. So GH really changed the parsing. Not sure why it says on the right of this PR the following:

image

PR #3578, opened 2023-06-13
Add mobile push device filter to non-null push uuid

### LINKED ISSUE #3577: Items Taking 15 Seconds to Save with PUSH_ENABLED=true
### Subject of the issue
With the PUSH_ENABLED=true environment varibale (merged in #3304) set (and PUSH_INSTALLATION_ID and PUSH_INSTALLATION_KEY), updating an item takes 15 seconds. When disabling the flag, the request takes only ~30ms
image

The log shows ~100 requests to bitwarden's push servers being made in that time.

* Vaultwarden version: v1.28.1-bd883de7
* OS/Arch: linux/aarch64
* Running within Docker: true (Base: Alpine)
* Uses a reverse proxy: true
* Internet access: true
* DNS Check: true
* Server/NTP Time Check: true
* HTTPS Check: true
* Database type: SQLite
* Database version: 3.41.2
* Clients used: Web
* Reverse proxy and version: Traefik v3.0 
* Other relevant information: 

```json
{
  "_duo_akey": null,
  "_enable_duo": false,
  "_enable_email_2fa": true,
  "_enable_smtp": true,
  "_enable_yubico": true,
  "_icon_service_csp": "",
  "_icon_service_url": "",
  "_ip_header_enabled": true,
  "_smtp_img_src": "cid:",
  "admin_session_lifetime": 20,
  "admin_token": "***",
  "data_folder": "data",
  "database_conn_init": "",
  "database_max_conns": 10,
  "database_timeout": 30,
  "db_connection_retries": 15,
  "disable_admin_token": false,
  "domain": "*****://*********************",
  "domain_origin": "*****://*********************",
  "domain_path": "",
  "domain_set": true,
  "duo_host": null,
  "duo_ikey": null,
  "duo_skey": null,
  "email_attempts_limit": 3,
  "email_expiration_time": 600,
  "email_token_size": 6,
  "enable_db_wal": true,
  "events_days_retain": null,
  "extended_logging": true,
  "helo_name": null,
  "hibp_api_key": null,
  "icon_blacklist_regex": null,
  "icon_cache_negttl": 259200,
  "icon_cache_ttl": 2592000,
  "icon_download_timeout": 10,
  "icon_redirect_code": 302,
  "icon_service": "internal",
  "invitation_org_name": "CENSORED",
  "invitations_allowed": true,
  "log_level": "debug",
  "org_attachment_limit": null,
  "org_creation_users": "",
  "org_events_enabled": false,
  "org_groups_enabled": false,
  "push_enabled": false,
  "reload_templates": false,
  "require_device_email": true,
  "sendmail_command": null,
  "sends_allowed": true,
  "sends_folder": "data/sends",
  "show_password_hint": true,
  "signups_allowed": true,
  "signups_verify": true,
  "smtp_auth_mechanism": null,
  "smtp_debug": false,
  "smtp_embed_images": true,
  "smtp_explicit_tls": false,
  "smtp_from": "*********************",
  "smtp_from_name": "Bitwarden",
  "smtp_host": "*******************",
  "smtp_password": "***",
  "smtp_port": 587,
  "smtp_security": "starttls",
  "smtp_ssl": true,
  "smtp_timeout": 15,
  "smtp_username": "***********",
  "tmp_folder": "data/tmp",
  "use_sendmail": false,
  "use_syslog": false,
  "web_vault_enabled": true,
  "websocket_enabled": true,
  "websocket_port": 3012,
  "yubico_client_id": "54614",
  "yubico_secret_key": "***",
  "yubico_server": null
}
```
</details>

push_enabled is false here since I disabled the feature temporarily.

### Steps to reproduce
Enable mobile push and edit/create an entry.

### Expected behaviour

Take the same amount of time.

### Actual behaviour

It takes >10 seconds, nearly a hundred connections are started.

### Troubleshooting data
On debug mode, a lot of connections to push.bitwarden.com are opened. I have attached the relevant part of the log.
[log.txt](https://github.com/dani-garcia/vaultwarden/files/11729185/log.txt)

If required I'm happy to do some network traces/debugging to further triage the issue.

quexten:
Partially fixes #3577.

Over time, old devices accumulate in the SQL device table. Even when the device push uuid is NULL, a push request is sent for them when mobile push is active. For my server, this lead to ~100 requests to the Bitwarden push servers without a single active mobile push client, and a request time of ~15 seconds.

This requests adds a filter to prevent mobile push notifications from being sent to clients who did not even register for mobile push.

### DIFF
--- a/src/api/push.rs
+++ b/src/api/push.rs
@@ -139,71 +139,52 @@ pub async fn push_cipher_update(
         }
     };
 
-    for device in Device::find_by_user(user_uuid, conn).await {
-        let data = json!({
+    if Device::check_user_has_push_device(user_uuid, conn).await {
+        send_to_push_relay(json!({
             "userId": user_uuid,
             "organizationId": (),
-            "deviceId": device.push_uuid,
+            "deviceId": acting_device_uuid,
             "identifier": acting_device_uuid,
             "type": ut as i32,
             "payload": {
-                "Id": cipher.uuid,
-                "UserId": cipher.user_uuid,
-                "OrganizationId": (),
-                "RevisionDate": cipher.updated_at
-            }
-        });
-
-        send_to_push_relay(data).await;
-    }
-}
-
-pub async fn push_logout(user: &User, acting_device_uuid: Option<String>, conn: &mut crate::db::DbConn) {
-    if let Some(d) = acting_device_uuid {
-        for device in Device::find_by_user(&user.uuid, conn).await {
-            let data = json!({
-                "userId": user.uuid,
+                "id": cipher.uuid,
+                "userId": cipher.user_uuid,
                 "organizationId": (),
-                "deviceId": device.push_uuid,
-                "identifier": d,
-                "type": UpdateType::LogOut as i32,
-                "payload": {
-                    "UserId": user.uuid,
-                    "Date": user.updated_at
-                }
-            });
-            send_to_push_relay(data).await;
-        }
-    } else {
-        let data = json!({
-            "userId": user.uuid,
-            "organizationId": (),
-            "deviceId": (),
-            "identifier": (),
-            "type": UpdateType::LogOut as i32,
-            "payload": {
-                "UserId": user.uuid,
-                "Date": user.updated_at
+                "revisionDate": cipher.updated_at
             }
-        });
-        send_to_push_relay(data).await;
+        }))
+        .await;
     }
 }
 
-pub async fn push_user_update(ut: UpdateType, user: &User) {
-    let data = json!({
+pub fn push_logout(user: &User, acting_device_uuid: Option<String>) {
+    let acting_device_uuid: Value = acting_device_uuid.map(|v| v.into()).unwrap_or_else(|| Value::Null);
+
+    tokio::task::spawn(send_to_push_relay(json!({
+        "userId": user.uuid,
+        "organizationId": (),
+        "deviceId": acting_device_uuid,
+        "identifier": acting_device_uuid,
+        "type": UpdateType::LogOut as i32,
+        "payload": {
+            "userId": user.uuid,
+            "date": user.updated_at
+        }
+    })));
+}
+
+pub fn push_user_update(ut: UpdateType, user: &User) {
+    tokio::task::spawn(send_to_push_relay(json!({
         "userId": user.uuid,
         "organizationId": (),
         "deviceId": (),
         "identifier": (),
         "type": ut as i32,
         "payload": {
-            "UserId": user.uuid,
-            "Date": user.updated_at
+            "userId": user.uuid,
+            "date": user.updated_at
         }
-    });
-
-    send_to_push_relay(data).await;
+    })));
 }
 
 pub async fn push_folder_update(
@@ -212,46 +193,42 @@ pub async fn push_folder_update(
     acting_device_uuid: &String,
     conn: &mut crate::db::DbConn,
 ) {
-    for device in Device::find_by_user(&folder.user_uuid, conn).await {
-        let data = json!({
+    if Device::check_user_has_push_device(&folder.user_uuid, conn).await {
+        tokio::task::spawn(send_to_push_relay(json!({
             "userId": folder.user_uuid,
             "organizationId": (),
-            "deviceId": device.push_uuid,
+            "deviceId": acting_device_uuid,
             "identifier": acting_device_uuid,
             "type": ut as i32,
             "payload": {
-                "Id": folder.uuid,
-                "UserId": folder.user_uuid,
-                "RevisionDate": folder.updated_at
+                "id": folder.uuid,
+                "userId": folder.user_uuid,
+                "revisionDate": folder.updated_at
             }
-        });
-
-        send_to_push_relay(data).await;
+        })));
     }
 }
 
-pub async fn push_send_update(ut: UpdateType, send: &Send, conn: &mut crate::db::DbConn) {
+pub async fn push_send_update(ut: UpdateType, send: &Send, acting_device_uuid: &String, conn: &mut crate::db::DbConn) {
     if let Some(s) = &send.user_uuid {
-        for device in Device::find_by_user(s, conn).await {
-            let data = json!({
+        if Device::check_user_has_push_device(s, conn).await {
+            tokio::task::spawn(send_to_push_relay(json!({
                 "userId": send.user_uuid,
                 "organizationId": (),
-                "deviceId": device.push_uuid,
-                "identifier": (),
+                "deviceId": acting_device_uuid,
+                "identifier": acting_device_uuid,
                 "type": ut as i32,
                 "payload": {
-                    "Id": send.uuid,
-                    "UserId": send.user_uuid,
-                    "RevisionDate": send.revision_date
+                    "id": send.uuid,
+                    "userId": send.user_uuid,
+                    "revisionDate": send.revision_date
                 }
-            });
-
-            send_to_push_relay(data).await;
+            })));
         }
     }
 }
 
-async fn send_to_push_relay(data: Value) {
+async fn send_to_push_relay(notification_data: Value) {
     if !CONFIG.push_enabled() {
         return;
     }
@@ -270,8 +247,8 @@ async fn send_to_push_relay(data: Value) {
         .post(CONFIG.push_relay_uri() + "/push/send")
         .header(ACCEPT, "application/json")
         .header(CONTENT_TYPE, "application/json")
-        .header(AUTHORIZATION, auth_header)
-        .json(&data)
+        .header(AUTHORIZATION, &auth_header)
+        .json(&notification_data)
         .send()
         .await
     {
--- a/src/db/models/device.rs
+++ b/src/db/models/device.rs
@@ -202,7 +202,7 @@ impl Device {
                 .from_db()
         }}
     }
-    pub async fn find_push_device_by_user(user_uuid: &str, conn: &mut DbConn) -> Vec<Self> {
+    pub async fn find_push_devices_by_user(user_uuid: &str, conn: &mut DbConn) -> Vec<Self> {
         db_run! { conn: {
             devices::table
                 .filter(devices::user_uuid.eq(user_uuid))
@@ -210,6 +210,18 @@ impl Device {
                 .load::<DeviceDb>(conn)
                 .expect("Error loading push devices")
                 .from_db()
+        }}
+    }
+
+    pub async fn check_user_has_push_device(user_uuid: &str, conn: &mut DbConn) -> bool {
+        db_run! { conn: {
+            devices::table
+            .filter(devices::user_uuid.eq(user_uuid))
+            .filter(devices::push_token.is_not_null())
+            .count()
+            .first::<i64>(conn)
+            .ok()
+            .unwrap_or(0) != 0
         }}
     }
 }
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -403,10 +403,10 @@ async fn delete_user(uuid: &str, token: AdminToken, mut conn: DbConn) -> EmptyRe
 async fn deauth_user(uuid: &str, _token: AdminToken, mut conn: DbConn, nt: Notify<'_>) -> EmptyResult {
     let mut user = get_user_or_404(uuid, &mut conn).await?;
 
-    nt.send_logout(&user, None, &mut conn).await;
+    nt.send_logout(&user, None).await;
 
     if CONFIG.push_enabled() {
-        for device in Device::find_push_device_by_user(&user.uuid, &mut conn).await {
+        for device in Device::find_push_devices_by_user(&user.uuid, &mut conn).await {
             match unregister_push_device(device.uuid).await {
                 Ok(r) => r,
                 Err(e) => error!("Unable to unregister devices from Bitwarden server: {}", e),
@@ -429,7 +429,7 @@ async fn disable_user(uuid: &str, _token: AdminToken, mut conn: DbConn, nt: Noti
 
     let save_result = user.save(&mut conn).await;
 
-    nt.send_logout(&user, None, &mut conn).await;
+    nt.send_logout(&user, None).await;
 
     save_result
 }
--- a/src/api/core/accounts.rs
+++ b/src/api/core/accounts.rs
@@ -343,7 +343,7 @@ async fn post_password(
     // Prevent loging out the client where the user requested this endpoint from.
     // If you do logout the user it will causes issues at the client side.
     // Adding the device uuid will prevent this.
-    nt.send_logout(&user, Some(headers.device.uuid), &mut conn).await;
+    nt.send_logout(&user, Some(headers.device.uuid)).await;
 
     save_result
 }
@@ -403,7 +403,7 @@ async fn post_kdf(data: JsonUpcase<ChangeKdfData>, headers: Headers, mut conn: D
     user.set_password(&data.NewMasterPasswordHash, Some(data.Key), true, None);
     let save_result = user.save(&mut conn).await;
 
-    nt.send_logout(&user, Some(headers.device.uuid), &mut conn).await;
+    nt.send_logout(&user, Some(headers.device.uuid)).await;
 
     save_result
 }
@@ -490,7 +490,7 @@ async fn post_rotatekey(data: JsonUpcase<KeyData>, headers: Headers, mut conn: D
     // Prevent loging out the client where the user requested this endpoint from.
     // If you do logout the user it will causes issues at the client side.
     // Adding the device uuid will prevent this.
-    nt.send_logout(&user, Some(headers.device.uuid), &mut conn).await;
+    nt.send_logout(&user, Some(headers.device.uuid)).await;
 
     save_result
 }
@@ -513,7 +513,7 @@ async fn post_sstamp(
     user.reset_security_stamp();
     let save_result = user.save(&mut conn).await;
 
-    nt.send_logout(&user, None, &mut conn).await;
+    nt.send_logout(&user, None).await;
 
     save_result
 }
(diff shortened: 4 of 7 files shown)


GeekCornerGH (changes requested):
I actually made a function filtring push devices, you'll want to use this

BlackDex:
This probably also needs to be done for the `push_logout`.
And using a function which runs the query that doesn't even returns those records would be better indeed.

quexten:
Ah, nice. I'll look into using that function.

BlackDex:
@quexten , if you check my comments on the other PR, it might be better to merge the two and move the `send_to_push_relay(data).await` call either outside of the for loop, and send a `vec![data]`, so that the `send_to_push_relay` function can handle all the calls within the same spawned call.

quexten:
Thank you both for the helpful suggestions. Updated the PR to have send_to_push_relay take a Vec, and updated all of the calls to send_to_push_relay to be wrapped in a spawned task. Furthermore, as per @GeekCornerGH's suggestion I used the find_push_device_by_user where possible, so that the SQL query does not even return those devices.

On my local system, there is now no noticeable difference in the time the PUT request takes (4-5 ms total whether it's enabled or not) with the full ~100 old devices from my database.

BlackDex:
@quexten Could you rebase your PR on the current main branch please :)

quexten:
Ah actually, seems like I need to do some more changes.

BlackDex:
Well, there are some clippy changes needed.

But i just tested it quickly, and it doesn't seem to update all the devices for me for some reason.

BlackDex:
I am using the v2023.5.0 version of the app on Android. And it looks like it doesn't send the device id's anymore.

I have not tested the v2023.4.0 again yet. But that might be a thing

src/api/push.rs
```diff
@@ -186,7 +195,7 @@ pub async fn push_logout(user: &User, acting_device_uuid: Option<String>, conn:
                 "Date": user.updated_at
             }
         });
-        send_to_push_relay(data).await;
+        tokio::task::spawn(send_to_push_relay(vec![data]));
     }
 }
```
GeekCornerGH, line 201: I would remove the `async` if you don't need, so it won't fail automated checks.

GeekCornerGH (changes requested):

quexten:
> I am using the v2023.5.0 version of the app on Android. And it looks like it doesn't send the device id's anymore.

Do you mean the app itself does not send the ids, whereas before it did?

src/api/push.rs
quexten: Removed async from the function where clippy complained, now the pipeline runs through.

BlackDex:
@quexten @GeekCornerGH 
Ok, i have figured it out i think by some debugging.

Instead of pushing to all those items, it in fact should be one, BUT, all the data in the payload should contain the information of the client creating the change.

You only need to check if there are devices which have a push token for that user, if that is the case, then call the Bitwarden push server with the data from the cipher and the data from the acting_device.

I probably need to check some other stuff, like organizations for example.

This also explains why it worked for me in the beginning, because i probably had multiple devices with different device id's and push_tokens already and the code back then didn't in time matched one of the devices i was using. Because there wasn't a match for the specific Android device, the clients processed the request.

The clients skip a call to refresh if they see there own device id in the call, and that is what probably is the clue here.

quexten:
So, deviceId and identifier should both be the UUID of the acting device? In that case why are there even different push uuids saved in the DB? As far as I can tell, if we use the acting UUID in the push notification, we would then only use the push UUID's in the registration, though I haven't studied the protocol closely.

I'll try this later today.

BlackDex:
> So, deviceId and identifier should both be the UUID of the acting device? In that case why are there even different push uuids saved in the DB? As far as I can tell, if we use the acting UUID in the push notification, we would then only use the push UUID's in the registration, though I haven't studied the protocol closely.
> 
> I'll try this later today.

Well it was an oversight i think. We still need to store them for user/device specific actions maybe like logout and the push token storage.

BlackDex:
> > So, deviceId and identifier should both be the UUID of the acting device? In that case why are there even different push uuids saved in the DB? As far as I can tell, if we use the acting UUID in the push notification, we would then only use the push UUID's in the registration, though I haven't studied the protocol closely.
> > I'll try this later today.
> 
> My best bet would be that Bitwarden is using uuids as a way to uniquely identify any device, instead of searching per mail then per identifier (as they call them)

That is what i think to, and we still need it during register

BlackDex:
@quexten , did you tested multiple devices? And what where you results?

quexten:
> @quexten , did you tested multiple devices? And what where you results?

I tested with 2 mobile devices, in one case with making the changes in my web extension, in one case on one of the mobile devices, and they did get updated correctly (for editing, deleting, adding a cipher).

I recommend to test the newest commit to see if your case is also resolved.

BlackDex:
Ok, I did some more testing. While these latest changes seem to be in line with what Bitwarden it self sends out, it still doesn't work with multiple devices for me.

But, if i test this on a Self-Hosted or even on there SaaS system, that doesn't even work.
It seems it round-robins the push requests, so, one update goes to device 1, then next update goes to device 2 etc.. etc...

When i force a logout, it seems to only reach one device.
So, how hard we will try, it will probably never get work no matter what we adjust on our side 😄 .

But since i think these current changes are better in line into upstream, and we are not spamming the Push server with these changes, that would be better.

If someone has other thoughts, please let me know.

tessus:
When I look at the current code in this PR, it calls `find_push_device_by_user` which returns a Vec, but then there is no loop, but just one call to `send_to_push_relay`. So maybe I am missing something here, but this seems weird.

Also, if this doesn't work with bw's self-hosted option, shouldn't there already be an issue open for this?

BlackDex:
> When I look at the current code in this PR, it calls `find_push_device_by_user` which returns a Vec, but then there is no loop, but just one call to `send_to_push_relay`. So maybe I am missing something here, but this seems weird.

My question was more technical in the sens of the workings. But this it would be much better if it would just return a bool, or int, by using a `COUNT` for the query for example, that is much quicker, and safes resources.

> 
> Also, if this doesn't work with bw's self-hosted option, shouldn't there already be an issue open for this?
> 
> Update:
> 
> Shouldn't this:
> 
> ```
> if !Device::find_push_device_by_user(user_uuid, conn).await.is_empty() {
> ```
> 
> be rather something like:
> 
> ```
> vec = find_push_device_by_user(user_uuid, conn).await;
> iterate over vec and push to all devices in vec
> ```

We determined that this was **not** the case, and that was reverted because of that.


> 
> Additionally, shouldn't the function be called `find_push_devices_by_user`? Isn't the function supposed to find all push devices for a given user?

As mentioned above, it should probably be renamed, or maybe a new function should be created.


But, what i wanted to establish here is that this is just the way it seems to work. At least for me currently for Vaultwarden, Self-Hosted and Bitwarden Hosted instances.

If that is the case, then I'm generally ok with the workings of this PR. And then i can review it as is with some of the above mentioned items :)

quexten:
Okay, it seems to work on the latest build, however I also encountered an issue with not  syncing. After a forced deauth, I signed my devices back in, but no changes were pushed because they did not have a push token / uuid anymore. The de-auth had cleared their push tokens in the database (as it should) but no re-registration happened. When I completely cleared the app-data and signed back in, the registration happened again and push worked again.

I suspect (but have not yet investigated) there is a problem somewhere in re-registration. Maybe just dropping all device entries for the user on a forced de-auth would be enough? Would this cause any other issues?

>     Device::delete_all_by_user(&user.uuid, &mut conn).await?;
There is this line in the deauth-function, but for some reason for me the table still contains my devices after a de-auth...

BlackDex:
It did removed all devices for me.
It also logged out at least one mobile device.

But, as mentioned, it looks like it's a strange issue with the push server. It might even be our registration or something. But it's hard to debug.
But since it also happens for me on vault.bitwarden.com, it probably is not something we can fix server side.

src/api/push.rs
```diff
@@ -185,13 +183,12 @@ pub async fn push_logout(user: &User, acting_device_uuid: Option<String>, conn:
                 "UserId": user.uuid,
                 "Date": user.updated_at
             }
-        });
-        send_to_push_relay(data).await;
+        })));
```
BlackDex, line 186: This can be shorter and easier to read like this i think.
Also, we probably always want to send a logout request i think, but if that is not the case, then the `if Device::check...` check can be added of course.


```suggestion
    let acting_device_uuid: Value = acting_device_uuid.map(|v| v.into()).unwrap_or_else(|| Value::Null);
    // Always send a logout command, doesn't matter if a user has a push enabled device or not.
    tokio::task::spawn(send_to_push_relay(json!({
        "userId": user.uuid,
        "organizationId": (),
        "deviceId": acting_device_uuid,
        "identifier": acting_device_uuid,
        "type": UpdateType::LogOut as i32,
        "payload": {
            "userId": user.uuid,
            "date": user.updated_at
        }
    })));
```

src/api/push.rs
```diff
@@ -270,8 +261,8 @@ async fn send_to_push_relay(data: Value) {
         .post(CONFIG.push_relay_uri() + "/push/send")
         .header(ACCEPT, "application/json")
         .header(CONTENT_TYPE, "application/json")
-        .header(AUTHORIZATION, auth_header)
-        .json(&data)
+        .header(AUTHORIZATION, auth_header.as_str())
```
BlackDex, line 264:
```suggestion
        .header(AUTHORIZATION, &auth_header)
```

src/api/push.rs
```diff
@@ -212,46 +207,42 @@ pub async fn push_folder_update(
...
 pub async fn push_send_update(ut: UpdateType, send: &Send, conn: &mut crate::db::DbConn) {
     if let Some(s) = &send.user_uuid {
-        for device in Device::find_by_user(s, conn).await {
-            let data = json!({
+        if Device::check_user_has_push_device(s, conn).await {
+            tokio::task::spawn(send_to_push_relay(json!({
                 "userId": send.user_uuid,
                 "organizationId": (),
-                "deviceId": device.push_uuid,
+                "deviceId": (),
```
BlackDex, line 232: This also needs to be the `acting_device_uuid`

src/db/models/device.rs
```diff
@@ -212,4 +212,16 @@ impl Device {
...
+
+    pub async fn check_user_has_push_device(user_uuid: &str, conn: &mut DbConn) -> bool {
+        db_run! { conn: {
+            diesel::select(diesel::dsl::exists(
+                devices::table
+                    .filter(devices::user_uuid.eq(user_uuid))
+                    .filter(devices::push_token.is_not_null())
+            ))
+            .get_result(conn)
+            .expect("Error loading push devices")
```
BlackDex, line 224: This is probably better, since it will return true if there is a count of 1 or higher, and false otherwise.
No need for errors here.

```suggestion
            .count()
            .first::<i64>(conn)
            .ok()
            .unwrap_or(0) != 0
```

src/db/models/device.rs
```diff
@@ -212,4 +212,16 @@ impl Device {
...
         }}
     }
+
+    pub async fn check_user_has_push_device(user_uuid: &str, conn: &mut DbConn) -> bool {
+        db_run! { conn: {
+            diesel::select(diesel::dsl::exists(
+                devices::table
+                    .filter(devices::user_uuid.eq(user_uuid))
+                    .filter(devices::push_token.is_not_null())
+            ))
```
BlackDex, line 222: This format is not how we build the queries. There is some macro magic done which might break here in the future or other database connections.

Best would be to use the same format as the function above.

```suggestion
            devices::table
                .filter(devices::user_uuid.eq(user_uuid))
                .filter(devices::push_token.is_not_null())
```

BlackDex (changes requested):
There are some small items to address i think.

Also, i noticed that all the `payload` keys are starting with a capital letter.
I'm not seeing that in the calls and code from Bitwarden it self.

Maybe best that we change those to lowercase also?

src/db/models/device.rs
quexten: GitHub shows the suggestions in a broken way, but I think this is implemented now.

src/api/push.rs
quexten: I passed through the data through a few functions as it wasn't previously being used for sends, but it does not seem to be working yet. Will have to further debug.

src/api/push.rs
quexten: Done.

src/api/push.rs
quexten: Done.

src/api/push.rs
GeekCornerGH: I tried on the official server to use send, push didn't worked too. Probably a problem Bitwarden will have to fix by themselves

src/api/push.rs
```diff
@@ -152,46 +152,29 @@ pub async fn push_cipher_update(
...
-                "deviceId": device.push_uuid,
-                "identifier": d,
-                "type": UpdateType::LogOut as i32,
-                "payload": {
-                    "UserId": user.uuid,
-                    "Date": user.updated_at
-                }
-            });
-            send_to_push_relay(data).await;
+pub async fn push_logout(user: &User, acting_device_uuid: Option<String>) {
```
GeekCornerGH, line 160:
```suggestion
pub fn push_logout(user: &User, acting_device_uuid: Option<String>) {
```
You'll want to avoid automated checks failing again :)

GeekCornerGH (changes requested):

src/api/push.rs
BlackDex: Or, enable `pre-commit`, which does all the needed checks so the pipeline will not fail :)

src/api/push.rs
quexten: Ah, I see. I think it is actually the mobile clients: https://github.com/bitwarden/mobile/blob/98705e443f321411dd3b7c8d54581fe4462cc3c5/src/App/Services/PushNotificationListenerService.cs#L76-L162 does not implement any of
```
SyncSendCreate = 12,
SyncSendUpdate = 13,
SyncSendDelete = 14,
```

src/api/push.rs
GeekCornerGH: Pre-commit? Never heard of it

BlackDex: https://pre-commit.com/

I install it via [pipx](https://pypa.github.io/pipx/) `pipx install pre-commit`

quexten:
FYI I have tested on bitwarden.com and sends are also not syncing there. I have created an upstream bug report in case they are not aware (there was no other issue tracking this).

src/api/push.rs
quexten: Yeah, sorry about that, I guess we can squash all the commits on merge so the fix formatting commits aren't cluttering the history. Anyways, should be fixed now.

GeekCornerGH: > https://pre-commit.com/
> 
> I install it via [pipx](https://pypa.github.io/pipx/) `pipx install pre-commit`

Thanks for the link, will take a look :)

quexten:
> Also, i noticed that all the payload keys are starting with a capital letter.
> I'm not seeing that in the calls and code from Bitwarden it self.

Changed them to lower-case now.

bw_web_builds
BlackDex, line 1: This needs to be removed. You probably used this during testing :).
And, this would also have been blocked by pre-commit :).

BlackDex (changes requested):
I think it all looks good.
Just one item which needs to be removed.

And, if possible if you can squash this PR that would be best i think.
Probably best way would be to run this i think.

```bash
# First reset to the beginning
git reset --soft HEAD~13

# Remove the bw_web_builds folder
rm -rf bw_web_builds

# Check if all is ok and do a force push by using a + sign infront of your branch-name
git push origin +my_branch_name
```

bw_web_builds
quexten: Ah, shucks, not sure how that happened. Thank you!

> git reset --soft HEAD~13

I just did `git reset --soft main`. That way there is no need to count the commits.

quexten:
Sorry for the confusion, I think it should be correctly reset now?

BlackDex:
> Sorry for the confusion, I think it should be correctly reset now?

Looks ok indeed.

GeekCornerGH (approved):
I guess that's good, didn't tried yet, but changes seems alright

tessus:
@BlackDex ok, so let me try to summarize the architecture for push notifications

1. the device is registered at the push server with `user_id` and `device_id` (but when?)
2. if there is a notification to be pushed, the device sends a ping to the push server with `user_id` and `device_id`
3. the push server searches for all `device_id`s registered for `user_id`
4. the push server sends out a ping to all previously found `device_id`s to refresh their data, except to the one device that sent the ping

Is this the current workflow?

If so, what do I have to do to get the devices registered after I upgrade to the code that includes this PR? De-auth all devices? logout/login?

BlackDex:
> @BlackDex ok, so let me try to summarize the architecture for push notifications
> 
> 1. the device is registered at the push server with `user_id` and `device_id` (but when?)

During login at the device.

> 2. if there is a notification to be pushed, the device sends a ping to the push server with `user_id` and `device_id`

The server sends a notification. So it will send a web-socket message and a message to the push server of Bitwarden.

> 3. the push server searches for all `device_id`s registered for `user_id`
> 4. the push server sends out a ping to all previously found `device_id`s to refresh their data, except to the one device that sent the ping

This part is not something i have looked into that deeply. And it doesn't seem that it will search and notify all devices which contain the users-id. Because for me multiple devices didn't received the update message. But this is the same for vault.bitwarden.com.

> 
> Is this the current workflow?
> 
> If so, what do I have to do to get the devices registered after I upgrade to the code that includes this PR? De-auth all devices? logout/login?

For me, during testing, i needed to fully clear the old devices and also clear the data on the mobile device and login again.


I know that Bitwarden is looking into new ways to handle these notifications, because the currently way is stressing there current infrastructure.

tessus:
Btw, the PR still shows that changes are requested. Isn't it ready to be merged?

BlackDex:
> Btw, the PR still shows that changes are requested. Isn't it ready to be merged?

I want to test it first. Didn't had time yet.

BlackDex (approved):

PR #3579, opened 2023-06-13
Run mobile push requests in background

quexten:
Another partial fix for #3577.

Combining this with PR #3578, this should solve #3577.

Mobile push requests are currently run blocking within the HTTP Put for updating a cipher request. In the Bitwarden UI this just shows as if updating the entry takes up to multiple seconds. This request moves the actual HTTP request to the Bitwarden Push servers into a tokio task. This way, sending the notifications does not block the request to update a cipher.

### DIFF
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -403,10 +403,10 @@ async fn delete_user(uuid: &str, token: AdminToken, mut conn: DbConn) -> EmptyRe
 async fn deauth_user(uuid: &str, _token: AdminToken, mut conn: DbConn, nt: Notify<'_>) -> EmptyResult {
     let mut user = get_user_or_404(uuid, &mut conn).await?;
 
-    nt.send_logout(&user, None, &mut conn).await;
+    nt.send_logout(&user, None).await;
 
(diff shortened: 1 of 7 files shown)


GeekCornerGH (approved):
Didn't thought of it, and didn't figured out, thanks for this catch also

BlackDex:
Wouldn't it even be better to run those push functions in the background? Now it spawns multiple threads per user, which could become an issue. But if you put this sounds the function that calls this function of would probably speed up more.

BlackDex:
Or, make this push functions allow to receive a `vec` of multiple values and then loop through those? That will even be better maybe.

quexten:
Okay, not sure why but GitHub seems to have closed my PR on force-push. Reopening with this comment.

BlackDex:
I think the reset to main was wrong actually.
But now your branch is out-of-date.

tessus:
this is the wrong PR

tessus:
The last 4 days everyone looked at PR #3578

quexten:
> The last 4 days everyone looked at PR #3578

Whoops.... my bad.

PR #3592, opened 2023-06-17
Implement "login with device"

quexten:
This is a **very WIP** pull-request for login-with-device. To run this, you need a new client version, for the web client this is 2023.06 (or latest master as of today) since login-with-device was disabled for self-hosted installations previously.

Basic login-with-device on WebSocket clients (desktop/web/webextension) works, but there is still quite some work to be done.
[Screencast from 2023-06-17 23-53-39.webm](https://github.com/dani-garcia/vaultwarden/assets/11866552/88a9360f-483a-4bcb-9e8f-58a264f2457e)

This PR implements a few components to make the login-with-device work, each still partially incomplete.
- A few new rocket endpoints for adding / updating requests (mostly done, some request validation / responses missing)
- A new database table for keeping the auth_requests (only done for sqlite)
- A new websocket endpoint for anonymous connections (needed for the login-with-device feature, this duplicates some code but I didn't know how to structure it better in this case)
- Some changes to the password_login. To finish the login-with-device, a regular password-login request to '/identity/connect/token' is made, with an additional field in the body, "authRequest", this contains the UUID of the authrequest, the password is the authCode from the passwordless login-request instead of the actual masterpasswordhash.

Aside from the points missing mentioned above, mobile push is not implemented yet, and a lot of code clean-up needs to be done.

Feel free to comment general comments, keep in mind the PR is very WIP so a lot of parts are missing and there is a lot of clean-up to be done.

I will squash the PR when it is done.

### DIFF
--- a/src/api/notifications.rs
+++ b/src/api/notifications.rs
@@ -36,10 +36,19 @@ static WS_USERS: Lazy<Arc<WebSocketUsers>> = Lazy::new(|| {
     })
 });
 
-use super::{push_cipher_update, push_folder_update, push_logout, push_send_update, push_user_update};
+pub static WS_ANONYMOUS_SUBSCRIPTIONS: Lazy<Arc<AnonymousWebSocketSubscriptions>> = Lazy::new(|| {
+    Arc::new(AnonymousWebSocketSubscriptions {
+        map: Arc::new(dashmap::DashMap::new()),
+    })
+});
+
+use super::{
+    push::push_auth_request, push::push_auth_response, push_cipher_update, push_folder_update, push_logout,
+    push_send_update, push_user_update,
+};
 
 pub fn routes() -> Vec<Route> {
-    routes![websockets_hub]
+    routes![websockets_hub, anonymous_websockets_hub]
 }
 
 #[derive(FromForm, Debug)]
@@ -74,6 +83,29 @@ impl Drop for WSEntryMapGuard {
     }
 }
 
+struct WSAnonymousEntryMapGuard {
+    subscriptions: Arc<AnonymousWebSocketSubscriptions>,
+    token: String,
+    addr: IpAddr,
+}
+
+impl WSAnonymousEntryMapGuard {
+    fn new(subscriptions: Arc<AnonymousWebSocketSubscriptions>, token: String, addr: IpAddr) -> Self {
+        Self {
+            subscriptions,
+            token,
+            addr,
+        }
+    }
+}
+
+impl Drop for WSAnonymousEntryMapGuard {
+    fn drop(&mut self) {
+        info!("Closing WS connection from {}", self.addr);
+        self.subscriptions.map.remove(&self.token);
+    }
+}
+
 #[get("/hub?<data..>")]
 fn websockets_hub<'r>(
     ws: rocket_ws::WebSocket,
@@ -96,6 +128,72 @@ fn websockets_hub<'r>(
 
         // Once the guard goes out of scope, the connection will have been closed and the entry will be deleted from the map
         (rx, WSEntryMapGuard::new(users, claims.sub, entry_uuid, addr))
+    };
+
+    Ok({
+        rocket_ws::Stream! { ws => {
+            let mut ws = ws;
+            let _guard = guard;
+            let mut interval = tokio::time::interval(Duration::from_secs(15));
+            loop {
+                tokio::select! {
+                    res = ws.next() =>  {
+                        match res {
+                            Some(Ok(message)) => {
+                                match message {
+                                    // Respond to any pings
+                                    Message::Ping(ping) => yield Message::Pong(ping),
+                                    Message::Pong(_) => {/* Ignored */},
+
+                                    // We should receive an initial message with the protocol and version, and we will reply to it
+                                    Message::Text(ref message) => {
+                                        let msg = message.strip_suffix(RECORD_SEPARATOR as char).unwrap_or(message);
+
+                                        if serde_json::from_str(msg).ok() == Some(INITIAL_MESSAGE) {
+                                            yield Message::binary(INITIAL_RESPONSE);
+                                            continue;
+                                        }
+                                    }
+                                    // Just echo anything else the client sends
+                                    _ => yield message,
+                                }
+                            }
+                            _ => break,
+                        }
+                    }
+
+                    res = rx.recv() => {
+                        match res {
+                            Some(res) => yield res,
+                            None => break,
+                        }
+                    }
+
+                    _ = interval.tick() => yield Message::Ping(create_ping())
+                }
+            }
+        }}
+    })
+}
+
+#[get("/anonymous-hub?<token..>")]
+fn anonymous_websockets_hub<'r>(
+    ws: rocket_ws::WebSocket,
+    token: String,
+    ip: ClientIp,
+) -> Result<rocket_ws::Stream!['r], Error> {
+    let addr = ip.ip;
+    info!("Accepting Anonymous Rocket WS connection from {addr}");
+
+    let (mut rx, guard) = {
+        let subscriptions = Arc::clone(&WS_ANONYMOUS_SUBSCRIPTIONS);
+
+        // Add a channel to send messages to this client to the map
+        let (tx, rx) = tokio::sync::mpsc::channel::<Message>(100);
+        subscriptions.map.insert(token.clone(), tx);
+
+        // Once the guard goes out of scope, the connection will have been closed and the entry will be deleted from the map
+        (rx, WSAnonymousEntryMapGuard::new(subscriptions, token, addr))
     };
 
     Ok({
@@ -352,6 +450,69 @@ impl WebSocketUsers {
             push_send_update(ut, send, acting_device_uuid, conn).await;
         }
     }
+
+    pub async fn send_auth_request(
+        &self,
+        user_uuid: &String,
+        auth_request_uuid: &String,
+        acting_device_uuid: &String,
+        conn: &mut DbConn,
+    ) {
+        let data = create_update(
+            vec![("Id".into(), auth_request_uuid.clone().into()), ("UserId".into(), user_uuid.clone().into())],
+            UpdateType::AuthRequest,
+            Some(acting_device_uuid.to_string()),
+        );
+        self.send_update(user_uuid, &data).await;
+
+        if CONFIG.push_enabled() {
+            push_auth_request(user_uuid.to_string(), auth_request_uuid.to_string(), conn).await;
+        }
+    }
+
+    pub async fn send_auth_response(
+        &self,
+        user_uuid: &String,
+        auth_response_uuid: &str,
+        approving_device_uuid: String,
+        conn: &mut DbConn,
+    ) {
+        let data = create_update(
+            vec![("Id".into(), auth_response_uuid.to_owned().into()), ("UserId".into(), user_uuid.clone().into())],
+            UpdateType::AuthRequestResponse,
+            approving_device_uuid.clone().into(),
+        );
+        self.send_update(auth_response_uuid, &data).await;
+
+        if CONFIG.push_enabled() {
+            push_auth_response(user_uuid.to_string(), auth_response_uuid.to_string(), approving_device_uuid, conn)
+                .await;
+        }
+    }
+}
+
+#[derive(Clone)]
+pub struct AnonymousWebSocketSubscriptions {
+    map: Arc<dashmap::DashMap<String, Sender<Message>>>,
+}
+
+impl AnonymousWebSocketSubscriptions {
+    async fn send_update(&self, token: &str, data: &[u8]) {
+        if let Some(sender) = self.map.get(token).map(|v| v.clone()) {
+            if let Err(e) = sender.send(Message::binary(data)).await {
+                error!("Error sending WS update {e}");
+            }
+        }
+    }
+
+    pub async fn send_auth_response(&self, user_uuid: &String, auth_response_uuid: &str) {
+        let data = create_anonymous_update(
+            vec![("Id".into(), auth_response_uuid.to_owned().into()), ("UserId".into(), user_uuid.clone().into())],
+            UpdateType::AuthRequestResponse,
+            user_uuid.to_string(),
+        );
+        self.send_update(auth_response_uuid, &data).await;
+    }
 }
 
 /* Message Structure
@@ -381,6 +542,24 @@ fn create_update(payload: Vec<(Value, Value)>, ut: UpdateType, acting_device_uui
             ("ContextId".into(), acting_device_uuid.map(|v| v.into()).unwrap_or_else(|| V::Nil)),
             ("Type".into(), (ut as i32).into()),
             ("Payload".into(), payload.into()),
+        ])]),
+    ]);
+
+    serialize(value)
+}
+
+fn create_anonymous_update(payload: Vec<(Value, Value)>, ut: UpdateType, user_id: String) -> Vec<u8> {
+    use rmpv::Value as V;
+
+    let value = V::Array(vec![
+        1.into(),
+        V::Map(vec![]),
+        V::Nil,
+        "AuthRequestResponseRecieved".into(),
+        V::Array(vec![V::Map(vec![
+            ("Type".into(), (ut as i32).into()),
+            ("Payload".into(), payload.into()),
+            ("UserId".into(), user_id.into()),
         ])]),
     ]);
 
@@ -420,6 +599,7 @@ pub enum UpdateType {
 }
 
 pub type Notify<'a> = &'a rocket::State<Arc<WebSocketUsers>>;
+pub type AnonymousNotify<'a> = &'a rocket::State<Arc<AnonymousWebSocketSubscriptions>>;
 
 pub fn start_notification_server() -> Arc<WebSocketUsers> {
     let users = Arc::clone(&WS_USERS);
--- a/src/api/push.rs
+++ b/src/api/push.rs
@@ -255,3 +255,40 @@ async fn send_to_push_relay(notification_data: Value) {
         error!("An error occured while sending a send update to the push relay: {}", e);
     };
 }
+
+pub async fn push_auth_request(user_uuid: String, auth_request_uuid: String, conn: &mut crate::db::DbConn) {
+    if Device::check_user_has_push_device(user_uuid.as_str(), conn).await {
+        tokio::task::spawn(send_to_push_relay(json!({
+            "userId": user_uuid,
+            "organizationId": (),
+            "deviceId": null,
+            "identifier": null,
+            "type": UpdateType::AuthRequest as i32,
+            "payload": {
+                "id": auth_request_uuid,
+                "userId": user_uuid,
+            }
+        })));
+    }
+}
+
+pub async fn push_auth_response(
+    user_uuid: String,
+    auth_request_uuid: String,
+    approving_device_uuid: String,
+    conn: &mut crate::db::DbConn,
+) {
+    if Device::check_user_has_push_device(user_uuid.as_str(), conn).await {
+        tokio::task::spawn(send_to_push_relay(json!({
+            "userId": user_uuid,
+            "organizationId": (),
+            "deviceId": approving_device_uuid,
+            "identifier": approving_device_uuid,
+            "type": UpdateType::AuthRequestResponse as i32,
+            "payload": {
+                "id": auth_request_uuid,
+                "userId": user_uuid,
+            }
+        })));
+    }
+}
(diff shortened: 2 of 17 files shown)


quexten:
Logging in on mobile, and approving on mobile also works fine now.

quexten:
Ok, most things are working now, still missing:
- Testing other databases (only tested sqlite, others are implemented but untested)
- Deleting old requests periodically

@BlackDex, as I understand, `start_notification_server` starts the old (port 3012) Websocket server and makes rocket manage the state. Additionally, rocket itself handles the same endpoint, but the old port 3012 server is kept for compatibility.

Since no user has the new anonymous hub endpoint in their proxy, there does not need to be a similar separate server for the anonymous hub right? Just using rocket to handle it should be enough. (Currently this PR only handles the anonymous hub connections over rocket).

tessus:
The current rewrite rule looks like this according to https://github.com/dani-garcia/vaultwarden/wiki/Proxy-examples:

```
RewriteRule /notifications/hub(.*) ws://<SERVER>:3012/$1 [P,L]
```

I believe this rule is incorrect, because 
- there is no separate rule for `/notifications/hub/negotiate` (shouldn't this go to http instead of ws)
- the websocket connection does not get the path info passed on, but only everything that is sent after `/notifications/hub`, which means only the data is sent to `ws://<SERVER>:3012/`

So how would the ws server differentiate between this connection and yours?

If we were to analogously use a rewrite rule

```
RewriteRule /notifications/anonymous-hub(.*) ws://<SERVER>:3012/$1 [P,L]
```

the web socket server would either retrieve `<data>` or `<token>` but without path.

BlackDex:
> The current rewrite rule looks like this according to https://github.com/dani-garcia/vaultwarden/wiki/Proxy-examples:
> 
> ```
> RewriteRule /notifications/hub(.*) ws://<SERVER>:3012/$1 [P,L]
> ```
> 
> I believe this rule is incorrect, because
> 
> * there is no separate rule for `/notifications/hub/negotiate` (shouldn't this go to http instead of ws)
> * the websocket connection does not get the path info passed on, but only everything that is sent after `/notifications/hub`, which means only the data is sent to `ws://<SERVER>:3012/`
> 
> So how would the ws server differentiate between this connection and yours?

The current main/testing version doesn't need a separate port anymore. So all can go via the main port now.

Also, the negotiate endpoint doesn't exist anymore for a long time already. We kept it for a while because of backwards compatibility, but it can be removed now i think.

> 
> If we were to analogously use a rewrite rule
> 
> ```
> RewriteRule /notifications/anonymous-hub(.*) ws://<SERVER>:3012/$1 [P,L]
> ```
> 
> the web socket server would either retrieve `<data>` or `<token>` but without path.
> 
> P.S.: It seems that all proxy examples only send data to the root of the websocket connection. So I have no idea how `#[get("/hub?<data..>")]` is even triggered, if only `?<data..>` is sent to `ws://<SERVER>:3012/` I have to look into this, because I apparently have no clue what is going on.

`<data...>` is all that comes after the `?` which is just a query string which comes after the URL.

In general, we tend to fully remove the old port. But not yet next release, at least not for the web-socket notifications just yet.

tessus:
Thanks for the reply.

> The current main/testing version doesn't need a separate port anymore. So all can go via the main port now.

Right, but I didn't see a description how to reverse proxy this, nor how this is activated in the config. Or maybe I missed it.

> Also, the negotiate endpoint doesn't exist anymore for a long time already.

Good to know. Thanks.

> <data...> is all that comes after the ? which is just a query string which comes after the URL.

Right, I understand that. So only the query string is "forwarded" to the `ws://...`. But this PR introduces another endpoint, which sends a token. So how does the websocket server know which path to follow, if the endpoint is the same (since the data (what was sent until now) and the token (from this PR)) is just forwarded to `/`.

quexten:
You do not need any proxy rewrite rules. Not for the regular WebSocket hub and not for the new, anonymous hub. The rewrite rule was needed for the WebSocket endpoint on port 3012 (and still work for compatibility) but you do not need it anymore. 

At least using Traefik, it just works out of the box for this PR. I'm not sure whether you have to enable WebSocket support on other proxies.

> So how does the websocket server know which path to follow

To establish the WebSocket connection, first a HTTP GET request is made to wss://my.bitwarden.server.com/notifications/anonymous-hub?Token=UUID
The HTTP request contains the info about which "path" the connection is made to, and the query. The response is then 101 Switching Protocols, and the server/client switch to WebSocket.

The Vaultwarden internal routing is done by rocket, thus `#[get("/anonymous-hub?<token..>")]` is enough.

tessus:
> To establish the WebSocket connection, first a HTTP GET request is made to wss://my.bitwarden.server.com/notifications/anonymous-hub?Token=UUID

This is exactly my point.

I am using `:3012` and the Apache rewrite rule I am using is the one that is in the wiki page for proxies.

```
RewriteRule /notifications/hub(.*) ws://<SERVER>:3012/$1 [P,L]
```

So what happens is, if a request is sent to the proxy

```
https://my.bitwarden.server.com/notifications/hub?data=whatever
```

The proxy makes this request to the backend:

```
ws:/127.0.0.1:3012/?data=whatever
```

quexten:
> I am using :3012 and the Apache rewrite rule I am using is the one that is in the wiki page for proxies.

> RewriteRule /notifications/hub(.*) ws://<SERVER>:3012/$1 [P,L]

Yes, this was necessary as rocket did not have built in websocket support before. The port 3012 endpoint is dedicated to only the websocket hub, that's why the path is not necessary.

> So what happens is, if a request is sent to the proxy
> https://my.bitwarden.server.com/notifications/hub?data=whatever
> The proxy makes this request to the backend:
> ws:/127.0.0.1:3012/?data=whatever

Exactly, since 127.0.0.1:3012 in this case is vaultwarden's dedicated TCP port, only for websockets, no path is needed. Now that Rocket Websocket support is merged, this rewrite is not required. 
Technically, something like:
> RewriteRule /notifications/hub(.*) ws://<SERVER>:<HTTP_PORT>/notifications/hub$1 [P,L] should work (not tested), and would route over rocket instead, but you probably don't need a rewrite rule at all anymore.

For the anonymous hub of this PR, only rocket notifications are supported. So if you want to rewrite, you would have to use something like my example, but as I said, I don't think any rewrite rules are necessary beyond what's needed to get the HTTP requests to vaultwarden.

tessus:
I am still not sure what the correct way is. So let's say I want to use the "new" way of things. How do I set:

```
ROCKET_ADDRESS=0.0.0.0
ROCKET_PORT=8000

WEBSOCKET_ENABLED=false
WEBSOCKET_ADDRESS=0.0.0.0
WEBSOCKET_PORT=8000
```

and then no more rewrite stuff in Apache?

Maybe we should add a wiki page for that, because these 2 ways of doing notifications is a bit strange when we use both of them at the same time.
e.g. I am running the main branch but haven't changed the vw config nor the Apache reverse proxy config. So this seems to me that I am running in a hybrid mode of sorts.
That is if rocket now automatically creates a ws connection on the same port as defined in `ROCKET_PORT`.

quexten:
I think what you describe should work, although I have not tested with Apache yet. I have removed my WebSocket proxy rule from traefik, and set WEBSOCKET_ENABLED to false and the notifications still work.

> Maybe we should add a wiki page for that, because these 2 ways of doing notifications is a bit strange when we use both of them at the same time.

Well, there is https://github.com/dani-garcia/vaultwarden/wiki/Enabling-WebSocket-notifications, but I think the page can be removed entirely in the future since WebSockets do not require special proxy configuration anymore.

BlackDex:
That page can indeed be removed in the future.
All proxy examples need a review so that they will work properly.

tessus:
> All proxy examples need a review so that they will work properly.

I think that is a good idea.

I also think we might still need rewrite rules: `ProxyPass / http://<SERVER>:80/` would pass traffic to `http://server:80/`, but web sockets use a different protocol. So what would happen, if a websoccket connection was routed to `http://server:80/`? Shouldn't it be proxied to `ws://server:80/`?

Anyway, I'll stop spamming this PR with my questions. I guess I should start a discussion instead...

GeekCornerGH (changes requested):
You'll probably want to remove the "copy" part from the migration folder, right?

quexten:
From my side this is mostly done. I didn't have time yet to test on postgresql or mysql yet though.

Also beware that on the current vaultwarden web-build it will not show the option (all other clients do still immediately benefit from it though, desktop, browser-extension, and mobile), that will only be the case in the next version of the web-build.

GeekCornerGH (changes requested):
I assume this is finished? Maybe rebase your changes? I'll gladly test your changes once this is done

quexten:
@GeekCornerGH, yes, pretty much. It is now up-to-date with the latest main branch. I still had to test on mariadb and postgres, which is now done. The postgres migration had an incorrect type, which is now fixed. Other than that it is also working on mariadb / postgres.

TheManchineel:
@dani-garcia does this look good to you?

qymab:
> I think what you describe should work, although I have not tested with Apache yet. I have removed my WebSocket proxy rule from traefik, and set WEBSOCKET_ENABLED to false and the notifications still work.
> 
> > Maybe we should add a wiki page for that, because these 2 ways of doing notifications is a bit strange when we use both of them at the same time.
> 
> Well, there is https://github.com/dani-garcia/vaultwarden/wiki/Enabling-WebSocket-notifications, but I think the page can be removed entirely in the future since WebSockets do not require special proxy configuration anymore.

mind sharing which traefik rules you remove ? is this enough ?

```
- traefik.http.routers.bwws.middlewares=redirect-to-https
- traefik.http.routers.bwws.service=bw-ws
- traefik.http.services.bw-ws.loadbalancer.server.port=3012
```

src/api/push.rs
```diff
@@ -255,3 +255,24 @@ async fn send_to_push_relay(notification_data: Value) {
...
+    auth_request_uuid: String,
+    acting_device_uuid: String,
+    conn: &mut crate::db::DbConn,
+) {
+    if Device::check_user_has_push_device(user_uuid.as_str(), conn).await {
+        tokio::task::spawn(send_to_push_relay(json!({
+            "userId": user_uuid,
+            "organizationId": (),
+            "deviceId": Some(acting_device_uuid.clone()),
+            "identifier": Some(acting_device_uuid.clone()),
```
BlackDex, line 270: The `deviceId` and `identifier` should both be `null` here.
At least that is how upstream seems to do it.

src/api/push.rs
BlackDex, line 1: Besides adding a `AuthRequest` type, there should also be `AuthRequestResponse` type for after the response.
The payload should be almost identical to the `AuthRequest`, except the `deviceId` and `identifier` should match the client which has authorized the request. So in the case of a Mobile client the values of that specific mobile client.

BlackDex (changes requested):
Ok, I have done some checks, and it seems to work for the most part.
There are some changes which can be removed, and there are some additions needed.

I have not looked yet at the layout of the code if there could be some improvements there.
But the changes i mentioned can be addressed already i think :).

src/api/push.rs
quexten: Is that not this part:
https://github.com/dani-garcia/vaultwarden/blob/85eba6f62ed3da8658585758b71bd1f9bd6d1a7a/src/api/notifications.rs#L493-L501

or am I confusing something?

src/config.rs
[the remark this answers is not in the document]
quexten: Must have been left over from the initial implementation when I added support for both a dedicated anonymous websocket port and rocket. I removed it now.

src/api/push.rs
BlackDex: It is, but it didn't seem to get triggered during my tests.
I'll have to check again then. As mentioned before i didn't checked the full code yet. Just the workings of it and where it failed i did some checks.

src/api/push.rs
quexten: Done.

src/api/push.rs
quexten: Well, I only tested it on the desktop client. The notification only gets send on the anonymous notification endpoint to the device initiating the request. Mobile clients don't even use websockets at all but use polling instead to verify the auth request.

BlackDex: During my tests with a self-hosted environment i saw it send type 16 also via mobile push after the request was validated.

So, i think we should do the same.

quexten: Huh. So not only the anonymous client gets the notification but all signed in devices too? In that case I will implement that behavior, but the mobile clients don't even currently make use of it (https://github.com/bitwarden/mobile/blob/0271a4db4ce6f18fda928c6328d64fb56ca32835/src/App/Services/PushNotificationListenerService.cs#L134).

BlackDex: Strange. It might also be that they do not differentiate between it? It doesn't hurt i guess to add it. In case they do add support for it in some way in the future.

src/config.rs
```diff
@@ -409,6 +409,10 @@ make_config! {
         /// Event cleanup schedule |> Cron schedule of the job that cleans old events from the event table.
         /// Defaults to daily. Set blank to disable this job.
         event_cleanup_schedule:   String, false,  def,    "0 10 0 * * *".to_string();
+        /// Auth Request cleanup schedule |> Cron schedule of the job that cleans old auth requests from the auth request.
+        /// Defaults to every minute. Set blank to disable this job.
+        auth_request_purge_schedule:   String, false,  def,    "30 * * * * *".to_string();
```
BlackDex, line 414: There needs to be a check for this setting to verify it is a valid cron expression.
See as an example:
https://github.com/dani-garcia/vaultwarden/blob/3fa502b864751e1401b36adf3da7acc24363e074/src/config.rs#L896-L898

src/api/notifications.rs
```diff
@@ -144,6 +175,72 @@ fn websockets_hub<'r>(
     })
 }
 
+#[get("/anonymous-hub?<token..>")]
+async fn anonymous_websockets_hub<'r>(
```
BlackDex, line 179: `async` can be removed here.
Future clippy checks will fail because of this.
The `websockets_hub` function also does not have `async` there.

BlackDex (changes requested):

BlackDex:
Ok, I found a few more items.
I also looked if we could change the custom `from_i32` to use `num_derive`, but that seems to only use `i64`, while that can also be casted to an i32 again, it might be a bit overkill. But it would safe manual work if there are new types.

Anyway, not that important i think.

Looks good overall from my opinion.
With the changes requested above i think it's a go for me.

src/api/notifications.rs
quexten: Done

src/config.rs
quexten: Done

src/api/push.rs
quexten: I did not further compare exact contents of push messages as I can't set up an official self hosted environment at the moment. Still, I implemented the messages in the latest commits.

BlackDex:
@quexten, Ok this looks good to me.

The only thing I would like is to have this PR squashed into one commit to keep the history a bit cleaner.
Once that is done, it has a go for me.

quexten:
> @quexten, Ok this looks good to me.
> 
> The only thing I would like is to have this PR squashed into one commit to keep the history a bit cleaner. Once that is done, it has a go for me.

Done.

BlackDex (approved):

fxzxmic:
I have no idea.

stefan0xC:
@fxzxmic No idea what your error message says either, but can you try `testing` as it includes this fix #3831 (which presumably fixes the issue).

fxzxmic:
@stefan0xC I don't know where the problem is either. There will be the error mentioned above on the phone, and there will be no error reported on the desktop, but the login has not been successful.

Tyree:
I'm just testing Login with Device. The first attempt to log in on my browser, no notification came up on my phone. I had to go into settings and into Pending Login Requests. I confirmed and it let me in.
All other tests have, whether I Confirm or Deny, it fails with an error, "We were unable to process the request. Please try again or contact us."
My phone is communicating with the server. Requests come in and I can sync the vault. But these approvals don't work anymore.

BobWs:
I apologize in advance before "someone" bites my head off. I don't know if this is the right place to point it out given the many duplicate reports surrounding this topic, but I'll try anyway.

I have noticed that "login with device" does not work if your device is still on iOS 14. In fact I still have my iPhone running iOS 14 and there "login with device" does not work. 
I do get the push notification but when I click "confirm login" I get the same error message as previously stated by others "We were unable...etc".

If I then try this with a device that has iOS 16, and with the Bitwarden app reinstalled it does work. But if I then want to use this again the next day it no longer works. 

When I log out of the iOS 16 app and log back in then it works again. By the way this is not the case with iOS 14, there this does not help, neither log out or reinstall the app.

BlackDex:
> I apologize in advance before "someone" bites my head off. I don't know if this is the right place to point it out given the many duplicate reports surrounding this topic, but I'll try anyway.
> 
> I have noticed that "login with device" does not work if your device is still on iOS 14. In fact I still have my iPhone running iOS 14 and there "login with device" does not work. I do get the push notification but when I click "confirm login" I get the same error message as previously stated by others "We were unable...etc".
> 
> If I then try this with a device that has iOS 16, and with the Bitwarden app reinstalled it does work. But if I then want to use this again the next day it no longer works.
> 
> When I log out of the iOS 16 app and log back in then it works again. By the way this is not the case with iOS 14, there this does not help, neither log out or reinstall the app.

If you would have read all those duplicate reports, you would have seen that updating to the `testing` tagged docker image will solve your issues.

BobWs:
> If you would have read all those duplicate reports, you would have seen that updating to the `testing` tagged docker image will solve your issues.

HAHAHA I didn't expect any other answer from you....

ados8:
Amazing work implementing device login so quick.
I did run into the approve bug and using master password to login once resolves the approve request issue.
Further testing shows that MFA is still required and I was hoping to have an option to bypass should the user approve via device.
It could be without device login: email + password + MFA = successful.
But should the user use device login it then would be: email + device login = successful
This shouldn't lower security as it only works for known devices and still uses a 2 factor method/physical device.
Currently the "workaround" is to have MFA remember allowed but I don't want that because without device login that lowers security.
Are there any plans for such a feature?

quexten:
> Amazing work implementing device login so quick.
I did run into the approve bug and using master password to login once resolves the approve request issue.
Further testing shows that MFA is still required and I was hoping to have an option to bypass should the user approve via device.
It could be without device login: email + password + MFA = successful.
But should the user use device login it then would be: email + device login = successful
This shouldn't lower security as it only works for known devices and still uses a 2 factor method/physical device.
Currently the "workaround" is to have MFA remember allowed but I don't want that because without device login that lowers security.
Are there any plans for such a feature?

This is behaviour is dictated by upstream (bitwarden). It works the same on vault.bitwarden.com. Since vaultwarden aims to be compatible with upstream, if you want this feature you would need to bring it up with bitwarden instead.

mada199122:
When I click "Confirm login" on the desktop app to approve the login request made for example from the web vault, it fails and in the log I see this line

`[2023-10-09 11:39:52.112][vaultwarden::api::core::accounts::_][WARN] Data guard `Json < AuthResponseRequest >` failed: Parse("{\"key\":\"4.dAa5C9CAzqysrFu7iQFjDyp9x0vzyZSCP/SNLNLQUYJSgCvqM1/qzidjUjBIsBSQPssLo4JuYB6lEHTCG0BAr1cBCexgvtqExzMi4mMbqVqAVNHMI2pbdMHQ/t09sWYLmEaTaYzZRxFiVlymLDApcKUErmvxVfgES3LbGXV9WIUdrdYOKcXu0KiNM1hSzIPlUU/XNKgwpegYDjKfeMrOnT+LKDQpgxJOba2Hh1I7Bmedd8BZI4lE5aOgkjzipkXuvx5fDwLX5yvgQ6qFeqA+/5eGBhfZuMpLEnVw/SBP65ich9qIKDtuw11OisGKhCs5KHME0kwt0iLo564dCNAR0Q==\",\"deviceIdentifier\":\"c2dcb8bc-82cd-485f-9cb3-ea0b45ad8659\",\"requestApproved\":true}", Error("missing field `masterPasswordHash`", line: 1, column: 437)).`

stefan0xC:
@mada199122 this has already been fixed in #3831 (currently available in `testing`)

mada199122:
Ah ok thank you!!!

zolakt:
> I'm just testing Login with Device. The first attempt to log in on my browser, no notification came up on my phone. I had to go into settings and into Pending Login Requests. I confirmed and it let me in.

I have this same issue, no notification, but it works when I manually go into the Android app and approve it. Does anyone know how to solve this?

michaelkebe:
Working on my side.
Maybe this helps?

https://github.com/dani-garcia/vaultwarden/wiki/Enabling-Mobile-Client-push-notification

stefan0xC:
Also please don't use old Pull Requests to ask for support. Just open a new discussion and provide more information about your issue.

PR #3606, opened 2023-06-22
Add group import on invite

### LINKED ISSUE #3599: Group Membership Information Not Retained on Organization Invite Creation
### Deployment environment

* Vaultwarden version: v1.28.1
* Web-vault version: v2023.3.0b
* OS/Arch: linux/x86_64
* Uses a reverse proxy: true
* Internet access: true
* DNS Check: true
* Server/NTP Time Check: true
* HTTPS Check: true
* Database type: SQLite
* Database version: 3.39.2
* Clients used: 
* Reverse proxy and version: 
* Other relevant information: 
* Install method:
* Reverse proxy and version: Caddy

### Steps to reproduce
1. Enable beta groups functionality
2. Create a group
3. Within an organization, send an invite to a user to join the organization with the group specified at the time of the invite
4. The invite will not retain the group information and it will need to be readded.

### Expected behaviour
With collections, an invite can be sent to a user with the collections set at the time of the invite so that when the join the org, the collection is visible. It would be nice if group membership information could mimic this behavior. Particularly, if the organization has a policy set to enforce 2FA before the user can be added to the org, you cannot go back and set the group membership after the invite goes out until they completely join.

### Actual behaviour
The invite is sent to the user without the group membership. You can only add it after the user is enrolled (with 2FA enforcement policy enabled).

farodin91:
Fixes #3599

### DIFF
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -849,6 +849,7 @@ struct CollectionData {
 #[allow(non_snake_case)]
 struct InviteData {
     Emails: Vec<String>,
+    Groups: Vec<String>,
     Type: NumberOrString,
     Collections: Option<Vec<CollectionData>>,
     AccessAll: Option<bool>,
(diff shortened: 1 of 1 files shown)


BlackDex (approved):

BlackDex:
Thanks @farodin91 !

PR #3609, opened 2023-06-23
add user to collection during creation

farodin91:


src/api/core/organizations.rs
```diff
@@ -424,6 +424,10 @@ async fn post_organization_collections(
             .await?;
     }
 
+    if !headers.org_user.access_all {
```
BlackDex, line 427: I think this should only happen if the user is a manager when they create the collection.
owners and admins do not have to be added to this list, Bitwarden also does not do that.

BlackDex (changes requested):
Thanks for this fix.
I do think it needs a small adjustment and it needs to check if the user is a manager, admins and owners do not need to be added to this.

src/api/core/organizations.rs
farodin91: I think this should behave more less as implement in bitwarden https://github.com/bitwarden/server/blob/master/src/Api/Controllers/CollectionsController.cs#L153. As Vaultwarden has no difference between manger and admin all access.

BlackDex: There is a difference, and i tested this with a Self-Hosted Bitwarden, and it does not seem to add a admin/owner to that list.
It only does it when the user creating a collection is a manager.

There is a difference, because a owner and admin always have access to all items, but.
If you disable the `access all` checkbox of a owner or admin, and select specific collections or groups for that owner or admin, then those items will appear on the main vault page, if they are not selected, they will not see those items.

At least, that is the behavior i found during my tests.

BlackDex: Also, if you look at the source you see that checks for admin or higher access too: https://github.com/bitwarden/server/blob/4c61d05b243fbdf8f230d1a2b6f3d2f88f53008e/src/Core/Context/CurrentContext.cs#L332C1-L332C1

farodin91: I check for manager.

src/api/core/organizations.rs
```diff
@@ -424,6 +424,10 @@ async fn post_organization_collections(
             .await?;
     }
 
+    if headers.org_user.atype == UserOrgType::Manager {
```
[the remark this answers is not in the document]
farodin91: changed.

BlackDex (approved):

dani-garcia (approved):

PR #3632, opened 2023-07-02
fix missing password check while manual reset password enrollment

sirux88:
While enrolling reset password as user the web interface forces to type the password to authenticate oneself.
In its current state the backend doesn't check the password. Therefore a random passord can be used for authentification.

This PR solves this issue.

### DIFF
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -2668,6 +2668,7 @@ async fn delete_group_user(
 #[allow(non_snake_case)]
 struct OrganizationUserResetPasswordEnrollmentRequest {
     ResetPasswordKey: Option<String>,
+    MasterPasswordHash: Option<String>,
 }
 
 #[derive(Deserialize)]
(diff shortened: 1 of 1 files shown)


src/api/core/organizations.rs
```diff
@@ -2849,6 +2850,19 @@ async fn put_reset_password_enrollment(
         err!("Reset password can't be withdrawed due to an enterprise policy");
     }
 
+    let user = headers.user;
+
+    if reset_request.ResetPasswordKey.is_some() {
+        match reset_request.MasterPasswordHash {
+            Some(password) => {
+                if !user.check_valid_password(&password) {
```
BlackDex, line 2858: No need to define `user` here as a new variable.
You can just use:

```suggestion
    if reset_request.ResetPasswordKey.is_some() {
        match reset_request.MasterPasswordHash {
            Some(password) => {
                if !headers.user.check_valid_password(&password) {
```

src/api/core/organizations.rs
```diff
@@ -2858,8 +2872,7 @@ async fn put_reset_password_enrollment(
         EventType::OrganizationUserResetPasswordWithdraw as i32
     };
 
-    log_event(log_id, org_user_id, org_id, headers.user.uuid.clone(), headers.device.atype, &headers.ip.ip, &mut conn)
-        .await;
+    log_event(log_id, org_user_id, org_id, user.uuid.clone(), headers.device.atype, &headers.ip.ip, &mut conn).await;
```
BlackDex, line 2875: This change is not needed by just using the `headers.user` above.

BlackDex (changes requested):

sirux88:
@BlackDex 
let me know if I shall squash the commits

BlackDex (approved):

dani-garcia (approved):

Issue #496, opened 2019-05-31
Prometheus metrics


Skeen:
Hi,

If I was gonna implement metrics importing using https://github.com/sd2k/rocket_prometheus or https://github.com/pingcap/rust-prometheus would such an addition be welcome to the project;

1. Behind a feature flag, or
2. In general build in, but enabled by an environmental variable, or
3. Always enabled, or
4. Not at all

mprasil:
Personally I'd love to see Prometheus metrics being available behind environment variable. (or generally as any other setting we have) Even better if there was some option to enable authorization.

dani-garcia:
Personally any addition is welcome as long as it's easy to enable/disable and doesn't complicate the code a ton or impact performance.

About how to activate the feature, the best way is hiding it behind an environment variable. The only exceptions I can think of are when:
 - the new features add external dependencies and link to system binaries
 - it's impossible to switch the feature on and off at runtime

Sadly both exceptions apply to adding new database support but I can't imagine they would for this case.

Out of curiosity, what kind of metrics are you interested in collecting?

Skeen:
Okay that sounds good, I'm thinking I'll add 4 environmental variables then, something like:

    METRICS_ENABLE
    METRICS_USERNAME
    METRICS_PASSWORD
    METRICS_URI

Where's enable will be the feature toggle with a default to false, username and password will be optional, and if provided will ask for http basic auth before providing metrics and finally metrics uri will be the suburl where metrics are served, defaulting to /metrics

The metrics I'm planning to expose are general http request cycle stuff, numbers of syncs potentially labeled by uuid and database metrics such as the total number of ciphers, ciphers by organization and user, etc. 

I'm very open for input on specific metrics you'd like to see included too though.

MarsWarrior:
I would like to see failed logins, so I can set an alarm on that event in Grafana!

And maybe:
- Nr of ogranizations
- Nr of users
- Nr of folders
- Nr of entries
- Nr of collections
- Nr of attachments
- Nr of lookups for username/password

Skeen:
Hi @MarsWarrior, I already have the database metrics set up. Including number of users per organization, and such, some of which may have to go behind an environmental variable or be toggleable for people with larger userbases, if they want to avoid too many metrics. 
- For instance the number of ciphers per user, which can produce a lot of metrics on a busy server. 

[...]

PR #3634, opened 2023-07-02
Provide Prometheus metrics endpoint

mtandersson:
This makes it possible to enable /metrics to provide Prometheus metrics (protected by PROMETHEUS_ENABLED flag).

I did not spend super much time on enable every possible metrics as I wanted to seek feedback before. It should be easy to add more metrics later.

This pull request also addresses issue #496

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -126,7 +126,7 @@ dependencies = [
  "log",
  "parking",
  "polling",
- "rustix",
+ "rustix 0.37.20",
  "slab",
  "socket2 0.4.9",
  "waker-fn",
@@ -154,7 +154,7 @@ dependencies = [
  "cfg-if",
  "event-listener",
  "futures-lite",
- "rustix",
+ "rustix 0.37.20",
  "signal-hook",
  "windows-sys 0.48.0",
 ]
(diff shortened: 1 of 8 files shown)


tessus:
I like metrics in general. I am not so sure how important they are in this case, since the server component is usually not very resource intensive. Afaik the most work is done on the client side.

I was also thinking about adding OpenTelemetry tracing to the server code, but haven't
- checked, whether such a PR would be merged
- found the time to do so

RealOrangeOne:
There was some discussion a while ago about adding metrics: https://github.com/dani-garcia/vaultwarden/discussions/2952

It resulted in an external project: https://github.com/Tricked-dev/vwmetrics

BlackDex:
As @RealOrangeOne already mentioned, it was discussed before. And I'm not really a fan of adding this built-in into Vaultwarden.
Also, this implementation exposes those metrics to the rest of the world when enabled.

As also mentioned, there already is an exporter of some kind created which does this, except maybe for the request duration.
My view of this is to not add it, but point users to the other project.

mtandersson:
Cool, no problem thanks for the consideration. And keep up the good work!

PR #3639, opened 2023-07-03
release optimization

tessus:
This optimization is not only for size but also runtime. Who doesn't want to use less CPU cycles? ;-)

Since releases are not done that often and usually via build machines and pipelines, I don't see it as an issue that the build time doubled.

That being said, if this is too extreme for a default release, I can also add another profile (e.g. called `production` or something similar).

| Description         | Change | Before   | After     |
|---------------------|--------|----------|-----------|
| Build time          | +110%  | 5min 31s | 11min 34s |
| Binary size (bytes) |  -42%  | 49466432 | 28576480  |

Build command: `cargo build --features sqlite,mysql,postgresql --release`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -168,10 +168,14 @@ rocket = { git = 'https://github.com/SergioBenitez/Rocket', rev = 'ce441b5f46fdf
 # rocket_ws = { git = 'https://github.com/SergioBenitez/Rocket', rev = 'ce441b5f46fdf5cd99cb32b8b8638835e4c2a5fa' } # v0.5 branch
 
 # Strip debuginfo from the release builds
-# Also enable thin LTO for some optimizations
+# Also enable fat LTO, opt-level 3, codegen-units 1 for best optimization
 [profile.release]
-strip = "debuginfo"
-lto = "thin"
+opt-level = 3
+debug = false
+panic = "unwind"
+lto = true
+codegen-units = 1
+strip = true
 
 # Always build argon2 using opt-level 3
 # This is a huge speed improvement during testing


Cargo.toml
```diff
@@ -168,10 +168,14 @@ rocket = { git = 'https://github.com/SergioBenitez/Rocket', rev = 'ce441b5f46fdf
...
 
 # Strip debuginfo from the release builds
-# Also enable thin LTO for some optimizations
+# Also enable fat LTO, opt-level 3, codegen-units 1 for best optimization
 [profile.release]
-strip = "debuginfo"
-lto = "thin"
+opt-level = 3
+debug = false
+panic = "unwind"
```
BlackDex, line 175: This is the default, no reason to define it

Cargo.toml
```diff
@@ -168,10 +168,14 @@ rocket = { git = 'https://github.com/SergioBenitez/Rocket', rev = 'ce441b5f46fdf
 # rocket_ws = { git = 'https://github.com/SergioBenitez/Rocket', rev = 'ce441b5f46fdf5cd99cb32b8b8638835e4c2a5fa' } # v0.5 branch
 
 # Strip debuginfo from the release builds
-# Also enable thin LTO for some optimizations
+# Also enable fat LTO, opt-level 3, codegen-units 1 for best optimization
 [profile.release]
-strip = "debuginfo"
-lto = "thin"
+opt-level = 3
+debug = false
```
BlackDex, line 174: This is default, no need to define it.

Cargo.toml
```diff
@@ -168,10 +168,14 @@ rocket = { git = 'https://github.com/SergioBenitez/Rocket', rev = 'ce441b5f46fdf
 # rocket_ws = { git = 'https://github.com/SergioBenitez/Rocket', rev = 'ce441b5f46fdf5cd99cb32b8b8638835e4c2a5fa' } # v0.5 branch
 
 # Strip debuginfo from the release builds
-# Also enable thin LTO for some optimizations
+# Also enable fat LTO, opt-level 3, codegen-units 1 for best optimization
 [profile.release]
-strip = "debuginfo"
-lto = "thin"
+opt-level = 3
```
BlackDex, line 173: This is default, no need to define it.

Cargo.toml
```diff
@@ -168,10 +168,14 @@ rocket = { git = 'https://github.com/SergioBenitez/Rocket', rev = 'ce441b5f46fdf
...
+debug = false
+panic = "unwind"
+lto = true
+codegen-units = 1
+strip = true
```
BlackDex, line 178: We actually want `symbols`, this removes all, `symbols` are needed for better backtrace output.

Cargo.toml
```diff
@@ -168,10 +168,14 @@ rocket = { git = 'https://github.com/SergioBenitez/Rocket', rev = 'ce441b5f46fdf
...
-# Also enable thin LTO for some optimizations
+# Also enable fat LTO, opt-level 3, codegen-units 1 for best optimization
 [profile.release]
-strip = "debuginfo"
-lto = "thin"
+opt-level = 3
+debug = false
+panic = "unwind"
+lto = true
+codegen-units = 1
```
BlackDex, line 177: This might be a good option to squeeze out some more performance, but I'm not sure if the gain rectifies the more then double build-time.

BlackDex (commented):
Thanks for thinking about this.
Though, I'm not sure if this will have much benefit in the end.

The binary size change is mostly because of stripping all metadata.

Cargo.toml
tessus: I have come to be very careful with defaults. I have run into issues several times because someone just changed the defaults. But I can remove them.

Cargo.toml
tessus: ok, in this case, I'll just add another profile that inherits release with those 2 options.

I think for package mantainers and also for people who don't care about build time but value performance this is a nice addition.

e.g. I don't care if a build takes 5 minutes, 10 minutes, or 30 minutes. I don't make builds every 10 minutes, so what gives? Also, when I run a build it is usually done on a Threadripper with 16 cores, 32 threads. My tests were done on a VM with 4 threads of the host.

Cargo.toml
tessus: see above

Cargo.toml
tessus: see above

Cargo.toml
BlackDex: If Rust changes the defaults, it will be mentioned in the changelogs, and if changed, then it might be for a good reason, if not, we can always adjust it. But in general i think it's better to keep using the defaults provided by Rust instead of re-defining them here.

Cargo.toml
tessus: backtraces are usually done with debug code and not production code. Well, it depends on the project of course.

But a panic can unwind just fine without symbols. At least you can see where it panicked. But I agree, for a `RUST_BACKTRACE=1`, it certainly makes sense to have symbols.

when using a separate profile for production I think it is ok to strip the symbols. People have to specify that profile explicitly anyway

Cargo.toml
BlackDex: I'm not a fan of adding arbitrary profiles. You might like this, but others might not.
You can override these options by setting environment variables, see: https://doc.rust-lang.org/cargo/reference/environment-variables.html#configuration-environment-variables

I suggest to use that for your build environment, as I think that the `Cargo.toml` needs to stay as independent as possible.

tessus: I guess that's it then. Closing.

Cargo.toml
BlackDex: We handle these backtraces internally our selfs and output it to stdout/logs if something like this happens.
People do not need to enable `RUST_BACKTRACE=1`. The reason is that we want to have as much as info during that specific call. Because, sometimes reproducing a panic could be complex if not almost impossible sometimes.

tessus:
> The binary size change is mostly because of stripping all metadata.

FYI, stripping the symbols is only about 3 MB.

With `lto = true` and `codegen-units = 1`, the file size is 31715048. Compared to 28576480 (strip symbols as well) and 49466432 (current state).

I don't really care about the file size. This is not why I am doing the optimization. But you are correct, env vars are possible, although `export CARGO_PROFILE_RELEASE_BUILD_OVERRIDE_STRIP=true` does not work, but I've decided to keep the symbols. All good.

BlackDex:
@tessus shouldn't that be `CARGO_PROFILE_RELEASE_STRIP=true`?

tessus:
This key does not exist according to the link you sent. However, there is the following:

```
CARGO_PROFILE_<name>_BUILD_OVERRIDE_<key>
```

BlackDex:
I would suggest to try it. Maybe they forgot to update the documentation

tessus:
I can try, but I've already decided to keep the symbols.

tessus:
You know, I am the curious type. ;-)

Yep, `CARGO_PROFILE_RELEASE_STRIP=true` works, so the documentation is missing that info.

I am still keeping the symbols.

PR #3651, opened 2023-07-04
fix version when compiled at a specific commit

tessus:
When a specific commit is checked out from the main branch, the vaultwarden version is reported as `vaultwarden x.y.z-githash (HEAD)`. This is a problem, because the admin interface reports this as a version from a branch called HEAD, while in reality the commit was from the main branch.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -72,7 +72,7 @@ fn version_from_git_info() -> Result<String, std::io::Error> {
     // Combined version
     if let Some(exact) = exact_tag {
         Ok(exact)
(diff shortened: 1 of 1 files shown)


BlackDex:
Could you provide the steps you took to checkout a specific commit?
So that we can mimic this and test it?

tessus:
Sure thing.

```
git clone https://github.com/dani-garcia/vaultwarden.git
cd vaultwarden
git checkout a6cfdddf
```

BlackDex (approved):

dani-garcia (approved):

PR #3715, opened 2023-07-19
Update Rust and Crates

tessus:
- Updated Rust to v1.71.0
- Updated MSRV to v1.69.0
- Updated crates


BlackDex:
This is also done in #3573, including some other updates.

tessus:
well, it doesn't seem this other PR is going to merged anytime soon.

BlackDex:
Why do you think that? We are waiting for the next patch release to be tagged before we want to merge that one, because we don't want all those changes in the current stable yet.

And since Rust v1.71 now also uses Musl v1.2.3 for all targets, I also want to wait for a merge like that to not end up in the next patch release.

tessus:
> Why do you think that?

because it's been open for over a month.

I understand your reasoning for not merging it. But the fact that musl targets hold up other stuff is rather annoying. Just my opinion. e.g. when I compile the latest master, I want to use the latest Rust. 
Well, I can't because the toolchain file still says 1.70.0. (Yea, I can change this locally.)

Anyway, feel free to close this PR then. It doesn't seem the new web-vault will be available anytime soon either. So it doesn't matter.

BlackDex:
Some changes just need some more chaining changes.
And i personally want to keep a current stable Vaultwarden which is tested a lot now on the same Rust version before releasing a new patch release.

The reason that PR there is for _months_, is because i saw this coming.

You don't need to patch there toolchain file, you just need to run cargo like this `cargo +stable build ....`

tessus:
> You don't need to patch there toolchain file, you just need to run cargo like this cargo +stable build ....

thanks. this is good to know

BlackDex:
I'm going to close this as a duplicate of #3573 and the other reasons given above.
If Rust v1.71.0 didn't had these changes I probably merged it my self. But for now i really want to release a new patch version first (maybe this weekend), and after that start with the next phase.

Besides that, still thanks for the PR! We appreciate all the help.

PR #3730, opened 2023-07-27
Update admin interface

BlackDex:
- Updated the admin interface dependencies.
- Replace bootstrap-native with bootstrap
- Added auto theme with an option to switch to dark/light
- Some small color changes
- Added a dev only function to always load static files from disk


src/static/templates/admin/base.hbs
```diff
@@ -33,22 +33,65 @@
...
-                {{/if}}
+                    {{/if}}
                     <li class="nav-item">
                         <a class="nav-link" href="{{urlpath}}/" target="_blank" rel="noreferrer">Vault</a>
                     </li>
                 </ul>
 
                 {{#if logged_in}}
-                    <a class="btn btn-sm btn-secondary" href="{{urlpath}}/admin/logout">Log Out</a>
+                <a class="btn btn-sm btn-secondary" href="{{urlpath}}/admin/logout">Log Out</a>
```
stefan0xC, line 43:
```suggestion
                <a class="btn btn-sm btn-secondary order-md-last" href="{{urlpath}}/admin/logout">Log Out</a>
```

It might just be my personal preference but I think it would be better if the logout button stays all the way on the right if you are logged in. (Using [`.order-md-last`](https://getbootstrap.com/docs/5.2/layout/columns/#order-classes) so it's not affecting the dropdown menu.)

src/static/templates/admin/base.hbs
```diff
@@ -33,22 +33,65 @@
...
+                <ul class="navbar-nav">
+                    <li class="nav-item dropdown">
+                        <button
+                            class="btn btn-link nav-link py-0 px-0 px-lg-2 dropdown-toggle d-flex align-items-center"
+                            id="bd-theme" type="button" aria-expanded="false" data-bs-toggle="dropdown"
+                            data-bs-display="static" aria-label="Toggle theme (auto)">
+                            <span class="my-1 fs-4 theme-icon-active">
+                                <use>&#9775;</use>
+                            </span>
+                            <span class="d-lg-none ms-2" id="bd-theme-text">Toggle theme</span>
```
stefan0xC, line 55:
```suggestion
                            <span class="d-md-none ms-2" id="bd-theme-text">Toggle theme</span>
```
I think this should be `.d-md-none` as we collapse at the `md`-breakpoint.

stefan0xC (approved):
Looks mostly good to me.

src/static/scripts/admin.js
```diff
@@ -37,36 +37,107 @@ function _post(url, successMsg, errMsg, body, reload_page = true) {
...
-    }).catch( e => {
+    }).catch(e => {
         if (e.error === false) { return true; }
         else { msg(`${errMsg}\n${e.body}`, reload_page); }
     });
 }
 
+// Bootstrap Theme Selector
+const getStoredTheme = () => localStorage.getItem("theme");
+const setStoredTheme = theme => localStorage.setItem("theme", theme);
```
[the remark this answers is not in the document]
BlackDex: I do not like it that the admin interface uses something from the web-vault, i would like to keep those separated.
Also, if the web-vault changes something in regards to the theme color, it would break the admin interface.

I think it is better to keep those apart.

src/static/templates/admin/base.hbs
BlackDex: It doesn't really matter to me, it looks nice either way. But, i have changed it :).
Thanks for the review.

src/static/templates/admin/base.hbs
BlackDex: It shouldn't matter in the end, since it's the exact same CSS. But it probably is better to use the correct flow indeed.
I also spotted an other one which needed a change too.

src/static/templates/admin/base.hbs
```diff
@@ -33,22 +33,66 @@
...
+                                    Auto
+                                </button>
+                            </li>
+                        </ul>
+                    </li>
+                </ul>
+
                 {{#if logged_in}}
-                    <a class="btn btn-sm btn-secondary" href="{{urlpath}}/admin/logout">Log Out</a>
+                <a class="btn btn-sm btn-secondary order-md-last" href="{{urlpath}}/admin/logout">Log Out</a>
```
stefan0xC, line 86:
```suggestion
                <a class="btn btn-sm btn-secondary" href="{{urlpath}}/admin/logout">Log Out</a>
```
Not important but we can remove this again.

stefan0xC (approved):

src/static/templates/admin/base.hbs
BlackDex: Well, lets make it pretty then.

dani-garcia (approved):

PR #3735, opened 2023-07-30
Fix typo in config.rs

eltociear:
Usefull -> Useful


BlackDex:
Thanks, but this is already covered in #3734 .

Issue #432, opened 2019-03-12
[Feature Request] Allow pre-defining user settings and disabling some


jvacek:
I'd like to make sure that all of the users on the server automatically get set values in the following areas (that are different to the ones that are given by default now).
Some of these should be allowed to be changed, and some should not

- Lock options (I want to make sure that the vault will always be locked within n amount of minutes)
- Disable using pin lock (Password only, ever)
- Disable Auto-fill (nuissance but something that I don't care for if it's changed)
- Default URI Match (should be changeable if desired but seeing as we use a lot fo subdomains on our intranet, suggestions are otherwise messy)

There are many others, but basically to boil down my request: I would like to be able to set some default values for newly created users, and I'd also like the option to have some of these options "locked" so that the user cannot change them.

mprasil:
We have #246 for feature requests, so maybe submit that there. However I think all of these are set and saved only on the client side, so I'm not sure what/if we actually can do something about these settings. It sounds like you'll need to modify and compile your own version of the client apps/plugins.

I'm going to close this one, if you still feel that there's something bitwarden_rs could do to help with this, submit feature request under #246.

dani-garcia:
To expand a bit on this, 2 and 3 are basically impossible from the server side. We could make an option to configure session length and to disable auto renewal for 1, but that will only kick the users out when they make a server call, so as long as they don't modify anything it won't work.

About 3, I think it would be possible to modify the ciphers url matching when they are being saved: using the default option, they send `match: null`, while the other options send a number, so it could be reasonable to implement, I think.

mprasil:
Good point about the 3rd thing @dani-garcia, I have forgotten that the match detection wasn't always configurable and there was just null used for backwards compatibility. I think it could be even more helpful if we could just return some set `DEFAULT_MATCH_DETECTION` for any `null` setting stored. That way you can set/change the default server-wide rather than having to re-save all items on any default change.

dani-garcia:
The problem with storing null and returning a different value is that when the user edits the cipher, they'll have the returned value selected in the dropdown, and they'll send it when submitting instead of the initial null, so we are basically in the same place, just a bit later.

I think it would be better to be consistent in this case and just save it all the time.

matlink:
What should be the behavior when the user has changed the matching algorithm in its settings? `match: null` can then either be user default or server default.

matlink:
[...]

PR #3748, opened 2023-08-03
implement server side default URI matching

matlink:
See item 4 of https://github.com/dani-garcia/vaultwarden/issues/432

Watch https://github.com/dani-garcia/vaultwarden/issues/432#issuecomment-1663854241 for a little different workflow


BlackDex:
From my point of view i still think this will cause more issues then it might solve for some people.
It is also error prone during handling or when you want to change the default to something else.

Going to close this one. If you think it still should be applied, please provide more detailed information or workflows on how this can work and ill recheck,

PR #3751, opened 2023-08-04
Optimized Favicon downloading

BlackDex:
Some optimizations in regards to downloading Favicon's.

I also encounterd some issues with accessing some sites where the connection got dropped or closed early. This seems a reqwest/hyper thingy, https://github.com/hyperium/hyper/issues/2136. This is now also fixed.

General:

- Decreased struct size (reduces memory usage)
- Decreased memory allocations (faster HTML parsing)
- Optimized tokenizer a bit more to only emit tags when all attributes are there and are valid.

reqwest/hyper connection issue:
The following changes helped solve the connection issues to some sites. The endresult is that some icons are now able to be downloaded always instead of sometimes.

- Enabled some extra reqwest features, `deflate` and `native-tls-alpn` (Which do not bring in any extra crates since other crates already enabled them, but they were not active for Vaultwarden it self)
- Configured reqwest to have a max amount of idle pool connections per host
- Configured reqwest to timeout the idle connections in 10 seconds

### DIFF
--- a/src/api/icons.rs
+++ b/src/api/icons.rs
@@ -19,7 +19,7 @@ use tokio::{
     net::lookup_host,
 };
 
-use html5gum::{Emitter, EndTag, HtmlString, InfallibleTokenizer, Readable, StartTag, StringReader, Tokenizer};
+use html5gum::{Emitter, HtmlString, InfallibleTokenizer, Readable, StringReader, Tokenizer};
 
 use crate::{
     error::Error,
@@ -46,10 +46,15 @@ static CLIENT: Lazy<Client> = Lazy::new(|| {
     // Generate the cookie store
     let cookie_store = Arc::new(Jar::default());
 
+    let icon_download_timeout = Duration::from_secs(CONFIG.icon_download_timeout());
+    let pool_idle_timeout = Duration::from_secs(10);
     // Reuse the client between requests
     let client = get_reqwest_client_builder()
         .cookie_provider(Arc::clone(&cookie_store))
-        .timeout(Duration::from_secs(CONFIG.icon_download_timeout()))
+        .timeout(icon_download_timeout)
+        .pool_max_idle_per_host(5) // Configure the Hyper Pool to only have max 5 idle connections
+        .pool_idle_timeout(pool_idle_timeout) // Configure the Hyper Pool to timeout after 10 seconds
+        .trust_dns(true)
         .default_headers(default_headers.clone());
 
     match client.build() {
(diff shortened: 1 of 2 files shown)


src/api/icons.rs
```diff
@@ -820,43 +819,65 @@ impl reqwest::cookie::CookieStore for Jar {
...
+    pub attributes: BTreeMap<HtmlString, HtmlString>,
 }
 
-#[derive(Default, Debug)]
+struct FaviconToken {
+    tag: Tag,
+    closing: bool,
+}
+
+// struct FaviconEmitter // size = 184 (0xB8), align = 0x8
```
dani-garcia, line 844:
```suggestion
```
This comment should be removed, otherwise it's going to end up going out of sync

src/api/icons.rs
```diff
@@ -871,87 +892,71 @@ impl Emitter for FaviconEmitter {
...
+            }
+            _ => None,
+        };
     }
 
     fn push_attribute_name(&mut self, s: &[u8]) {
-        self.current_attribute.as_mut().unwrap().0.extend(s);
+        if self.current_attribute.is_some() {
+            self.current_attribute.as_mut().unwrap().0.extend(s);
+        }
```
dani-garcia, line 948:
```suggestion
        if let Some(attr) = &mut self.current_attribute {
            attr.0.extend(s)
        }
```

src/api/icons.rs
```diff
@@ -871,87 +892,71 @@ impl Emitter for FaviconEmitter {
...
     }
 
     fn push_attribute_value(&mut self, s: &[u8]) {
-        self.current_attribute.as_mut().unwrap().1.extend(s);
+        if self.current_attribute.is_some() {
+            self.current_attribute.as_mut().unwrap().1.extend(s);
+        }
```
dani-garcia, line 954:
```suggestion
        if let Some(attr) = &mut self.current_attribute {
            attr.1.extend(s)
        }
```

src/api/icons.rs
```diff
@@ -415,39 +422,31 @@ fn get_favicons_node(
...
+        let tag_name: &[u8] = &token.tag.name;
+        match tag_name {
+            TAG_LINK => {
+                icon_tags.push(token.tag);
             }
-            FaviconToken::EndTag(tag) => {
-                if *tag.name == TAG_HEAD {
-                    break;
+            TAG_BASE => {
+                let href = std::str::from_utf8(token.tag.attributes.get(ATTR_HREF).unwrap()).unwrap_or_default();
```
dani-garcia, line 437: Should we handle this being None instead of calling unwrap?: `token.tag.attributes.get(ATTR_HREF).unwrap()`

src/api/icons.rs
BlackDex: It's already out of sync haha. That was the old size of I'm correct. That should have been removed indeed.

src/api/icons.rs
BlackDex: Well, it can not ever be None since the tokenizer only emits if there is some for this specific item. So that should in theory never happen.

BlackDex: But, i have made some changes to catch it, and also remove a clone which wasn't needed in the end.

dani-garcia (approved):

PR #3752, opened 2023-08-04
US or EU Data Region Selection

toto-xoxo:
Following https://github.com/dani-garcia/vaultwarden/discussions/3731

Add the possibility of choosing between data region us or eu for push notifications

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -80,8 +80,10 @@
 # PUSH_ENABLED=true
 # PUSH_INSTALLATION_ID=CHANGEME
 # PUSH_INSTALLATION_KEY=CHANGEME
+# PUSH_RELAY_REGION=us
 ## Don't change this unless you know what you're doing.
 # PUSH_RELAY_URI=https://push.bitwarden.com
+# IDENTITY_URI=https://identity.bitwarden.com
 
 ## Controls whether users are allowed to create Bitwarden Sends.
 ## This setting applies globally to all users.
--- a/.env.template
+++ b/.env.template
@@ -81,9 +81,6 @@
 # PUSH_INSTALLATION_ID=CHANGEME
 # PUSH_INSTALLATION_KEY=CHANGEME
 # PUSH_RELAY_REGION=us
-## Don't change this unless you know what you're doing.
-# PUSH_RELAY_URI=https://push.bitwarden.com
-# IDENTITY_URI=https://identity.bitwarden.com
 
 ## Controls whether users are allowed to create Bitwarden Sends.
 ## This setting applies globally to all users.
--- a/.env.template
+++ b/.env.template
@@ -30,10 +30,6 @@
 ## Define the size of the connection pool used for connecting to the database.
 # DATABASE_MAX_CONNS=10
 
-## Database timeout
-## Timeout when acquiring database connection
-# DATABASE_TIMEOUT=30
-
 ## Database connection initialization
 ## Allows SQL statements to be run whenever a new database connection is created.
 ## This is mainly useful for connection-scoped pragmas.
--- a/.env.template
+++ b/.env.template
@@ -30,6 +30,10 @@
 ## Define the size of the connection pool used for connecting to the database.
 # DATABASE_MAX_CONNS=10
 
+## Database timeout
+## Timeout when acquiring database connection
+# DATABASE_TIMEOUT=30
+
 ## Database connection initialization
 ## Allows SQL statements to be run whenever a new database connection is created.
 ## This is mainly useful for connection-scoped pragmas.
--- a/.env.template
+++ b/.env.template
@@ -95,6 +95,10 @@
 ## Disabled by default. Also check the EVENT_CLEANUP_SCHEDULE and EVENTS_DAYS_RETAIN settings.
 # ORG_EVENTS_ENABLED=false
 
+## Controls whether users can change their email.
+## This setting applies globally to all users
+# EMAIL_CHANGE_ALLOWED=true
+
 ## Number of days to retain events stored in the database.
 ## If unset (the default), events are kept indefinitely and the scheduled job is disabled!
 # EVENTS_DAYS_RETAIN=
--- a/.env.template
+++ b/.env.template
@@ -95,10 +95,6 @@
 ## Disabled by default. Also check the EVENT_CLEANUP_SCHEDULE and EVENTS_DAYS_RETAIN settings.
 # ORG_EVENTS_ENABLED=false
 
-## Controls whether users can change their email.
-## This setting applies globally to all users
-# EMAIL_CHANGE_ALLOWED=true
-
 ## Number of days to retain events stored in the database.
 ## If unset (the default), events are kept indefinitely and the scheduled job is disabled!
 # EVENTS_DAYS_RETAIN=
--- a/.env.template
+++ b/.env.template
@@ -77,10 +77,13 @@
 # WEBSOCKET_PORT=3012
 
 ## Enables push notifications (requires key and id from https://bitwarden.com/host)
+## If you choose "European Union" Data Region, uncomment PUSH_RELAY_URI and PUSH_IDENTITY_URI then replace .com by .eu
 # PUSH_ENABLED=true
 # PUSH_INSTALLATION_ID=CHANGEME
 # PUSH_INSTALLATION_KEY=CHANGEME
-# PUSH_RELAY_REGION=us
+## Don't change this unless you know what you're doing.
+# PUSH_RELAY_URI=https://push.bitwarden.com
+# PUSH_IDENTITY_URI=https://identity.bitwarden.com
 
 ## Controls whether users are allowed to create Bitwarden Sends.
 ## This setting applies globally to all users.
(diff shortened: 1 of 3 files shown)


src/config.rs
```diff
@@ -380,8 +380,26 @@ make_config! {
...
+        };
+        /// Identity uri
+        identity_uri:           String, false,  auto,   |c| {
+            let relay_region = match c.push_relay_region.as_str() {
+                "us" => "com",
+                "eu" => "eu",
+                _ => "com", // Default to US if the region is not recognized
+            };
+            format!("https://identity.bitwarden.{}", relay_region)
+        };
```
RealOrangeOne, line 402: Suggestion: Instead of switching just the TLD, switch the entire hostname. It's a little more code duplication, but much easier to read, and easier to maintain should upstream decide to switch it in future on something other than TLD.

BlackDex: I also think it is better to have two manually configurable options such as `push_relay_uri` and `push_identity_uri` which both default to there `.com` defaults like it is for Bitwarden Self-Hosted, and have people configure it them selfs.
As @RealOrangeOne already mentioned, they could change both items, or even add an Asian region for that matter.
Keeping these options a bit more flexible would be better.

Maybe some validation could be added to make sure they start with `https` and are a valid URI/URL.

toto-xoxo: the default variables are already defined to .com but I don't rewrite the full URL every time.

Default
.com

.env to region us
.com

.env to region eu
.eu

.env to invalid region
.com

users can still write their own variable with `push_relay_uri` and `push_identity_uri` (actually identity_uri but i can rename it)

The question is should I duplicate the URL each time or change the domain extension?

BlackDex: This would add more complexity by allowing custom URI's and region selection.

Better to just have one way.
And yes i would suggest to rename it to `push_identity_uir` to keep all together.

toto-xoxo: you mean I should give up the idea of region? I was going for it because of the discussion we had. https://github.com/dani-garcia/vaultwarden/discussions/3731#discussioncomment-6598973

toto-xoxo:
Hey I've done some change let me know what do u think !

.env.template
```diff
@@ -76,8 +76,7 @@
 # PUSH_ENABLED=true
 # PUSH_INSTALLATION_ID=CHANGEME
 # PUSH_INSTALLATION_KEY=CHANGEME
-## Don't change this unless you know what you're doing.
-# PUSH_RELAY_BASE_URI=https://push.bitwarden.com
```
williamdes, line 80: You could keep this ENV for users that want to have a custom push server, that said I am not sure my advice is good.

src/config.rs
```diff
@@ -745,6 +763,26 @@ fn validate_config(cfg: &ConfigItems) -> Result<(), Error> {
         )
     }
 
+    if cfg.push_enabled {
+        let push_relay_uri = cfg.push_relay_uri.to_lowercase();
```
lnchan, line 767: URIs are case-sensitive, this could cause issues for case-sensitive HTTP servers/reverse-proxies if a path is specified

src/config.rs
```diff
@@ -745,6 +763,26 @@ fn validate_config(cfg: &ConfigItems) -> Result<(), Error> {
...
+        if !push_relay_uri.starts_with("https://") {
+            err!("`PUSH_RELAY_URI` must start with 'https://'.")
+        }
+
+        if Url::parse(&push_relay_uri).is_err() {
+            err!("Invalid URL format for `PUSH_RELAY_URI`.");
+        }
+
+        let push_identity_uri = cfg.push_identity_uri.to_lowercase();
```
lnchan, line 776: Same here

src/config.rs
```diff
@@ -380,8 +380,26 @@ make_config! {
...
-        /// Push relay base uri
-        push_relay_uri:         String, false,  def,    "https://push.bitwarden.com".to_string();
+        /// Push relay region
+        push_relay_region:      String, false,  def,    "us".to_string();
+        /// Push relay uri
+        push_relay_uri:         String, false,  auto,   |c| {
+            let push_relay_uri = match c.push_relay_region.as_str() {
+                "us" => "https://push.bitwarden.com".to_string(),
+                "eu" => "https://push.bitwarden.eu".to_string(),
+                _ => "https://push.bitwarden.com".to_string(), // Default to "us" region
```
lnchan, line 390: Why default to the US region?

BlackDex:
Having thought about this a bit further.

Either we force the URI's and not have anything configurable for the users except the region, thus `EU` or `US`.
Or, we keep it a bit more flexible, and only provide the `PUSH_RELAY_URI` and `PUSH_IDENITY_URI` including the `/connect/token`.
So no `PUSH_RELAY_REGION`.

The default would be `.com` and not `.eu` since `.com` was the only option before.
There should be good comments into the `.env.template` and the `config.rs` so it will be clear on what to fill in.
Probably best if we supply the options people can fill in which are the defaults from Bitwarden, and if they want to use something else, it's up to them.

I actually tend a bit more towards the latter option, because of the flexibility, also in case Bitwarden is going to change.
Or, maybe someone even builds there own Bitwarden client with there own Android/iOS messaging endpoints.

src/config.rs
mightyBroccoli: URI are actually case-insensitive, at least if you're going by the RFC:
https://www.rfc-editor.org/rfc/rfc3986#section-3.1
I would assume servers that rely on the URI casing, are in the wrong on this one.

sorcix: You're linking to the **scheme** part, which is case insensitive. You can use `https://` or `HTTPS://` as you wish. The **path** is case sensitive.

.env.template
williamdes: @toto-xoxo 
In fact my proposal to use a custom server was discussed here: https://github.com/dani-garcia/vaultwarden/discussions/2469

williamdes (changes requested):
I am also with @BlackDex to keep it fully flexible in the `.env` file

Other related discussion about custom push notifications servers: https://github.com/dani-garcia/vaultwarden/discussions/2469

> I actually tend a bit more towards the latter option, because of the flexibility, also in case Bitwarden is going to change.
Or, maybe someone even builds there own Bitwarden client with there own Android/iOS messaging endpoints.

Exactly !

toto-xoxo:
Implemented requested change. Not sure if this is what you want @BlackDex waiting for your feedback

BlackDex:
@toto-xoxo  There seems to be a conflict with the .env.template.
Looks like the rebase/merge with main is not done correctly.

williamdes (approved):
Looks much nicer and easier to maintain

Any thoughts ~~@BlackDex~~ @dani-garcia ?

.env.template
```diff
@@ -30,6 +30,10 @@
 ## Define the size of the connection pool used for connecting to the database.
 # DATABASE_MAX_CONNS=10
 
+## Database timeout
+## Timeout when acquiring database connection
+# DATABASE_TIMEOUT=30
+
```
williamdes, line 36:
```suggestion
```

You can directly apply this with GitHub

.env.template
```diff
@@ -93,6 +95,10 @@
 ## Disabled by default. Also check the EVENT_CLEANUP_SCHEDULE and EVENTS_DAYS_RETAIN settings.
 # ORG_EVENTS_ENABLED=false
 
+## Controls whether users can change their email.
+## This setting applies globally to all users
+# EMAIL_CHANGE_ALLOWED=true
+
```
williamdes, line 101:
```suggestion
```
I guess this one too is not related to this PR

.env.template
```diff
@@ -30,10 +30,6 @@
 ## Define the size of the connection pool used for connecting to the database.
 # DATABASE_MAX_CONNS=10
 
-## Database timeout
-## Timeout when acquiring database connection
-# DATABASE_TIMEOUT=30
-
```
williamdes, line 36: now that you rebased the last commits should be dropped..

toto-xoxo:
everything should be good now check @BlackDex @williamdes

BlackDex (approved):
Looks good to me this way.

src/config.rs
williamdes: @BlackDex should the to lower be removed?
Anyway, if users put wrong URLs it's their wish and problem

BlackDex: Well, if it doesn't matter if users put in with our without capitals, then this check is fine too be sure it starts with https i think. It doesn't change the value

williamdes: Ah yes okay it's a check
Sorry I  missed that

williamdes (approved):

.env.template
```diff
@@ -451,4 +453,4 @@
 ## HaveIBeenPwned API Key, request it here: https://haveibeenpwned.com/API/Key
 # HIBP_API_KEY=
 
-# vim: syntax=ini
+# vim: syntax=ini
```
SuperSandro2000, line 456:
```suggestion
# vim: syntax=ini

```

BlackDex (approved):

dani-garcia (approved):

dani-garcia (approved):
Had to solve some merge conflicts, thanks!

PR #3754, opened 2023-08-04
Fix #3624: fix manager permission within groups

### LINKED ISSUE #3624: Issue with group-level access permissions and collection management for managers
Tested with the most recent testing docker image (Digest:sha256:78f4cf6c42004d70afb8673ef55bd88f25b62094b41275e935947e4ed6e8db17)

### Subject of the issue

Group-level access permissions are not working as intended with regards to collection management (for members with the manager role).

### Deployment environment (Generated via diagnostics page)

* Vaultwarden version: v1.28.1-e7f083de
* OS/Arch: linux/x86_64
* Uses a reverse proxy: true
* Internet access: false
* DNS Check: true
* Server/NTP Time Check: n/a
* HTTPS Check: true
* Database type: PostgreSQL
* Database version: PostgreSQL 14.2 (Ubuntu 14.2-1.pgdg20.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0, 64-bit
* Clients used: Issue is seen with the web-vault. Not using any other clients.
* Reverse proxy and version: nginx/1.18.0 (Ubuntu)
* Other relevant information: 

Vaultwarden is started like this: docker run -d --env-file /opt/docker/vw-data-test/.env --name vw-test -v /opt/docker/vw-data-test:/data -p 0.0.0.0:8081:81 -p 0.0.0.0:3013:3013 --restart on-failure harbor/mirror/docker.io/vaultwarden/server@sha256:78f4cf6c42004d70afb8673ef55bd88f25b62094b41275e935947e4ed6e8db17

### Steps to reproduce the issue

As an admin assign the manager role to Member.
Add Member to Group that has 'Can edit' on Collection.

Log in as Member. Go to Organizations (top menu). Collection cannot be edited/modified as one would expect with the manager role and 'Can edit'.

The small pop-up menu with 'Edit info','Access',Delete' is simply not accessible. Normally this small pop-up menu can be accessed by clicking the 3 small dots to the far right of a collection or by clicking the "arrow" (pointing down) right next to the collection name once you're already looking inside the collection in question.  Neither the 3 dots, nor the arrow pointing down is shown in the web UI.

However, new collections can without problems be created. As such, create new NestedCollection with Collection as "parent" and give 'Can edit' to Group.

Now NestedCollection has been created, but Member also cannot edit/modify this one.

The ability to modify/edit collections only works, if Member gets 'Can edit' applied directly as a user-level access permission (which of course defeats the whole purpose of utilizing group-level access permissions which are highly convenient in many scenarios with several users).

Now comes the funny/puzzling part... If Member gets even just 'Can view' applied as a user-level access permission, the 'Can edit' access permission from the Group starts to work immediately.

matlink:
Fix #3624 by checking if the user is member of a group that has access to the collection.

### DIFF
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -337,6 +337,9 @@ async fn get_org_collections_details(org_id: &str, headers: ManagerHeadersLoose,
             Vec::with_capacity(0)
         };
 
+        // uuids of users belonging to a group of this collection
+        let group_users = GroupUser::get_collection_group_users_uuid(&col.uuid, &mut conn).await;
+
         let mut assigned = false;
         let users: Vec<Value> = coll_users
             .iter()
@@ -350,6 +353,11 @@ async fn get_org_collections_details(org_id: &str, headers: ManagerHeadersLoose,
                 SelectionReadOnly::to_collection_user_details_read_only(collection_user).to_json()
             })
             .collect();
+
+        // if current user is in any collection-assigned group
+        if group_users.contains(&user_org.uuid) {
+            assigned = true;
+        }
 
         if user_org.access_all {
             assigned = true;
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -345,9 +345,6 @@ async fn get_org_collections_details(org_id: &str, headers: ManagerHeadersLoose,
             Vec::with_capacity(0)
         };
 
-        // uuids of users belonging to a group of this collection
-        let has_collection_access_via_group = GroupUser::get_group_members_for_collection(&col.uuid, &mut conn).await;
-
         let mut assigned = has_full_access;
         let users: Vec<Value> = coll_users
             .iter()
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -324,7 +324,9 @@ async fn get_org_collections_details(org_id: &str, headers: ManagerHeadersLoose,
     let coll_users = CollectionUser::find_by_organization(org_id, &mut conn).await;
 
     // check if current user has full access to the organization (either directly or via any group)
-    let has_full_access_to_org = user_org.access_all || (CONFIG.org_groups_enabled() && GroupUser::has_full_access_by_member(org_id, &user_org.uuid, &mut conn).await);
+    let has_full_access_to_org = user_org.access_all
+        || (CONFIG.org_groups_enabled()
+            && GroupUser::has_full_access_by_member(org_id, &user_org.uuid, &mut conn).await);
 
     for col in Collection::find_by_organization(org_id, &mut conn).await {
         // assigned indicates whether the current user has access to the given collection
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -321,6 +321,8 @@ async fn get_org_collections_details(org_id: &str, headers: ManagerHeadersLoose,
     };
 
     let coll_users = CollectionUser::find_by_organization(org_id, &mut conn).await;
+    // uuids of users in groups having access to all collections
+    let all_access_group_uuids = GroupUser::get_all_access_group_users_uuid(org_id, &mut conn).await;
 
     for col in Collection::find_by_organization(org_id, &mut conn).await {
         let groups: Vec<Value> = if CONFIG.org_groups_enabled() {
@@ -354,8 +356,8 @@ async fn get_org_collections_details(org_id: &str, headers: ManagerHeadersLoose,
             })
             .collect();
 
-        // if current user is in any collection-assigned group
-        if group_users.contains(&user_org.uuid) {
+        // if current user is in any collection-assigned group or in a group having access to all collections
+        if group_users.contains(&user_org.uuid) || all_access_group_uuids.contains(&user_org.uuid) {
             assigned = true;
         }
 
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -326,24 +326,11 @@ async fn get_org_collections_details(org_id: &str, headers: ManagerHeadersLoose,
     // check if current user has full access to the organization (either directly or via any group)
     let has_full_access_via_group =
         CONFIG.org_groups_enabled() && GroupUser::has_full_access_by_member(org_id, &user_org.uuid, &mut conn).await;
-    let has_full_access = user_org.access_all || has_full_access_via_group;
+    let has_full_access_to_org = user_org.access_all || has_full_access_via_group;
 
     for col in Collection::find_by_organization(org_id, &mut conn).await {
-        // get the group details for the given collection
-        let groups: Vec<Value> = if CONFIG.org_groups_enabled() {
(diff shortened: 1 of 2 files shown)


stefan0xC:
What if a Manager is member of a group with access to all collections?

matlink:
> What if a Manager is member of a group with access to all collections?

you are write, we lack something since in this case the manager cannot edit the collection. I'm looking into it, thanks =)

matlink:
@stefan0xC This d282b32d79dea870a54f7020a764b277702c4e54 should do the job.

farodin91:
Any progress?

BlackDex:
It has to be checked first. Did not had time yet to validate. Working on other issues when i have the time.

stefan0xC (changes requested):
Sorry, it took me a bit but I've finally had time to review your changes.

I think the comments and the naming of some variables/functions could be improved to make it more obvious why something is done so the changes are easier to review. I've added some suggestions for your changes but this probably applies to other variables and comments as well.

With the renamed variables I think it is easier to see how your approach could be improved upon. I.e. instead of returning the two lists of member ids and later checking if they contain the uuid of the current user, we really only care if `user_org` has full access via a group or if not, if they have access to a given collection via a group. So we could make two functions that check

1. if the given user_org has full access to an organization (either directly or via a group) 
2. if the given user_org is a member of a group assigned to the given collection.

src/api/core/organizations.rs
```diff
@@ -353,7 +358,13 @@ async fn get_org_collections_details(org_id: &str, headers: ManagerHeadersLoose,
             })
             .collect();
 
-        if user_org.access_all {
+        // if current user is in any collection-assigned group
+        // or in a group having access to all collections
+        // or itself has access to all collections
+        if group_users.contains(&user_org.uuid)
+            || all_access_group_uuids.contains(&user_org.uuid)
+            || user_org.access_all
```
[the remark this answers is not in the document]
stefan0xC: Also it's not necessary to check `all_access_group_uuids.contains(&user_org.uuid)` in each for loop as the result would not change. So you can start with combining it beforehand (i.e. by setting `let has_full_access = user_org.access_all || all_access_group_uuids.contains(&user_org.uuid)` and then inside the for loop change the starting value to `let mut assigned = has_full_access` and if it's not assigned at this point check if the user is a member of a group assigned to the given collection.

BlackDex:
@matlink, could you rebase and address the items mentioned by @stefan0xC?
Looks ok for the rest of it i think.

matlink:
@BlackDex will do ;)

matlink:
@BlackDex @stefan0xC I've made the changes you requested and rebased. Needs further testing ;)

src/api/core/organizations.rs
```diff
@@ -352,9 +363,10 @@ async fn get_org_collections_details(org_id: &str, headers: ManagerHeadersLoose,
             })
             .collect();
 
-        if user_org.access_all {
-            assigned = true;
-        }
+        // if current user is in any collection-assigned group
+        // or in a group having access to all collections
+        // or itself has access to all collections
+        assigned = !assigned && has_collection_access_via_group.contains(&user_org.uuid);
```
stefan0xC, line 369: I think this is wrong. It should be 
```suggestion
        if (!assigned) {
            assigned = has_collection_access_via_group.contains(&user_org.uuid);
        }
```

stefan0xC: As one liner it should be `A || B` (which is only false if both A and B are false) not `!A && B` (which is only true if A is false and B is true) so you could write it as
```suggestion
        assigned = assigned || has_collection_access_via_group.contains(&user_org.uuid);
```

stefan0xC (changes requested):
Thanks for taking the time. After taking another look at the whole change set again I found one thing (besides the formatting which you can fix by simply running `cargo fmt`) that should really be changed in this PR, because I think it's important to not do more work than is necessary and also not have side effects from a disabled feature.

Another thing that should be changed too is the one comment I suggested you remove previously because we do the checks earlier in the code:

1. check if the current user has direct access to all collections either via their membership
2. or (if groups are enabled) if the current user is a member of a group with direct access to all collections
3. and then for each collection check if the current user is assigned to the given collection
4. and (if groups are enabled and only if it's necessary) we also check if the current user has access to the given collection via a group

Otherwise it looks good to me. (I mean I think that 2. and 4. could be further improved upon but that can just as well be done in a separate PR.)

stefan0xC:
I've taken the liberty to create a PR of my own https://github.com/matlink/vaultwarden/pull/1

stefan0xC (approved):
@BlackDex Can you review/test the changes? 🙈

src/api/core/organizations.rs
```diff
@@ -321,41 +321,49 @@ async fn get_org_collections_details(org_id: &str, headers: ManagerHeadersLoose,
         None => err!("User is not part of organization"),
     };
 
+    // get all collection memberships for the current organization
     let coll_users = CollectionUser::find_by_organization(org_id, &mut conn).await;
 
+    // check if current user has full access to the organization (either directly or via any group)
+    let has_full_access_via_group =
+        CONFIG.org_groups_enabled() && GroupUser::has_full_access_by_member(org_id, &user_org.uuid, &mut conn).await;
+    let has_full_access_to_org = user_org.access_all || has_full_access_via_group;
```
BlackDex, line 330: This can probably be a bit more efficient.
Currently there is a query done to check if the user has full access via the group.
But, that is not needed when the user already has full access.

Probably this should work (not tested):
```suggestion
    let has_full_access_to_org = user_org.access_all || (CONFIG.org_groups_enabled() && GroupUser::has_full_access_by_member(org_id, &user_org.uuid, &mut conn).await);
```

BlackDex (changes requested):

BlackDex:
And, it needs a rebase of course.
I tested it btw, including my suggestion, that seems to work fine.

fthobe:
Hey Guys, I don't want to put pressure on you people, but do you think we can nudge this one over the finish line? I would literally send you pizza :))))

matlink:
So I just need to rebase?

BlackDex:
> So I just need to rebase?

@matlink no see my request.
And a rebase too of course.

matlink:
@BlackDex I've applied you suggestion.
it looks like the rebase went well, no conflict.
Keep me in touch if needed.
Cheers,

BlackDex:
> @BlackDex I've applied you suggestion. it looks like the rebase went well, no conflict. Keep me in touch if needed. Cheers,

@matlink well, a small item. The code isn't formatted correctly.
Run a `cargo fmt` and that should be fixed.

matlink:
> @matlink well, a small item. The code isn't formatted correctly. Run a `cargo fmt` and that should be fixed.

Done.

BlackDex (approved):

fthobe:
Hey, can this be merged now? The issue seems to be resolved to everybody's satisfaction. If testing is needed I would volunteer.

PR #3757, opened 2023-08-05
Fix #3413: push to users accessing the collections using groups

### LINKED ISSUE #3413: Live sync for organization items only working when not using roles / groups
### Subject of the issue
* live sync for organization items **does not work** if you are using roles / groups. 
* live sync **does not work** if you are adding members via group to a collection
* live sync **works** only if you are adding a member directly to a collection

### Deployment environment
* vaultwarden version: 1.28.1-alpine
* Install method: docker
* Clients used: all (tested: Windows Desktop, Browser, Firefox, Chrome)
* Reverse proxy and version: traefik-2.9.9
* MySQL/MariaDB or PostgreSQL version: 8.X
* Other relevant details: websocket works as without using groups everything runs as expected

### Steps to reproduce
1. Create an organisation
2. Create a collection
3. Create a group
4. Create two test accounts
5. assign one account to the group (3) and then the group (3) to the collection (2)
6. assign one account directly to the collection (2)
7. add / update / delete items in the collection (2)

**Result**
* account from 5: will only see the changes via reload
* account from 6: will see the changes immediately

### Expected behaviour
Live sync all the items to all the group members not only to the "hardcoded" directly assigned collection members.

### Actual behaviour
Changes from organization members (independent whether using groups or not) will be reflected immediately to the organization owner but if the owner / somebody else creates a new entry in the collection this change will only be available after a "hard" reload (ctrl+shift+r).

This issue is resolved by no using groups! By not using them everything works as expected.

BR wucherpfennig

matlink:
I bet it can be done with a single query within the `find_by_cipher_and_org` function but I didn't managed to get it work (the `filter` gets complicated, adding `groups::access_all.eq(true)` works but I can't figure out how to check if the user is in a group that has access to the collection).

Basically it retrieves the users having access to the cipher with a group that either has access to all collections or is bound to the cipher's collection. Then users uuid are used to push changes to clients.

### DIFF
--- a/src/db/models/cipher.rs
+++ b/src/db/models/cipher.rs
@@ -273,15 +273,18 @@ impl Cipher {
             None => {
                 // Belongs to Organization, need to update affected users
                 if let Some(ref org_uuid) = self.organization_uuid {
-                    for user_org in UserOrganization::find_by_cipher_and_org(&self.uuid, org_uuid, conn).await.iter() {
-                        User::update_uuid_revision(&user_org.user_uuid, conn).await;
-                        user_uuids.push(user_org.user_uuid.clone());
+                    // users having access to the collection
+                    let mut collection_users =
+                        UserOrganization::find_by_cipher_and_org(&self.uuid, org_uuid, conn).await;
+                    if CONFIG.org_groups_enabled() {
+                        // members of a group having access to the collection
+                        let group_users =
+                            UserOrganization::find_by_cipher_and_org_with_group(&self.uuid, org_uuid, conn).await;
+                        collection_users.extend(group_users);
                     }
-                    for user_org in
-                        UserOrganization::find_by_cipher_and_org_with_group(&self.uuid, org_uuid, conn).await.iter()
-                    {
+                    for user_org in collection_users {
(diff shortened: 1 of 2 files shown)


mightyBroccoli:
What needs to be done to get this feature on its way?

BlackDex:
> What needs to be done to get this feature on its way?

I have to test it first, like with all PR's, unless it's an easy change, which this is not in my opinion.

mightyBroccoli:
I tested a build with this patch added onto the master.
My testsetup consisted of a docker-compose setup in addition to a "normal bare metal" installation, in both cases connecting to a MariaDB instance.

What I can tell is that functionality wise, this patch works. Testing with just 2 active users and some dummies, one updating a cipher and one "recieving" the update, I was able to immediatly see the changed, added entry.
I did not test the involved mysql query though, I suppose it would be neccessary to verify that the query isn't wasting SQL resources.

BlackDex:
Seems to work. While i would liked this to be able in just one query, it seems to be difficult and would make it too complex probably.

BlackDex (approved):

dani-garcia (approved):

Issue #3785, opened 2023-08-18
Health check failing when IPv6 enabled


Caros2017:
I think I messed up with my pull request, so now I am making an issue for it.

### Subject of the issue
When I enabled IPv6 I later realized that the healthcheck of the container was failing.

Setting `ROCKET_ADDRESS=::` in docker-compose gives the error:
```
$ docker-compose up -d --build
services.vaultwarden.environment.13 must be a string
```

So to be able to use IPv6 `ROCKET_ADDRESS` needs to be set to `::0`
`ROCKET_ADDRESS=::0`

In bash `::` doesn't equal `::0` and hence my healthcheck is failing, since it isn't rewritten to localhost

### Solution
In `docker/healtcheck.sh` change this:

```
if [ -z "${addr}" ] || [ "${addr}" = '0.0.0.0' ] || [ "${addr}" = '::' ]; then
    addr='localhost'
fi

```

To this:

```
if [ -z "${addr}" ] || [ "${addr}" = '0.0.0.0' ] || [ "${addr}" = '::0' ]; then
    addr='localhost'
fi
```

stefan0xC:
You could also escape the string. I.e. if you use a YAML list, you can quote the whole list item in your `docker-compose.yml` like this:

```yaml
    environment:
      - 'ROCKET_ADDRESS=::'
```

Alternatively you could also specify the environment variables as a dictionary

```
    environment:
      ROCKET_ADDRESS: "::"
```

Just be aware that `docker-compose` [will interpret them differently](https://docs.docker.com/compose/compose-file/05-services/#environment), which is why I suggest you check your config with: `docker-compose config`.

Caros2017:
> You could also escape the string. I.e. if you use a YAML list, you can quote the whole line in your `docker-compose.yml`:
> 
> ```yaml
>     environment:
>       - 'ROCKET_ADDRESS=::'
> ```
> 
> Alternatively you could also specify the environment variables as a dictionary like this:
> 
> ```
>     environment:
>       ROCKET_ADDRESS: "::"
> ```
> 
> Just be aware that `docker-compose` [will interpret them differently](https://docs.docker.com/compose/compose-file/05-services/#environment), which is why I suggest you check your config with: `docker-compose config`.

Thanks! Will use this instead!

PR #3786, opened 2023-08-18
IPv6 healthcheck.sh failing fix

Caros2017:
Solution for #3785

### DIFF
--- a/docker/healthcheck.sh
+++ b/docker/healthcheck.sh
@@ -46,7 +46,7 @@ if [ -r "${CONFIG_FILE}" ]; then
 fi
 
 addr="${ROCKET_ADDRESS}"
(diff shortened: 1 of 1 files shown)


williamdes (approved):

stefan0xC (commented):
As explained in #3785 I don't think this fix is needed as it's not really an issue with the `healthcheck.sh` but the way you can configure environment variables in ` docker-compose.yml`.

Because the short hand version `::' would be more common than an unnecessary 0, I think this change might also possible break some users existing configuration.

Caros2017:
> As explained in #3785 I don't think this fix is needed as it's not really an issue with the `healthcheck.sh` but the way you can configure environment variables in ` docker-compose.yml`.
> 
> Because the short hand version `::` would be more common than adding an arbitray (and unnecessary) 0, I think this change might also possible break some users existing configuration.

I think you are right especially about the part that it might possibly break other their environment. Will close this, thanks!

PR #3796, opened 2023-08-24
Use Diesels MultiConnections Derive

BlackDex:
With this PR we remove almost all custom macro's to create the multiple database type code. This is now handled by Diesel it self.

This removed the need of the following functions/macro's:
- `db_object!`
- `::to_db`
- `.from_db()`

It is also possible to just use one schema instead of multiple per type.

---

**NOTE:** There are just few items which cause some issues with models and schemas having the exact same name, and that causes some ambiguity. I have solved that currently by adding `schema::` in-front of those specific items. But it would be nicer if we didn't need that though.

Not sure what the best way is to address this. Probably renaming the tables is the best option i think.

I still want to do some testing with the different database types with an empty database, an existing database which needs migrations etc.. etc.. 

Also, we probably want to wait until Diesel releases v.2.1.1, since there are currently some bugs and missing features which got fixed, but not yet merged.

And if someone else wants to help test, please do!, The more the better.


BlackDex:
> Is there anything from the diesel side of things I can to do move this forward?

Not really, but thanks for asking.

We don't want to break a pending PR for SSO for example, since this is a huge change.

I do think that is the only one btw.

BlackDex:
Also, during this i encountered a few naming-scheme errors, which i probably need to fix too.
Best to do that before merging something like this i think.
Same for some other optimizations like removing the `mut` for all the `DbConn` parameters.
Something on my ToDo list :)

alvinpeters:
This is great! I'm in the process of adding non-diesel database backends (key-value stores like RocksDB for single-node setup and FoundationDB for clusters). This will make my life much easy. Is there anything I can help (on the coding side) to get this merged?

BlackDex:
> This is great! I'm in the process of adding non-diesel database backends (key-value stores like RocksDB for single-node setup and FoundationDB for clusters). This will make my life much easy. Is there anything I can help (on the coding side) to get this merged?

Not really a.t.m. we were waiting for some other PR's to be merged to prevent big changes for those PR's. Mainly the SSO one.

BlackDex:
Closing this as it's superseded by #6279

PR #3804, opened 2023-08-27
Fix External ID not set during DC Sync

### LINKED ISSUE #3777: No external user ID created when using bwdc and user already has an account
### Subject of the issue

I user Bitwarden directory connector (v2022.11.0) to sync my AD users and groups into my Organisation.
Thereby I noticed that users who already have an account do not get an external id. 
This seems to be independent of whether the account already exists because the user is using the private vault and is not yet part of an organization or the user has already been manually invited to the organization.

* Vaultwarden version: v1.29.1
* OS/Arch: linux/x86_64
* Uses a reverse proxy: true
* Internet access: true
* DNS Check: true
* Server/NTP Time Check: true
* HTTPS Check: true
* Database type: SQLite
* Database version: 3.41.2
* Clients used: 
* Reverse proxy and version: 
* Other relevant information: 

```json
{
  "_duo_akey": null,
  "_enable_duo": false,
  "_enable_email_2fa": true,
  "_enable_smtp": true,
  "_enable_yubico": false,
  "_icon_service_csp": "",
  "_icon_service_url": "",
  "_ip_header_enabled": true,
  "_smtp_img_src": "cid:",
  "admin_session_lifetime": 20,
  "admin_token": "***",
  "data_folder": "data",
  "database_conn_init": "",
  "database_max_conns": 10,
  "database_timeout": 30,
  "db_connection_retries": 15,
  "disable_admin_token": false,
  "domain": "*****://**********************",
  "domain_origin": "*****://**********************",
  "domain_path": "",
  "domain_set": true,
  "duo_host": null,
  "duo_ikey": null,
  "duo_skey": null,
  "email_attempts_limit": 3,
  "email_expiration_time": 600,
  "email_token_size": 6,
  "enable_db_wal": true,
  "events_days_retain": 60,
  "extended_logging": true,
  "helo_name": null,
  "hibp_api_key": null,
  "icon_blacklist_non_global_ips": false,
  "icon_blacklist_regex": null,
  "icon_cache_negttl": 259200,
  "icon_cache_ttl": 2592000,
  "icon_download_timeout": 10,
  "icon_redirect_code": 302,
  "icon_service": "internal",
  "invitation_expiration_hours": 336,
  "invitation_org_name": "Bitwarden",
  "invitations_allowed": true,
  "ip_header": "X-Real-IP",
  "log_level": "warn",
  "org_attachment_limit": null,
  "org_creation_users": "***********************",
  "org_events_enabled": true,
  "org_groups_enabled": true,
  "push_enabled": false,
  "reload_templates": false,
  "sendmail_command": null,
  "sends_allowed": false,
  "sends_folder": "data/sends",
  "show_password_hint": false,
  "signups_allowed": false,
  "signups_verify": true,
  "smtp_auth_mechanism": null,
  "smtp_debug": false,
  "smtp_embed_images": true,
  "smtp_explicit_tls": null,
  "smtp_from": "**********************",
  "smtp_from_name": "Bitwarden",
  "smtp_host": "*************",
  "smtp_password": null,
  "smtp_port": 25,
  "smtp_security": "off",
  "smtp_ssl": null,
  "smtp_timeout": 15,
  "smtp_username": null,
  "tmp_folder": "data/tmp",
  "trash_auto_delete_days": 14,
  "use_sendmail": false,
  "use_syslog": false,
  "web_vault_enabled": true,
  "websocket_enabled": false,
  "websocket_port": 3012,
  "yubico_client_id": null,
  "yubico_secret_key": null,
  "yubico_server": null
}
```
</details>

### Steps to reproduce

Case 1: A user without an account in this Vaultwarden instance is added to an organization via bwdc.
The user is invited and External ID is set successfully.

Case 2: A user with an existing account in this Vaultwarden instance is added to an organization via bwdc.
The user is invited and added to the organization, but no External ID is set.

Case 3: A user who is already part of my organization is to be subsequently managed by bwdc and given an External ID. After synchronization the user seems to be untouched. No External ID has been set.

### Expected behaviour
In every case an External ID should be assigned to the user. 
Even if the user already exists, an external ID should be assigned, otherwise the permission assignment between Vaultwarden and AD groups will not work.

If this is an issue in bwdc, is there a way to manually assign the External ID?

BlackDex:
While working on the fix for #3777 I realized the location where the `external_id`
is stored was wrong. It was stored in the `users` table, but it actually
should have been stored in the `users_organizations` table.

This will move the column to the right table. It will not move the
values of the `external_id` column, because if there are more
organizations, there is no way to really know which organization it is
linked to. Setups using the Directory Connector can clear the sync
cache, and sync again, that will store all the `external_id` values at
the right location.

Also changed the function to revoke,restore an org-user and set_external_id to return a boolean.
It will state if the value has been changed or not, and if not, we can
prevent a `save` call to the database.

The `users` table is not changed to remove the `external_id` column, thi
to prevent issue when users want to revert back to an earlier version
for some reason. We can do this after a few minor release i think.

Fixes #3777

### DIFF
--- a/src/api/core/public.rs
+++ b/src/api/core/public.rs
@@ -56,16 +56,34 @@ async fn ldap_import(data: JsonUpcase<OrgImportData>, token: PublicToken, mut co
             if let Some(mut user_org) =
                 UserOrganization::find_by_email_and_org(&user_data.Email, &org_id, &mut conn).await
             {
-                user_org.revoke();
-                user_org.save(&mut conn).await?;
-            }
+                // Only revoke a user if it is not the last confirmed owner
+                let revoked = if user_org.atype == UserOrgType::Owner
+                    && user_org.status == UserOrgStatus::Confirmed as i32
+                {
+                    if UserOrganization::count_confirmed_by_org_and_type(&org_id, UserOrgType::Owner, &mut conn).await
+                        <= 1
+                    {
+                        warn!("Can't revoke the last owner");
+                        false
+                    } else {
+                        user_org.revoke()
+                    }
+                } else {
+                    user_org.revoke()
+                };
 
+                let ext_modified = user_org.set_external_id(Some(user_data.ExternalId.clone()));
+                if revoked || ext_modified {
+                    user_org.save(&mut conn).await?;
+                }
+            }
         // If user is part of the organization, restore it
         } else if let Some(mut user_org) =
             UserOrganization::find_by_email_and_org(&user_data.Email, &org_id, &mut conn).await
         {
-            if user_org.status < UserOrgStatus::Revoked as i32 {
-                user_org.restore();
+            let restored = user_org.restore();
+            let ext_modified = user_org.set_external_id(Some(user_data.ExternalId.clone()));
+            if restored || ext_modified {
                 user_org.save(&mut conn).await?;
             }
         } else {
@@ -73,9 +91,8 @@ async fn ldap_import(data: JsonUpcase<OrgImportData>, token: PublicToken, mut co
             let user = match User::find_by_mail(&user_data.Email, &mut conn).await {
                 Some(user) => user, // exists in vaultwarden
                 None => {
-                    // doesn't exist in vaultwarden
+                    // User does not exist yet
                     let mut new_user = User::new(user_data.Email.clone());
-                    new_user.set_external_id(Some(user_data.ExternalId.clone()));
                     new_user.save(&mut conn).await?;
 
                     if !CONFIG.mail_enabled() {
@@ -92,6 +109,7 @@ async fn ldap_import(data: JsonUpcase<OrgImportData>, token: PublicToken, mut co
             };
 
             let mut new_org_user = UserOrganization::new(user.uuid.clone(), org_id.clone());
+            new_org_user.set_external_id(Some(user_data.ExternalId.clone()));
             new_org_user.access_all = false;
             new_org_user.atype = UserOrgType::User as i32;
             new_org_user.status = user_org_status;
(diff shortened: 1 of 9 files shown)


src/api/core/public.rs
```diff
@@ -68,12 +70,23 @@ async fn ldap_import(data: JsonUpcase<OrgImportData>, token: PublicToken, mut co
                 user_org.restore();
                 user_org.save(&mut conn).await?;
             }
+
+            // Set external_id if the user is restored.
+            if let Some(mut user) = user {
+                user.set_external_id(Some(user_data.ExternalId.clone()));
+                user.save(&mut conn).await?;
```
dani-garcia, line 77: I imagine most users are going to keep the same external_id if they already have it, would it be worth it to only save the user when the external_id is different to save some db operations?

BlackDex: Well, i did some looking just quick, and we store it in the wrong place it looks like.
This actually needs to be stored in the user_org, which i already had a feeling of being better but it is also how Bitwarden does it.

It is of course possible to use the Directory Connector for multiple Org entities within the same company and a user could be part of multiple, this way it would overwrite all items no matter which org/directory-connector does the sync.

Maybe we need to change this. But the problem here will be that we can't really migrate the external_id to the correct org, since we don't know which org did the sync, and which is correct.

While, technically it shouldn't matter, since it is only for display purpose as far as i know at the moment.
Nothing yet which uses this value in the code.

BlackDex:
Converted to a draft. I have the code fixed, but i need to do the SQL migrations and testing before this can be merged.

I'm moving the external_id into the user_org table.

src/api/core/public.rs
BlackDex: @dani-garcia, i have modified the PR so it will move the `external_id` column to the `users_organizations` table instead of the `users` table. This fixes issues with multiple different orgs synced to the same instance.

I also made it so that it will only save the record when needed.

BlackDex:
The only thing i encountered is, that you can't revert back that easily after this PR has been merged to a previous version.
The reason is we have removed a column from the users table which causes an issue with older versions.

This makes me wonder what we should do here?
Leave the column there for a while, and create some tech-debt ticket/todo to remove it in the future say, x new minor release, or x date or something?

@dani-garcia what do you think?

stefan0xC:
Since we in general don't support downgrading I think we could remove the column the next release after the release with this PR (so if there should be an issue they can roll back to the latest working release). Provided that the next release fixes the `armv6` issue which currently compels users to downgrade back to `1.29.1`. Otherwise we should probably delay removing the column for another release (or warn affected users to use the `-alpine` images images instead because downgrading to a previous version - especially one without this change - might require a manual restore of the column).

BlackDex:
I agree, lets remove the deletion of the column for now.

dani-garcia (approved):

PR #3806, opened 2023-08-27
Allow Authorization header for Web Sockets

### LINKED ISSUE #3776: WebSocket notifications require `access_token` query parameter
### Subject of the issue
Looking at the updated WebSocket notifications implementation, we explicitly check for the `access_token` parameter here: https://github.com/BlackDex/vaultwarden/blob/3dbfc484a54c41d1759646444b439da06445060b/src/api/notifications.rs#L86

While this is correct most of the time, the token can actually be passed with Authorization Bearer token!
ts: https://github.com/dotnet/aspnetcore/blob/721cd72b627bb43f4f9d7988662fbc5f26c46167/src/SignalR/clients/ts/signalr/src/WebSocketTransport.ts#L55
java: https://github.com/dotnet/aspnetcore/blob/721cd72b627bb43f4f9d7988662fbc5f26c46167/src/SignalR/clients/java/signalr/core/src/main/java/com/microsoft/signalr/WebSocketTransport.java#L48

### Expected behaviour
Access token is derived from either query params or bearer header.

### Actual behaviour
Access token is derived from only query params.

BlackDex:
Some clients (Thirdparty) might use the `Authorization` header instead of a query param. We didn't supported this since all the official clients do not seem to use this way of working. But Bitwarden does check both ways.

This PR adds an extra check for this header which can be optional.

Fixes #3776

### DIFF
--- a/src/api/notifications.rs
+++ b/src/api/notifications.rs
@@ -20,7 +20,7 @@ use tokio_tungstenite::{
 };
 
 use crate::{
-    auth::ClientIp,
+    auth::{ClientIp, WsAccessTokenHeader},
     db::{
         models::{Cipher, Folder, Send as DbSend, User},
         DbConn,
@@ -111,11 +111,19 @@ fn websockets_hub<'r>(
     ws: rocket_ws::WebSocket,
     data: WsAccessToken,
     ip: ClientIp,
+    header_token: WsAccessTokenHeader,
 ) -> Result<rocket_ws::Stream!['r], Error> {
     let addr = ip.ip;
     info!("Accepting Rocket WS connection from {addr}");
 
-    let Some(token) = data.access_token else { err_code!("Invalid claim", 401) };
+    let token = if let Some(token) = data.access_token {
+        token
+    } else if let Some(token) = header_token.access_token {
+        token
+    } else {
+        err_code!("Invalid claim", 401)
+    };
+
     let Ok(claims) = crate::auth::decode_login(&token) else { err_code!("Invalid token", 401) };
 
     let (mut rx, guard) = {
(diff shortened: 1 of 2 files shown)


src/api/notifications.rs
```diff
@@ -111,11 +111,19 @@ fn websockets_hub<'r>(
...
+    header_token: WsAccessTokenHeader,
 ) -> Result<rocket_ws::Stream!['r], Error> {
     let addr = ip.ip;
     info!("Accepting Rocket WS connection from {addr}");
 
-    let Some(token) = data.access_token else { err_code!("Invalid claim", 401) };
+    let token = if let Some(token) = data.access_token {
+        token
+    } else if let Some(token) = header_token.access_token {
+        token
```
RealOrangeOne, line 122: Question: Should this comparison be constant-time? It looks like all the data is from the client, so "no", but worth confirming.

BlackDex: I don't think constant-time would matter here, as you mentioned, the data is already sent by the client.

But I'm happy to convert it. But I'm not seeing any issues in leaving it as is. It's only retrieving the token, not verifying if it is valid.

But if you think it would be better please educate me why, and I'll convert it to constant-time way.

sorcix: These `if let` constructs don't actually compare.

The verificarion of the jwt token is using crypto functions from the `ring` crate [via the `jsonwebtoken` crate](https://docs.rs/jsonwebtoken/latest/src/jsonwebtoken/crypto/mod.rs.html#92), called from [the `auth` module in this crate](https://github.com/dani-garcia/vaultwarden/blob/main/src/auth.rs#L70).

No constant time comparison is needed in Vaultwarden's code.

dani-garcia (approved):

PR #3808, opened 2023-08-28
Update Rust and Crates

BlackDex:
- Updated Rust to v1.72.0
- Updated all the crates Including a CVE https://github.com/dani-garcia/vaultwarden/security/dependabot/21
- Updated GitHub Workflows
- Run `cargo fmt` which has some new fmt's
- Moved from `rust-toolchain` to `rust-toolchain.toml`


tessus:
@BlackDex I remember you mentioned that the following crates are not updated due to breaking changes: webauthn-rs, cookie, cookie_store. Wouldn't it still makes sense to update them and adjust the vw code accordingly? Or would the changes to the code be that complex?

BlackDex:
> @BlackDex I remember you mentioned that the following crates are not updated due to breaking changes: webauthn-rs, cookie, cookie_store. Wouldn't it still makes sense to update them and adjust the vw code accordingly? Or would the changes to the code be that complex?

The cookie crates are tied to Reqwest, so that has to be updated first by that crate before we can. Same btw for thungstintie, that needs to be update by Rocket first.

For webauthn we need to create a migration like we did before, we even need to be sure we can migrate both versions to the new way.

So, all of those are currently not an option to be updated right now.

But, thanks for mentioning.

dani-garcia (approved):

PR #3831, opened 2023-09-01
Fix Login With Device without MasterPassword

### LINKED ISSUE #3819: Login with device not working
### Subject of the issue
I have some questions regarding logging in with another device

* Vaultwarden version: v1.29.2
* Web-vault version: v2023.7.1
* OS/Arch: linux/x86_64
* Uses a reverse proxy: true
* Internet access: true
* DNS Check: true
* Server/NTP Time Check: true
* HTTPS Check: true
* Database type: MySQL
* Database version: 10.11.5-MariaDB-1:10.11.5+maria~ubu2204
* Clients used: 
* Reverse proxy and version: 
* Other relevant information: 

```json
{
  "_duo_akey": null,
  "_enable_duo": false,
  "_enable_email_2fa": false,
  "_enable_smtp": true,
  "_enable_yubico": true,
  "_icon_service_csp": "",
  "_icon_service_url": "",
  "_ip_header_enabled": true,
  "_smtp_img_src": "cid:",
  "admin_session_lifetime": 20,
  "admin_token": "***",
  "data_folder": "data",
  "database_conn_init": "",
  "database_max_conns": 10,
  "database_timeout": 30,
  "database_url": "*****://*******************************************",
  "db_connection_retries": 15,
  "disable_admin_token": false,
  "domain_path": "",
  "domain_set": true,
  "duo_host": null,
  "duo_ikey": null,
  "duo_skey": null,
  "email_attempts_limit": 3,
  "email_expiration_time": 600,
  "email_token_size": 6,
  "enable_db_wal": true,
  "events_days_retain": null,
  "extended_logging": true,
  "helo_name": null,
  "hibp_api_key": "***",
  "icon_blacklist_regex": null,
  "icon_cache_negttl": 259200,
  "icon_cache_ttl": 2592000,
  "icon_download_timeout": 10,
  "icon_redirect_code": 302,
  "icon_service": "internal",
  "invitations_allowed": true,
  "ip_header": "X-Real-IP",
  "log_file": null,
  "log_level": "Info",
  "org_attachment_limit": null,
  "org_creation_users": "",
  "org_events_enabled": false,
  "org_groups_enabled": false,
  "push_enabled": false,
  "reload_templates": false,
  "sendmail_command": null,
  "sends_allowed": true,
  "sends_folder": "data/sends",
  "show_password_hint": false,
  "signups_allowed": false,
  "signups_verify": false,
  "smtp_accept_invalid_certs": true,
  "smtp_accept_invalid_hostnames": true,
  "smtp_auth_mechanism": null,
  "smtp_debug": false,
  "smtp_embed_images": true,
  "smtp_explicit_tls": null,
  "smtp_from": "******************",
  "smtp_from_name": "BitWarden",
  "smtp_host": "**********",
  "smtp_password": "***",
  "smtp_port": 587,
  "smtp_security": "starttls",
  "smtp_ssl": null,
  "smtp_timeout": 15,
  "smtp_username": "***************",
  "tmp_folder": "data/tmp",
  "use_sendmail": false,
  "use_syslog": false,
  "web_vault_enabled": true,
  "websocket_enabled": false,
  "websocket_port": 3012,
  "yubico_client_id": null,
  "yubico_secret_key": null,
  "yubico_server": null
}
```
</details>

### Steps to reproduce
Opening the login page of my vaultwarden instance and typing in my mail address

### Expected behaviour
Clicking on login with device, accepting the request and successfully logging in

### Actual behaviour
I can only initiate logging in with another device out of a non-incognito Chrome tab. Neither an incognito-tab on Chrome nor any other browser (incognito or not) show me this option.
When I manage to send out a notification I receive this notification on my iPhone or on the Desktop App. I can accept the request but I am not logged in after accepting it.

BlackDex:
It looks like either something changed in the latest v2023.8.x versions, or when using Biometrics to login, but the MasterPasswordHash is an optional value during the Approve action.

This PR makes the MasterPasswordHash an optional value which resolves this issues. Bitwarden works the same way.

I also changed the EncKey to an Option in the database since empty strings as a default value is not nice in databases, better to use `null` in these cases.

Fixes #3819


migrations/mysql/2023-09-01-170620_update_auth_request_table/down.sql
stefan0xC, line 1: Might be OT but I have a question about migrations: When do we provide something in a `down.sql` file? I noticed that we rarely provide a way to undo a change (and in most cases only if we CREATE a new table, we add a corresponding statement that will DROP the table again).

The `auth_requests` table is still missing that DROP statement. Do we care about this? I think it might be useful if we are testing changes to the database to be able to easily undo them again so our database would not be in a corrupt state. (Conversely, I think DROP might be something we won't really need because it's easy to come up with the command ourselves, but I think we might just underutilize this feature?)

stefan0xC (approved):
LGTM. (And as far as I've tested it works without isues.) :+1:

migrations/mysql/2023-09-01-170620_update_auth_request_table/down.sql
BlackDex: > The `auth_requests` table is still missing that DROP statement. Do we care about this? I think it might be useful if we are testing changes to the database to be able to easily undo them again so our database would not be in a corrupt state. (Conversely, I think DROP might be something we won't really need because it's easy to come up with the command ourselves, but I think we might just underutilize this feature?)

I'm not sure that I understand what you mean?
All migrations ensure that both columns remove the `NOT NULL` on all databases.

Also, if I'm correct, the migrations are done within transactions, and that should reverse the queries, but not sure how that works with create/delete of tables actually.

BlackDex: Ah, you mean if the `DROP` / `MODIFY` statement fails, we should revert that. But i think that will be done by the transactions in those cases. So, that shouldn't be a problem. I, never encountered that a migration failed, and it partially did it anyway i think.

dani-garcia (approved):

migrations/mysql/2023-09-01-170620_update_auth_request_table/down.sql
stefan0xC: I mean that if I make a PR that changes the database drastically and you check it out run the migrations, without providing the correct `down.sql` you would have to delete the `data/db.sqlite3` to have a working database again or restore from a backup to undo the changes to your database. It's not something that should affect production in practice (unless you downgrade which we don't recommend).

stefan0xC: From the [`diesel_migrations` docs](https://docs.rs/diesel_migrations/latest/diesel_migrations/):

> Individual migrations should be a folder containing two files, `up.sql` and `down.sql`. `up.sql` will be used to run the migration, while `down.sql` will be used for reverting it.

So `diesel_cli` commands like `diesel migration revert` and `diesel migration redo` will not work without actually specifying how to revert a change. Cf. this [StackOverflow answer](https://stackoverflow.com/a/65859325/6058571) how this could be useful.

BlackDex: That is only useful using the cli.
I think, but I'm not sure, that the migrations done through the API are enclosed within a transaction. That will only fully commit if the queries worked, and thus prevent breaking.

We do not provide a way downgrade or something by the users.

BlackDex: See https://docs.rs/diesel/2.1.1/diesel/migration/trait.MigrationMetadata.html#method.run_in_transaction

stefan0xC: I am aware. I think you are missing my point which is that it might be useful for **development** to provide a way to undo and redo migrations. But I also think that it's not strictly necessary and it would be just a quality of life thing. In practice I would most likely not remember to use the `diesel_cli` and just start over with a new database if I notice that my database is not in a valid state.

BlackDex: Ah, sorry for the confusion, i suffer from some sleep deprivation hehe.

For dev it would be useful. But, I'm not sure it will fully work for sqlite, that will probably lose the data, since we do not know where the migrations failed. So, manual intervention is probably always needed there. While it had no issues i think with the mysql and postgresql queries.

PR #3909, opened 2023-09-22
Reopen log file on SIGHUP

tobiasmboelz:
This change makes Vaultwarden reopen the log file on SIGHUP, enabling proper log rotation without logrotate's copytruncate hack.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,6 @@ unstable = []
 [target."cfg(not(windows))".dependencies]
 # Logging
 syslog = "6.1.0"
-reopen = "1.0.3"
 
 [dependencies]
 # Logging
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,11 +37,12 @@ unstable = []
 [target."cfg(not(windows))".dependencies]
 # Logging
 syslog = "6.1.0"
+reopen = "1.0.3"
 
 [dependencies]
 # Logging
 log = "0.4.20"
-fern = { version = "0.6.2", features = ["syslog-6"] }
+fern = { version = "0.6.2", features = ["syslog-6", "reopen-1"] }
 tracing = { version = "0.1.37", features = ["log"] } # Needed to have lettre and webauthn-rs trace logging to work
 
 # A `dotenv` implementation for Rust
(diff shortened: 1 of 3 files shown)


williamdes (approved):
Thank you, I learned something

More reading at https://github.com/tinyproxy/tinyproxy/issues/25#issuecomment-268849637
and Wikipedia

src/main.rs
```diff
@@ -317,7 +317,16 @@ fn init_logging(level: log::LevelFilter) -> Result<(), fern::InitError> {
...
 
     if let Some(log_file) = CONFIG.log_file() {
-        logger = logger.chain(fern::log_file(log_file)?);
+        #[cfg(windows)]
+        {
+            logger = logger.chain(fern::log_file(log_file)?);
+        }
+        #[cfg(not(windows))]
+        {
+            const SIGHUP: i32 = 1;
```
BlackDex, line 326: It might be a bit long, but maybe removing the custom `const SIGHUP` with `tokio::signal::unix::SignalKind::hangup().as_raw_value()` would be an option?
Not that the signal value will ever change, but still.

So something like:
```suggestion
            const SIGHUP: i32 = tokio::signal::unix::SignalKind::hangup().as_raw_value();
```

Cargo.toml
```diff
@@ -37,11 +37,12 @@ unstable = []
 [target."cfg(not(windows))".dependencies]
 # Logging
 syslog = "6.1.0"
+reopen = "1.0.3"
```
BlackDex, line 40: I don't think this is needed, it will be included by fern it self.

BlackDex (changes requested):
I have not yet tested this my self, but i think this should work.
Thanks for the contribution.

src/main.rs
tobiasmboelz: That's some wrapper around `libc::SIGHUP`, so IMO it would be cleaner to use `libc::SIGHUP` directly. 

I just didn't use `libc::SIGHUP` or `signal_hook::consts::signal::SIGHUP` (which is mentioned in [the reopen documentation](https://docs.rs/reopen/latest/reopen/)) because I didn't want to add additional dependencies and imports for a single value.

BlackDex: But that would mean an extra dependency in our Cargo.toml, this doesn't.

tobiasmboelz: I don't think it's pretty, but fine by me.

BlackDex (approved):
Works great, thanks!

dani-garcia (approved):

PR #3951, opened 2023-10-03
Adds LastActive on /admin/users API route

mvalois:


src/api/admin.rs
```diff
@@ -326,6 +326,10 @@ async fn get_users_json(_token: AdminToken, mut conn: DbConn) -> Json<Value> {
         let mut usr = u.to_json(&mut conn).await;
         usr["UserEnabled"] = json!(u.enabled);
         usr["CreatedAt"] = json!(format_naive_datetime_local(&u.created_at, DT_FMT));
+        usr["LastActive"] = match u.last_active(&mut conn).await {
+            Some(dt) => json!(format_naive_datetime_local(&dt, DT_FMT)),
+            None => json!("Never"),
```
stefan0xC, line 331: Wouldn't it be better to either not return a `LastActive` value or set it to `null`? (So consumers of the API don't have to check for a `Never` and can assume that the format of the value if it exists is correct.)

mvalois: I just took the code from https://github.com/dani-garcia/vaultwarden/blob/main/src/api/admin.rs#L346. Do you think consumers are not the same ?

stefan0xC: The `/admin/users/overview` does not use that endpoint but uses the function (that you have linked) to immediately generate an HTML page using the `src/static/templates/admin/users.hbs` template. I'm not aware of any consumers of the `/admin/users` endpoint myself, but I think it would be better simply from an API design standpoint to leave the display of a missing value to other consumers of this endpoint as they might not just want to display the string as we do.

mvalois: I've set it to 'null' then.

dani-garcia (approved):

PR #3959, opened 2023-10-05
Fix typos

tuhanayim:
Fixes typos in the code base, mostly comments.

There's also a typo at https://github.com/dani-garcia/vaultwarden/blob/bc26bfa589c007da9b9be37e1172060f38a948b9/src/api/notifications.rs#L568 but I'm not sure if it is intentional or not, so I left it unfixed for now.

### DIFF
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -1520,9 +1520,9 @@ async fn bulk_public_keys(
     let data: OrgBulkIds = data.into_inner().data;
 
     let mut bulk_response = Vec::new();
-    // Check all received UserOrg UUID's and find the matching User to retreive the public-key.
+    // Check all received UserOrg UUID's and find the matching User to retrieve the public-key.
     // If the user does not exists, just ignore it, and do not return any information regarding that UserOrg UUID.
-    // The web-vault will then ignore that user for the folowing steps.
+    // The web-vault will then ignore that user for the following steps.
     for user_org_id in data.Ids {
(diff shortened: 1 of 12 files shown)


stefan0xC:
> not sure if it is intentional

It's also upstream (in [bitwarden/clients](https://github.com/bitwarden/clients/blob/7190bad1b0e81c6178b468dcdd8b6e7533ffe7ee/libs/common/src/services/anonymousHub.service.ts#L42) and [bitwarden/server](https://github.com/bitwarden/server/blob/4dcd467a5fb8f99e9644da12dcafa36b14d870d2/src/Notifications/HubHelpers.cs#L76)). So intentional by us, probably not intentional by them. :wink:

src/api/core/organizations.rs
```diff
@@ -2897,7 +2897,7 @@ async fn put_reset_password_enrollment(
 
 // This is a new function active since the v2022.9.x clients.
 // It combines the previous two calls done before.
-// We call those two functions here and combine them our selfs.
+// We call those two functions here and combine them our selves.
```
BlackDex, line 2900: Shouldn't this be `ourselves`?

BlackDex (changes requested):

src/api/core/organizations.rs
tuhanayim: Correct, my bad.

BlackDex (approved):

dani-garcia (approved):

PR #3986, opened 2023-10-20
New config option disable email change

admav:
Added new configuration option to prevent users from changing their email.
By default is set to allow it.


BlackDex (approved):
Needs a rebase now, but looks fine be me.

dani-garcia (approved):

PR #3989, opened 2023-10-21
update web-vault to v2023.9.1

reysonk:


BlackDex:
Thanks, but this probably is going to be outdated when #3958 gets merged.

tessus:
can web-vault 2023.9.1 be used with current master? Or better yet, what is the earliest commit point that supports 2023.9.1?

reysonk:
> can web-vault 2023.9.1 be used with current master? Or better yet, what is the earliest commit point that supports 2023.9.1?

Now master bes

dani-garcia:
The web vault has been updated to 2023.9.1 with the rest of the container changes, so this won't be needed anymore, thanks for the help!

I don't think 2023.9.1 had any special server requirements, and should work with any recent vaultwarden commits. 2023.9.2 and newer versions of the web vault on the other hand will require the changes I merged a couple of hours ago #3990

PR #3990, opened 2023-10-21
Implement cipher key encryption

dani-garcia:
New versions of the bitwarden clients will very soon start using separate encryption keys for each cipher, so we need to handle those.

At the same time, I've noticed that the clients have started depending on the version value of the `/config` endpoint to do backwards compatibility checks, so I think the simplest solution for everyone would be for us to return the server version that most closely matches the features we support.

They are doing the semver checks here: 
https://github.com/bitwarden/clients/blob/3e495ab082e2152284a36bb5d2b2fb7a4060cfb3/libs/common/src/platform/services/config/config.service.ts#L116-L127

For now the only check they are doing is for version 2023.9.1 for the cipher key encryption feature that this PR implements, so we report that we support that version: 
https://github.com/bitwarden/clients/blob/3e495ab082e2152284a36bb5d2b2fb7a4060cfb3/libs/common/src/vault/services/cipher.service.ts#L56

Also added the section for feature flags in the config, the current ones reported on the official vault are: 
```
"display-eu-environment": true,
"display-kdf-iteration-warning": true,
"trusted-device-encryption": true,
"sm-ga-billing": true,
"autofill-v2": true
```

For now as a test I've enabled autofill-v2 as that is an entirely client side change, so it won't affect us

### DIFF
--- a/src/api/core/mod.rs
+++ b/src/api/core/mod.rs
@@ -194,7 +194,12 @@ fn version() -> Json<&'static str> {
 fn config() -> Json<Value> {
     let domain = crate::CONFIG.domain();
     Json(json!({
-        "version": crate::VERSION,
+        // Note: The clients use this version to handle backwards compatibility concerns
+        // This means they expect a version that closely matches the Bitwarden server version
+        // We should make sure that we keep this updated when we support the new server features
+        // Version history:
+        // - Individual cipher key encryption: 2023.9.1
+        "version": "2023.9.1",
         "gitHash": option_env!("GIT_REV"),
         "server": {
           "name": "Vaultwarden",
@@ -206,6 +211,12 @@ fn config() -> Json<Value> {
           "identity": format!("{domain}/identity"),
           "notifications": format!("{domain}/notifications"),
           "sso": "",
+        },
+        "featureStates": {
+          // Any feature flags that we want the clients to use
+          // Can check the enabled ones at:
+          // https://vault.bitwarden.com/api/config
+          "autofill-v2": true
         },
         "object": "config",
     }))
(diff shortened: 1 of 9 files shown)


src/api/core/mod.rs
```diff
@@ -194,7 +194,12 @@ fn version() -> Json<&'static str> {
 fn config() -> Json<Value> {
     let domain = crate::CONFIG.domain();
     Json(json!({
-        "version": crate::VERSION,
+        // Note: The clients use this version to handle backwards compatibility concerns
+        // This means they expect a version that closely matches the Bitwarden server version
+        // We should make sure that we keep this updated when we support the new server features
+        // Version history:
+        // - Individual cipher key encryption: 2023.9.1
+        "version": "2023.9.1",
```
BlackDex, line 202: Do we want to provide the Vaultwarden Version here also? with something like `vw_version`?
Or should we leave it like this?

For some not adding the exact version of Vaultwarden is a plus, by not revealing it, so it might be fine like this.

BlackDex (commented):
Looks ok too me. Not yet tested though.
But simple changes if this is all to be done.

Is there some background information for this change from upstream?

BlackDex:
Doesn't seem to be in the master branch yet.
Have not looked for the specific PR yet to fully test this.

src/api/core/mod.rs
dani-garcia: We are providing the git hash already, so I think with that we are already giving pretty much the same information, so I didn't add another field, but we could if we wanted.

dani-garcia:
The changes are in master but you need to enable them in the web vault flags before building the project:

https://github.com/bitwarden/clients/blob/3e495ab082e2152284a36bb5d2b2fb7a4060cfb3/apps/web/config/selfhosted.json#L12

After that, the new build will create new ciphers with individual keys, but it won't update the old ones just yet.

The clients PR is here, and it also links to the server changes: https://github.com/bitwarden/clients/pull/6241

src/api/core/mod.rs
BlackDex: Yea, the hash should be good enough.

BlackDex (approved):

PR #3994, opened 2023-10-23
Fix issue with MariaDB/MySQL migrations

### LINKED ISSUE #3993: Error running migrations Docker won't start
### Subject of the issue
On pulling the latest testing image I can't get it to start it shows the following in the logs

[23-10-2023 09:19:26.270][panic][ERROR] thread 'main' panicked at 'Error running migrations: QueryError(DieselMigrationName { name: "2023-10-21-221242_add_cipher_key", version: MigrationVersion("20231021221242") }, DatabaseError(Unknown, "You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '\"key\" TEXT' at line 2"))': src/db/mod.rs:471

   1: std::panicking::rust_panic_with_hook

   2: std::panicking::begin_panic_handler::{{closure}}

   3: std::sys_common::backtrace::__rust_end_short_backtrace

   4: rust_begin_unwind

   5: core::panicking::panic_fmt

   7: vaultwarden::db::mysql_migrations::run_migrations

   8: vaultwarden::db::DbPool::from_config

   9: vaultwarden::main::{{closure}}

  10: rocket::async_main

  11: vaultwarden::main

  12: std::sys_common::backtrace::__rust_begin_short_backtrace

  13: std::rt::lang_start::{{closure}}

  14: std::rt::lang_start_internal

  15: main

  16: <unknown>

  17: __libc_start_main

  18: _start

### Deployment environment

Synology DS918+     =========================================================================================
    Preferably, use the `Generate Support String` button on the admin page's Diagnostics tab.
    That will auto-generate most of the info requested in this section.
    =========================================================================================
-->

* vaultwarden version: Testing build as of 1027 on 23rd October 2023. Unable to start to get version.

* Install method: Docker on Synology NAS

* MySQL/MariaDB or PostgreSQL version: MariaDB latest

* Other relevant details:

### Steps to reproduce

Pulled latest testing image and issue happens. 

### Expected behaviour

Docker container starts up
### Troubleshooting data

Logs are above

BlackDex:
MariaDB/MySQL doesn't like the normal `"` quotes around the column name. This needs to be a backtick **`**.

This PR changes the migration script to fix this issue.

Fixes #3993

### DIFF
--- a/migrations/mysql/2023-10-21-221242_add_cipher_key/up.sql
+++ b/migrations/mysql/2023-10-21-221242_add_cipher_key/up.sql
@@ -1,2 +1,2 @@
-ALTER TABLE ciphers 
-ADD COLUMN "key" TEXT;
+ALTER TABLE ciphers
+ADD COLUMN `key` TEXT;


dani-garcia (approved):
That’s what I get for assuming there would be no way to screw up such a simple migration lol. Did you check that postgres runs fine with the quotes just in case?

BlackDex:
> That’s what I get for assuming there would be no way to screw up such a simple migration lol. Did you check that postgres runs fine with the quotes just in case?

Yes i did, and that breaks when using backtick's ;)

migrations/mysql/2023-10-21-221242_add_cipher_key/up.sql
```diff
@@ -1,2 +1,2 @@
-ALTER TABLE ciphers 
-ADD COLUMN "key" TEXT;
+ALTER TABLE ciphers
+ADD COLUMN `key` TEXT;
```
williamdes, line 2: Probably except if the server has ANSI_QUOTES mode enabled
https://dev.mysql.com/doc/refman/8.0/en/sql-mode.html#sqlmode_ansi_quotes

Commenting just for fun and for people to know this mode 😄

BlackDex: Ah good to post. Strange feature, as that would break a lot of other stuff using `"` haha.

PR #3997, opened 2023-10-24
ci: add trivy workflow

mightyBroccoli:
to ensure critical and high vulnerabilties are detected quickly

### DIFF
--- a/.github/workflows/trivy.yml
+++ b/.github/workflows/trivy.yml
@@ -1,0 +1,43 @@
+name: trivy
+
+on:
+  push:
+    branches:
+      - main
+      - release-build-revision
+    tags:
+      - '*'
+  pull_request:
+    branches: [ "main" ]
+  schedule:
+    - cron: '00 12 * * *'
+
+permissions:
+  contents: read
+
+jobs:
+  trivy-scan:
+    name: Check
+    runs-on: ubuntu-22.04
+    timeout-minutes: 30
+    permissions:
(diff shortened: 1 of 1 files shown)


mightyBroccoli:
@dani-garcia would you mind reviewing this pr?

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,43 @@
...
+    permissions:
+      contents: read
+      security-events: write
+      actions: read
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 #v4.1.1
+
+      - name: Run Trivy vulnerability scanner
+        uses: aquasecurity/trivy-action@fbd16365eb88e12433951383f5e99bd901fc618f # v0.12.0
```
dani-garcia, line 32: This can be updated to 0.13.1: https://github.com/aquasecurity/trivy-action/releases/tag/0.13.1

dani-garcia (commented):
We haven't used code scanning much in the past because in my experience we end up dealing with more false positives and non-actionable reports than good reports, but I think it's worth trying at least. Just gotta update the trivy action as there was a new release last week

.github/workflows/trivy.yml
mightyBroccoli: done :+1:

dani-garcia (approved):

PR #4001, opened 2023-10-25
Remove custom WebSocket code

BlackDex:
Remove our custom WebSocket code and only use the Rocket code. Removed all options in regards to WebSockets
Added a new option `WEBSOCKET_DISABLED` which defaults too `false`. This can be used to disable WebSockets if you really do not want to use it.

Also:

- Updated crates, including Rocket to the latest merged v0.5 changes
- Removed an extra header which should not be sent for websocket connections

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -84,12 +84,8 @@
 ### WebSocket ###
 #################
 
-## Enables websocket notifications
-# WEBSOCKET_ENABLED=false
-
-## Controls the WebSocket server address and port
-# WEBSOCKET_ADDRESS=0.0.0.0
-# WEBSOCKET_PORT=3012
+## Disable websocket notifications
+# WEBSOCKET_DISABLED=false
 
 ##########################
 ### Push notifications ###
--- a/.env.template
+++ b/.env.template
@@ -84,8 +84,8 @@
 ### WebSocket ###
 #################
 
-## Disable websocket notifications
-# WEBSOCKET_DISABLED=false
+## Enable websocket notifications
+# ENABLE_WEBSOCKET=true
 
 ##########################
 ### Push notifications ###
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -316,7 +316,7 @@ async fn test_smtp(data: Json<InviteData>, _token: AdminToken) -> EmptyResult {
 
 #[get("/logout")]
 fn logout(cookies: &CookieJar<'_>) -> Redirect {
-    cookies.remove(Cookie::build(COOKIE_NAME).path(admin_path()));
+    cookies.remove(Cookie::build(COOKIE_NAME, "").path(admin_path()).finish()); 
     Redirect::to(admin_path())
 }
 
@@ -797,7 +797,7 @@ impl<'r> FromRequest<'r> for AdminToken {
 
             if decode_admin(access_token).is_err() {
                 // Remove admin cookie
-                cookies.remove(Cookie::build(COOKIE_NAME).path(admin_path()));
+                cookies.remove(Cookie::build(COOKIE_NAME, "").path(admin_path()).finish());
                 error!("Invalid or expired admin JWT. IP: {}.", &ip.ip);
                 return Outcome::Error((Status::Unauthorized, "Session expired"));
             }
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -316,7 +316,7 @@ async fn test_smtp(data: Json<InviteData>, _token: AdminToken) -> EmptyResult {
 
 #[get("/logout")]
 fn logout(cookies: &CookieJar<'_>) -> Redirect {
-    cookies.remove(Cookie::build(COOKIE_NAME, "").path(admin_path()).finish()); 
+    cookies.remove(Cookie::build(COOKIE_NAME).path(admin_path()));
     Redirect::to(admin_path())
 }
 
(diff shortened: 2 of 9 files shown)


stefan0xC (changes requested):
Looks mostly good to me. Not sure I'm a fan of the new configuration option name though.

If we don't want to keep the old name (for which I guess there are reasons, i.e. some users would have to change their configuration if they have set the value to false explicitly or saved a `data/config.json` with the old default), I'd rather call it `WEBSOCKETS_ENABLED` instead. I think it would be better to have an affirmative name for booleans as they tend to be easier to reason about (i.e. a feature is enabled when it's true, not when a disable toggle is on).

If we want to communicate that this is an option to disable the feature (to discourage the use) I think we should use the active form `DISABLE_WEBSOCKETS` instead (as we do with a few other configuration options`).

.env.template
```diff
@@ -69,12 +69,8 @@
...
 # WEB_VAULT_ENABLED=true
 
-## Enables websocket notifications
-# WEBSOCKET_ENABLED=false
-
-## Controls the WebSocket server address and port
-# WEBSOCKET_ADDRESS=0.0.0.0
-# WEBSOCKET_PORT=3012
+## Disable websocket notifications
+# WEBSOCKET_DISABLED=false
```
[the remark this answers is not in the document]
BlackDex: Reuse is a bit strange, since we encouraged people to disable it in the current versions. That would mean instantly breaking websocket support.

Changing it to add a `S` would be fine by me, though I don't think it's syntactically correct.

Also, i think `DISABLE_WEBSOCKET` would be a good alternative.

src/api/notifications.rs
```diff
@@ -521,7 +552,9 @@ impl AnonymousWebSocketSubscriptions {
             UpdateType::AuthRequestResponse,
             user_uuid.to_string(),
         );
-        self.send_update(auth_response_uuid, &data).await;
+        if !CONFIG.websocket_disabled() {
+            self.send_update(auth_response_uuid, &data).await;
```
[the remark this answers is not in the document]
BlackDex: Overlooked this I think thx.

src/api/notifications.rs
```diff
@@ -339,27 +328,39 @@ impl WebSocketUsers {
 
     // NOTE: The last modified date needs to be updated before calling these methods
     pub async fn send_user_update(&self, ut: UpdateType, user: &User) {
+        // Skip any processing if both WebSockets and Push are not active
+        if CONFIG.websocket_disabled() && !CONFIG.push_enabled() {
```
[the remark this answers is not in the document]
BlackDex: Yea i thought about that too. I even was thinking about a dummy crate/mod to load that instead.

But maybe a single function which takes care of this is a better option indeed.

src/api/notifications.rs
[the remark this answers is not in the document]
BlackDex: That function does send a push notification, so it should still make a difference between them i think.

tessus:
The way this is implemented now means that this is a breaking change. If vw follows semantic versioning, the next release will have to be `2.0.0`.

Even if vw didn't follow semantic versioning, a proper BREAKING changelog entry should be added.

src/api/admin.rs
```diff
@@ -313,7 +312,7 @@ async fn test_smtp(data: Json<InviteData>, _token: AdminToken) -> EmptyResult {
 
 #[get("/logout")]
 fn logout(cookies: &CookieJar<'_>) -> Redirect {
-    cookies.remove(Cookie::build(COOKIE_NAME, "").path(admin_path()).finish());
+    cookies.remove(Cookie::build((COOKIE_NAME, "")).path(admin_path()));
```
SergioBenitez, line 315:
```suggestion
    cookies.remove(Cookie::build(COOKIE_NAME).path(admin_path()));
```

src/api/admin.rs
```diff
@@ -793,7 +792,7 @@ impl<'r> FromRequest<'r> for AdminToken {
 
             if decode_admin(access_token).is_err() {
                 // Remove admin cookie
-                cookies.remove(Cookie::build(COOKIE_NAME, "").path(admin_path()).finish());
+                cookies.remove(Cookie::build((COOKIE_NAME, "")).path(admin_path()));
```
SergioBenitez, line 795:
```suggestion
                cookies.remove(Cookie::build(COOKIE_NAME).path(admin_path()));
```

BlackDex: Thx

src/api/notifications.rs
```diff
@@ -108,7 +103,7 @@ impl Drop for WSAnonymousEntryMapGuard {
 
 #[get("/hub?<data..>")]
 fn websockets_hub<'r>(
-    ws: rocket_ws::WebSocket,
+    ws: WebSocket,
```
SergioBenitez, line 106: I think if you make this `mut`, you don't need to rebind it as `mut` inside the `Stream!`. And if that's not the case, we can probably make it so in Rocket.

Also, since I can't leave a comment on code that wasn't changed: there's a lot of rightward drift on this one. I'd probably do bind the stream instead of returning it directly to save one indentation level, something like:

```rust
let stream = rocket_ws::Stream!(ws => {
    ....
});

Ok(stream)
```

But that's just a matter of preference.

SergioBenitez (commented):
Nothing too interesting. Looks like a straightforward migration to rc4!

src/api/notifications.rs
BlackDex: Thx for your comments.
Ill see what looks better. The mut is good one. Ill test it out.

tessus:
Can this be merged? I think it makes sense to have this "break" reverse-proxy settings in testing **before** the 1.31.0 is released.

BlackDex:
We (might) need to release a 1.30.4 first.
Because of #4307, which is fixed now.

But after that (if nothing else pops up), that was the plan.

tessus:
Ping. ;-) Can a few PRs be merged? The list of PRs id getting longer and longer...

dani-garcia (approved):

PR #3593, opened 2023-06-19
feat: Support for storing passkeys in the vault

GeekCornerGH:
# Implements the ability to store passkeys in the vault

## How to test
1. Checkout [this branch](https://github.com/bitwarden/server/tree/EC-598-beeep-properly-store-passkeys-in-bitwarden), then install dependencies with `npm i` (`npm ci` didn't worked, where `npm i` somehow worked in my case)
2. Enable feature flag for the extension. As I didn't figured out how to do this properly, I changed the values [there](https://github.com/bitwarden/clients/pull/4715/files#diff-ee9271bfbbddd589f955418cf1fc8a360ea76819afe2400d96d2f8a79dc3dd6dR41-R43) and [there](https://github.com/bitwarden/clients/pull/4715/files#diff-ee9271bfbbddd589f955418cf1fc8a360ea76819afe2400d96d2f8a79dc3dd6dR194-R196) to make them always return true
3. Follow the [docs](https://contributing.bitwarden.com/getting-started/clients/browser/#build-instructions) to build and install the modified extension in your browser (I installed it in a custom profile)
4. Go to a website supporting passkeys. I used [passkeys.io](https://passkeys.io), and also tested against [webauthn.io](https://webauthn.io) but any website supporting webauthn should actually work, then register your key

## Demo
https://github.com/dani-garcia/vaultwarden/assets/45696571/4c5bbfd6-3302-4ce2-9a2f-146e5653ce1b

## Notes
Upstream PR: bitwarden/server#2679
Tested on `Brave 1.54.65 Chromium: 114.0.5735.133 (Official build) nightly (64 bits) `
The feature isn't released yet, but there shouldn't be any big changes as Vaultwarden doesn't type the properties of ciphers the way upstream server does

PR #4025, opened 2023-11-03
feat: Working passkeys storage

GeekCornerGH:
Revert changes in #3593 and make them work with Browser plugin 2023.10.1

### DIFF
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -212,8 +212,7 @@ pub struct CipherData {
     Login = 1,
     SecureNote = 2,
     Card = 3,
-    Identity = 4,
-    Fido2Key = 5
+    Identity = 4
     */
     pub Type: i32,
     pub Name: String,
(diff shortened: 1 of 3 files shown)


Justus-D:
I just wanted to confirm, that the current stable version of Vaultwarden (1.29.2) seems to be working with passkeys. I modified the response of `/api/config` to include `"fido2-vault-credentials": true` and I was able to store passkeys with the current version of the browser extension (2023.10.1).

tessus:
Interesting. I am wondering how they are stored in the db. Maybe they are reusing the password field.

Justus-D:
> Interesting. I am wondering how they are stored in the db. Maybe they are reusing the password field.

Passkeys are stored in the login cipher. So in the same field as the password and the URL of the login.

dani-garcia (approved):
Sometimes the fact that I decided to dump the cipher data from the clients directly into the column as JSON ends up being really convenient, nice haha

PR #4045, opened 2023-11-07
Update version for clients as well

k0gen:
I've noticed an old version reference in the Firefox extension.
Screenshot from 2023-11-07 18 54 26

Close if this is intentional


BlackDex:
This is intentional, see https://github.com/dani-garcia/vaultwarden/discussions/4034#discussioncomment-7487746

We don't know the clients check for 2023.9.1 or higher to enable the current features we support. It could be that future clients check for 2023.10.x and that could cause issues.

So we keep it to the lowest version number as possible for the features we support.

PR #4085, opened 2023-11-19
Improve JWT RSA key initialization and avoid saving public key

dani-garcia:
Based on the discussion on #4083

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -2,9 +2,10 @@
 //
 use chrono::{Duration, Utc};
 use num_traits::FromPrimitive;
-use once_cell::sync::Lazy;
+use once_cell::sync::{Lazy, OnceCell};
 
 use jsonwebtoken::{self, errors::ErrorKind, Algorithm, DecodingKey, EncodingKey, Header};
+use openssl::rsa::Rsa;
 use serde::de::DeserializeOwned;
 use serde::ser::Serialize;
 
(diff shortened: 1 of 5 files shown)


williamdes (approved):

tessus:
Nice change.

BlackDex:
> Nice change.
> 
> P.S.: Just to clarify, this means I can delete the `rsa_key.pub.pem` file from my vw data directory, correct?

Yes it does.

BlackDex:
@dani-garcia It looks ok too me.

The only thing i encountered is that if i generate a RSA key manually like `openssl genrsa -out rsa_key.pem 3072` that works, but if i go heigher, like `5120`, it will load, but it will also fail generating a JWT since the key size is too large.

Maybe we want to put a limit on the size or at least check the size? Or maybe even allow a different key type? ECDSA for example? That is smaller, though not sure if that the generated JWT's are also smaller though.

tessus:
In theory that makes sense, but I see a case where you'd run into an issue.

You start vw and the key is generated and used. You invite someone or add someone to an org. Then you restart the vw server. What now? The tokens are no longer valid, because they were created with a key that no longer exists. Neither on disk, nor in memory.

Or did I miss something?

tessus:
May I ask what the status of this PR is? I am asking because maybe I can help with something.

dani-garcia:
We can merge this as is and leave the discussion of different ciphers for another PR. 

I think migrating to ed25519 would make the most sense, as it would also cut down on the generated JWTs signature sizes (a quick test with the admin page JWT reduced it's size from 499 bytes with RSA2048 to 243 bytes with ED25519). The clients aren't checking the JWT signature so this change shouldn't cause any compatibility issues.

We might need to consider keeping the RSA decoding at least for one release to still be able to handle older invites etc, but in my opinion this shouldn't need to be configurable as long as we use a secure algorithm and key size. If any admins want to use a different key size, well they are on their own for that.

BlackDex:
> We can merge this as is and leave the discussion of different ciphers for another PR.
> 
> I think migrating to ed25519 would make the most sense, as it would also cut down on the generated JWTs signature sizes (a quick test with the admin page JWT reduced it's size from 499 bytes with RSA2048 to 243 bytes with ED25519). The clients aren't checking the JWT signature so this change shouldn't cause any compatibility issues.
> 
> We might need to consider keeping the RSA decoding at least for one release to still be able to handle older invites etc, but in my opinion this shouldn't need to be configurable as long as we use a secure algorithm and key size. If any admins want to use a different key size, well they are on their own for that.

I agree. Merging as is seems good to me. The rest would be extra for later.

BlackDex (approved):

PR #4168, opened 2023-12-14
Allow customizing the featureStates

PKizzle:
Instead of statically disabling autofill-v2 and enabling fido2-vault-credentials let the user decide which features to en-/disable.

By default the above mentioned standard values are used but can be changed by specifying a comma separated list of any of the currently available in the bitwarden clients: https://github.com/bitwarden/clients/blob/00fd45a678956c76dde6caa76e8ba266c204471c/libs/common/src/enums/feature-flag.enum.ts

~To disable a feature just prepend an exclamation mark (!). The default value is: `!autofill-v2,fido2-vault-credentials`~

### DIFF
--- a/src/api/core/mod.rs
+++ b/src/api/core/mod.rs
@@ -213,7 +213,7 @@ fn config() -> Json<Value> {
           "notifications": format!("{domain}/notifications"),
           "sso": "",
         },
-        "feature_states": feature_states,
+        "featureStates": feature_states,
         "object": "config",
     }))
 }
--- a/src/api/core/mod.rs
+++ b/src/api/core/mod.rs
@@ -54,7 +54,7 @@ use crate::{
     auth::Headers,
     db::DbConn,
     error::Error,
-    util::{get_reqwest_client, parse_feature_flags},
+    util::{get_reqwest_client, parse_experimental_client_feature_flags},
 };
 
 #[derive(Serialize, Deserialize, Debug)]
@@ -192,8 +192,8 @@ fn version() -> Json<&'static str> {
 #[get("/config")]
 fn config() -> Json<Value> {
     let domain = crate::CONFIG.domain();
-    let feature_states = parse_feature_flags(&crate::CONFIG.feature_flags());
-    let mut config = json!({
+    let feature_states = parse_experimental_client_feature_flags(&crate::CONFIG.experimental_client_feature_flags());
+    Json(json!({
         // Note: The clients use this version to handle backwards compatibility concerns
         // This means they expect a version that closely matches the Bitwarden server version
         // We should make sure that we keep this updated when we support the new server features
@@ -213,10 +213,9 @@ fn config() -> Json<Value> {
           "notifications": format!("{domain}/notifications"),
           "sso": "",
         },
+        "feature_states": feature_states,
         "object": "config",
-    });
-    config["featureStates"] = serde_json::to_value(feature_states).unwrap();
-    Json(config)
+    }))
 }
 
 pub fn catchers() -> Vec<Catcher> {
--- a/src/api/core/mod.rs
+++ b/src/api/core/mod.rs
@@ -47,15 +47,14 @@ pub fn events_routes() -> Vec<Route> {
 //
 // Move this somewhere else
 //
-use rocket::{serde::json::Json, Catcher, Route};
-use serde_json::Value;
+use rocket::{serde::json::Json, serde::json::Value, Catcher, Route};
 
 use crate::{
     api::{JsonResult, JsonUpcase, Notify, UpdateType},
     auth::Headers,
     db::DbConn,
     error::Error,
-    util::get_reqwest_client,
+    util::{get_reqwest_client, parse_feature_flags},
 };
 
 #[derive(Serialize, Deserialize, Debug)]
@@ -193,9 +192,7 @@ fn version() -> Json<&'static str> {
 #[get("/config")]
 fn config() -> Json<Value> {
     let domain = crate::CONFIG.domain();
-    let feature_flags = crate::CONFIG.feature_flags().to_lowercase();
-    let features = feature_flags.split(',').map(|f| f.trim()).collect::<Vec<_>>();
-    let mut feature_states = json!({});
+    let feature_states = parse_feature_flags(&crate::CONFIG.feature_flags());
     let mut config = json!({
         // Note: The clients use this version to handle backwards compatibility concerns
         // This means they expect a version that closely matches the Bitwarden server version
@@ -218,12 +215,7 @@ fn config() -> Json<Value> {
         },
         "object": "config",
     });
-    // Disabled because it is causing issues https://github.com/dani-garcia/vaultwarden/discussions/4052
-    feature_states["autofill-v2"] = serde_json::Value::Bool(false);
-    for feature in features {
-        feature_states[feature.to_string()] = serde_json::Value::Bool(true);
-    }
-    config["featureStates"] = feature_states;
+    config["featureStates"] = serde_json::to_value(feature_states).unwrap();
     Json(config)
 }
 
--- a/src/api/core/mod.rs
+++ b/src/api/core/mod.rs
@@ -193,7 +193,10 @@ fn version() -> Json<&'static str> {
 #[get("/config")]
 fn config() -> Json<Value> {
     let domain = crate::CONFIG.domain();
-    Json(json!({
+    let feature_flags = crate::CONFIG.feature_flags().to_lowercase();
+    let features = feature_flags.split(',').map(|f| f.trim()).collect::<Vec<_>>();
+    let mut feature_states = json!({});
+    let mut config = json!({
         // Note: The clients use this version to handle backwards compatibility concerns
         // This means they expect a version that closely matches the Bitwarden server version
         // We should make sure that we keep this updated when we support the new server features
@@ -213,15 +216,15 @@ fn config() -> Json<Value> {
           "notifications": format!("{domain}/notifications"),
           "sso": "",
         },
-        "featureStates": {
-          // Any feature flags that we want the clients to use
-          // Can check the enabled ones at:
-          // https://vault.bitwarden.com/api/config
-          "fido2-vault-credentials": true,  // Passkey support
-          "autofill-v2": false,             // Disabled because it is causing issues https://github.com/dani-garcia/vaultwarden/discussions/4052
-        },
         "object": "config",
-    }))
+    });
+    // Disabled because it is causing issues https://github.com/dani-garcia/vaultwarden/discussions/4052
+    feature_states["autofill-v2"] = serde_json::Value::Bool(false);
+    for feature in features {
+        feature_states[feature.to_string()] = serde_json::Value::Bool(true);
+    }
+    config["featureStates"] = feature_states;
+    Json(config)
 }
 
 pub fn catchers() -> Vec<Catcher> {
--- a/src/config.rs
+++ b/src/config.rs
@@ -757,17 +757,11 @@ fn validate_config(cfg: &ConfigItems) -> Result<(), Error> {
     }
 
     const SUPPORTED_FLAGS: &[&str] = &[
+        "autofill-overlay",
+        "autofill-v2",
+        "browser-fileless-import",
         "display-kdf-iteration-warning",
         "fido2-vault-credentials",
-        "trusted-device-encryption",
-        "passwordless-login",
-        "autofill-v2",
-        "autofill-overlay",
-        "browser-fileless-import",
-        "item-share",
-        "flexible-collections",
-        "flexible-collections-v-1",
-        "bulk-collection-access",
     ];
     for flag in parse_feature_flags(&cfg.feature_flags).keys() {
         if !SUPPORTED_FLAGS.contains(&flag.as_str()) {
--- a/src/config.rs
+++ b/src/config.rs
@@ -756,12 +756,8 @@ fn validate_config(cfg: &ConfigItems) -> Result<(), Error> {
         )
     }
 
-    const KNOWN_FLAGS: &[&str] = &[
-        "autofill-overlay",
-        "autofill-v2",
-        "browser-fileless-import",
-        "fido2-vault-credentials",
-    ];
+    const KNOWN_FLAGS: &[&str] =
+        &["autofill-overlay", "autofill-v2", "browser-fileless-import", "fido2-vault-credentials"];
     for flag in parse_experimental_client_feature_flags(&cfg.experimental_client_feature_flags).keys() {
         if !KNOWN_FLAGS.contains(&flag.as_str()) {
             err!(format!("The experimental client feature flag {flag:?} is unrecognized. Please ensure the feature flag is spelled correctly, a caret (^) is used for disabling and that it is supported in this version."));
--- a/src/config.rs
+++ b/src/config.rs
@@ -758,7 +758,7 @@ fn validate_config(cfg: &ConfigItems) -> Result<(), Error> {
         &["autofill-overlay", "autofill-v2", "browser-fileless-import", "fido2-vault-credentials"];
     for flag in parse_experimental_client_feature_flags(&cfg.experimental_client_feature_flags).keys() {
         if !KNOWN_FLAGS.contains(&flag.as_str()) {
-            err!(format!("The experimental client feature flag {flag:?} is unrecognized. Please ensure the feature flag is spelled correctly and that it is supported in this version."));
+            warn!("The experimental client feature flag {flag:?} is unrecognized. Please ensure the feature flag is spelled correctly and that it is supported in this version.");
         }
     }
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -9,7 +9,7 @@ use reqwest::Url;
 use crate::{
     db::DbConnType,
     error::Error,
-    util::{get_env, get_env_bool, parse_feature_flags},
+    util::{get_env, get_env_bool, parse_experimental_client_feature_flags},
 };
 
 static CONFIG_FILE: Lazy<String> = Lazy::new(|| {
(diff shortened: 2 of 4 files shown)


BlackDex:
I'm not very fond of adding this.
The main reason is that people might enable something which either needs server changes or is not fully supported yet on all the clients if not updated.

One item which we do not (yet) support is flexible collections for example (and i do not even know what it does too).
And there are several others which are pointless when using Vaultwarden.

So, this will be a no-go from me.

PKizzle:
I understand your concerns regarding possible issues with unsupported server-side features. However, I would like to suggest that those features could simply be removed from the list of allowed features.

When it comes to old client versions that do not support all features enabled in the server, they simply ignore those features. I have already tested it with the autofill-overlay feature flag, which is only available in the latest client release, and older clients simply ignore it.

In the end, it is up to the user to make sure their configuration is stable and secure, as there are countless other ways to screw up your configuration. As the default values are exactly the same as before, a warning could be added to the documentation making sure that users are aware of the implications that en- or disabling of feature flags has.

I respectfully disagree with your decision to reject this feature, as I believe it would provide more flexibility and customization for the users. I hope you will reconsider your position and give this feature a chance.

PS: As already mentioned above I have enabled the autofill-overlay feature which for me is one of the biggest improvements to Bitwarden and has been demanded by the community [since 2018](https://community.bitwarden.com/t/overlay-popup-interface/14). It is currently hidden behind this feature flag so users wanting to test it require a way to enable it.

BlackDex:
I meant to say _as-is_ actually.

But we currently only support:
- `autofill-v2`
- `fido2-vault-credentials`

I think a few others might also work. So if you can prune the list to the bare minimum support items, that would be cool.

PKizzle:
I guess the following feature flags are purely client related if I am not mistaken?

- `autofill-overlay` (requires `autofill-v2`)
- `autofill-v2`
- `browser-fileless-import`
- `display-kdf-iteration-warning`
- `passwordless-login` (requires `trusted-device-encryption`)

And these might require additional changes in Vaultwarden:

- `bulk-collection-access`
- `flexible-collections`
- `flexible-collections-v-1`
- `item-share`
- `trusted-device-encryption`

BlackDex:
I think `browser-fileless-import` is used to import from LastPass via a client without using uploaded files. I suppose they use the same import endpoints so that shouldn't be an issue.

PKizzle:
Then the list of available featureStates in Vaultwarden would look like:

- `autofill-overlay` (requires `autofill-v2`)
- `autofill-v2`
- `browser-fileless-import`
- `display-kdf-iteration-warning`
- `fido2-vault-credentials`

By the way is something like `trusted-device-encryption` ([Bitwarden doc](https://bitwarden.com/help/about-trusted-devices/#:~:text=The%20Device%20Public%20Key%20is,Device%20Key%2DEncrypted%20Private%20Key.)) and `passwordless-login` (WebAuthn client login) planned for Vaultwarden as well?

BlackDex:
@PKizzle, the `trusted-device-encryption` is linked to SSO. There is a PR open for SSO (Which i still need to look better at) so if, it should be part of that, or can be worked upon after it has been merged.

The `passwordless-login` probably needs some new endpoints or login validation.
While we already support Webauthn as 2FA/MFA, using it as a passwordless type probably needs some adjustments.

Keep in mind that the Bitwarden Self Hosted currently only supports
- `trusted-device-encryption` (Because it has SSO support)
- `fido2-vault-credentials`

PKizzle:
This is how Vaultwarden currently looks when enabling `passwordless-login` (the spinner never stops as the API call fails):
Screenshot passwordless login

stefan0xC:
[Last time I checked](https://github.com/dani-garcia/bw_web_builds/pull/141#issuecomment-1741935735) the `display-kdf-iteration-warning` feature flag is not used anymore, so enabling it should not make a difference.

src/api/core/mod.rs
```diff
@@ -213,15 +213,10 @@ fn config() -> Json<Value> {
...
-          // Can check the enabled ones at:
-          // https://vault.bitwarden.com/api/config
-          "fido2-vault-credentials": true,  // Passkey support
-          "autofill-v2": false,             // Disabled because it is causing issues https://github.com/dani-garcia/vaultwarden/discussions/4052
-        },
         "object": "config",
-    }))
+    });
+    config["featureStates"] = serde_json::to_value(feature_states).unwrap();
+    Json(config)
```
BlackDex, line 219: I would suggest to remove the `let mut` here and let `json!` handle it all automatically. No reason to do it manually.

```suggestion
    Json(json!({
        // Note: The clients use this version to handle backwards compatibility concerns
        // This means they expect a version that closely matches the Bitwarden server version
        // We should make sure that we keep this updated when we support the new server features
        // Version history:
        // - Individual cipher key encryption: 2023.9.1
        "version": "2023.9.1",
        "gitHash": option_env!("GIT_REV"),
        "server": {
          "name": "Vaultwarden",
          "url": "https://github.com/dani-garcia/vaultwarden",
          "version": crate::VERSION
        },
        "environment": {
          "vault": domain,
          "api": format!("{domain}/api"),
          "identity": format!("{domain}/identity"),
          "notifications": format!("{domain}/notifications"),
          "sso": "",
        },
        "featureStates": feature_states,
        "object": "config",
    }))
```

src/config.rs
```diff
@@ -547,6 +547,11 @@ make_config! {
         /// TOTP codes of the previous and next 30 seconds will be invalid.
         authenticator_disable_time_drift: bool, true, def, false;
 
+        /// Customize the enabled feature flags on the clients |> This is a comma separated list of feature flags to en-/disable.
+        /// Features are enabled by default which can be overridden by prefixing the feature with a `!`.
+        /// Autofill v2 is disabled by default because it is causing issues https://github.com/dani-garcia/vaultwarden/discussions/4052
+        feature_flags: String, false, def, "!autofill-v2,fido2-vault-credentials".to_string();
```
BlackDex, line 553: Maybe we should call this `client_feature_flags`?
This kinda looks like enabling features for Vaultwarden it self.

BlackDex (changes requested):
Looks almost ok to me.
There are a few items i left a comment about.

I'm also missing the new feature in the `.env.template` with comments so that needs to be added.

And, there might be an issue with using `!` as an exclusion option.
In bash this will cause issues when used between `"` double quotes.

We can either add a good note regarding this in the config/template/wiki or think of a different character or other way to negate it? Maybe a `^` could be an option? While i do like the `!` though, it is kinda universal.

stefan0xC (changes requested):
I think if we let users decide which features should be enabled/disabled, we should use a conservative approach (i.d. the features should be disabled by default).

The problem with making it configurable by users at all is that we can't provide the defaults anymore for all users because whether or not we enable or disable them by default, once a user has the default saved in the `config.json` they'd need to update the feature flag list accordingly, even if we decided that it is (not) supported anymore.

So if this is implemented I'd rather tell users that they'll have to enable a known feature flag to help us test a new feature instead of telling users that they'll have to disable features which we have set as a default and have noticed that they were not ready yet.

Regarding the question of whether or not we should also give users a way to disable features I think this would depend on the clients. Currently if a feature flag is missing the clients should assume that any given feature flag defaults to `false` (however that might not be the case given that the `getFeatureFlag()` function technically allows for providing a different `defaultValue` in the client).

Last point I'd prefer if users only had to list features they want enabled to keep it as simple as possible. (And only implement a way to disable a feature explicitly, when it is really mandatory, which I personally would consider a bug in the clients repository...)

src/config.rs
BlackDex: But we already have some flags set to either explicitly to true or false. That counts as a default already too me.

How would you address those?

On one hand I'm more comfortable that we decide these flags and there state, on the other hand i think people want to tune these items.

stefan0xC: I mean yeah, we could set the default for a known feature flag that is already supported can be set to `true` like we do for `fido2-vault-credentials`. Setting a feature flag to `false` or not having the flag should be treated the same IMHO (so we could also remove the `autofill-v2` while it's not supported).

Wether the user should be able to decide and how is the concern I wanted to address by explicitly discussing the consequences. If we let users decide some of them will have to update the list of enabled feature flags if they manually set this new environment variable or if they are using the `config.json` with this value set (so they might not know that they have set a feature flag). The defaults will still be decided by us for users that don't use the `config.json` and don't set the environment variable.

But if we opt to let users decide we should be aware of what this means for them and for us. Making this configurable instead of hardcoding the supported features, some users will have to keep track of our release notes to enable new features (or wait until the client does not check it anymore) variable. And in this case I'd rather use feature flags as a way to test it.

If a feature flag is removed we'd have to implement the functionality anyways to stay compatible with the clients.

PKizzle: Sadly for some feature flags like `autofill-v2` it is not possible to deactivate them in the client by leaving them out of the `featureStates` as their default value is `true` (to be fair I have not yet tested this). This is why the exclamation mark notation was added.

Yes, hardcoding of supported `featureStates` is required as not all of those client features are supported by Vaultwarden. However, any changes to the list will cause incompatible configurations to fail the validation and make users aware of necessary changes to perform (i.e. check the release notes). Users not wanting to experiment with the `featureStates` themselves will not have any complications as sensible default values can be provided.

PKizzle:
@BlackDex It seems like Bitwarden is now enabling `autofill-v2` by default: https://api.bitwarden.com/config

src/config.rs
stefan0xC: As far as I can tell the value should be `false` if the flag is not provided. C.f. https://contributing.bitwarden.com/architecture/adr/feature-management/

We have disabled the feature flag in #4056 because users reported issues with the feature (which had nothing to do with Vaultwarden I think because the behavior is not dependent on any server side code as far as I can tell) but from my understanding removing the feature flag from the `/api/config` should have had the same effect (unless the `defaultValue` in the client is really set explicitly to `true`, which could happen but I don't think is the case, i.e. `web-v2023.12.0` [just checks if the feature flag is provided](https://github.com/bitwarden/clients/blob/8d90085607341deab976952bce8d8937cf3eefb1/apps/browser/src/background/runtime.background.ts#L136) and does not assume that it's true).

But since there were some issues they have (hopefully) already fixed the issue in `main`, c.f. https://github.com/bitwarden/clients/commit/bf60711efec33fd3b07227a58856dec6396e8e18 (so we could also soon enable the flag again (maybe) but that's also why I'm saying that we should take a more conservative approach regarding those feature flags because with the acceptance of this change the users could test the new feature that is currently in development upstream as soon as it's available but we either don't set it or only set it as default once it's more or less known to work without issues - because there's a demand to enable some new features but currently as far as I checked/understand only a few will be provided by [the bitwarden/server itself](https://github.com/bitwarden/server/blob/828566d87999acd1281dcfaea8761f7f027c95a0/src/Core/Constants.cs#L108-L117)).

stefan0xC:
@PKizzle feature flags as I understand them are not enabled by default. So we should add those features that might be safe to enable to the `KNOWN_FLAGS` (and not add something that is known that it is not supported in the backend like `trusted-device-encryptions` - i.e. the PR that implements this feature would have to also add such feature flags and once the feature is stable, we could also add this to a default if it makes sense)

PKizzle:
@stefan0xC This is sadly not always the case. As mentioned earlier `autofill-v2` was enabled by default in several client releases before being disabled again using Bitwarden's `featureStates`. Fore more information please have a look at the discussion in https://github.com/dani-garcia/vaultwarden/discussions/4052

src/util.rs
```diff
@@ -747,3 +748,22 @@ pub fn convert_json_key_lcase_first(src_json: Value) -> Value {
...
+    let features = experimental_client_feature_flags_lowercase.split(',').map(|f| f.trim()).collect::<Vec<_>>();
+    let mut feature_states: HashMap<String, bool> = HashMap::new();
+
+    for feature in features {
+        let is_enabled = !feature.starts_with('^');
+        let flag = if is_enabled {
+            feature
+        } else {
+            &feature[1..]
+        };
```
stefan0xC, line 764: After looking up the relevant [Bitwarden architecture decision record](https://contributing.bitwarden.com/architecture/adr/feature-management/) I think it is unnecessary to provide a way to explicitly disable a feature as simply removing an enabled feature from the list would also disable that feature. (If this were not the case it would mean that it's a bug in the client as the feature flag should either be removed if a feature became active or this would have to be separately configurable somehow on the client side, e.g. by [client side flags](https://github.com/bitwarden/clients/blob/6199e585326949120350e396ab5bce64d64aba18/apps/web/config/selfhosted.json#L9-L13). I.e. the clients should never assume a feature flag to be enabled by default as that would be against the point of their feature flag system, at least as far as I understand it.) 

Not having this logic should simplify the way we can configure the enabled `featureStates` and this would make maintaining the list for us and our users easier as well as there would be no additional overhead.

Other than that we'd also have to add new possible features to the `KNOWN_FLAGS` of course (which should be easier for client-side only features that don't require any changes in the server backend). But as I've already argued I'd take the more conservative approach and only enable feature flags by default that are more or less stable and known to work (e.g. if they are also set to `true` for self-hosted Bitwarden instances) and let users decide if they really want to enable the feature flag early (i.e. to test new features or because the upstream Bitwarden cloud services have them enabled too and users don't want to wait until the feature flags are removed from the clients).

PKizzle:
@stefan0xC I see. I misunderstood some of [the comments in the Bitwarden repo](https://github.com/bitwarden/clients/issues/6832#issuecomment-1802579736) regarding the issue hinting at the client not resetting a feature flag to `false` once it was enabled. However, it seems like the self-hosted server version had the `autofill-v2` feature flag enabled and not the client by default.

src/config.rs
```diff
@@ -751,6 +754,14 @@ fn validate_config(cfg: &ConfigItems) -> Result<(), Error> {
         )
     }
 
+    const KNOWN_FLAGS: &[&str] =
+        &["autofill-overlay", "autofill-v2", "browser-fileless-import", "fido2-vault-credentials"];
+    for flag in parse_experimental_client_feature_flags(&cfg.experimental_client_feature_flags).keys() {
+        if !KNOWN_FLAGS.contains(&flag.as_str()) {
+            err!(format!("The experimental client feature flag {flag:?} is unrecognized. Please ensure the feature flag is spelled correctly and that it is supported in this version."));
```
dani-garcia, line 761:
```suggestion
            warn!("The experimental client feature flag {flag:?} is unrecognized. Please ensure the feature flag is spelled correctly and that it is supported in this version.");
```

I don't think the validation should return an error, if we ever update the allowed flags and remove some, we don't want that to break all the users that have enabled it before.

This could potentially happen if we find a big error in the implementation of one, or also when the feature flag code is stabilized, the flag might be deleted once all the clients support it.

src/util.rs
```diff
@@ -747,3 +748,16 @@ pub fn convert_json_key_lcase_first(src_json: Value) -> Value {
...
+    let experimental_client_feature_flags_lowercase = experimental_client_feature_flags.to_lowercase();
+    let features = experimental_client_feature_flags_lowercase.split(',').map(|f| f.trim()).collect::<Vec<_>>();
+    let mut feature_states: HashMap<String, bool> = HashMap::new();
+
+    for feature in features {
+        feature_states.insert(feature.to_string(), true);
+    }
+
+    feature_states
+}
```
dani-garcia, line 763:
```suggestion
pub fn parse_experimental_client_feature_flags(flags: &str) -> HashMap<String, bool> {
    let feature_states: HashMap<String, bool> = flags.to_lowercase().split(',').map(|f| (f.trim().to_owned(), true)).collect();

    feature_states
}
```

This can be simplified if you know that you can `collect()` into a `HashMap` :)

dani-garcia (commented):
Some minor comments, I agree about having a whitelist of features to enable, to avoid someone accidentally breaking their vault if the clients think that the server supports a feature that it doesn't.

How would we handle a change in the defaults for users that have already configured their flags? For example, by default we now set `fido2-vault-credentials`. If I decide on my instance that I don't care about passkeys but want the new autofill, I might set this setting to `autofill-v2`.

Now, if we release support for a new feature, we won't have any way to get this user updated, beyond announcing on the README or the changelog that this new feature is available. I think we need a way to forcefully enable some feature flags, regardless of user config. Of course we'd need to be conservative about it to make sure we don't break anything, but we can't afford to let users stay on old feature flags forever.

PKizzle:
@dani-garcia A possible way to inform users about new features is to display a notice in the UI alongside the recommended default configuration.
Since these features are experimental, users who want to try them should be aware of the potential risks (such as bugs or removal of the feature in the future). It might also be helpful to show a general warning if the user's configuration differs from the default values in terms of the feature flags.

dani-garcia:
My concern is some future feature flag informing the clients to use some new API endpoint instead of the existing one, eventually we might want to deprecate and then remove the old API endpoints, which can't be done if some older clients haven't enabled that flag.

Honestly this is mostly a future proofing issue and might turn out to be not a big deal, so we can cross that bridge when we get to it.

Yeah showing a warning with the defaults and maybe all the available values in the UI might be nice, but that can be done separately from this change, I think. We can also reserve a section at the release notes to mention any added/removed/updated flags too.

BlackDex:
As a note, they released a bug fix release which disables the overlay by default.

https://github.com/bitwarden/clients/compare/web-v2023.12.0...browser-v2023.12.1

PKizzle:
The default value is now off, but the visibility in the client's settings is still controlled by the feature flag, which remains unchanged. I have already enabled the feature and adjusted the settings in multiple clients, so I don’t think this change affects me. Thank you for the information anyway!

src/config.rs
stefan0xC: @dani-garcia I'm not sure I agree. I think it's better to return an error so that users can't enable unsupported flags (or at least immediately notice that they have misspelled a feature flag when changing the flag via the admin panel - which I'm not sure would happen if this were a mere warning) and if we remove a feature flag from `KNOWN_FLAGS` (because it's not supported anymore or was added prematurely) I'd probably want this validation to fail. Furthermore I don't think that it's an issue enabling feature flags that are not in use anymore (like `display-kdf-iteration-warning`), so we won't have to remove them as soon as possible. (Users might also want to clean up removed feature flags that they have enabled to reduce the size of the returned `/api/config` - without immediate feedback I don't think this will happen.)

stefan0xC: In order not to break too many installations we probably should not remove features from `KNOWN_FLAGS` that we have enabled by default (like `fido2-vault-credentials`) or only if there's an important reason to do so or we really want to remove it. (I think the other flags which we don't enable by default - would be fair game as users would be responsible if they change the defaults).

BlackDex (approved):

dani-garcia (approved):

cl1ent:
thank you for this PR! do you have any ETA when this will be part of the next (minor) release? once #4263 is merged maybe?

bibicadotnet:
> thank you for this PR! do you have any ETA when this will be part of the next (minor) release? once #4263 is merged maybe?

If you like, can use `image:vaultwarden/server:testing`
Add environment `- EXPERIMENTAL_CLIENT_FEATURE_FLAGS=autofill-overlay,autofill-v2,browser-fileless-import,fido2-vault-credentials`

PR #4173, opened 2023-12-15
Update crates

BlackDex:
Update all crates instead of only the zerocopy from dependabot.
Closes #4170

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -140,13 +140,13 @@ dependencies = [
 
 [[package]]
 name = "async-global-executor"
-version = "2.4.0"
+version = "2.4.1"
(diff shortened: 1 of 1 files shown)


dani-garcia (approved):

tessus:
@BlackDex when do you change the `.toml` file and when the `.lock` file? I thought the lock file is generated when the `.toml` file changes. Aren't dependencies permanently updated in the `.toml` file?

BlackDex:
I change the `.toml` when versions change which we defined our self. The `.lock` is changed when running `cargo update`.

I generally run `cargo update --dry-run` first to see if there are packages update which we defined our self, and then also check those changelogs.

I also use a crate `cargo-outdated` which also can check some crates updates.

Issue #4266, opened 2024-01-22
Docker compose don't parse DATABASE_URL_FILE


thejoelinux:
### Subject of the issue
We are trying to deploy the docker container in a Docker Swarm.
As the DATABASE_URL contains a password, we try to pass it via a DATABASE_URL_FILE env variable, but it is not parsed.
At the same time, when DATABASE_URL_FILE and DATABASE_URL are both defined, it throws an error.

### Deployment environment

Docker swarm, vault-warden on MariaDB.

### Steps to reproduce

Put the DATABASE_URL in a secret file (i.e. /run/secrets/database_url_secret), or a config...
Set DATABASE_URL_FILE=/run/secrets/database_url_secret 

### Expected behaviour

Vault connect to the database using the URL in the DATABASE_URL_FILE

### Actual behaviour

The container don't start, complaining the database is not found.

BlackDex:
Works just fine for me. Can you share your docker-compose.yml relevant parts?

This is what worked for me:
```yaml
---
version: '3.8'

services:
  vaultwarden:
    image: ghcr.io/dani-garcia/vaultwarden:latest-alpine
    network_mode: host
    container_name: vaultwarden
    restart: unless-stopped
    environment:
      TZ: Europe/Amsterdam
      ROCKET_PORT: 8888
      DATABASE_URL_FILE: /run/secrets/database_url
      I_REALLY_WANT_VOLATILE_STORAGE: true
      EXTENDED_LOGGING: true
      LOG_LEVEL: info
      DISABLE_ADMIN_TOKEN: true
    secrets:
      - database_url

secrets:
  database_url:
    file: mariadb_url.txt
```

BlackDex:
Also, the error message when  you define both specifically states to not define both.
```
You should not define both DATABASE_URL and DATABASE_URL_FILE!
```
That is as intended, and the message seems to be clearly stating what the issue is and how to resolve it.

PR #4267, opened 2024-01-22
DATABASE_URL_PARAM_FILE for docker

thejoelinux:
Permit the usage of a DATABASE_URL_PARAM_FILE variable, pointing to a file containing the actual DATABASE_URL.


BlackDex:
This is not needed as you can just use `DATABASE_URL_FILE`.
See the issue you created #4266, as it works just fine as it should.

If you want to have this in your own repo/fork that is fine, but we are not going to merge this.

PR #4276, opened 2024-01-27
Update env template file

gzfrozen:
- Fix some default parameters which not match the config.rs.
- Categorize settings into sevral sections. Mainly followed the config.rs's structure.
- Add a missing setting: AUTH_REQUEST_PURGE_SCHEDULE

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -17,12 +17,12 @@
 ## When using SQLite, this is the path to the DB file, default to %DATA_FOLDER%/db.sqlite3
 # DATABASE_URL=data/db.sqlite3
 ## When using MySQL, specify an appropriate connection URI.
-## Details: https://docs.diesel.rs/diesel/mysql/struct.MysqlConnection.html
+## Details: https://docs.diesel.rs/2.1.x/diesel/mysql/struct.MysqlConnection.html
 # DATABASE_URL=mysql://user:password@host[:port]/database_name
 ## When using PostgreSQL, specify an appropriate connection URI (recommended)
 ## or keyword/value connection string.
 ## Details:
-## - https://docs.diesel.rs/diesel/pg/struct.PgConnection.html
+## - https://docs.diesel.rs/2.1.x/diesel/pg/struct.PgConnection.html
 ## - https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING
 # DATABASE_URL=postgresql://user:password@host[:port]/database_name
 
@@ -52,7 +52,7 @@
 
 ## Templates data folder, by default uses embedded templates
 ## Check source code to see the format
-# TEMPLATES_FOLDER=/path/to/templates
+# TEMPLATES_FOLDER=data/templates
 ## Automatically reload the templates for every request, slow, use only for development
 # RELOAD_TEMPLATES=false
 
@@ -78,7 +78,7 @@
 
 ## Enables push notifications (requires key and id from https://bitwarden.com/host)
 ## If you choose "European Union" Data Region, uncomment PUSH_RELAY_URI and PUSH_IDENTITY_URI then replace .com by .eu
-# PUSH_ENABLED=true
+# PUSH_ENABLED=false
 # PUSH_INSTALLATION_ID=CHANGEME
 # PUSH_INSTALLATION_KEY=CHANGEME
 ## Don't change this unless you know what you're doing.
@@ -327,7 +327,7 @@
 
 ## Number of server-side passwords hashing iterations for the password hash.
 ## The default for new users. If changed, it will be updated during login for existing users.
-# PASSWORD_ITERATIONS=350000
+# PASSWORD_ITERATIONS=600000
 
 ## Controls whether users can set password hints. This setting applies globally to all users.
 # PASSWORD_HINTS_ALLOWED=true
@@ -342,6 +342,9 @@
 ## It's recommended to configure this value, otherwise certain functionality might not work,
 ## like attachment downloads, email links and U2F.
 ## For U2F to work, the server must use HTTPS, you can use Let's Encrypt for free certs
+## For development
+# DOMAIN=http://localhost
+## For public server
 # DOMAIN=https://vw.domain.tld:8443
 
 ## Allowed iframe ancestors (Know the risks!)
@@ -405,7 +408,7 @@
 ## - "autofill-v2": Use the new autofill implementation.
 ## - "browser-fileless-import": Directly import credentials from other providers without a file.
 ## - "fido2-vault-credentials": Enable the use of FIDO2 security keys as second factor.
-## EXPERIMENTAL_CLIENT_FEATURE_FLAGS=fido2-vault-credentials
+# EXPERIMENTAL_CLIENT_FEATURE_FLAGS=fido2-vault-credentials
 
 ## Rocket specific settings
 ## See https://rocket.rs/v0.4/guide/configuration/ for more details.
@@ -434,7 +437,7 @@
 ## Defaults for SSL is "Plain" and "Login" and nothing for Non-SSL connections.
 ## Possible values: ["Plain", "Login", "Xoauth2"].
 ## Multiple options need to be separated by a comma ','.
-# SMTP_AUTH_MECHANISM="Plain"
+# SMTP_AUTH_MECHANISM=
 
 ## Server name sent during the SMTP HELO
 ## By default this value should be is on the machine's hostname,
@@ -442,7 +445,7 @@
 # HELO_NAME=
 
 ## Embed images as email attachments
-# SMTP_EMBED_IMAGES=false
+# SMTP_EMBED_IMAGES=true
 
 ## SMTP debugging
 ## When set to true this will output very detailed SMTP messages.
--- a/.env.template
+++ b/.env.template
@@ -321,8 +321,8 @@
 
 ## Icon download timeout
 ## Configure the timeout value when downloading the favicons.
-## The default is 9 seconds, but this could be to low on slower network connections
-# ICON_DOWNLOAD_TIMEOUT=9
+## The default is 10 seconds, but this could be to low on slower network connections
+# ICON_DOWNLOAD_TIMEOUT=10
 
 ## Icon blacklist Regex
 ## Any domains or IPs that match this regex won't be fetched by the icon service.
--- a/.env.template
+++ b/.env.template
@@ -10,8 +10,30 @@
 ## variable ENV_FILE can be set to the location of this file prior to starting
 ## Vaultwarden.
 
+####################
+### Data folders ###
+####################
+
 ## Main data folder
 # DATA_FOLDER=data
+
+## Individual folders, these override %DATA_FOLDER%
+# RSA_KEY_FILENAME=data/rsa_key
+# ICON_CACHE_FOLDER=data/icon_cache
+# ATTACHMENTS_FOLDER=data/attachments
+# SENDS_FOLDER=data/sends
+# TMP_FOLDER=data/tmp
+
+## Templates data folder, by default uses embedded templates
+## Check source code to see the format
+# TEMPLATES_FOLDER=data/templates
+
+## Web vault folder
+# WEB_VAULT_FOLDER=web-vault/
+
+#########################
+### Database settings ###
+#########################
 
 ## Database URL
 ## When using SQLite, this is the path to the DB file, default to %DATA_FOLDER%/db.sqlite3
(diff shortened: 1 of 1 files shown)


.env.template
```diff
@@ -337,12 +270,129 @@
...
+## If an external service is set, an icon request to Vaultwarden will return an HTTP
+## redirect to the corresponding icon at the external service. An external service may
+## be useful if your Vaultwarden instance has no external network connectivity, or if
+## you are concerned that someone may probe your instance to try to detect whether icons
+## for certain sites have been cached.
+# ICON_SERVICE=internal
+
+## Icon redirect code
+## The HTTP status code to use for redirects to an external icon service.
+## The supported codes are 300 (legacy permanent), 302 (legacy temporary), 307 (temporary), and 308 (permanent).
```
dfunkt, line 311: Should be 301 instead of 300.

.env.template
```diff
@@ -337,12 +270,129 @@
...
+## Temporary redirects are useful while testing different icon services, but once a service
+## has been decided on, consider using permanent redirects for cacheability. The legacy codes
+## are currently better supported by the Bitwarden clients.
+# ICON_REDIRECT_CODE=301
```
dfunkt, line 315: Default is 302 (https://github.com/dani-garcia/vaultwarden/blob/main/src/config.rs#L532)

dfunkt (changes requested):

.env.template
gzfrozen: @dfunkt 
Wow, thank you for find out this mistake.
It's weird why I edit this number 😆

.env.template
```diff
@@ -337,12 +270,129 @@
...
+
+## Cache time-to-live for successfully obtained icons, in seconds (0 is "forever")
+# ICON_CACHE_TTL=2592000
+## Cache time-to-live for icons which weren't available, in seconds (0 is "forever")
+# ICON_CACHE_NEGTTL=259200
+
+## Icon download timeout
+## Configure the timeout value when downloading the favicons.
+## The default is 9 seconds, but this could be to low on slower network connections
+# ICON_DOWNLOAD_TIMEOUT=9
```
dfunkt, line 325:
```suggestion
## The default is 10 seconds, but this could be to low on slower network connections
# ICON_DOWNLOAD_TIMEOUT=10
```
https://github.com/dani-garcia/vaultwarden/blob/main/src/config.rs#L538

dfunkt (changes requested):

dfunkt (approved):

gzfrozen:
@dfunkt 
I think this wired number decrease mistake is cause by my editor vim binding, which seems Ctrl-X is decrease number.
I will check out the file again.

.env.template
```diff
@@ -434,38 +473,58 @@
...
+
+##########################
+### Rocket settings ###
+##########################
+
+## Rocket specific settings
+## See https://rocket.rs/v0.4/guide/configuration/ for more details.
+# ROCKET_ADDRESS=0.0.0.0
+# ROCKET_PORT=80  # Defaults to 80 in the Docker images, or 8000 otherwise.
+# ROCKET_WORKERS=10
```
BlackDex, line 526: The workers default isn't 10. It will match `num_cores` by default. Also, best to leave it as default instead of setting this manually. So, maybe best to remove this setting, as it will only be confusing i think.

.env.template
```diff
@@ -434,38 +473,58 @@
...
+
+## Maximum attempts before an email token is reset and a new email will need to be sent.
+# EMAIL_ATTEMPTS_LIMIT=3
+
+##########################
+### Rocket settings ###
+##########################
+
+## Rocket specific settings
+## See https://rocket.rs/v0.4/guide/configuration/ for more details.
```
BlackDex, line 523: This link can be updated to https://rocket.rs/v0.5/guide/configuration/

.env.template
```diff
@@ -434,38 +473,58 @@
...
+# ROCKET_ADDRESS=0.0.0.0
+# ROCKET_PORT=80  # Defaults to 80 in the Docker images, or 8000 otherwise.
+# ROCKET_WORKERS=10
+# ROCKET_TLS={certs="/path/to/certs.pem",key="/path/to/key.pem"}
 
-## HIBP Api Key
-## HaveIBeenPwned API Key, request it here: https://haveibeenpwned.com/API/Key
-# HIBP_API_KEY=
 
 # vim: syntax=ini
```
BlackDex, line 530: Please add a empty new-line at the end of the file, that best practice in general.

.env.template
```diff
@@ -337,12 +270,129 @@
...
+## Disable authenticator time drifted codes to be valid.
+## TOTP codes of the previous and next 30 seconds will be invalid
+##
+## According to the RFC6238 (https://tools.ietf.org/html/rfc6238),
+## we allow by default the TOTP code which was valid one step back and one in the future.
+## This can however allow attackers to be a bit more lucky with there attempts because there are 3 valid codes.
+## You can disable this, so that only the current TOTP Code is allowed.
+## Keep in mind that when a sever drifts out of time, valid codes could be marked as invalid.
+## In any case, if a code has been used it can not be used again, also codes which predates it will be invalid.
+# AUTHENTICATOR_DISABLE_TIME_DRIFT=false
```
BlackDex, line 352: This is best placed together with the Yubico and DUO settings, since they all fall under the same section, MFA/2FA.

.env.template
```diff
@@ -434,38 +473,58 @@
...
+## Number of digits in an email 2FA token (min: 6, max: 255).
+## Note that the Bitwarden clients are hardcoded to mention 6 digit codes regardless of this setting!
+# EMAIL_TOKEN_SIZE=6
+
+## Token expiration time
+## Maximum time in seconds a token is valid. The time the user has to open email client and copy token.
+# EMAIL_EXPIRATION_TIME=600
+
+## Maximum attempts before an email token is reset and a new email will need to be sent.
+# EMAIL_ATTEMPTS_LIMIT=3
```
BlackDex, line 516: This is best placed together with the Yubico and DUO settings, since they all fall under the same section, MFA/2FA.

gzfrozen:
> @dfunkt I think this wired number decrease mistake is cause by my editor vim binding, which seems Ctrl-X is decrease number. I will check out the file again.

OK I think it is safe for digital derement mistakes.

.env.template
```diff
@@ -43,31 +77,9 @@
 ## - PostgreSQL: ""
```
BlackDex, line 77: This comment has nothing to do specifically with the setting bellow.

Also, the defaults are different per Database engine which is used.
MariaDB/Mysql and PostgreSQL are both empty, and for SQLite it is `PRAGMA busy_timeout = 5000; PRAGMA synchronous = NORMAL;`

See: https://github.com/BlackDex/vaultwarden/blob/8b66e3441571fdb71f0610882d1ee47dccc8c0a3/src/db/mod.rs#L231-L238

BlackDex (changes requested):
Looks ok in general, some sections might be better merged like i suggested in my comments. And some values seem to be missing.

I have not yet checked for other missing values which are currently in the config code.

.env.template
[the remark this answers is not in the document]
BlackDex: Probably a `CTRL+x` issue also ;)

.env.template
```diff
@@ -365,6 +415,17 @@
...
+## Controls whether group support is enabled for organizations
+## This setting applies to organizations.
+## Disabled by default because this is a beta feature, it contains known issues!
+## KNOW WHAT YOU ARE DOING!
+# ORG_GROUPS_ENABLED=false
+
+########################
+### Yubikey settings ###
+########################
+
```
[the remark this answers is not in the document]
gzfrozen: @BlackDex 
`YUBICO_CLIENT_ID` setting is not missing. 
Line 432, 433 is automatically hiddened by the github review UI I guss.

.env.template
```diff
@@ -373,6 +434,10 @@
 # YUBICO_SECRET_KEY=AAAAAAAAAAAAAAAAAAAAAAAA
 # YUBICO_SERVER=http://yourdomain.com/wsapi/2.0/verify
 
+####################
+### Duo settings ###
+####################
+
```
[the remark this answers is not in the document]
gzfrozen: @BlackDex 
Also these settings is not missing. Just hidden by Github review UI.
They are in line 446~448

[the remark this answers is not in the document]
BlackDex: Ah, indeed, sorry.

.env.template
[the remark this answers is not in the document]
BlackDex: Indeed sorry.

.env.template
gzfrozen: @dfunkt 
This is also an Github review UI accident.
Line 77~79 is folded.

gzfrozen:
@BlackDex 
Thank you for your review!
Fixed all metioned problems above. 

A big change is that I merged the Yubico, Duo, Email 2FA settings and other related settings into a big "MFA/2FA settings" section.

.env.template
```diff
@@ -150,42 +146,143 @@
...
+# SIGNUPS_DOMAINS_WHITELIST=example.com,example.net,example.org
+
+## Controls whether event logging is enabled for organizations
+## This setting applies to organizations.
+## Disabled by default. Also check the EVENT_CLEANUP_SCHEDULE and EVENTS_DAYS_RETAIN settings.
+# ORG_EVENTS_ENABLED=false
+
 ## Cron schedule of the job that cleans old events from the event table.
 ## Defaults to daily. Set blank to disable this job. Also without EVENTS_DAYS_RETAIN set, this job will not start.
 # EVENT_CLEANUP_SCHEDULE="0 10 0 * * *"
```
BlackDex, line 236: Now that i think of it, it might be better to keep `EVENTS_DAYS_RETAIN` and `EVENT_CLEANUP_SCHEDULE` together.
But still keep the `EVENT_CLEANUP_SCHEDULE` option near (or maybe the last one) in the list of `SCHEDULE` settings.

Else this will be the only one not listed there.
Then we can put the `DAYS` setting just right after the `SCHEDULE` option, since they are related.

BlackDex (changes requested):
Looks like almost done :)

.env.template
gzfrozen: OK, fixed👍

.env.template
```diff
@@ -153,39 +149,139 @@
...
-# LOG_TIMESTAMP_FORMAT="%Y-%m-%d %H:%M:%S.%3f"
+## Domain settings
+## The domain must match the address from where you access the server
+## It's recommended to configure this value, otherwise certain functionality might not work,
+## like attachment downloads, email links and U2F.
+## For U2F to work, the server must use HTTPS, you can use Let's Encrypt for free certs
+## For development
+# DOMAIN=http://localhost
+## For public server
+# DOMAIN=https://vw.domain.tld:8443
```
BlackDex, line 172: It might be a good idea to also add a link to the reverse proxy examples on the wiki.
`https://github.com/dani-garcia/vaultwarden/wiki/Proxy-examples`

And, the example with a port is maybe not the best one, while it is useful though to indicate that it is possible.
With that in mind, an example with a sub-path is also useful.

So, maybe add a `https://vw.domain.tld` and `https://domain.tld/vw` also with a small comment, and make the one without a port the default `For public server` example.

.env.template
```diff
@@ -208,12 +304,10 @@
...
-## Disable icon downloading
-## Set to true to disable icon downloading in the internal icon service.
-## This still serves existing icons from $ICON_CACHE_FOLDER, without generating any external
-## network requests. $ICON_CACHE_TTL must also be set to 0; otherwise, the existing icons
-## will be deleted eventually, but won't be downloaded again.
-# DISABLE_ICON_DOWNLOAD=false
+## Cache time-to-live for successfully obtained icons, in seconds (0 is "forever")
+# ICON_CACHE_TTL=2592000
+## Cache time-to-live for icons which weren't available, in seconds (0 is "forever")
+# ICON_CACHE_NEGTTL=259200
```
BlackDex, line 310: Maybe add here that 2592000 seconds is 30 days? That makes it more clear.

BlackDex (changes requested):
Just some small extra changes to make it a bit better.
But i think this is a good improvement :)
Thanks!

.env.template
gzfrozen: Do I also need to add these to wiki pages? Seems related.
https://github.com/dani-garcia/vaultwarden/wiki/Enabling-HTTPS
https://github.com/dani-garcia/vaultwarden/wiki/Running-a-private-vaultwarden-instance-with-Let%27s-Encrypt-certs

BlackDex: The Enable HTTPS is good i think, the other not so much, and is linked in the first.

gzfrozen:
> Just some small extra changes to make it a bit better. But i think this is a good improvement :) Thanks!

Thank you for your review! Changes been applied!

.env.template
```diff
@@ -10,30 +10,67 @@
...
+
+## Templates data folder, by default uses embedded templates
+## Check source code to see the format
+# TEMPLATES_FOLDER=data/templates
+## Automatically reload the templates for every request, slow, use only for development
+# RELOAD_TEMPLATES=false
+
+## Web vault settings
+# WEB_VAULT_FOLDER=web-vault/
+# WEB_VAULT_ENABLED=tru
```
BlackDex, line 35: Spotted a buggy here. Should be `true` 😄

BlackDex (changes requested):

.env.template
gzfrozen: Sorry, my bad😆
Fixed it.

dani-garcia (approved):
LGTM, thanks!

PR #4299, opened 2024-01-31
Fix healthcheck when using .env file

BlackDex:
It seems Debian based images see the `.env` file in the `pwd` path, but sourcing it via `. .env` breaks. It does work if you provide the full path `/.env`. Changed the default to `/.env`.

Alpine does not have an issue with both ways.

Fixes #4295

### DIFF
--- a/docker/healthcheck.sh
+++ b/docker/healthcheck.sh
@@ -1,17 +1,15 @@
-#!/bin/sh
+#!/usr/bin/env sh
 
 # Use the value of the corresponding env var (if present),
 # or a default value otherwise.
-: "${DATA_FOLDER:="data"}"
+: "${DATA_FOLDER:="/data"}"
 : "${ROCKET_PORT:="80"}"
+: "${ENV_FILE:="/.env"}"
 
 CONFIG_FILE="${DATA_FOLDER}"/config.json
 
-# Check if there is a .env file configured
+# Check if the $ENV_FILE file exist and is readable
 # If that is the case, load it into the environment before running any check
-if [ -z "${ENV_FILE}" ]; then
-    ENV_FILE=".env"
-fi
 if [ -r "${ENV_FILE}" ]; then
     # shellcheck disable=SC1090
     . "${ENV_FILE}"


dani-garcia (approved):

docker/healthcheck.sh
```diff
@@ -1,17 +1,15 @@
-#!/bin/sh
+#!/usr/bin/env sh
 
 # Use the value of the corresponding env var (if present),
 # or a default value otherwise.
-: "${DATA_FOLDER:="data"}"
+: "${DATA_FOLDER:="/data"}"
```
jjlin, line 5: I don't think it makes any difference given the way the Dockerfile is set up, but it probably would've been better to keep this as it was to match the actual default value

https://github.com/dani-garcia/vaultwarden/blob/4438da39f9c9aa9d83d1dddbf06616ad287f70eb/src/config.rs#L354

BlackDex: The same can be said for `.env`.
I just wanted to make sure it used the full path instead of somehow not pointing to the exact path.
It doesn't harm, and it is more specific.

docker/healthcheck.sh
```diff
@@ -1,17 +1,15 @@
-#!/bin/sh
+#!/usr/bin/env sh
```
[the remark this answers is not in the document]
BlackDex: While we know our images have the sh located on `/bin/sh`, i do find it more idiomatic to use this way.
The *extra* overhead is negligible i think.

docker/healthcheck.sh
jjlin: If someone were to change `WORKDIR` in their Dockerfile, then things wouldn't work the same. Arguably, it would be better to check if `ENV_FILE` contains a slash, and if not, use `${PWD}/${ENV_FILE}` instead.

docker/healthcheck.sh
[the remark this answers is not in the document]
jjlin: In my experience, `/bin/sh` is much more idiomatic than `/usr/bin/env sh`, though `env` is commonly used for less standard programs.

PR #4305, opened 2024-02-01
Allow to override log level for specific target

Timshel:
Useful for either silencing some target when running in `debug` or switching `debug` only for a specific target,

With this a good chunk of the default overrides could be moved to the conf, cf:

```rust
        // Hide unknown certificate errors if using self-signed
        .level_for("rustls::session", log::LevelFilter::Off)
        // Hide failed to close stream messages
        .level_for("hyper::server", log::LevelFilter::Warn)
        // Silence Rocket `_` logs
        .level_for("_", rocket_underscore_level)
        .level_for("rocket::response::responder::_", rocket_underscore_level)
        .level_for("rocket::server::_", rocket_underscore_level)
        .level_for("vaultwarden::api::admin::_", rocket_underscore_level)
        .level_for("vaultwarden::api::notifications::_", rocket_underscore_level)
        // Silence Rocket logs
        .level_for("rocket::launch", log::LevelFilter::Error)
        .level_for("rocket::launch_", log::LevelFilter::Error)
        .level_for("rocket::rocket", log::LevelFilter::Warn)
        .level_for("rocket::server", log::LevelFilter::Warn)
        .level_for("rocket::fairing::fairings", log::LevelFilter::Warn)
        .level_for("rocket::shield::shield", log::LevelFilter::Warn)
        .level_for("hyper::proto", log::LevelFilter::Off)
        .level_for("hyper::client", log::LevelFilter::Off)
        // Filter handlebars logs
        .level_for("handlebars::render", handlebars_level)
        // Prevent cookie_store logs
        .level_for("cookie_store", log::LevelFilter::Off)
        // Variable level for trust-dns used by reqwest
        .level_for("trust_dns_resolver::name_server::name_server", trust_dns_level)
        .level_for("trust_dns_proto::xfer", trust_dns_level)
        .level_for("diesel_logger", diesel_logger_level)
```

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -362,8 +362,9 @@
 ## Log level
 ## Change the verbosity of the log output
 ## Valid values are "trace", "debug", "info", "warn", "error" and "off"
-## Setting it to "trace" or "debug" would also show logs for mounted
-## routes and static file, websocket and alive requests
+## Setting it to "trace" or "debug" would also show logs for mounted routes and static file, websocket and alive requests
+## For a specific module append a comma separated `path::to::module=log_level`
+## For example, to only see debug logs for icons use: LOG_LEVEL="info,vaultwarden::api::icons=debug"
 # LOG_LEVEL=info
 
 ## Token for the admin interface, preferably an Argon2 PCH string
--- a/src/config.rs
+++ b/src/config.rs
@@ -576,8 +576,9 @@ make_config! {
         use_syslog:             bool,   false,  def,    false;
         /// Log file path
         log_file:               String, false,  option;
-        /// Log level
-        log_level:              String, false,  def,    "Info".to_string();
+        /// Log level |> Valid values are "trace", "debug", "info", "warn", "error" and "off"
+        /// For a specific module append it as a comma separated value "info,path::to::module=debug"
+        log_level:              String, false,  def,    "info".to_string();
 
         /// Enable DB WAL |> Turning this off might lead to worse performance, but might help if using vaultwarden on some exotic filesystems,
         /// that do not support WAL. Please make sure you read project wiki on the topic before changing this setting.
--- a/src/main.rs
+++ b/src/main.rs
@@ -26,6 +26,7 @@ extern crate diesel;
 extern crate diesel_migrations;
 
 use std::{
+    collections::HashMap,
     fs::{canonicalize, create_dir_all},
     panic,
     path::Path,
@@ -65,13 +66,7 @@ async fn main() -> Result<(), Error> {
     parse_args();
     launch_info();
 
-    use log::LevelFilter as LF;
-    let level = LF::from_str(&CONFIG.log_level()).unwrap_or_else(|_| {
-        let valid_log_levels = LF::iter().map(|lvl| lvl.as_str().to_lowercase()).collect::<Vec<String>>().join(", ");
-        println!("Log level must be one of the following: {valid_log_levels}");
-        exit(1);
-    });
-    init_logging(level).ok();
+    let level = init_logging()?;
 
     check_data_folder().await;
     auth::initialize_keys().unwrap_or_else(|e| {
@@ -89,7 +84,7 @@ async fn main() -> Result<(), Error> {
     schedule_jobs(pool.clone());
     crate::db::models::TwoFactor::migrate_u2f_to_webauthn(&mut pool.get().await.unwrap()).await.unwrap();
 
-    let extra_debug = matches!(level, LF::Trace | LF::Debug);
+    let extra_debug = matches!(level, log::LevelFilter::Trace | log::LevelFilter::Debug);
     launch_rocket(pool, extra_debug).await // Blocks until program termination.
 }
 
@@ -210,7 +205,38 @@ fn launch_info() {
     );
 }
 
-fn init_logging(level: log::LevelFilter) -> Result<(), fern::InitError> {
+fn init_logging() -> Result<log::LevelFilter, Error> {
+    let levels = log::LevelFilter::iter().map(|lvl| lvl.as_str().to_lowercase()).collect::<Vec<String>>().join("|");
+    let log_level_rgx_str = format!("^({levels})((,[^,=]+=({levels}))*)$");
+    let log_level_rgx = regex::Regex::new(&log_level_rgx_str)?;
+    let config_str = CONFIG.log_level().to_lowercase();
+
+    let (level, levels_override) = if let Some(caps) = log_level_rgx.captures(&config_str) {
+        let level = caps
+            .get(1)
+            .and_then(|m| log::LevelFilter::from_str(m.as_str()).ok())
+            .ok_or(Error::new("Failed to parse global log level".to_string(), ""))?;
+
+        let levels_override: Vec<(&str, log::LevelFilter)> = caps
+            .get(2)
+            .map(|m| {
+                m.as_str()
+                    .split(',')
+                    .collect::<Vec<&str>>()
+                    .into_iter()
+                    .flat_map(|s| match s.split('=').collect::<Vec<&str>>()[..] {
+                        [log, lvl_str] => log::LevelFilter::from_str(lvl_str).ok().map(|lvl| (log, lvl)),
+                        _ => None,
+                    })
+                    .collect()
+            })
+            .ok_or(Error::new("Failed to parse overrides".to_string(), ""))?;
+
+        (level, levels_override)
+    } else {
+        err!(format!("LOG_LEVEL should follow the format info,vaultwarden::api::icons=debug, invalid: {config_str}"))
+    };
+
     // Depending on the main log level we either want to disable or enable logging for hickory.
     // Else if there are timeouts it will clutter the logs since hickory uses warn for this.
     let hickory_level = if level >= log::LevelFilter::Debug {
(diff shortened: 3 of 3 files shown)


Timshel:
Any interest in the feature ?

BlackDex:
It sounds like a nice feature. The only issue i have is, moving all defaults to config also means that either someone has to add all the defaults them self, or we someone need to provide a default and allow them to be overridden via the settings.

I'm also not sure on the multi-line formatting of the config. I would then rather use the standard of Rust it self.
https://rust-lang-nursery.github.io/rust-cookbook/development_tools/debugging/config_log.html

That allows for a general log_level settings as first item (or key less item) and are all comma separated.
That would allow you to keep using the `LOG_LEVEL` env for example and that will not break other instances.
It would then look like `LOG_LEVEL=info,hyper::client=info,handlebars::render=debug` for example.
We could have all defaults in a HashMap, and add or overwrite this with the configured items.

What do you think?

I btw also think this is very useful during development and troubleshooting issues reported by users :).

Timshel:
I think it all depends on if it's used for the current defaults.

Thinking on it, I believe it might be acceptable if all defaults are set (in `config.rs`) and in `env.template`.
I would expect that if someone then start customizing then they would copy and paste from the template; and if they don't do it then it's not really complicated for them to replicate it by adding the lines that generate too many logs. Additionally, a warning could be added to the release notes.

If used this way then it means it would contain at least 10 values and that's why I decided to make it multiline because I believe it makes it way more legible than a comma separated list.

If it's decided not to use it for the defaults then I believe your solution is better since with less values the comma separated solution would make more sense and risk less conflict due to the multiline.

But since I wrote it initially ^^ I believe the multiline is better since it allows the user to easily override the defaults while keeping a very simple logic in the rust code.

Timshel:
I was a bit fast with my response, the `Hashmap` you suggest would keep things quite simple while allowing override.

But I think I still prefer the multiline and default included in `env.template` because it makes those default more accessible and visible, as opposed at having to check in the code.

BlackDex:
The new line has some strange behavior in some environments, that is why I'm a bit against that way. While dotenvy does it correctly now, it might be systemd or other OS environments make it more error prone.

Cargo.toml
```diff
@@ -128,7 +128,7 @@ hickory-resolver = "0.24.1"
 
 # Favicon extraction libraries
 html5gum = "0.5.7"
-regex = { version = "1.10.4", features = ["std", "perf", "unicode-perl"], default-features = false }
+regex = { version = "1.10.4", features = ["std", "perf", "unicode-case", "unicode-perl"], default-features = false }
```
BlackDex, line 131: Is the `unicode-case` feature really needed?
Isn't it better to parse the `LOG_LEVEL` config to a lowercase by default in the `config.rs` upon loading.
Also, since you seem to convert every part to lowercase anyway in the `init_logging` functon.
If you do it once, then it will save the repeating steps, and probably a faster regex too.

src/main.rs
```diff
@@ -210,7 +205,41 @@ fn launch_info() {
     );
 }
 
-fn init_logging(level: log::LevelFilter) -> Result<(), fern::InitError> {
+fn init_logging() -> Result<log::LevelFilter, Error> {
+    let levels = log::LevelFilter::iter().map(|lvl| lvl.as_str().to_lowercase()).collect::<Vec<String>>().join("|");
+    let log_level_rgx_str = format!("^({levels})((,[^,=]+=({levels}))*)$");
+    let log_leve_rgx = regex::RegexBuilder::new(&log_level_rgx_str).case_insensitive(true).build()?;
```
BlackDex, line 211: I think adding this regex here as a var is fine, unless it is loaded multiple times. If that is the case, then i would suggest to convert it to a lazy_static. If it is called just once during load, then no problem.

BlackDex (commented):
Looks nice :).
Just a few questions

src/main.rs
Timshel: Yes `init_logging()` is called only once in `main()`.

BlackDex: Cool, then loading the regex once is fine here :).

Cargo.toml
Timshel: The `unicode-case` is needed to use the `case_insensitive(true)` on the regex which will apply almost only on the `log_level` and I believe it's better to make it a bit more forgiving.

As for converting to lower case it's done only on the `log_level` part not the whole string. 
I believe module name in `rust` should always be snake case but still I think it's better not to risk transforming it.

And I just realized that the `LevelFilter::from_str` is in fact already case-insensitive, so they are not needed.

BlackDex: I disagree here. You convert the levels to lowercase your self to match on. All crate names need to be lowercase as kinda mentioned here: https://blog.rust-lang.org/2023/10/27/crates-io-non-canonical-downloads.html

So, being forgiving can also be done by fully converting the `LOG_LEVEL` string to a lowercase value and match on that.
I also checked several log crates on how they handle this, and they just split the string on `=` and passes the split[1] element as level, since the log crate checks it case-insensitive anyway.

Also, converting everything to lowercase also makes it more forgiving on the crate-names if someone uses a capital there.

Timshel:
Hey, any chance to merge this ?

BlackDex:
Ill see if i have some time shortly to review again

src/config.rs
```diff
@@ -564,7 +564,7 @@ make_config! {
         use_syslog:             bool,   false,  def,    false;
         /// Log file path
         log_file:               String, false,  option;
-        /// Log level
+        /// Log level: `info,path::to::module=warn`
```
BlackDex, line 567: This is the wrong format. It now looks like this in the admin interface.
image

It probably should be something like this (wording or grammar not check)
```rust
/// Log level |> Valid values are "trace", "debug", "info", "warn", "error" and "off"
/// For a specific module append it as a comma separated value "info,path::to::module=debug"
```

.env.template
```diff
@@ -362,9 +362,9 @@
 ## Log level
 ## Change the verbosity of the log output
 ## Valid values are "trace", "debug", "info", "warn", "error" and "off"
-## Setting it to "trace" or "debug" would also show logs for mounted
-## routes and static file, websocket and alive requests
-# LOG_LEVEL=info
+## Setting it to "trace" or "debug" would also show logs for mounted routes and static file, websocket and alive requests
+## For a specific module append a comma separated `path::to::module=log_level`
+# LOG_LEVEL=info,path::to::module=debug
```
BlackDex, line 367: We want the values in the template file to match with the default values mostly.
So probably best to keep it as `info`. But add a real working example maybe?
Something like this maybe?
```
## For example, to only see debug logs for icons use: LOG_LEVEL="info,vaultwarden::api::icons=debug"
```

src/config.rs
```diff
@@ -564,7 +564,7 @@ make_config! {
...
         log_level:              String, false,  def,    "Info".to_string();
```
BlackDex, line 568: Maybe the default can also be changed to a lowercase `info`. I do not think that will hurt anything.

src/main.rs
```diff
@@ -241,47 +268,64 @@ fn init_logging(level: log::LevelFilter) -> Result<(), fern::InitError> {
...
-        logger = logger.level_for("lettre::transport::smtp", log::LevelFilter::Off)
+    }
+
+    let mut logger = fern::Dispatch::new().level(level).chain(std::io::stdout());
+
+    let mut sorted_levels = default_levels.into_iter().collect::<Vec<(&str, log::LevelFilter)>>();
+    sorted_levels.sort_by_key(|a| a.0);
+
+    for (path, level) in sorted_levels {
+        logger = logger.level_for(path.to_string(), level);
```
BlackDex, line 328: I have not tested it, but why is this sorted? Is there any benefit doing this?

I do not think it will matter in which order you set them. Fern will try to match the most specific path entry anyway, doesn't matter if it is added at the top or bottom of the list.

BlackDex (changes requested):
Looks nice, and seems to work great :).

Just some small items to address and i have questions about.

BlackDex (changes requested):

src/main.rs
Timshel: Honestly it was simpler to sort than to test that the sort was not necessary :sweat_smile:.

src/config.rs
Timshel: I think it's a leftover of when I made it editable for a time.

src/main.rs
BlackDex: But sorting when not needed at all is useless of course.

BlackDex (dismissed):
Thanks, looks like a nice addition :)

src/main.rs
```diff
@@ -241,47 +267,63 @@ fn init_logging(level: log::LevelFilter) -> Result<(), fern::InitError> {
...
-        logger = logger.level_for("lettre::transport::smtp", log::LevelFilter::Off)
+    }
+
+    let mut logger = fern::Dispatch::new().level(level).chain(std::io::stdout());
+
+    let levels = default_levels.into_iter().collect::<Vec<(&str, log::LevelFilter)>>();
+
+    for (path, level) in levels {
+        logger = logger.level_for(path.to_string(), level);
     }
```
dani-garcia, line 327: I don't think this collect is necessary, you can just iterate over `default_levels` directly I think:
```suggestion
    for (path, level) in default_levels {
        logger = logger.level_for(path.to_string(), level);
    }
```

Timshel: O yes a leftover of when I removed the sort.

BlackDex (approved):

dani-garcia (approved):

PR #4309, opened 2024-02-02
Update GHA Workflows

### LINKED ISSUE #4226: Testing version differ from latest commit in diagnostic admin panel
### Subject of the issue
The diagnostic page is showing an update from `latest` commit (https://github.com/dani-garcia/vaultwarden/commit/ae3a153bdb062414402ec94901fabb1105838674), but this commit doesn't actually cause the CI to build a new testing image that I could update to.

I think the diagnostic panel is checking the latest git commit, instead of the latest CI build that is available.

### Deployment environment

* Vaultwarden version: v1.30.1-2c369937
* Web-vault version: v2023.12.0
* OS/Arch: linux/x86_64
* Running within Docker: false (Base: Not applicable)

### Steps to reproduce

Use the testing build.

### Expected behaviour

Diagnostic page should show there is no update available.

### Actual behaviour

There seems to be an update as indicated on the page, but there is actually none.

### Troubleshooting data

Actual behaviour

An image comparing the commit between deployment environment version versus the latest commit

BlackDex:
- Update the workflow GH Actions.
- Configured the release workflow to always run on main/tag as discussed in #4226

Closes #4226

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,20 +2,10 @@ name: Release
 
 on:
   push:
-    paths:
-      - ".github/workflows/release.yml"
-      - "src/**"
-      - "migrations/**"
-      - "docker/**"
-      - "Cargo.*"
-      - "build.rs"
-      - "diesel.toml"
-      - "rust-toolchain.toml"
-
-    branches: # Only on paths above
+    branches:
       - main
 
-    tags: # Always, regardless of paths above
+    tags:
       - '*'
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -2,16 +2,6 @@ name: Release
...
-      - ".github/workflows/release.yml"
-      - "src/**"
-      - "migrations/**"
-      - "docker/**"
-      - "Cargo.*"
-      - "build.rs"
-      - "diesel.toml"
-      - "rust-toolchain.toml"
-
     branches: # Only on paths above
```
stefan0xC, line 5: this comment could be removed too

dani-garcia (approved):

stefan0xC (approved):

PR #4315, opened 2024-02-03
Fix env templateto ensure compatibility with systemd's EnvironmentFile parsing

seiuneko:
This commit addresses an issue in the Vaultwarden project's `.env.template` file where comments explaining environment variable options were incorrectly placed on the same line as the variable definitions. Systemd's `EnvironmentFile` directive only recognizes comment symbols (#) at the beginning of a line. Therefore, explanatory comments placed on the same line as environment variable definitions can lead to misinterpretations or errors when the file is used in a systemd service file.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -477,13 +477,18 @@
 # SMTP_HOST=smtp.domain.tld
 # SMTP_FROM=vaultwarden@domain.tld
 # SMTP_FROM_NAME=Vaultwarden
-## ("starttls", "force_tls", "off") Enable a secure connection. Default is "starttls" (Explicit - ports 587 or 25), "force_tls" (Implicit - port 465) or "off", no encryption (port 25)
-# SMTP_SECURITY=starttls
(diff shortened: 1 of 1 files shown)


stefan0xC:
I'm not happy with just moving the comments on their own line. Can you improve the wording of the smtp security settings? Since the `SMTP_PORT` default depends on the `SMTP_SECURITY` setting this could be probably put more succinctly. 
https://github.com/dani-garcia/vaultwarden/blob/897bdf8343148ee4e4f1a937cbc501ad3f743b4a/src/config.rs#L656-L657

And regarding the `ROCKET_PORT` I'd say that we should change the default to `8000` and explain that we use port `80` in the container image.

seiuneko:
Thank you for your feedback. Please review the new commit.

BlackDex (approved):

dani-garcia (approved):

PR #4328, opened 2024-02-08
Update Rust, crates and web-vault

BlackDex:
- Updated Rust to v1.76.0
- Updated crates
- Updated web-vault to v2024.1.2b
- Fixed some Clippy lints
- Moved lint check configuration to Cargo.toml
- Fixed issue with Reset Password Enrollment when logged-in via device
- Added a `release-low` profile to build Vaultwarden on systems with low system resources

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ name = "vaultwarden"
 version = "1.0.0"
 authors = ["Daniel García <dani-garcia@users.noreply.github.com>"]
 edition = "2021"
-rust-version = "1.73.0"
+rust-version = "1.74.0"
 resolver = "2"
 
 repository = "https://github.com/dani-garcia/vaultwarden"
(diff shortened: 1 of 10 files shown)


Cargo.toml
```diff
@@ -190,3 +190,60 @@ strip = "symbols"
...
 codegen-units = 1
 panic = "abort"
+
+# Profile for systems with low resources
+# It will use less resources during build
+[profile.release-low]
+inherits = "release"
+strip = "symbols"
+lto = "thin"
+codegen-units = 1
```
dani-garcia, line 200: This should be already in `release` so we don't need to specify it again.

dani-garcia (commented):
LGTM, just a small thing on the new profile

Cargo.toml
BlackDex: the `codegen-units` you mean?
The others we have derived from release also have this.
My main reason for this is to make sure it keeps this and not changes when we change the main `release` profile for example.

dani-garcia: Yeah, that's a fair point

dani-garcia (approved):

PR #4355, opened 2024-02-15
Change timestamp data type.

### LINKED ISSUE #4177: Year 2038 problem
This is not a "vulnerability" due to the fact that the issue is 14 years away*; furthermore Vaultwarden may not even be a thing then. Anyway, I am not sure why [32-bit timestamps](https://github.com/dani-garcia/vaultwarden/blob/main/src/api/core/two_factor/authenticator.rs#L121) are used for the TOTP code considering most OSes running on 64-bit systems have been using 64-bit timestamps for a decade or more (even 32-bit Linux has support since 5.6), `chrono` uses `i64`s, and `totp-lite` uses `u64`s. The change to `u64`/`i64`s is easy in code—I've done so on my personal fork—and SQLite has no issues either since `INTEGER` is a flexible-width data type (up to 64 bits) and `DATETIME` has type affinity `NUMERIC` which will store values as `INTEGER` so long as they are valid 64-bit signed integers. I'm not sure what if anything would be needed to "fix" PostgreSQL and MariaDB/MySQL.

\* _Technically a contrived setup where servers and clients have their clocks in-sync but set to beyond 2038 are vulnerable._

gzfrozen:
Fix the "Year 2038 problem".

- Change the data type of `last_used` in `twofactor` table.
- Create migration files.
- Delete unnessary type convertion.
- Tested in SQlite, MySQL, Postgres.
  - Test method: add the two step authenticator to a account using the original code, then update the code to see if it can still login

I am not confident if it is the only "Year 2038 problem". If there are still other problems please let me know.

Fixes #4177

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -84,12 +84,8 @@
 ### WebSocket ###
 #################
 
-## Enables websocket notifications
-# WEBSOCKET_ENABLED=false
-
-## Controls the WebSocket server address and port
-# WEBSOCKET_ADDRESS=0.0.0.0
-# WEBSOCKET_PORT=3012
+## Enable websocket notifications
+# ENABLE_WEBSOCKET=true
 
 ##########################
 ### Push notifications ###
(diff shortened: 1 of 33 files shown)


BlackDex (approved):

dani-garcia (approved):
Thanks! Had to resolve a small merge conflict but LGTM!

PR #4364, opened 2024-02-18
Roll the refresh_token

Timshel:
Hey,

We recently discussed the `refresh_token` in the SSO PR and during this discussion it occurred to me that the long lived `refresh_token` could be rolled as is done by some SSO client.

This is handled / expected by the Bitwarden clients since the response `_refresh_login` contain the potentially updated [refresh_token](https://github.com/dani-garcia/vaultwarden/blob/bb2412d0339e1da5dee99fc566a2b2aab5d2808c/src/api/identity.rs#L122).

### DIFF
--- a/src/db/models/device.rs
+++ b/src/db/models/device.rs
@@ -1,4 +1,5 @@
 use chrono::{NaiveDateTime, Utc};
+use data_encoding::BASE64URL;
 
 use crate::{crypto, CONFIG};
 use core::fmt;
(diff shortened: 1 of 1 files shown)


BlackDex:
@Timshel, regarding the Rolling you just pushed.
Have you tried it in the web-vault?

You probably need to set the token/refresh validity lower, else you have to wait a long time hehe.

But i think it will break. It would be great of ot will not, but i have my doubts.

Timshel:
Yep, trick I found to trigger the refresh token without waiting are :

 - In `web` create a collection, it will trigger multiple `refresh_token`
 - In mobile it's more complicated, need to lower the `access_token` validity (like 6min) then a `Synchronisation` will trigger a refresh if the `access_token` is expired

BlackDex:
I have not had any time yet to take a good look at it.

Timshel:
Had issues with this, too many cases where the client call the refresh token endpoint in parallel.
So rolled it back out for now.

PR #4370, opened 2024-02-20
Making vaultwarden run in linux/386

amedeorizzo:
I was trying to make vaultwarden work on a 32-bit linux server of a friend. After some tries with the official image I decided to build a docker image specifically for linux/386 systems. These are the changes I made.
Builded image can be found here: https://hub.docker.com/r/amedeorizzo/vaultwarden386

### DIFF
--- a/docker/Dockerfile.debian
+++ b/docker/Dockerfile.debian
@@ -63,7 +63,8 @@ RUN apt-get update && \
         git \
         "libc6-$(xx-info debian-arch)-cross" \
         "libc6-dev-$(xx-info debian-arch)-cross" \
-        "linux-libc-dev-$(xx-info debian-arch)-cross" && \
+        "linux-libc-dev-$(xx-info debian-arch)-cross" \
+        "cpp-i686-linux-gnu" && \
     # Run xx-cargo early, since it sometimes seems to break when run at a later stage
     echo "export CARGO_TARGET=$(xx-cargo --print-target-triple)" >> /env-cargo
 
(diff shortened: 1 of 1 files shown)


docker/Dockerfile.debian
```diff
@@ -193,4 +194,4 @@ COPY --from=build /app/target/final/vaultwarden .
 
 HEALTHCHECK --interval=60s --timeout=10s CMD ["/healthcheck.sh"]
 
-CMD ["/start.sh"]
+CMD ["sh", "/start.sh"]
```
williamdes, line 197: is this change really required for 386 ?

amedeorizzo:
> This will totally break our other builds, so it will not be implemented like this in the current way.
> 
> I also have not seen any requests before to build for this architecture, so not sure if we should try to add support for it in our repo. I'm not fully against it, and i might be able to add this in a better way so you do not have to make these changes all the time.
> 
> But, it will be landing on the not fully supported list.

Yes, I'm aware that needs to be implemented in another way to not break the other builds. This could be an idea for the future.

docker/Dockerfile.debian
williamdes: Let's trim down this patch to get it merged

BlackDex: It won't be merged like this hehe.
Should be handled totally different.

But I'm fine with this for now to keep it as a reminder.

Ornias1993:
> This will totally break our other builds, so it will not be implemented like this in the current way.
> 
> I also have not seen any requests before to build for this architecture, so not sure if we should try to add support for it in our repo. I'm not fully against it, and i might be able to add this in a better way so you do not have to make these changes all the time.
> 
> But, it will be landing on the fully supported list.

I personally see no usecase for this, is there even any x86 non x86_64 hardware being released anymore?

BlackDex:
> > This will totally break our other builds, so it will not be implemented like this in the current way.
> > I also have not seen any requests before to build for this architecture, so not sure if we should try to add support for it in our repo. I'm not fully against it, and i might be able to add this in a better way so you do not have to make these changes all the time.
> > But, it will be landing on the fully supported list.
> 
> I personally see no usecase for this, is there even any x86 non x86_64 hardware being released anymore?

That doesn't matter of course, I personally also so no use-case for SSO for Vaultwarden, that doesn't mean it might be useful for others.

I'm not against adding a simple way to allow building it. We will just not give support for it. If it doesn't work, it just doesn't work. Same currently for MIPS, might be someone is using that already in some way, but we will not support it as one of the main architectures i think.

BlackDex:
i386/i686 is still supported by Debian until at least 2028

Ornias1993:
> i386/i686 is still supported by Debian until at least 2028

The important note on this, is that Debian has current customers/clients/users using it, because they offered it before.
I doubt they would add support for it in 2024 either, if it wasn't already there...

BlackDex:
Closing this already in favor of #4524

PR #3899, opened 2023-09-18
SSO using OpenID Connect

Timshel:
This is based on previous PR (#2787, #2449 and #3154) with work done by @pinpox, @m4w0lf, @Sheap, @bmunro-peralex, @tribut and others I probably missed sorry.

This PR add support for OpenId Connect to handle authentication to an external SSO.
This introduce another way to control who can use the vault without having to use invitation or an LDAP.

A master password is still required and not controlled by the SSO (depending on your point of view this might be a feature ;).  

Bitwarden [key connector](https://bitwarden.com/help/about-key-connector) is not supported and due to the [license](https://github.com/bitwarden/key-connector/blob/main/LICENSE.txt) it's highly unlikely that it will ever be:

> 2.1 Commercial Module License. Subject to Your compliance with this Agreement, Bitwarden hereby grants to You a limited, non-exclusive, non-transferable, royalty-free license to use the Commercial Modules for the sole purposes of internal development and internal testing, and only in a non-production environment.

## Usage 

This should be agnostic to the SSO used as long as it supports client secret authentication and expose an OpenID Connect Discovery endpoint. (I'm testing it with Keycloak at the moment, a demo test stack is available [README.md](https://github.com/Timshel/vaultwarden/blob/sso-support/test/oidc/README.md))

Added some documentation at the root of the project [SSO.md](https://github.com/Timshel/vaultwarden/blob/sso-support/SSO.md) that could be later moved to the wiki.

I made some additional modification in my [main branch](https://github.com/Timshel/vaultwarden/blob/main/README.md) to allow for easier testing (modified Docker image to use prebuilt patched front-end).

On front-end modification, I made patched versions available at [Timshel/oidc_web_builds](https://github.com/Timshel/oidc_web_builds). Two versions are available : 
 
 - One contains the change expected to be merged (named `button`); all change needs to be compatible with the non-sso version.
 - Second one set `#sso` as the default redirect url.

## Issues

As mentioned in the previous PR one of the main issue is the inability for the organization invitation to work with the SSO redirection. To fix it a [patch](https://github.com/Timshel/oidc_web_builds/blob/master/oidc_invite.patch) to the front-end is needed.

## **:warning::warning: :warning:  If you have issues or need help testing the PR  :warning: :warning:  :warning:**
Please open issues in [Timshel/vaultwarden](https://github.com/Timshel/vaultwarden/issues) in order to keep the discussion here focused on merging this work. 
Of course if you believe your issue is important mention this PR so a reference will be visible.

[...]

PR #4371, opened 2024-02-20
JWT Refresh Token

Timshel:
To facilitate  review decided to move out the switch to a JWT refresh_token from the sso PR https://github.com/dani-garcia/vaultwarden/pull/3899.

Without the SSO logic it's not the most useful still :
 - Add an expiration on the `refresh_token` (work like an idle timer reset when a new access_token is generated).
 - Store the information of the `AuthMethod` in the token (`Password` ...).

### DIFF
--- a/src/api/identity.rs
+++ b/src/api/identity.rs
@@ -3,6 +3,7 @@ use num_traits::FromPrimitive;
 use rocket::serde::json::Json;
 use rocket::{
     form::{Form, FromForm},
+    http::Status,
     Route,
 };
 use serde_json::Value;
@@ -17,7 +18,8 @@ use crate::{
         push::register_push_device,
         ApiResult, EmptyResult, JsonResult,
     },
-    auth::{generate_organization_api_key_login_claims, ClientHeaders, ClientIp},
+    auth,
+    auth::{generate_organization_api_key_login_claims, AuthMethod, AuthMethodScope, ClientHeaders, ClientIp},
     db::{models::*, DbConn},
     error::MapResult,
     mail, util, CONFIG,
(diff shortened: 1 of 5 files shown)


Timshel:
When discussed with @BlackDex we thought that 7 days might be an appropriate value for the `refresh_token` expiration.

Had some feedback on it, and it might be a bit short, since many actions in the application will not trigger a refresh (reading password is done as "offline"`).

Might make sense to raise it to something around 30 days as mentioned in this [documentation](https://bitwarden.com/blog/configuring-bitwarden-clients-for-offline-access/#staying-logged-in-to-browser-extension-desktop-and-mobile-clients).

BlackDex:
I agree, lets keep it the same as Bitwarden, and 7 days is a bit short indeed.
Also, how is this linked to the refresh roll PR? Could that potentially break offline logins too?

Timshel:
> Also, how is this linked to the refresh roll PR? 

This is a superset of the refresh roll PR, the same logic of changing the device token is used to prevent reusing twice the same `refresh_token`.

> Could that potentially break offline logins too?

I'm unsure what are the offline logins ?

BlackDex:
> > Also, how is this linked to the refresh roll PR?
> 
> This is a superset of the refresh roll PR, the same logic of changing the device token is used to prevent reusing twice the same `refresh_token`.
> 
> > Could that potentially break offline logins too?
> 
> I'm unsure what are the offline logins ?

I mean offline unlock in some way. But i doubt it.

Timshel:
A yes when unlocking I don't believe any call is done to the `refresh` endpoint, so there should be no issue/change.

Timshel:
Removed the rolling of the device token since it caused issues when the web client called the refresh token endpoint in parallel.

dfunkt:
Maybe something can be done about the logging, I think it's a little bit too verbose. (this happened when I was logging in after starting Vaultwarden with this PR applied)
```
vaultwarden  | [2024-09-01 23:46:02.959][vaultwarden::auth][ERROR] Token is invalid
vaultwarden  | [2024-09-01 23:46:02.962][vaultwarden::auth][ERROR] Impossible to read refresh_token: {"error":"","errorModel":{"message":"Token is invalid","object":"error"},"error_description":"","exceptionMessage":null,"exceptionStackTrace":null,"innerExceptionMessage":null,"message":"Token is invalid","object":"error","validationErrors":{"":["Token is invalid"]}}
vaultwarden  | [2024-09-01 23:46:02.962][vaultwarden::api::identity][ERROR] {"error":"","errorModel":{"message":"Impossible to read refresh_token: {\"error\":\"\",\"errorModel\":{\"message\":\"Token is invalid\",\"object\":\"error\"},\"error_description\":\"\",\"exceptionMessage\":null,\"exceptionStackTrace\":null,\"innerExceptionMessage\":null,\"message\":\"Token is invalid\",\"object\":\"error\",\"validationErrors\":{\"\":[\"Token is invalid\"]}}","object":"error"},"error_description":"","exceptionMessage":null,"exceptionStackTrace":null,"innerExceptionMessage":null,"message":"Impossible to read refresh_token: {\"error\":\"\",\"errorModel\":{\"message\":\"Token is invalid\",\"object\":\"error\"},\"error_description\":\"\",\"exceptionMessage\":null,\"exceptionStackTrace\":null,\"innerExceptionMessage\":null,\"message\":\"Token is invalid\",\"object\":\"error\",\"validationErrors\":{\"\":[\"Token is invalid\"]}}","object":"error","validationErrors":{"":["Impossible to read refresh_token: {\"error\":\"\",\"errorModel\":{\"message\":\"Token is invalid\",\"object\":\"error\"},\"error_description\":\"\",\"exceptionMessage\":null,\"exceptionStackTrace\":null,\"innerExceptionMessage\":null,\"message\":\"Token is invalid\",\"object\":\"error\",\"validationErrors\":{\"\":[\"Token is invalid\"]}}"]}}
```

src/api/identity.rs
```diff
@@ -96,43 +98,43 @@ async fn login(data: Form<ConnectData>, client_header: ClientHeaders, mut conn:
...
-        "KdfMemory": user.client_kdf_memory,
-        "KdfParallelism": user.client_kdf_parallelism,
-        "ResetMasterPassword": false, // TODO: according to official server seems something like: user.password_hash.is_empty(), but would need testing
-        "scope": scope,
-        "unofficialServer": true,
-    });
-
-    Ok(Json(result))
+    match auth::refresh_tokens(&refresh_token, conn).await {
+        Err(err) => err_code!(err.to_string(), Status::Unauthorized.code),
```
dfunkt, line 113:
```suggestion
        Err(err) => err_code!("Unable to refresh login credentials", Status::Unauthorized.code),
```

src/auth.rs
```diff
@@ -900,3 +968,132 @@ impl<'r> FromRequest<'r> for WsAccessTokenHeader {
...
+            access_claims,
+        }
+    }
+}
+
+pub async fn refresh_tokens(refresh_token: &str, conn: &mut DbConn) -> ApiResult<(Device, User, AuthTokens)> {
+    let time_now = Utc::now();
+
+    let refresh_claims = match decode_refresh(refresh_token) {
+        Err(err) => err!(format!("Impossible to read refresh_token: {err}")),
```
dfunkt, line 1071:
```suggestion
        Err(err) => err!(format!("Impossible to read refresh_token")),
```

src/auth.rs
```diff
@@ -900,3 +968,132 @@ impl<'r> FromRequest<'r> for WsAccessTokenHeader {
...
+        Some(user) => user,
+    };
+
+    if refresh_claims.exp < time_now.timestamp() {
+        err!("Expired refresh token");
+    }
+
+    let auth_tokens = match refresh_claims.sub {
+        AuthMethod::Password => AuthTokens::new(&device, &user, refresh_claims.sub),
+        _ => err!("Invalid auth method cannot refresh token"),
```
dfunkt, line 1095:
```suggestion
        _ => err!("Invalid auth method, cannot refresh token"),
```

dfunkt (commented):
Made a few suggestions, I think the end result would look decent:
```
[vaultwarden::auth][ERROR] Token is invalid
[vaultwarden::auth][ERROR] Impossible to read refresh_token
[vaultwarden::api::identity][ERROR] Unable to refresh login credentials
```

Timshel:
Made the change differently with the following logic:

- I believe the root error should be logged so no change here.
- I don't think it makes sense to log again an error when just rewrapping it so switched to `err_silent`.
- When rewrapping an error I want to be able to keep the original error message but the current `to_string` is way too verbose, so added a `fn message(): &str` to `Error`.

With those change we end up with the following logs:

```log
> curl 'http://127.0.0.1:8000/identity/connect/token' -X POST --data-raw 'grant_type=refresh_token&refresh_token=toto'

[2024-09-03 14:25:59.160][request][INFO] POST /identity/connect/token
[2024-09-03 14:25:59.161][vaultwarden::auth][ERROR] Token is invalid
[2024-09-03 14:25:59.161][vaultwarden::api::identity][ERROR] Unable to refresh login credentials: Impossible to read refresh_token: Token is invalid
[2024-09-03 14:25:59.161][response][INFO] (login) POST /identity/connect/token => 401 Unauthorized
````

PR #4385, opened 2024-02-27
Finer SMTP TLS certificate control

JosefSchoenberger:
This PR adds two new SMTP configurations regarding TLS:
- `SMTP_ADDITIONAL_ROOT_CERTS` allows an admin to add new root certificates that Vaultwarden accepts for the SMTP server. This can be useful if the SMTP server only offers a self-signed certificate. Example: `SMTP_ADDITIONAL_ROOT_CERTS=/etc/ssl/certs/mail1.pem;/etc/ssl/certs/mail2.pem`
- `SMTP_USE_SYSTEM_ROOT_CERTS` disables the system's root certificate store for TLS. This can be used in combination with `SMTP_ADDITIONAL_ROOT_CERTS` for certificate pinning.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -528,6 +528,9 @@
 ## Paths to PEM files, separated by semicolons
 # SMTP_ADDITIONAL_ROOT_CERTS=
 
+## Use system root certificate store for TLS host verification
+# SMTP_USE_SYSTEM_ROOT_CERTS=true
+
 ##########################
 ### Rocket settings ###
 ##########################
--- a/.env.template
+++ b/.env.template
@@ -524,6 +524,10 @@
 ## Only use this as a last resort if you are not able to use a valid certificate.
 # SMTP_ACCEPT_INVALID_HOSTNAMES=false
 
+## Accept additional root certs
+## Paths to PEM files, separated by semicolons
+# SMTP_ADDITIONAL_ROOT_CERTS=
+
 ##########################
 ### Rocket settings ###
 ##########################
--- a/src/mail.rs
+++ b/src/mail.rs
@@ -7,7 +7,7 @@ use percent_encoding::{percent_encode, NON_ALPHANUMERIC};
 use lettre::{
     message::{Attachment, Body, Mailbox, Message, MultiPart, SinglePart},
     transport::smtp::authentication::{Credentials, Mechanism as SmtpAuthMechanism},
-    transport::smtp::client::{Certificate, Tls, TlsParameters},
+    transport::smtp::client::{Certificate, CertificateStore, Tls, TlsParameters},
     transport::smtp::extension::ClientId,
     Address, AsyncSendmailTransport, AsyncSmtpTransport, AsyncTransport, Tokio1Executor,
 };
@@ -66,6 +66,9 @@ fn smtp_transport() -> AsyncSmtpTransport<Tokio1Executor> {
             for cert in certs {
                 tls_parameters = tls_parameters.add_root_certificate(cert.clone());
             }
+        }
+        if !CONFIG.smtp_use_system_root_certs() {
+            tls_parameters = tls_parameters.certificate_store(CertificateStore::None);
         }
         let tls_parameters = tls_parameters.build().unwrap();
 
--- a/src/mail.rs
+++ b/src/mail.rs
@@ -1,12 +1,13 @@
 use std::str::FromStr;
 
 use chrono::NaiveDateTime;
+use once_cell::sync::Lazy;
 use percent_encoding::{percent_encode, NON_ALPHANUMERIC};
 
 use lettre::{
     message::{Attachment, Body, Mailbox, Message, MultiPart, SinglePart},
     transport::smtp::authentication::{Credentials, Mechanism as SmtpAuthMechanism},
-    transport::smtp::client::{Tls, TlsParameters},
+    transport::smtp::client::{Certificate, Tls, TlsParameters},
     transport::smtp::extension::ClientId,
     Address, AsyncSendmailTransport, AsyncSmtpTransport, AsyncTransport, Tokio1Executor,
 };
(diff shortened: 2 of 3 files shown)


.env.template
```diff
@@ -525,6 +525,13 @@
 ## Only use this as a last resort if you are not able to use a valid certificate.
 # SMTP_ACCEPT_INVALID_HOSTNAMES=false
 
+## Accept additional root certs
+## Paths to PEM files, separated by semicolons
+# SMTP_ADDITIONAL_ROOT_CERTS=
+
+## Use system root certificate store for TLS host verification
+# SMTP_USE_SYSTEM_ROOT_CERTS=true
```
BlackDex, line 533: Is the addition of `ROOT` really needed for these environment variables?
Or can id do without. Looks a bit nicer without it I think. But if it really explains it better then maybe it should stay like this.

BlackDex (changes requested):
Looks fine by me. This would even work if we want to switch to `rustls` instead of native.
Just a few comments.

src/mail.rs
```diff
@@ -29,6 +30,21 @@ fn sendmail_transport() -> AsyncSendmailTransport<Tokio1Executor> {
...
+            .split(';')
+            .filter(|path| !path.is_empty())
+            .map(|path| {
+                let cert = std::fs::read(path)
+                    .unwrap_or_else(|e| panic!("Error loading additional SMTP root certificate file {path}.\n{e}"));
+                Certificate::from_pem(&cert).unwrap_or_else(|e| panic!("Error decoding certificate file {path}.\n{e}"))
+            })
+            .collect(),
+    )
+});
```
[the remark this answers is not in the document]
JosefSchoenberger: I agree that panicing only upon sending the first mail is not appropriate.

I'm not sure if I understand you correctly. I'm not familiar with Vaultwarden's codebase, but how should I proceed? I currently see three different options. Which do you prefer? Do you have something else in mind?

1. `SMTP_ADDITIONAL_ROOT_CERTS` remains remain a static global variable, just that it is moved into `config.rs` and that `verify_config` reads it first, such that the `Lazy` evaluates instantly upon config verification. I dislike this option since (a) it uses a global variable and (b) any calls to `update_config` never actually reload the certificate (something that I missed in this pull request, too).
1. I can also move the reading and parsing into `verify_config` and store the resulting certificate list in an internal `Option` config item upon success. That's fine considering that `SMTP_ADDITIONAL_ROOT_CERTS` is an `Option` anyway. The only disadvantage is that the name `verify_config` becomes misleading as a verification function typically has no side effects.
1. I can also create a new internal `gen` config option of type `Result<Option<Vec<Certificate>>, Box<dyn Error>>` in which the generating function read and parse the certificates. Any errors that occur there are then stored as an `Box<dyn Error>` which is then caught in `validate_config`. The disadvantage of this approach is that the config is of type `Result`, yet it is never `Err` during runtime, so any uses of this config option will always need an additional call to `unwrap` first.

What is your opinion on that matter? Am I missing something totally obvious?

.env.template
JosefSchoenberger: I don't really have a strong opinion on that matter.

The advantage of keeping it that way is that any user instantly knows that one does not have to provide the used certificate directly – any parent signing certificate is fine, too. In other words, it's clearer that this provides a trust anchor and any subsequent certificates are also accepted.
I admit that this is only a minor advantage. However, I don't really see the advantage of shortening that config name, either.

In the end I think this is a question of taste. I personally would keep the `ROOT` part in there, but I leave that decision up to you.

BlackDex: Sounds reasonable. Let's keep it.

BlackDex:
@JosefSchoenberger any interest in following up on the questions? Else i tend to close this PR as it also is possible to add/overwrite allowed certificates as a whole within a container for example.

If you want to continue, then i think it would even make sense to add this same kind of functionality to the `reqwest` part of the code, and not only link it to SMTP and make it more general.

This could help with company wide transparent proxies maybe. But again, someone can also add custom certs and use the `vaultwarden.sh.d` startup functionality to add custom CA's to the global CA Store of the container before Vaultwarden actually starts.

JosefSchoenberger:
Yes, I do. I thought I ended this conversation with a question, hence no further work from me. Anyway, I'm trying the third option now, but there might by some difficulties with complex types in the configuration.

> Else i tend to close this PR as it also is possible to add/overwrite allowed certificates as a whole within a container for example.

I think its still better to have these configuration options, setting it up that way is far easier (just setting one or two envs and ro-bind-mounting the certificate file into the container – both works fine in a compose file, for example).

> If you want to continue, then i think it would even make sense to add this same kind of functionality to the `reqwest` part of the code, and not only link it to SMTP and make it more general.

I don't have a lot of time right now. For now, I'd just like to finish the SMTP part. They should have separate configuration variables (the certificates might differ between the mail server and other servers) and the internal setup for `reqwest` is a bit different, so that works fine as a separate PR, IMHO.

JosefSchoenberger:
Alright. Option 3 did not work, as lettre's `Certificate` is not (de)serializable. Anyway, I went with Option 2 now. Feel free to review the first commit again. The second commit is identical to the previous version.

Neustradamus:
@JosefSchoenberger: Any progress on your PR?

JosefSchoenberger:
@Neustradamus The current state is ready for a review, as far as I'm concerned.

Neustradamus:
@JosefSchoenberger: Good, thanks :)

@BlackDex: Can you look it?

BlackDex:
I'm looking at this again, and I wonder why not just use a global default like `SSL_CERT_DIR`?
Setting that environment variable to a directory which holds the `mail2.pem` will work just fine. Also for Reqwest without any issue.

The only problem currently with this, is that rustls currently doesn't load certificates with any name, and only does that by hashed name. That can be solved by running `openssl rehash` inside the `SSL_CERT_DIR` directory though.

BlackDex:
So, the newer versions of Vaultwarden currently in `testing` allow for certs placed in `SSL_CERT_DIR` without the need to rehash. In order to not make everything more complex code-wise, and have the same functionality count for both mail and https calls, I'm going to close this PR in favor of placing the needed certs into a specific directory for all SSL/TLS traffic.

These kind of tweaks only makes it difficult to troubleshoot issues in my opinion.

Even though, thanks for the contribution!

PR #4422, opened 2024-03-13
chore: remove repetitive words

one230six:


stefan0xC:
The `datatables.js` is a third party library. Cf. https://github.com/dani-garcia/vaultwarden/blob/d0c22b9fc90ae52aedbbccdafe9915bd0bdb0d84/src/static/scripts/datatables.js#L2-L3 So to fix the comments in that file you should probably file a PR [upstream](https://github.com/DataTables/DataTablesSrc) otherwise the change will be undone when we next update the file.

dani-garcia (approved):
Thanks!

PR #4436, opened 2024-03-17
Update chrono and sqlite

BlackDex:
- Updated sqlite crate
- Updated chrono crate

The latter needed a lot of changes done, mostly `Duration` to `TimeDelta`. And some changes on how to use Naive.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -533,9 +533,9 @@ checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
 name = "chrono"
-version = "0.4.34"
+version = "0.4.35"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5bc015644b92d5890fab7489e49d21f879d5c990186827d42ec511919404f38b"
+checksum = "8eaf5903dcbc0a39312feb77df2ff4c76387d591b9fc7b04a238dcf8bb62639a"
 dependencies = [
  "android-tzdata",
  "iana-time-zone",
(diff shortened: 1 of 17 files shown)


tessus:
I find it rather brutal that chrono had breaking changes in a patch release.
Additionally, now there are unwrap calls where there haven't been before.

I know that this is not your fault. I am just expressing my opinion towards chrono.

dani-garcia:
They aren't breaking changes just yet, just deprecation notices. The deprecated functions will be removed in chrono 0.5

tessus:
@dani-garcia Ah, I see. Thanks for the info.

Btw, there are 3 more "approved" PRs. Sorry for using this PR. Is there a better way to remind you of approved PRs? The problem is that they will have to be maintained as well during the time they are open and approved. I am really sorry to bother you with this, but I am talking from my own experience. If there are too many open PRs, it can be quite overwhelming...

dani-garcia:
Hey @tessus, it's absolutely not a bother at all, please feel free to reach out here or at matrix any time. I haven't had much time lately to actually review PRs, but I've gone through a few of them today and hopefully I'll have some time to check the rest soon.

tessus:
Thanks again for the reply @dani-garcia. I also know from experience that sometimes it's hard to invest time in reviewing PRs - just because there aren't enough hours in a day. ;-) We all have to work, and after work it's also nice to disconnect. I certainly do not want to annoy and pressure you with my asking to merge PRs.

I'm unfortunately still not an expert in Rust, otherwise I could help out more.

dani-garcia (approved):
The previous impl already panicked on invalid values, so the fact that we're adding `.unwrap()`s all over the place pretty much just maintains the status-quo, but I'd love for us to handle those errors better.

Maybe when chrono 0.5 is released we can look over the new APIs and improve the situation.

BlackDex:
It's indeed a shift in where it would panic now. But in a lot of cases we use a constant value to get the duration from. Those will never be out of bounds of course, creating a whole error catching around those seem extra work too.

In would like to get rid of all `unwraps()`, but it shouldn't be at the cost of everything.

And indeed, maybe they change some stuff again in 0.5.
Maybe we need to create a wrapper around it, or if we have common values make those constants.

BlackDex:
I took a quick look at the `0.5` version, and it seems they remove the `try_` and `_opt` versions and removed the need for the `unwrap()` if i look at there test code. Not sure though, but we will see when it will be released.

PR #4445, opened 2024-03-22
Update Rust and crates

BlackDex:
- Updated Rust to v1.77.0
- Updated several crates
  The `reqwest` update included `trust-dns` > `hickory-dns` changes.
  Also, `reqwest` v0.12 is not working correctly for us, that is something to investigate.
- Fixed a new clippy warning

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ name = "vaultwarden"
 version = "1.0.0"
 authors = ["Daniel García <dani-garcia@users.noreply.github.com>"]
 edition = "2021"
-rust-version = "1.74.0"
+rust-version = "1.75.0"
 resolver = "2"
 
 repository = "https://github.com/dani-garcia/vaultwarden"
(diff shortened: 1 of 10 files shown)


dani-garcia (approved):
LGTM, what's the issue with reqwest 0.12?

Cargo.toml
```diff
@@ -123,7 +123,7 @@ email_address = "0.2.4"
 handlebars = { version = "5.1.0", features = ["dir_source"] }
 
 # HTTP client (Used for favicons, version check, DUO and HIBP API)
-reqwest = { version = "0.11.26", features = ["stream", "json", "gzip", "brotli", "socks", "cookies", "trust-dns", "native-tls-alpn"] }
+reqwest = { version = "0.11.27", features = ["default-tls", "native-tls-alpn", "stream", "json", "gzip", "brotli", "socks", "cookies", "hickory-dns"], default-features = false}
```
dani-garcia, line 126: I don't think default-tls needs to be specified, as it already is enabled by native-tls-alpn, but no big deal either way:

https://github.com/seanmonstar/reqwest/blob/cf69fd4bfe22855d576497eb94e9eb549e742475/Cargo.toml#L37-L39

BlackDex: Ah, yes, that is true indeed.

BlackDex:
> LGTM, what's the issue with reqwest 0.12?

It doesn't seem to download almost any icon.
I suspect http2 maybe, or some other changes.

Also, other crates we use still use reqwest 0.11.x, so we would introduce duplicate dependencies/crates.

PR #4446, opened 2024-03-22
Update Key Rotation web-vault v2024.3.x

BlackDex:
Key rotation was changed since 2024.1.x.
Multiple other items were added to be rotated like password-reset and emergency-access data to be part of just one POST instead of having multiple.

See: https://github.com/dani-garcia/bw_web_builds/pull/157

### DIFF
--- a/src/api/core/accounts.rs
+++ b/src/api/core/accounts.rs
@@ -438,24 +438,46 @@ async fn post_kdf(data: JsonUpcase<ChangeKdfData>, headers: Headers, mut conn: D
 #[derive(Deserialize)]
 #[allow(non_snake_case)]
 struct UpdateFolderData {
-    Id: String,
+    // There is a bug in 2024.3.x which adds a `null` item.
+    // To bypass this we allow a Option here, but skip it during the updates
+    // See: https://github.com/bitwarden/clients/issues/8453
+    Id: Option<String>,
     Name: String,
 }
 
+#[derive(Deserialize)]
+#[allow(non_snake_case)]
+struct UpdateEmergencyAccessData {
+    Id: String,
+    KeyEncrypted: String,
+}
+
+#[derive(Deserialize)]
+#[allow(non_snake_case)]
+struct UpdateResetPasswordData {
+    OrganizationId: String,
+    ResetPasswordKey: String,
+}
+
 use super::ciphers::CipherData;
+use super::sends::{update_send_from_data, SendData};
 
 #[derive(Deserialize)]
 #[allow(non_snake_case)]
 struct KeyData {
     Ciphers: Vec<CipherData>,
     Folders: Vec<UpdateFolderData>,
+    Sends: Vec<SendData>,
+    EmergencyAccessKeys: Vec<UpdateEmergencyAccessData>,
+    ResetPasswordKeys: Vec<UpdateResetPasswordData>,
     Key: String,
-    PrivateKey: String,
     MasterPasswordHash: String,
+    PrivateKey: String,
 }
 
 #[post("/accounts/key", data = "<data>")]
 async fn post_rotatekey(data: JsonUpcase<KeyData>, headers: Headers, mut conn: DbConn, nt: Notify<'_>) -> EmptyResult {
+    // TODO: See if we can wrap everything within a SQL Transaction. If something fails it should revert everything.
     let data: KeyData = data.into_inner().data;
 
     if !headers.user.check_valid_password(&data.MasterPasswordHash) {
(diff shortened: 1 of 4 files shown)


BlackDex:
There seems to be a bug in 2024.3.0 (and probably the latest 2024.2.x also).

There is a phantom folder without an `id`, that causes an issue. This also happens on a Self-Hosted Bitwarden.

stefan0xC:
> There seems to be a bug in 2024.3.0 (and probably the latest 2024.2.x also).
> 
> There is a phantom folder without an `id`, that causes an issue. This also happens on a Self-Hosted Bitwarden.

What do you mean?

BlackDex:
> > There seems to be a bug in 2024.3.0 (and probably the latest 2024.2.x also).
> > There is a phantom folder without an `id`, that causes an issue. This also happens on a Self-Hosted Bitwarden.
> 
> What do you mean?

That during key rotation there is a folder in the folders array without an id, but with a name. Which isn't allowed.
It also breaks on there self-hosted instance, and probably also on the cloud environment. I have not yet tested that.

stefan0xC:
Yep. Looking into it I think that it seems like the empty folder is sent by the client as it is added to the folderViews but not filtered out anymore (since https://github.com/bitwarden/clients/pull/6881)
Cf. https://github.com/bitwarden/clients/blob/web-v2023.12.0/apps/web/src/app/auth/settings/change-password.component.ts#L235-L242 vs. https://github.com/bitwarden/clients/blob/web-v2024.3.0/apps/web/src/app/auth/key-rotation/user-key-rotation.service.ts#L118-L130

stefan0xC:
A workaround would be that we change https://github.com/dani-garcia/vaultwarden/blob/2d98aa304501b0f710d3a97ec854acdfc115228f/src/api/core/accounts.rs#L441 to `Option<String>` and skip the processing of it accordingly https://github.com/dani-garcia/vaultwarden/blob/2d98aa304501b0f710d3a97ec854acdfc115228f/src/api/core/accounts.rs#L474-L475

BlackDex:
> A workaround would be that we change
> 
> https://github.com/dani-garcia/vaultwarden/blob/2d98aa304501b0f710d3a97ec854acdfc115228f/src/api/core/accounts.rs#L440-L441
> to `Option<String>` and skip the processing of it accordingly
> 
> https://github.com/dani-garcia/vaultwarden/blob/2d98aa304501b0f710d3a97ec854acdfc115228f/src/api/core/accounts.rs#L474-L475

I'm working on that indeed. But bumping into other strange issues.
Like they started sending org ciphers too, which we probably do not want to update i think.
So we need to skip those too.

stefan0xC:
>  It also breaks on there Self-Hosted images. Strangely it doesn't on the cloud, even with the same versions.

Not familiar with bitwarden self-hosted. Might be related to whether you have enabled the `key-rotation-improvements` feature flag? At least [the validation seems to differ when it's enabled](https://github.com/bitwarden/server/blob/6a0f6e1dac85f1940226735bf7c5907189c2ef9b/src/Api/Auth/Controllers/AccountsController.cs#L442).

BlackDex:
> > It also breaks on there Self-Hosted images. Strangely it doesn't on the cloud, even with the same versions.
> 
> Not familiar with bitwarden self-hosted. Might be related to whether you have enabled the `key-rotation-improvements` feature flag? At least [the validation seems to differ when it's enabled](https://github.com/bitwarden/server/blob/6a0f6e1dac85f1940226735bf7c5907189c2ef9b/src/Api/Auth/Controllers/AccountsController.cs#L442).

Indeed, that is enabled on Cloud, but not on the Self-Hosted Bitwarden, that might be the difference indeed.

BlackDex:
Ah, that flag does make it better for us. But it doesn't change the layout.
I think it already sends it in this format because of a version check maybe?

Not sure since we had 2023.9.1 there, and i just now changed it to 2024.2.0, but that didn't make a difference really.
The flag does. It prevents some other calls to be done afterwards :) which are handled already with the new format.

BlackDex:
Hmm. Maybe the org ciphers are needed for the flexible collections feature maybe?
Not sure how those work actually.

Gerardv514:
Curious if the title of this Pr is a type, should v2023.3 be v2024.3?

BlackDex:
Technically this PR works. But i want to check for some more optimizations.

Gerardv514 (commented):
Change title on commit to 2024.3 as well?

src/api/core/accounts.rs
```diff
@@ -438,24 +438,46 @@ async fn post_kdf(data: JsonUpcase<ChangeKdfData>, headers: Headers, mut conn: D
...
+    ResetPasswordKeys: Vec<UpdateResetPasswordData>,
     Key: String,
-    PrivateKey: String,
     MasterPasswordHash: String,
+    PrivateKey: String,
 }
 
 #[post("/accounts/key", data = "<data>")]
 async fn post_rotatekey(data: JsonUpcase<KeyData>, headers: Headers, mut conn: DbConn, nt: Notify<'_>) -> EmptyResult {
+    // TODO: See if we can wrap everything within a SQL Transaction. If something fails it should reverte everything.
```
Gerardv514, line 480: spelling on revert

Gerardv514 (changes requested):

dani-garcia (approved):

PR #4524, opened 2024-04-27
Add extra (unsupported) container build arch's

BlackDex:
There was a PR (#4370) to add i686/i386 support for Vaultwarden. That specific PR was not a viable way of adding this.

This PR adds extra architectures for Debian based containers which we will not support by default. Those images will not be build and pushed to our container registries.

Added the following architectures:
 - linux/386
 - linux/ppc64le
 - linux/s390x

Again, there will be no major support for these architectures, but it will allow people who use these architectures to build a Debian based binary more easily

Closes #4370

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -11,6 +11,11 @@ With just these two files we can build both Debian and Alpine images for the fol
  - armv7 (linux/arm/v7)
  - armv6 (linux/arm/v6)
 
+Some unsupported platforms for Debian based images. These are not built and tested by default and are only provided to make it easier for users to build for these architectures.
+- 386     (linux/386)
+- ppc64le (linux/ppc64le)
+- s390x   (linux/s390x)
+
 To build these containers you need to enable QEMU binfmt support to be able to run/emulate architectures which are different then your host.<br>
 This ensures the container build process can run binaries from other architectures.<br>
 
--- a/docker/docker-bake.hcl
+++ b/docker/docker-bake.hcl
@@ -125,6 +125,40 @@ target "debian-armv6" {
   tags = generate_tags("", "-armv6")
 }
 
+// ==== Start of unsupported Debian architecture targets ===
+// These are provided just to help users build for these rare platforms
+// They will not be built by default
+target "debian-386" {
+  inherits = ["debian"]
+  platforms = ["linux/386"]
+  tags = generate_tags("", "-386")
+  args = {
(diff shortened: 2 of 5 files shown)


docker/README.md
```diff
@@ -11,6 +11,11 @@ With just these two files we can build both Debian and Alpine images for the fol
  - armv7 (linux/arm/v7)
  - armv6 (linux/arm/v6)
 
+Some unsupported platforms for Debian based images. These are not build and tested by default and are only provided to make it easier for users to build for these architectures.
```
dani-garcia, line 14:
```suggestion
Some unsupported platforms for Debian based images. These are not built and tested by default and are only provided to make it easier for users to build for these architectures.
```

docker/docker-bake.hcl
```diff
@@ -125,6 +125,40 @@ target "debian-armv6" {
   tags = generate_tags("", "-armv6")
 }
 
+// ==== Start of unsupported Debian architecture targets ===
+// These are provided just to help users build for these rare platforms
+// This will not be build by default
```
dani-garcia, line 130:
```suggestion
// They will not be built by default
```

dani-garcia (approved):
LGTM, just some typos.

Do these new architectures need to be running on their specific hosts for building or can they be cross-compiled?

BlackDex:
> LGTM, just some typos.

Thanks, fixed.

> Do these new architectures need to be running on their specific hosts for building or can they be cross-compiled?

These will be cross-compiled, the same as the others. The only thing for these is that they needed some adjustment on the OpenSSL variables, the others are configured correctly without needing these.

tessus (commented):
LGTM. Thanks for the update. Even though I don't use these platforms it is nice to have a starting point to make builds for them easier. Cheers.

senpro-ingwersenk:
So, simply put, if I added the following to `docker-bake.hcl`, I could create a PR for RISC-V?

```hcl
target "debian-riscv64" {
  inherits = ["debian"]
  platforms = ["linux/riscv64"]
  tags = generate_tags("", "-riscv64")
  args = {
    ARCH_OPENSSL_LIB_DIR = "/usr/lib/riscv64-linux-gnu"
    ARCH_OPENSSL_INCLUDE_DIR = "/usr/include/riscv64-linux-gnu"
  }
}
```

Or is there something else to note?
My VisionFive2 is currently off after I moved homes; will turn it on later to make sure the args actually match.

BlackDex:
I doubt this will work for all database types actually.
In theory that should be enough.

PR #4386, opened 2024-02-28
Change API and structs to camelCase

dani-garcia:
This PR changes all the project's structs and API inputs and outputs from the old `PascalCase` format to the new `camelCase` format. At the moment the clients support both but that won't be a thing forever.

I haven't had the time to fully test this, but at least the basic things work. There may be some things missing, and I still need to review what we store in the db to make sure that there aren't any backwards compat issues.

This might also cause problems with other pending PRs so we can wait to merge it until they are dealt with.

Some fields that have special casing:
- `OTP` in `/accounts/verify-otp`, as that's what the clients send. That said we also check `otp`
- `Keys` and `Nfc` in yubikey, this is for backwards compat with older values stored on the server, now we serialize new versions to `keys` and `nfc`.
- `AttestationObject` and `clientDataJson` in webauthn, as that's what the clients send. This is a bigger annoyance as it complicates the use of the webauthn crate.
- Organization policies, this is for backwards compat with older values stored on the server, now we serialize new versions as `camelCase`.

 
Fixes #4656

PR #4526, opened 2024-04-27
Some fixes for the new mobile apps

dani-garcia:
The new mobile apps are going to send some values slightly differently and will expect some other values to be there, while we previously weren't sending them. This should fix those issues, at least for Android.

Note that we will still need to go through with #4386 for the apps to work correctly.

### DIFF
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -10,6 +10,7 @@ use rocket::{
 };
 use serde_json::Value;
 
+use crate::util::NumberOrString;
 use crate::{
     api::{self, core::log_event, EmptyResult, JsonResult, JsonUpcase, Notify, PasswordOrOtpData, UpdateType},
     auth::Headers,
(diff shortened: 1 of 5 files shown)


BlackDex (approved):

BlackDex:
Is #4386 mandatory for the new Android/iOS apps to work on the new releases?
Or will they still work for now, but will become an issue in a few months?

dani-garcia:
It will be required for the new native mobile apps, whenever they get released, as they don't have the backwards compatibility support for server responses using PascalCase. So not super urgent but we probably don't want to wait too long.

BlackDex:
Indeed. But at least good to know it's not needed for the current versions.
Ill see if i can verify that PR soon.

AkechiShiro:
Is this going to be released soon ? I cannot use the mobile app with my vaultwarden instance at the moment due to version mismatch I think.

farfromrefug:
@AkechiShiro where can we get those new apps ? Is the source code on github?

BlackDex:
What i know is that it is still in a alpha stage, and not publicly available. But when ready it will be visible on GitHub as far as i known. We just have to wait.

AkechiShiro:
@farfromrefug there is no vaultwarden app I believe, I think the changes here are made in order to be compatible with an updated Bitwarden app ? But the release containing this new pull request hasn't gone out yet AFAIK

farfromrefug:
@AkechiShiro you arr right I was talkin about thé new bitwarden apps

AkechiShiro:
@farfromrefug the new one is most likely released in the Bitwarden repository

PR #4532, opened 2024-05-01
Optimize Dockerfiles

dfunkt:
Move some ARGs closer to the build stage (potentially improving caching):
Example: I usually set VW_VERSION to be the latest commit hash, without this change I'd have to also rebuild the dependencies even if they didn't change since the previous layers would get invalidated.

Remove redundant COPY commands:
They can be combined, no point in using extra layers if it can be avoided.

Remove redundant RUN command:
apt-get and xx-apt-get commands can be moved into the same RUN command, thus saving another layer.

Move CARGO_HOME's "&&" operator to the first line (improves consistency)

### DIFF
--- a/docker/Dockerfile.j2
+++ b/docker/Dockerfile.j2
@@ -82,10 +82,7 @@ RUN apt-get update && \
         "libc6-$(xx-info debian-arch)-cross" \
         "libc6-dev-$(xx-info debian-arch)-cross" \
         "linux-libc-dev-$(xx-info debian-arch)-cross" && \
-    # Run xx-cargo early, since it sometimes seems to break when run at a later stage
-    echo "export CARGO_TARGET=$(xx-cargo --print-target-triple)" >> /env-cargo
-
-RUN xx-apt-get install -y \
+    xx-apt-get install -y \
         --no-install-recommends \
         gcc \
         libmariadb3 \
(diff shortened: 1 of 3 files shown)


GaetanJeff (approved):

BlackDex:
Saving a lot of layers at the top is not that efficiënt i think.
Ill have to check, there were some reasons i didn't combined the, but not sure.

dfunkt:
Fair enough, if there's stuff here that you don't like just tell me and I'll remove it.

I've been running my builds for a while with these changes and I figured other people might benefit from them as well.

BlackDex:
> Fair enough, if there's stuff here that you don't like just tell me and I'll remove it.
> 
> I've been running my builds for a while with these changes and I figured other people might benefit from them as well.

Do you build all arch's?

dfunkt:
amd64 and arm64, both for personal use

williamdes:
A good tool to check this PR is https://github.com/wagoodman/dive

BlackDex:
> A good tool to check this PR is https://github.com/wagoodman/dive

I use that too indeed. But that will only work for the final image, not for the build layer.

dfunkt:
What do you guys think about including this as well?
https://docs.docker.com/build/cache/#use-the-dedicated-run-cache

BlackDex:
It doesn't really add anything useful i think. 
The main reason is that we went fresh images which have the latest packages available to keep up-to-date and all CVE's are gone as good as possible.

Also, docker buildx does some nice caching by it self. And we already use build caches during our release pipeline.

We used that before btw, which was useful back then, and not building in parallel. When building in parallel it gave some issues, which can be solved, but then it loses its benefits.

BlackDex:
While optimizing for layers does not have any benefit for the build part, it also doesn't hurt in this case.
The only time it will hurt is when you are testing/debugging the last apt commands in the same RUN. But that will not be often i hope.

docker/Dockerfile.j2
```diff
@@ -141,21 +138,23 @@ RUN echo "export CARGO_TARGET=${RUST_MUSL_CROSS_TARGET}" >> /env-cargo && \
...
-COPY ./rust-toolchain.toml ./rust-toolchain.toml
-COPY ./build.rs ./build.rs
+{% if base == "debian" %}
+# Configure the DB ARG as late as possible to not invalidate the cached layers above
+ARG DB=sqlite,mysql,postgresql
+{% elif base == "alpine" %}
+# Configure the DB ARG as late as possible to not invalidate the cached layers above
+# Enable MiMalloc to improve performance on Alpine builds
+ARG DB=sqlite,mysql,postgresql,enable_mimalloc
+{% endif %}
```
BlackDex, line 157: The comment `# Configure the DB ARG as late as possible to not invalidate the cached layers above` can be placed above the `if` and then also be removed from the `else`.

BlackDex (changes requested):
Looks fine by me.

BlackDex (approved):

dani-garcia (approved):

PR #4539, opened 2024-05-05
Check regex blacklist in internal icon resolver

1fexd:


BlackDex:
This is already done here https://github.com/dani-garcia/vaultwarden/blob/main/src/util.rs#L864...L872

So there should be no need for this PR.
Are you experiencing some issues?

1fexd:
Oh, you are absolutely right, it never actually looks up the url - my bad. 

Do you think it would make sense to exclude `.onion`/`.i2p` from icon fetching by default (and allow enabling it via an environment variable)? Also, I feel like logging each icon request with the domain on the info log level is a little invasive

PR #4575, opened 2024-05-20
Fix web-vault version in Docker(files/Settings)

dfunkt:
Seems to have been missed in https://github.com/dani-garcia/vaultwarden/pull/4558


BlackDex:
Indeed. Though it luckily doesn't do anything actually. It's only used for visual display and clarity. 

But thanks!

BlackDex (approved):

dani-garcia (approved):

PR #4577, opened 2024-05-21
Disable Login with device if push is not enabled

Timshel:
I believe the `knowndevice` endpoint is only used to toggle the `Login with device` feature (Could not find any other reference in Bitwarden clients).

And unless I'm mistaken the `Login with device` only work if push is enabled.

Additionally, I removed the compatibility method since the migration to the endpoint was merged on March 7 2023 in the clients (github.com/bitwarden/clients/pull/4710).

### DIFF
--- a/src/api/core/accounts.rs
+++ b/src/api/core/accounts.rs
@@ -50,7 +50,6 @@ pub fn routes() -> Vec<rocket::Route> {
         api_key,
         rotate_api_key,
         get_known_device,
-        get_known_device_from_path,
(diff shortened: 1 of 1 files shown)


BlackDex:
Login with device also works without push enabled.
You can manually fetch it within the client.

Timshel:
Oups couldn't find it, will check again :).

BlackDex:
@Timshel is located at the settings > account security. 
Screenshot_20240521-200745~2

PR #4578, opened 2024-05-21
Remove old knowndevice route

Timshel:
I believe the migration to the endpoint was merged on March 7 2023 in the clients (github.com/bitwarden/clients/pull/4710).

So it probably makes sense to remove it.

### DIFF
--- a/src/api/core/accounts.rs
+++ b/src/api/core/accounts.rs
@@ -50,7 +50,6 @@ pub fn routes() -> Vec<rocket::Route> {
         api_key,
         rotate_api_key,
         get_known_device,
-        get_known_device_from_path,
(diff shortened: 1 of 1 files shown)


BlackDex:
Thanks, cleaning up is a good thing todo.
I think i know a few more place/parts which have some old tech debt

BlackDex (approved):
While this endpoint is still available in the Bitwarden server code-base, it probably isn't used for a long time anymore by the available clients.

Thanks!

dani-garcia (approved):
LGTM, checked all the clients and the new mobile apps and they are all using the header variant, so the other can be removed safely.

PR #4583, opened 2024-05-23
Update Alpine to version 3.20

dfunkt:
- needed to add double quotes, otherwise it was parsed as 3.2 instead of 3.20


BlackDex (approved):

dani-garcia (approved):
Oh how much I love yaml's loose typing hah

PR #4585, opened 2024-05-23
Switch to Github Actions cache

dfunkt:
- remove releasecache-cleanup.yml, GHA cache auto removes oldest cache files after 10GB usage or files not being accessed for more than 7 days
- remove unnecessary step in release.yml, GHA cache doesn't require GHCR login

More info:
https://github.com/moby/buildkit?tab=readme-ov-file#github-actions-cache-experimental


BlackDex:
First, thanks for the PR.

The reason we do not use GHA cache is because our builds take-up a lot of storage. In the end more then the permitted 10GB.
That is why we use the GitHub Container Registry.

Before we used GHA Cache, but that in the end was almost never fully used because cache invalidation.
If you think this is not the case then please elaborate on it, else I'm inclined to closing this.

dfunkt:
Fair enough, sounds like a reasonable explanation.
This was one of the things I was using in my builds since I didn't want to bother with manually cleaning the caches and it seemed to work well enough.

PR #4589, opened 2024-05-25
Prepare for repo to org move

BlackDex:
This PR prepares this repo to be moved to the Vaultwarden organization.


calvin-li-developer:
I'm curious on when is this repo planned to be moved to organization?

Ornias1993:
> Will the sponsors be migrated automagically, or will it have to be manual?

I can say from experience sponsors are always on the user/org level, not on the repo level.
So moving this repo will not move sponsors, because sponsors aren't linked to a repo ;-)

BlackDex:
And in this case they are even linked to a user and not a repo :)

Guiorgy:
In that case, do you plan to notify all sponsors to migrate to the org, or will you keep it on _dani-garcia_?

williamdes (approved):

micah:
Is there any reason why this is still in draft? Could it be resolved now?

BlackDex:
Any reason it should be resolved now?

Guiorgy:
> Is there any reason why this is still in draft? Could it be resolved now?

> Any reason it should be resolved now?

Insert "Both Is Good" meme here.

PrivatePuffin:
> Any reason it should be resolved now?

Only one I can personally think of:
With the recent close-sourcing of part of the Bitwarden client codebase (which, in contrast to the PR campaign by bitwarden, is **not** fully reverted), it seems more-and-more important to distance Vaultwarden from the obviously interest-conflicted leadership by a Bitwarden Employee.

Personally, no need to rush it.
But thats the only reason I can think of that this might be suddenly important.

BlackDex:
Which conflict are you talking about? There is no conflict. There are no restrictions applied as far as i know.
While i understand it would indicate a bit more disconnection, it doesn't prevent that specific employee from doing anything on the new project. So technically there will be not much of a difference.

PrivatePuffin:
> Which conflict are you talking about? There is no conflict. There are no restrictions applied as far as i know.

I think you dont fully understand what "conflict of interest" means...
It doesn't refer to an actual conflict or actually applied influence.

It means you're *prone* to "being influenced", due to having multiple interests that may have conflicting goals.
When it comes to employers, it's generally considered "conflict of interest" when you develop a competing software solution. Regardless on how your employer thinks about it.

As "conflict of interest" is not about how things are "now", it's about how things might be "tomorrow".

Conflict of interest, is something that is defined to protect against the fact that outsides might not know details about how much someone is influenced or not.

---

Though, again, I only play devils advocate here. It's not my personal opinion that this changes the urgency of distancing Vaultwarden from Dani-Garcia. This urgency has been significant for a while, even though its just optics.

BlackDex:
I do understand the meaning of it. But I just do not see the relevance/importance in this specific case. And also not regarding the whole license debate which was mentioned.

Neustradamus:
@BlackDex: I have discovered this PR by chance in a talk with @stefan0xC, good job!

If @dani-garcia will do this transfer, there is no lost, all are redirected before this PR merging :)

BlackDex:
There also is no urgency either.
It will be moved when it's ok to move.

PR #4606, opened 2024-06-02
Remove mimalloc workaround

dfunkt:
- libatomic linking for armv6 has been fixed in mimalloc v0.1.42

For reference: https://github.com/purpleprotocol/mimalloc_rust/commit/446fc231151cfab485b0ac7d5178baf797799dda

### DIFF
--- a/docker/Dockerfile.alpine
+++ b/docker/Dockerfile.alpine
@@ -67,8 +67,6 @@ WORKDIR /app
 
 # Environment variables for Cargo on Alpine based builds
 RUN echo "export CARGO_TARGET=${RUST_MUSL_CROSS_TARGET}" >> /env-cargo && \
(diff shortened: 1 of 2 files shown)


williamdes (approved):

BlackDex (changes requested):
This unfortunately is not true.
The new mimalloc-sys crate actually breaks everything.
It also seems it has been broken for a longer time already.
So, this is something we need to look into.

The binaries do work within an Alpine environment, but not as a true static binary anymore.

dfunkt:
Are we waiting for a new crate release or should I override it with commit https://github.com/purpleprotocol/mimalloc_rust/commit/992c9da4c5afba7fbf4c5815c43c8f0fbd2a8da6?

BlackDex:
I have not yet checked if the removal works for all arch's.

BlackDex (approved):

PR #4639, opened 2024-06-12
add docker compose to readme.md

TheRsKing:
since most people prefer docker compose to docker run, it is useful to have a good docker compose example in the readme. this is also useful since many folks discover vaultwarden at hub.docker.com and so do not need to search the wiki for compose.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,38 @@ Basically full implementation of Bitwarden API is provided including:
  * Emergency Access
 
 ## Installation
+
+## Docker compose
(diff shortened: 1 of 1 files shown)


BlackDex:
See #4243 where we actually prefer to not have this there, but have a link to a good Wiki article instead.

Also, the README at Docker Hub isn't synced, so this will not help.

TheRsKing:
https://github.com/dani-garcia/vaultwarden/pull/4243#issuecomment-2162794434

PR #4645, opened 2024-06-16
Update rust and remove unused header values

dani-garcia:
Updated rust to 1.79, and removed some unused values in the Header guards that Clippy was complaining about not being used anywhere.


BlackDex (dismissed):

BlackDex:
There seem to be some recursieve variables which need to be removed after the removal of the unused variables.

dani-garcia:
Yep just noticed I missed one, should be fixed now

PR #4657, opened 2024-06-19
Fix some nightly build errors

dani-garcia:
I was building vaultwarden with the latest rust nightly and was met with a couple of build errors/warnings.
- Some `unknown cfg` warnings about the cfg's that we define based on features, which I disabled in the build.rs file
- Some type inference errors, where a couple of functions were inferring the `!` type instead of `()`. I'm not sure if this is a transient error or will become a hard error in a future stable release, but it was easy enough to solve.

This was using nightly `2024-06-18`

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -17,6 +17,13 @@ fn main() {
         "You need to enable one DB backend. To build with previous defaults do: cargo build --features sqlite"
     );
 
+    // Use check-cfg to let cargo know which cfg's we define,
+    // and avoid warnings when they are used in the code.
+    println!("cargo::rustc-check-cfg=cfg(sqlite)");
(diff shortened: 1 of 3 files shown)


BlackDex:
It will become a hard error stated by the warning:
`
warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
`

Probably only when we switch to the 2024 edition.

BlackDex:
@dani-garcia i just noticed that a few are missed, because you probably only tested the sqlite feature ;).

dani-garcia:
Missed those hah, should be fixed in https://github.com/dani-garcia/vaultwarden/pull/4661

PR #4678, opened 2024-06-27
Reduce healthcheck internal while keep performance

hellodword:
Currently, it takes 60 seconds for `vaultwarden/server` to become healthy:

```ymal
version: "3"

services:
  vaultwarden:
    image: vaultwarden/server:1.30.5-alpine
    # image: vaultwarden/server:1.30.5
    restart: no
    environment:
      I_REALLY_WANT_VOLATILE_STORAGE: true

  nginx:
    image: nginx
    restart: no
    depends_on:
      vaultwarden:
        condition: service_healthy
```

By implementing a cache mechanism, we can reduce this interval.

### DIFF
--- a/docker/Dockerfile.alpine
+++ b/docker/Dockerfile.alpine
@@ -153,6 +153,6 @@ COPY docker/healthcheck.sh docker/start.sh /
 COPY --from=vault /web-vault ./web-vault
 COPY --from=build /app/target/final/vaultwarden .
 
-HEALTHCHECK --interval=60s --timeout=10s CMD ["/healthcheck.sh"]
+HEALTHCHECK --interval=3s --timeout=2s CMD ["/healthcheck.sh"]
 
 CMD ["/start.sh"]
--- a/docker/Dockerfile.debian
+++ b/docker/Dockerfile.debian
@@ -196,6 +196,6 @@ COPY docker/healthcheck.sh docker/start.sh /
 COPY --from=vault /web-vault ./web-vault
 COPY --from=build /app/target/final/vaultwarden .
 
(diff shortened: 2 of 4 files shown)


BlackDex:
This would make the container runtime check for every 3 seconds. That could be an issue for slower less resource systems.

Also, when the log_level is set to debug or trace, this will flood the logs and make it harder to troubleshoot.

If you are really bothered by this i would suggest to use at least docker v25 and a recent docker compose and use start_interval and start_period in your compose file.

Also i see you use `I_REALLY_WANT_VOLATILE_STORAGE` i must say that is not really save. Not sure if that is only done for this example. It could make all your data disappear when stopping and removing the container.

jjlin:
It doesn't seem like there's any real benefit to this. `nginx` doesn't depend on `vaultwarden`, so there's no need to add a dependency there. Both services will start up just fine without each other, and `vaultwarden` presumably will not be reachable until both services are up anyway, so it's best to let them start up in parallel so they each reach an operational state as soon as possible.

jjlin:
I think it's a rather niche use case to have other services that actually will not start up properly without vaultwarden being ready, and I would rather there not be a weird caching mechanism just to accommodate that. Anyway, vaultwarden shouldn't take more than 2-3 seconds to start up.

BlackDex:
I would actually not add those dependencies as @jjlin mentioned.
And i agree with @jjlin regarding the healtcheck script customization.

If you really want it to be healthier faster, i suggest to use the options i mentioned in my previous post.

Going to close this. Still thinks for wanting to contribute :), and please do not hesitate to do so in the future.

PR #4704, opened 2024-07-04
Switch registry cache compression algorithm to zstd

dfunkt:
- faster builds than with gzip (the default)

Quick comparison I did (vaultwarden built for 3 platforms: linux/amd64, linux/amd64/v3 and linux/arm64):
gzip
• No cache: 22m 7s
• Cached: 11m 20s

zstd:
• No cache: 19m 36s
• Cached: 10m 48s


tessus:
I like zstd. It's used more and more these days. It's also used as the default compression algorithm in Proxmox for backups. I ran several perf tests a while ago and zstd is way faster than gzip with the same or better compression (default values for both).

BlackDex (approved):

BlackDex:
Thanks @dfunkt!

dani-garcia (approved):

PR #4723, opened 2024-07-09
Installation Procedure and Database Export Script

NicolasW-7:
I have tried to create an installation procedure that is easily reusable and has already been tested in a professional environment.
The goal is to try to enhance the Vaultwarden experience and provide a script for database export in case of emergency recovery. I leave it to you to assess the relevance of my proposal.

Nicolas

### DIFF
--- a/docker/Installation Procedure Vaultwarden on Docker.md
+++ b/docker/Installation Procedure Vaultwarden on Docker.md
@@ -1,0 +1,277 @@
+# Vaultwarden Installation
+
+The goal of this procedure is to simplify the installation of Vaultwarden using Docker.
+
+Prerequisites:
+A Debian machine virtual or physique; *the method works on multiple distributions, but commands may need to be adapted for Docker installation.*
+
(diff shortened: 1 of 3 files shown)


BlackDex:
First of all, thanks for taking the effort of creating these articles.

I'm not a fan of adding this to the repo.

This is more of a Wiki article of which we already have a few of.
It might be better to update existing articles or create new ones using this information.

NicolasW-7:
Thank you for your response and for the time you have given me. Do you have a link or a repository to suggest where I can upload this procedure and script ?

BlackDex:
It isn't a repo in that sense. You can just add or adjust articles here: https://github.com/dani-garcia/vaultwarden/wiki

smainz:
A very comprehensive and detailed guide, but unfortunately no longer up to date.

A web socket is still configured and the admin token is not an aragon2 token.

BlackDex:
Closing this in favor of adding these to the wiki

PR #4725, opened 2024-07-09
chore: Dockerfile to Remove port 3012

calvin-li-developer:
as of [v1.31.0](https://github.com/dani-garcia/vaultwarden/releases/tag/1.31.0), this is no longer needed


BlackDex (approved):

dani-garcia (approved):
Thanks!

PR #4730, opened 2024-07-10
Fix bug where secureNotes is empty

### LINKED ISSUE #4719: When using vaultwarden version 1.31.0 and the beta version of bitwarden, synchronization fails.
### Subject of the issue
When I upgrade from an older version to the latest version of Vaultwarden, I can successfully log in to the user, but I cannot sync the password vault.

### Deployment environment

* vaultwarden version: 

* Clients used: Android Beta

* Reverse proxy and version: nginx version: nginx/1.26.1

* MySQL/MariaDB or PostgreSQL version: null

* Other relevant details:

### Steps to reproduce
1. update docker images
2. redeploy vaultwarden
3. login bitwarden Android Beta
4. Enter my vault interface.

### Expected behaviour

Show my password vault.

### Actual behaviour

In my password vault interface, it displays "We were unable to process your request. Please try again or contact us.".
However, when I create a new account and log in with the new account, it does not display "We were unable to process your request. Please try again or contact us." and I can normally create logins, sends, etc.

### Troubleshooting data

```log
vaultwarden  | [2024-07-09 17:09:37.610][request][INFO] GET /api/accounts/revision-date
vaultwarden  | [2024-07-09 17:09:37.610][response][INFO] (revision_date) GET /api/accounts/revision-date => 200 OK
vaultwarden  | [2024-07-09 17:09:37.674][request][INFO] GET /api/sync
vaultwarden  | [2024-07-09 17:09:37.689][response][INFO] (sync) GET /api/sync?<data..> => 200 OK
```

cobyge:
Fixes #4719 
Fixes bug where the value of my notes is an empty string, and thus doesn't pass the check for an empty array.
This adds empty strings to the if condition.  Tested with the mobile app and this fixes my issue

### DIFF
--- a/src/db/models/cipher.rs
+++ b/src/db/models/cipher.rs
@@ -189,9 +189,11 @@ impl Cipher {
             }
         }
 
-        // Fix secure note issues when data is `{}`
+        // Fix secure note issues when data is invalid
         // This breaks at least the native mobile clients
-        if self.atype == 2 && (self.data.eq("{}") || self.data.to_ascii_lowercase().eq("{\"type\":null}")) {
+        if self.atype == 2
+            && (self.data.is_empty() || self.data.eq("{}") || self.data.to_ascii_lowercase().eq("{\"type\":null}"))
+        {
             type_data_json = json!({"type": 0});
         }
 


src/db/models/cipher.rs
```diff
@@ -191,7 +191,7 @@ impl Cipher {
 
         // Fix secure note issues when data is `{}`
```
jjlin, line 192:
```suggestion
        // Fix secure note issues when data is invalid
```

src/db/models/cipher.rs
```diff
@@ -191,7 +191,7 @@ impl Cipher {
...
         // This breaks at least the native mobile clients
-        if self.atype == 2 && (self.data.eq("{}") || self.data.to_ascii_lowercase().eq("{\"type\":null}")) {
+        if self.atype == 2 && (self.data.eq("{}") || self.data.is_empty() || self.data.to_ascii_lowercase().eq("{\"type\":null}")) {
```
jjlin, line 194: Nitpick, but it seems more logical for the `self.data.is_empty()` to be listed first

cobyge: Done

src/db/models/cipher.rs
cobyge: Done

cobyge:
Any chance this could get pushed out as a bugfix? (1.31.1)
I think it would be nice to not have to wait until the next release, because for now the beta mobile app is unusable for some users.

BlackDex:
We will probably release a patch version somewhere soon. But not instantly. There might be some other issues and fixes we want to merge before releasing a new patch version.

BlackDex (approved):

PR #4737, opened 2024-07-11
Update admin interface

BlackDex:
- Updated datatables
- Set Cookie Secure flag if the connection is https
- Prevent possible XSS via Organization Name
  Converted all `innerHTML` and `innerText` to the Safe Sink version `textContent`
- Removed `jsesc` function as handlebars escapes all these chars already and more by default

### DIFF
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -18,7 +18,7 @@ use crate::{
         core::{log_event, two_factor},
         unregister_push_device, ApiResult, EmptyResult, JsonResult, Notify,
     },
-    auth::{decode_admin, encode_jwt, generate_admin_claims, ClientIp},
+    auth::{decode_admin, encode_jwt, generate_admin_claims, ClientIp, Secure},
     config::ConfigBuilder,
     db::{backup_database, get_sql_server_version, models::*, DbConn, DbConnType},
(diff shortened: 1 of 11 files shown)


lukasj98 (approved):

tessus:
> Set Cookie Secure flag if the connection is https

What happens, if https is terminated at the reverse proxy (and vw is running as http)?

BlackDex:
That is where the header check is for.
`X-Forwarded-Proto`, which is already used in an other function.

BlackDex:
It is a commonly used header https://docs.aws.amazon.com/elasticloadbalancing/latest/application/x-forwarded-headers.html#x-forwarded-proto

tessus:
I know that header, I am just asking what happens, if the server thinks it's running under TLS, but in reality it is running as http. e.g. http won't understand TLS traffic.
In a reverse proxy situation with vw not using TLS, vw is running as a http server.

BlackDex:
Then it will not set the secure flag, same as it is now.
So nothing changes in that case.
But again, that is where that header is for, that is something the reverse proxy needs to sent and provide the correct value.

dani-garcia (approved):

PR #4740, opened 2024-07-12
Improved HTTP client

dani-garcia:
In an effort to defend against rebinding attacks more thoroughly, I've done some changes to the HTTP client.
- Extracted it from `util` to a separate file, as it was getting a bit too crowded there
- Removed `get_reqwest_client` and replaced it with `make_http_request`, which will check the provided URL for possible blacklisted domains/ips or non-global IPs, when relevant.
- Added the custom DNS resolver to all HTTP clients, not just `icons`
- Added a redirect policy that checks the targets for possible blacklisted domains/ips or non-global IPs. This is enabled for all HTTP clients as well
- Renamed both `ICON_BLACKLIST_` settings to `HTTP_REQUEST_BLACKLIST_*`, to better reflect that this should affect more than just the `icons` endpoints. We're still reading the old values for backwards compatibility.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -148,9 +148,8 @@ macro_rules! make_config {
 
 
                 // Copy the values from the deprecated flags to the new ones
-                if config.http_request_blacklist_regex.is_none() {
-                    config.http_request_blacklist_non_global_ips = config.icon_blacklist_non_global_ips;
-                    config.http_request_blacklist_regex = config.icon_blacklist_regex.clone();
+                if config.http_request_block_regex.is_none() {
+                    config.http_request_block_regex = config.icon_blacklist_regex.clone();
                 }
 
                 config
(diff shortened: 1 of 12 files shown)


BlackDex (approved):
Nice!

src/config.rs
```diff
@@ -146,6 +146,13 @@ macro_rules! make_config {
                 config.signups_domains_whitelist = config.signups_domains_whitelist.trim().to_lowercase();
                 config.org_creation_users = config.org_creation_users.trim().to_lowercase();
 
+
+                // Copy the values from the deprecated flags to the new ones
+                if config.http_request_blacklist_regex.is_none() {
+                    config.http_request_blacklist_non_global_ips = config.icon_blacklist_non_global_ips;
```
BlackDex, line 152: Shouldn't this be separated btw?
Both items could be either configured or not separately.

BlackDex (changes requested):

src/config.rs
```diff
@@ -531,12 +537,18 @@ make_config! {
         icon_cache_negttl:      u64,    true,   def,    259_200;
         /// Icon download timeout |> Number of seconds when to stop attempting to download an icon.
         icon_download_timeout:  u64,    true,   def,    10;
-        /// Icon blacklist Regex |> Any domains or IPs that match this regex won't be fetched by the icon service.
+
+        /// [Deprecated] Icon blacklist Regex |> Use `icon_blacklist_regex` instead
```
BlackDex, line 541: Should be revering `http_request_block_regex`

src/config.rs
```diff
@@ -531,12 +537,18 @@ make_config! {
...
+        icon_blacklist_regex:   String, false,   option;
+        /// [Deprecated] Icon blacklist non global IPs |> Use `http_request_blacklist_non_global_ips` instead
```
BlackDex, line 543: Should be revering too `http_request_block_non_global_ips`

BlackDex (changes requested):

BlackDex (approved):

PR #4763, opened 2024-07-17
Fix Dockerfile linter warnings

dfunkt:
- they seem to have started appearing with buildx v0.16.0

Alpine warnings:
```
13 warnings found (use --debug to expand):
 - FromPlatformFlagConstDisallowed: FROM --platform flag should not use constant value "linux/amd64" (line 37)
 - FromPlatformFlagConstDisallowed: FROM --platform flag should not use constant value "linux/amd64" (line 38)
 - RedundantTargetPlatform: Setting platform to predefined $TARGETPLATFORM in FROM is redundant as this is the default behavior (line 133)
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 35)
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 36)
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 42)
 - FromPlatformFlagConstDisallowed: FROM --platform flag should not use constant value "linux/amd64" (line 29)
 - FromPlatformFlagConstDisallowed: FROM --platform flag should not use constant value "linux/amd64" (line 36)
 - FromPlatformFlagConstDisallowed: FROM --platform flag should not use constant value "linux/amd64" (line 42)
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 29)
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 34)
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 37)
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 38)
```
Debian warnings:
```
5 warnings found (use --debug to expand):
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 29)
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 38)
 - FromPlatformFlagConstDisallowed: FROM --platform flag should not use constant value "linux/amd64" (line 29)
 - FromPlatformFlagConstDisallowed: FROM --platform flag should not use constant value "linux/amd64" (line 34)
 - RedundantTargetPlatform: Setting platform to predefined $TARGETPLATFORM in FROM is redundant as this is the default behavior (line 171)
```

For ```FromPlatformFlagConstDisallowed``` I decided to use the 3rd variant from here:
https://github.com/moby/buildkit/blob/master/frontend/dockerfile/linter/docs/FromPlatformFlagConstDisallowed.md#description

### DIFF
--- a/docker/Dockerfile.j2
+++ b/docker/Dockerfile.j2
@@ -1,4 +1,5 @@
 # syntax=docker/dockerfile:1
+# check=skip=FromPlatformFlagConstDisallowed,RedundantTargetPlatform
 
 # This file was generated using a Jinja2 template.
 # Please make your changes in `DockerSettings.yaml` or `Dockerfile.j2` and then `make`
@@ -26,7 +27,7 @@
 #     $ docker image inspect --format "{{ '{{' }}.RepoTags}}" docker.io/vaultwarden/web-vault@{{ vault_image_digest }}
 #     [docker.io/vaultwarden/web-vault:{{ vault_version }}]
 #
-FROM --platform=linux/amd64 docker.io/vaultwarden/web-vault@{{ vault_image_digest }} as vault
+FROM --platform=linux/amd64 docker.io/vaultwarden/web-vault@{{ vault_image_digest }} AS vault
 
 {% if base == "debian" %}
 ########################## Cross Compile Docker Helper Scripts ##########################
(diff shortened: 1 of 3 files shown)


dfunkt:
That was needed to fix the FromPlatformFlagConstDisallowed warning.
"Stage name should include the platform, OS, or architecture name to indicate that it only contains platform-specific instructions."
I used the OS variant in order to fix the warnings.

docker/Dockerfile.j2
```diff
@@ -196,7 +196,7 @@ RUN source /env-cargo && \
 # To uninstall: docker run --privileged --rm tonistiigi/binfmt --uninstall 'qemu-*'
 #
 # We need to add `--platform` here, because of a podman bug: https://github.com/containers/buildah/issues/4742
-FROM --platform=$TARGETPLATFORM {{ runtime_stage_image[base] }}
+FROM {{ runtime_stage_image[base] }}
```
BlackDex, line 199: Not sure why you removed this here as the comments mention we need this.
I understand that docker might complain here, maybe we can add an ignore or something?

dfunkt: That's the default already, there's no need to set it. Check the warning message:
> RedundantTargetPlatform: Setting platform to predefined $TARGETPLATFORM in FROM is redundant as this is the default behavior (line 133)

BlackDex: Have you tried podman?

dfunkt: Oops, completly forgot about the other container programs since I only use docker. How should I proceed with this commit?

BlackDex: Im not sure if we can ignore/exclude some of these warnings. That would be my preference in this case.

I can test it for podman, that is no problem.

dfunkt:
I added a way to skip the lint check for ```FromPlatformFlagConstDisallowed``` and ```RedundantTargetPlatform``` in the new revision.
Only change that I left in is the ```as``` to ```AS``` one.

BlackDex (approved):

dani-garcia (approved):

PR #4779, opened 2024-07-25
Duo: use the formatted db email

Timshel:
Was rebasing and had a conflict with the latest addition of `validate_duo_login`.
Had made the fix for the OpenID PR and never back-ported it (Don't exactly remember the issue and it's buried in the comments, but it was linked to using the non lowercased email).

I think in general it makes more sense if available to use the reference email from the user instead of the non formatted value from the form.
I don't have a Duo account anymore so can't really test it :(.

### DIFF
--- a/src/api/core/two_factor/duo_oidc.rs
+++ b/src/api/core/two_factor/duo_oidc.rs
@@ -423,8 +423,6 @@ pub async fn validate_duo_login(
     device_identifier: &str,
     conn: &mut DbConn,
 ) -> EmptyResult {
-    let email = &email.to_lowercase();
-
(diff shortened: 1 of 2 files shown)


BlackDex:
Seems logical. Practically nothing changes, since if you convert it to lowercase and trim it will be exactly the same as we store in the database.

Also, DUO stores username and email in lowercase format anyway, so that should be no issue from that side either.

But, this is probably a nicer way too indeed.

And, it works just fine, i just tested it :).

BlackDex (approved):

dani-garcia (approved):

PR #4786, opened 2024-07-26
Add Read-Only Mode

ssendev:
Implementation of https://github.com/dani-garcia/vaultwarden/discussions/4780

This allows to set ENV variable `READONLY=1` which forbids `POST` `PUT` `DELETE` requests with exceptions for stuff that's required to login.

First i wanted to create a middleware but rocket only has fairings which cannot respond on their own so i instead put a guard in all `POST` `PUT` `DELETE` requests.

I tested it with authenticator and email 2fa which still work the others i'm only 80% sure (for webauthn only 50%).

`post("/accounts/api-key")` is only forbidden if `api_key.is_none()` since that would cause a write.

In organisations some responses returned `Json<Value>` which i changed to `JsonResult`

Using a recovery token to login is not allowed since that would permanently disable 2fa which is a write. The rationale for disallowing 2fa changes is that it may cause people to think "ok i don't need that 2nd factor anymore" and then delete it only to come back and find the 2fa is still enabled since the data on that instance was overwritten from the primary one which still hat 2fa enabled.

On the admin interface there is an info popup announcing the Read-Only Mode but config changes are still allowed.

I don't know how to get a popup for the Vault especially one that's visible on arbitrary clients.

### DIFF
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -19,6 +19,7 @@ use crate::{
         unregister_push_device, ApiResult, EmptyResult, JsonResult, Notify,
     },
     auth::{decode_admin, encode_jwt, generate_admin_claims, ClientIp, Secure},
+    config::not_readonly,
     config::ConfigBuilder,
     db::{backup_database, get_sql_server_version, models::*, DbConn, DbConnType},
     error::{Error, MapResult},
@@ -257,6 +258,7 @@ fn render_admin_page() -> ApiResult<Html<String>> {
     let settings_json = json!({
         "config": CONFIG.prepare_json(),
         "can_backup": *CAN_BACKUP,
+        "readonly": CONFIG.readonly(),
     });
     let text = AdminTemplateData::new("admin/settings", settings_json).render()?;
     Ok(Html(text))
@@ -288,6 +290,8 @@ async fn get_user_or_404(uuid: &str, conn: &mut DbConn) -> ApiResult<User> {
 
 #[post("/invite", data = "<data>")]
 async fn invite_user(data: Json<InviteData>, _token: AdminToken, mut conn: DbConn) -> JsonResult {
+    not_readonly()?;
+
     let data: InviteData = data.into_inner();
     if User::find_by_mail(&data.email, &mut conn).await.is_some() {
         err_code!("User already exists", Status::Conflict.code)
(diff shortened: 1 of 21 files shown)


jjlin:
This seems like a pretty invasive change for a pretty niche use case. I'm not sure what you mean by "HA hassles which make the whole setup only more brittle", but you might consider using something like [Litestream](https://litestream.io/) to be able to quickly bring up a replacement instance.

ssendev:
What i meant with HA hassles is that it makes the system more complex and adds more points of failures especially in smaller organisations where operations is just a "side gig" and not a full time job. The other reason to have a readonly replica is that the credentials for the server can then be saved inside Vaultwarden since it being down means it's still possible to get the password. Yes that means the password of the backup server probably shouldn't also only be in the vault.

The primary motivation for this pr was that in case of the server being down the first thing you would need are the passwords and when they too become unavailable at the same time it would require first bringing up Vaultwarden from a backup and only then being able to solve the actual problem which is valuable time that is lost and could be prevented by having setup the second vault in advance.

Yes it touches a lot of code since i couldn't figure out a way to add a middleware just in one place but it's nothing mind-bending and i think wouldn't complicate ongoing development too much.

While i hope you merge this pr i could understand if you think this feature is too niche but i still think it is a worthwhile addition to Vaultwarden.

BlackDex:
I kinda agree with @jjlin here. This is very invasive and also cause a lot of overhead. Every function will call this, and all future functions would need to add something like this. Doesn't matter if it could be catched more globally, it would still need to check then if it has something to do with login or not.

Maybe using triggers within the SQLite database which you use as readonly might be a better idea. That is probably easier to add on every copy of the database and shouldn't need all these checks.

Maybe it would be a better option to handle readonly database, like a readonly sqlite file if that is possible somehow. 

But this is i think too much to add to the codebase.

jjlin:
> The primary motivation for this pr was that in case of the server being down the first thing you would need are the passwords and when they too become unavailable at the same time it would require first bringing up Vaultwarden from a backup and only then being able to solve the actual problem which is valuable time that is lost and could be prevented by having setup the second vault in advance.

If this is your primary concern, as long as you keep at least one of your clients logged in (i.e., lock the client instead of logging out), it will load its cached copy of the database even if the server is unreachable.

ssendev:
Well the other solution i could imagine is having a second db for the login tokens but that would probably lead to similar maintenance concerns it would also mean an annoying user experience since there wouldn't be an error message telling users that updating passwords not working is intentional.

I guess that's a reason to start using a dedicated client instead of only the web interface.

Issue #4528, opened 2024-04-29
Panic if temporary SMTP-Problem


Subito:
### Subject of the issue

Sometimes on trying to send a 2FA-Mail, the SMTP Server replies with `454: 4.7.0 Temporary authentication failure`, which should indicate to the sender "please try later, this is a temporary problem". However vaultwarden panics if it receives this error and the server stops.

### Deployment environment

* vaultwarden version: 1.30.3

* Install method: FreeBSD pkg

* Clients used: happens on all clients on Login 

* Reverse proxy and version: nginx 

* Other relevant details:

### Expected behaviour

Server does not crash if it receives a temporary SMTP-Failure.

### Actual behaviour

Server crashes and needs restarting.

### Troubleshooting data
```
Apr 29 16:13:58 vault1 vaultwarden[66287]: [2024-04-29 16:13:58.251][vaultwarden::mail][ERROR] SMTP 4xx error: transient error (454): 4.7.0 Temporary authentication failure: Connection lost to authentication server
Apr 29 16:13:58 vault1 vaultwarden[66287]: [2024-04-29 16:13:58.252][panic][ERROR] thread 'tokio-runtime-worker' panicked at 'Error sending incomplete 2FA email: SMTP 4xx error: transient error (454): 4.7.0 Temporary authentication failure: Connection lost to authentication server': src/api/core/two_factor/mod.rs:273
```

BlackDex:
Vaultwarden does not have a retry mechanisch for these kind of scenarios. There message indicates a connection error. 

Most of the time temp errors should be tried again after a few minutes, but this is not something Vaultwarden handles. I also do not thing Vaultwarden should handle those and try again actually. That kind of functionality belongs to mail servers, not clients. 

We probably can catch this error and return an error instead of a panic though, that is probably nicer then a panic.

Subito:
Of course, just catching the error would be great. As you said: Any sort of queue-and-retry should be handled on a mailserver. But panicing on a simple SMTP-Error is probably not great. Just logging an error would be perfect.

PR #4792, opened 2024-07-28
Fix mail::send_incomplete_2fa_login panic issue

dfunkt:
- fixes https://github.com/dani-garcia/vaultwarden/issues/4528

### DIFF
--- a/src/api/core/two_factor/mod.rs
+++ b/src/api/core/two_factor/mod.rs
@@ -269,10 +269,18 @@ pub async fn send_incomplete_2fa_notifications(pool: DbPool) {
             "User {} did not complete a 2FA login within the configured time limit. IP: {}",
             user.email, login.ip_address
         );
-        mail::send_incomplete_2fa_login(&user.email, &login.ip_address, &login.login_time, &login.device_name)
+        match mail::send_incomplete_2fa_login(&user.email, &login.ip_address, &login.login_time, &login.device_name)
             .await
-            .expect("Error sending incomplete 2FA email");
-        login.delete(&mut conn).await.expect("Error deleting incomplete 2FA record");
+        {
+            Ok(_) => {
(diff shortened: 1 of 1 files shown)


src/api/core/two_factor/mod.rs
```diff
@@ -269,9 +269,11 @@ pub async fn send_incomplete_2fa_notifications(pool: DbPool) {
...
         );
-        mail::send_incomplete_2fa_login(&user.email, &login.ip_address, &login.login_time, &login.device_name)
-            .await
-            .expect("Error sending incomplete 2FA email");
+        if let Err(e) =
+            mail::send_incomplete_2fa_login(&user.email, &login.ip_address, &login.login_time, &login.device_name).await
+        {
+            error!("Error sending incomplete 2FA email: {e:#?}");
+        }
         login.delete(&mut conn).await.expect("Error deleting incomplete 2FA record");
```
dani-garcia, line 277: Should we make sure that we only delete the incomplete 2FA record when the email got sent correctly? 

Currently we delete them regardless of if the email got sent, so the notification gets lost forever. If we instead only call `login.delete(&mut conn)` when the email gets sent correctly, then it will be retried again during the next cycle.

We could also convert the `login.delete(&mut conn)` to print the error instead of panicking.

Something like this:
```rust

match  mail::send_incomplete_2fa_login(&user.email, &login.ip_address, &login.login_time, &login.device_name).await {
Ok(_) => {
        if let Err(e) = login.delete(&mut conn).await {
            error!("Error deleting incomplete 2FA record: {e:#?}");
        }
},
Err(e) => {
    error!("Error sending incomplete 2FA email: {e:#?}");
}
```

dfunkt: Sounds good to me.

dani-garcia (approved):

PR #4823, opened 2024-08-07
Update crates, web-vault and fixes

### LINKED ISSUE #4628: v2024.5.0 doesn't allow one to enroll in account recovery
Running vaultwarden 1.30.5-f05398a6 with bw_web build 2024.5.0 produces the following when trying to enroll in account recovery:

```
Jun 05 13:02:39 [vaultwarden] [2024-06-05 13:02:39.003][response][INFO] (get_organization_keys) GET /api/organizations/<org_id>/keys => 200 OK
Jun 05 13:02:39 [vaultwarden] [2024-06-05 13:02:39.763][request][INFO] PUT /api/organizations/39e583bf-50fe-41a0-86db-bfeefe01904d/users/246a5
106-ac41-4830-83ec-b64b07511886/reset-password-enrollment
Jun 05 13:02:39 [vaultwarden] [2024-06-05 13:02:39.763][vaultwarden::api][ERROR] No validation provided
Jun 05 13:02:39 [vaultwarden] [2024-06-05 13:02:39.763][response][INFO] (put_reset_password_enrollment) PUT /api/organizations/<org_id>/users/
<org_user_id>/reset-password-enrollment => 400 Bad Request
```

Switched to bw_web build v2024.3.1 and it works fine.

BlackDex:
- Updated crates
- Updated web-vault to v2024.6.2 This version is currently the latest version compatible with our API implementation. For newer versions we need more code updates to make it compatible. Thanks to @stefan0xC this version fixes #4628
- Added a small fix to prevent errors in the Vaultwarden and Client logs. The v2024.6.2 web-vault calls an endpoint with invalid arguments. If this happens we ignore the call and just return an Ok.
- Added the bulk-collection endpoint (Though not yet available in v2024.6.2)

Fixes #4628

### DIFF
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -2,6 +2,7 @@ use num_traits::FromPrimitive;
 use rocket::serde::json::Json;
 use rocket::Route;
 use serde_json::Value;
+use std::collections::{HashMap, HashSet};
 
 use crate::{
     api::{
@@ -39,6 +40,7 @@ pub fn routes() -> Vec<Route> {
         delete_organization_collection,
         post_organization_collection_delete,
         bulk_delete_organization_collections,
+        post_bulk_collections,
         get_org_details,
         get_org_users,
         send_invite,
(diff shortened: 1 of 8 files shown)


src/api/core/organizations.rs
```diff
@@ -1631,6 +1633,60 @@ async fn post_org_import(
...
 }
 
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+#[allow(dead_code)]
+struct BulkCollectionsData {
+    organization_id: String,
+    cipher_ids: Vec<String>,
+    collection_ids: HashSet<String>,
+    // remove_collections: Bool, // This is not used in v2024.6.3 a.t.m.
```
dani-garcia, line 1643: When this value is true, this endpoint seems to do basically the reverse of what it does, and delete the collections instead of adding them. I don't mind leaving it unimplemented if the clients aren't using it yet, but it might be a good idea to return an error if it's true. Otherwise the operation will succeed and do the opposite of what the user intended.

https://github.com/bitwarden/server/blob/92eac5b59f5969bb0e6f35d6058710cfaf662776/src/Api/Vault/Controllers/CiphersController.cs#L682-L689

BlackDex: There is no way to trigger this from any client as far as i can see.
Better to not implement it and be safe, then implement it in the wrong way if they change something in the future.

It even isn't in v2024.7.3 as far as i can tell.

dani-garcia: Yeah I'm okay with leaving it unimplemented, my worry is that a future client that sends `remove_collections: true` will expect some ciphers to be removed from the collections and instead they will be added to them. The request will succeed but do something quite different.

If we just return an error when the value is true then it's more obvious that something's going on from the perspective of a client.

BlackDex: Ah, that is true

BlackDex: @dani-garcia Done

dani-garcia (approved):

PR #4827, opened 2024-08-08
Add orgUserHasExistingUser parameters to org invite

Timshel:
With the new web-vault `web-v2024.6.2` the organization invitation logic was rewritten (had a look since organization invitation is now working even with sso :tada:).

Discovered that there were additional parameters that were not used such as `orgUserHasExistingUser` which allows to control if the user is redirected to the create account or login page (Cf  [web-v2024.6.2/accept-organization.component.ts](https://github.com/bitwarden/clients/blob/web-v2024.6.2/apps/web/src/app/auth/organization-invite/accept-organization.component.ts#L69), but this already existed before).

The additional `orgSsoIdentifier` parameter could then be added to the SSO PR.

### DIFF
--- a/src/mail.rs
+++ b/src/mail.rs
@@ -17,6 +17,7 @@ use crate::{
         encode_jwt, generate_delete_claims, generate_emergency_access_invite_claims, generate_invite_claims,
         generate_verify_email_claims,
     },
+    db::models::User,
     error::Error,
     CONFIG,
 };
@@ -229,37 +230,55 @@ pub async fn send_single_org_removed_from_org(address: &str, org_name: &str) -> 
 }
 
 pub async fn send_invite(
-    address: &str,
-    uuid: &str,
+    user: &User,
     org_id: Option<String>,
     org_user_id: Option<String>,
     org_name: &str,
     invited_by_email: Option<String>,
 ) -> EmptyResult {
     let claims = generate_invite_claims(
-        uuid.to_string(),
-        String::from(address),
+        user.uuid.clone(),
+        user.email.clone(),
         org_id.clone(),
         org_user_id.clone(),
         invited_by_email,
     );
     let invite_token = encode_jwt(&claims);
+    let mut query = url::Url::parse("https://query.builder").unwrap();
+    {
+        let mut query_params = query.query_pairs_mut();
+        query_params
+            .append_pair("email", &user.email)
+            .append_pair("organizationName", org_name)
+            .append_pair("token", &invite_token);
+        if let Some(id) = org_id {
+            query_params.append_pair("organizationId", &id);
+        };
+        if let Some(id) = org_user_id {
+            query_params.append_pair("organizationUserId", &id);
+        };
+        if user.private_key.is_some() {
+            query_params.append_pair("orgUserHasExistingUser", "true");
+        }
+    }
 
+    let query_string = match query.query() {
+        None => err!(format!("Failed to build invite URL query parameters")),
+        Some(query) => query,
+    };
+
+    // `url.Url` would place the anchor `#` after the query parameters
+    let url = format!("{}/#/accept-organization/?{}", CONFIG.domain(), query_string);
     let (subject, body_html, body_text) = get_text(
         "email/send_org_invite",
         json!({
-            "url": CONFIG.domain(),
+            "url": url,
             "img_src": CONFIG._smtp_img_src(),
-            "org_id": org_id.as_deref().unwrap_or("_"),
-            "org_user_id": org_user_id.as_deref().unwrap_or("_"),
-            "email": percent_encode(address.as_bytes(), NON_ALPHANUMERIC).to_string(),
-            "org_name_encoded": percent_encode(org_name.as_bytes(), NON_ALPHANUMERIC).to_string(),
             "org_name": org_name,
-            "token": invite_token,
         }),
     )?;
 
-    send_email(address, &subject, body_html, body_text).await
+    send_email(&user.email, &subject, body_html, body_text).await
 }
 
 pub async fn send_emergency_access_invite(
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -297,7 +297,7 @@ async fn invite_user(data: Json<InviteData>, _token: AdminToken, mut conn: DbCon
 
     async fn _generate_invite(user: &User, conn: &mut DbConn) -> EmptyResult {
         if CONFIG.mail_enabled() {
-            mail::send_invite(&user.email, &user.uuid, None, None, &CONFIG.invitation_org_name(), None).await
+            mail::send_invite(user, None, None, &CONFIG.invitation_org_name(), None).await
         } else {
             let invitation = Invitation::new(&user.email);
             invitation.save(conn).await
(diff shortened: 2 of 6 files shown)


src/mail.rs
```diff
@@ -228,38 +229,55 @@ pub async fn send_single_org_removed_from_org(address: &str, org_name: &str) ->
...
+        if user.private_key.is_some() {
+            query_params.append_pair("orgUserHasExistingUser", "true");
+        }
+    }
+
     let (subject, body_html, body_text) = get_text(
         "email/send_org_invite",
         json!({
-            "url": CONFIG.domain(),
+            "url": String::from(url).replace("/?email", "/#/accept-organization/?email"),
```
stefan0xC, line 274: Would it be possible to [join](https://docs.rs/url/latest/url/struct.Url.html#method.join) `/#/accept-organization/` to the base url, when constructing the `url`, instead of replacing it here? Or is there a reason for doing it like that?

src/mail.rs
```diff
@@ -228,38 +229,55 @@ pub async fn send_single_org_removed_from_org(address: &str, org_name: &str) ->
...
+    let mut url = match url::Url::parse(&CONFIG.domain()) {
+        Ok(url) => url.clone(),
+        Err(err) => err!(format!("Failed to build org invite url: {err}")),
+    };
+
+    {
+        let mut query_params = url.query_pairs_mut();
+        query_params
+            .append_pair("email", &user.email)
+            .append_pair("organizationName", org_name)
```
stefan0xC, line 258: Am I correct in assuming that doing it this way will automatically url encode the parameters?

src/mail.rs
```diff
@@ -228,38 +229,55 @@ pub async fn send_single_org_removed_from_org(address: &str, org_name: &str) ->
...
+            .append_pair("token", &invite_token);
+        if let Some(id) = org_id {
+            query_params.append_pair("organizationId", &id);
+        };
+        if let Some(id) = org_user_id {
+            query_params.append_pair("organizationUserId", &id);
+        };
+        if user.private_key.is_some() {
+            query_params.append_pair("orgUserHasExistingUser", "true");
```
stefan0xC, line 267: Do you know when this would be false?

src/mail.rs
```diff
@@ -228,38 +229,55 @@ pub async fn send_single_org_removed_from_org(address: &str, org_name: &str) ->
     send_email(address, &subject, body_html, body_text).await
 }
 
+// The url cannot be fully built using `url.Url` since the anchor `#` should be after the query parameters
```
stefan0xC, line 232: What does this comment mean exactly? Aren't you replacing the `/?email` with `/#/accept-organization/?email below? (Is this related to my other question? If so I would move the comment down before the replace.)

stefan0xC (commented):
Looks mostly good to me. I do have some questions though.

src/mail.rs
Timshel: It is, by definition the anchor `#` should be after the query parameters.
So if set at the start you'll end up with an url: `domain.com/?token=toto#/accept-organization/`

src/mail.rs
Timshel: when using the `sso` if you want to redirect to the SSO login flow.

src/mail.rs
Timshel: I believe it does: cf: https://docs.rs/url/latest/url/struct.Url.html#method.query_pairs_mut

src/mail.rs
stefan0xC: Ah, makes sense.

src/mail.rs
stefan0xC: This comment should be moved down then (and maybe also reworded a bit as it's saying the opposite to me).
```suggestion
// The url cannot be fully built using `url.Url` since the anchor `#` would be after the query parameters so we use `replace` as a work around here
```

Timshel: Do you prefer it just before the `Url::parse` or lower down before the `replace` ?

stefan0xC: If possible I would add the comment directly above the replace (as this is where I was wondering why it's done this way and where I think it would be most relevant).

Timshel: Realized that the `replace` was invalid if the domain included a subpath.
Could have just removed the `/  but I think this new version should be safer.

src/mail.rs
```diff
@@ -229,37 +230,50 @@ pub async fn send_single_org_removed_from_org(address: &str, org_name: &str) ->
...
+        if let Some(id) = org_user_id {
+            query_params.append_pair("organizationUserId", &id);
+        };
+        if user.private_key.is_some() {
+            query_params.append_pair("orgUserHasExistingUser", "true");
+        }
+    }
 
+    // `url.Url` would place the anchor `#` after the query parameters
+    let url = format!("{}/#/accept-organization/?{}", CONFIG.domain(), query.query().unwrap_or(""));
```
BlackDex, line 266: I think it is best to handle this unwrap manually instead of using `unwrap_or()`.
This could cause issues when something with the unwrap goes wrong here.

Best to check if unwrap works or not and error out.
Something like this maybe?
```rust
if let query_string = match query.query() {
  Ok(query) => query,
  Err(e) => err!("Unable to build invite url", format!("Query Build Error {e:?}")),
};
```

BlackDex (changes requested):
Overall i think it looks nice.
Just a small remark.

src/mail.rs
Timshel: I had agreed at first but the `unwrap` is here because of an `Option` not an `Error`.

That's why I believed it's ok to `unwrap` since I'm guessing it returns `None` only if there is no query parameters which should never happen.

Timshel: I'm all for enforcing type constraint and proper checking but in some cases I believe the handling just add bloat and lower code legibility. 
Same reason I did not handle the potential `Error` and used an even riskier `unwrap` earlier: `url::Url::parse("https://query.builder").unwrap()`.

But if you think it's worth it I'll make the change.

BlackDex: Well, either use plain `unwrap()` if you are sure it never happens, but that could panic if not. Else i would suggest to use my suggestion. But i would at least not use `unwrap_or()`, since then we would never know.

BlackDex: Ow, `expect("Invite query string")` is also an option.

Timshel: Pushed the `match` version.

src/mail.rs
```diff
@@ -229,37 +230,55 @@ pub async fn send_single_org_removed_from_org(address: &str, org_name: &str) ->
...
+        if let Some(id) = org_user_id {
+            query_params.append_pair("organizationUserId", &id);
+        };
+        if user.private_key.is_some() {
+            query_params.append_pair("orgUserHasExistingUser", "true");
+        }
+    }
 
+    let query_string = match query.query() {
+        None => err!(format!("Failed to build invite url query parameters")),
```
dfunkt, line 266:
```suggestion
        None => err!(format!("Failed to build invite URL query parameters")),
```
I feel like this would be better from a stylistic point of view.

BlackDex (approved):

BlackDex:
Nice, works well.

dani-garcia (approved):

PR #4845, opened 2024-08-12
Include sqlite3 binaries in Docker images

capi:
I understand that the runtime docker images are trimmed down for size. Regardless, I think that it makes sense to have the sqlite3 binaries as part of the images because it makes backups way easier without having to resort to other images/containers or having sqlite installed on the docker host.

A cron-driven backup on the Docker host would then just be similar to the following cronjob (assuming a running container named `vaultwarden`)
```
0 0 * * * root /usr/bin/docker exec vaultwarden sqlite3 /data/db.sqlite3 "VACUUM INTO '/backup/db-$(date '+%Y%m%d-%H%M').sqlite3'"
```

If size of the images is more important, I understand, please decline the PR.

### DIFF
--- a/docker/Dockerfile.alpine
+++ b/docker/Dockerfile.alpine
@@ -139,7 +139,8 @@ RUN mkdir /data && \
         ca-certificates \
         curl \
         openssl \
-        tzdata
+        tzdata \
+        sqlite
 
 VOLUME /data
 EXPOSE 80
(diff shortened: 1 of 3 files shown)


capi:
Ok, makes sense, especially w.r.t. the embedde sqlite. Just pointing out though, that [this backup documentation](https://github.com/dani-garcia/vaultwarden/wiki/Backing-up-your-vault) would point one towards using the host's sqlite version, though.

I'll close the PR and look into the API or build a derivative image that installs the missing binaries if I think it will fit my use-case.

PR #4860, opened 2024-08-14
Fix manager in web-vault v2024.6.2 for collections

### LINKED ISSUE #4844: Manager Role no long has ability to manage collection access
### Subject of the issue
Manager Role no longer has ability to manage collection access

* Vaultwarden version: v1.32.0
* Web-vault version: v2024.6.2b
* OS/Arch: linux/x86_64
* Uses a reverse proxy: true
* Internet access: true
* DNS Check: true
* Server/NTP Time Check: true
* HTTPS Check: true
* Database type: SQLite
* Database version: 3.46.0

### Steps to reproduce
We recently noticed (after updating to version 1.32.0, but could have existed prior to this version unnoticed) that the Manager role no longer has the ability to view/manage the members of an existing collection. It does appear that this role does permit the creation of a new collection and the assignment of members during creation, but then loses all ability to view/manager post-creation as well.

### Expected behaviour
Manager role would be able to view/manager collection members as described in the role description.

### Actual behaviour
Each collection under an organization no longer as the "..." 3-dots menu to navigate to the access settings of the collection when signed in with a manager role for the organization

BlackDex:
The web-vault v2024.6.2 we use needs some extra information to allow managers to actually be able to manage collections.

The v2024.6.2 web-vault has somewhat of a mixture of the newer roles and older manager roles. To at least fix this for the web-vault we bundle these changes will make the manager able to manage.

For future web-vaults we would need a lot more changes to be done to fix this in a better way though.

Fixes #4844

### DIFF
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -509,7 +509,7 @@ async fn post_organization_collection_update(
         CollectionUser::save(&org_user.user_uuid, col_id, user.read_only, user.hide_passwords, &mut conn).await?;
     }
 
-    Ok(Json(collection.to_json()))
+    Ok(Json(collection.to_json_details(&headers.user.uuid, None, &mut conn).await))
 }
 
 #[delete("/organizations/<org_id>/collections/<col_id>/user/<org_user_id>")]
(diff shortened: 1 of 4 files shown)


lukasj98 (approved):

dani-garcia (approved):

OdyX:
It would be really awesome if this could be released soon. The bug is putting a lot of burden on our instance administrators, who have to do all the collections' memberships updates.

Is there anything I could do to help get a new release out?

dwestness:
It appears this has been merged into main, but no new build available yet. Do i need to compile from source in order to get this fix in my environment?

BlackDex:
Use the `testing` tagged containers

vKnmnn:
Any hints on when this will be released to non-testing tagged containers?

PR #4869, opened 2024-08-18
Fix Vaultwarden Admin page error messages

### LINKED ISSUE #4834: Bad request if config is not read/writeable
I migrated from a docker container to an LCX container. Therefore, I moved the data to the new system.
Everything was updated with the correct owner, just the config.json was not changed to the correct owner.

During changing data in the admin-area, the save throws and "400 Bad Request", but the UI shows, that the changed are there (also after a force refresh). The changed settings are active, until I update to the next version.

After chaning the owner to the correct user, saving is working.

I would recommend, that the response contain an information, that the config file is not writeable.

BlackDex:
Since the change to camelCase variables the error messages in the Vaultwarden Admin were not shown correctly anymore.

This PR fixes this by changing the case of the json key's. Also updated the save and delete of the config to provide a more descriptive error instead of only `Io` or which ever other error might occure.

Fixes #4834

### DIFF
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -750,12 +750,18 @@ fn get_diagnostics_config(_token: AdminToken) -> Json<Value> {
 #[post("/config", data = "<data>")]
 fn post_config(data: Json<ConfigBuilder>, _token: AdminToken) -> EmptyResult {
     let data: ConfigBuilder = data.into_inner();
-    CONFIG.update_config(data)
+    if let Err(e) = CONFIG.update_config(data) {
+        err!(format!("Unable to save config: {e:?}"))
+    }
+    Ok(())
 }
 
 #[post("/config/delete")]
 fn delete_config(_token: AdminToken) -> EmptyResult {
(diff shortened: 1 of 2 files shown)


src/api/admin.rs
```diff
@@ -750,12 +750,19 @@ fn get_diagnostics_config(_token: AdminToken) -> Json<Value> {
...
+        err!(format!("Unable to save config: {e:?}"))
+    }
+    Ok(())
 }
 
 #[post("/config/delete")]
 fn delete_config(_token: AdminToken) -> EmptyResult {
-    CONFIG.delete_user_config()
+    if let Err(e) = CONFIG.delete_user_config() {
+        dbg!(e.to_string());
```
tessus, line 762:
```suggestion
```

This seems like an oversight. e.g. there is no `dbg!` in `post_config`.

BlackDex: good catch.

dani-garcia (approved):

PR #4876, opened 2024-08-20
Update issue template

BlackDex:
Updated the issue template to use a form and guide users to provide all information useful to troublshoot issues

Als updated links to prefer the usage of GitHub Discussions.


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,140 @@
...
+        3. Click on `Copy To Clipboard`
+        4. Replace this text by pasting it into this textarea without any modifications
+    validations:
+      required: true
+  - type: textarea
+    id: reproduce
+    attributes:
+      label: Steps To Reproduce
+      description: How can we reproduce the behavior.
+      value: |
```
dani-garcia, line 33: Should this use `placeholder` too?

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,140 @@
...
+      render: text
+  - type: textarea
+    id: additional-context
+    attributes:
+      label: Additional Context
+      description: Add any other context about the problem here.
+  - type: dropdown
+    id: os
+    attributes:
+      label: Operating System
```
dani-garcia, line 74: Should we mention in these OS sections that we mean the OS where the _server_ is running as opposed to where the clients run?

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,140 @@
...
+  - type: input
+    id: client-version
+    attributes:
+      label: Client Version
+      description: What version(s) of the client(s) are you seeing the problem on?
+      placeholder: ex. v2024.7.2
+  - type: input
+    id: version
+    attributes:
+      label: Build Version
```
dani-garcia, line 136: Can we move this entry before the clients?
So it goes:
- OS
- Deployment
- Server build version
- Clients
- Client version

That way we have all the server questions first and the clients last

dani-garcia (commented):
Nice looks pretty great, just some comments.

Also some of the fields are required while others aren't, but in the form they all kinda look the same to me? Should we add `(required)` or `(optional)` after the names so it's more clear?

.github/ISSUE_TEMPLATE/bug_report.yml
BlackDex: Nea, this is easy for the users to have something to modify instead of the placeholder which disappears when typing.

.github/ISSUE_TEMPLATE/bug_report.yml
BlackDex: That might be a good one :)

.github/ISSUE_TEMPLATE/bug_report.yml
BlackDex: sure

dani-garcia:
Also along the required/optional distinction, maybe it would make more sense to have all the required fields first and then the optionals last? To make it easier to know what's been filled and what's missing.

BlackDex:
All required options will have a red asterisk behind them, so that should be fine.

BlackDex:
Some fields are optional but still kinda belong to a previous required item like `id: os` and `id: os-version` or `id: deployment` and `id: deployment-other`.

Same goes for the others i think.

dani-garcia:
I don't seem to be getting any red asterisks, weird:

image

BlackDex:
> I don't seem to be getting any red asterisks, weird:
> 
> image

That is true, those will not show on private repo's, i was confused too haha.

dani-garcia (approved):
Well, then we can merge this and see how it looks here hah

BlackDex:
As a reference see here: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms#about-yaml-syntax-for-issue-forms

.github/ISSUE_TEMPLATE/bug_report.yml
stefan0xC: To me the questions about the OS sound more like questions about the client environment. I mean, I have no idea if you can run Vaultwarden on Android but I wouldn't assume that's a common runtime OS. Or why would you need to specify multiple entries for the server you run Vaultwarden on? I'd assume you only run it in one. :raised_eyebrow: 

And what's maybe missing are the fields you are expected to fill out yourself from the generated support string:
* Reverse proxy and version: 
* Other relevant information:

BlackDex: Well, i can run Vaultwarden in my Android, not that i think someone will do something like that though. But it is possible.

Good point of the reverse proxy, forgot that point. 
Maybe you want to change since if these items to better suite the reports?

PR #4889, opened 2024-08-23
Update crates (GHSA-wq9x-qwcq-mmgf)

BlackDex:
- Updated crates
- Fixed MSRV to actually be N-2
- Changed some features to use the `dep:` prefix. This is needed for edition-2024 anyway although that will be a while before we can use that.

This also updates Diesel which seems to have a possible CVE.


dfunkt:
> This also updates Diesel which seems to have a possible CVE.

It does appear so on GitHub (GHSA-wq9x-qwcq-mmgf)

dani-garcia (approved):

PR #4894, opened 2024-08-25
Non-interactive Argon2id PHC hash generation

buengese:
This is a convenience feature I added mostly for my own sake.

It adds a non interactive way of creating the Agon2id PHC hash from a password which allows for better scriptability of the setup process. While this can technically already be done using the argon2 cli tool, thats still an extra tool that needs to be installed.

This PR just modifies the `vaultwarden hash` command to check if stdin is a tty and if not will directly read an input password from stdin and output the PHC hash without any confirmation or extra output.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -33,6 +33,7 @@ use std::{
     process::exit,
     str::FromStr,
     thread,
+    io::{self, IsTerminal},
 };
 
 use tokio::{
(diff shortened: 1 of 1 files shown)


BlackDex:
Thx for your contribution.

We didn't add this feature because we want to prevent people saving there password into the shell history or showing up in the process list during execution. 

That is why I'm not really for adding this feature.

buengese:
I understand that reasoning but it's already something many other tools allow. Allowing the password to be read from stdin does not necessarily expose the password to shell history or process list. I.e. for my usecase vaultwarden is called from ansible which uses popen to directly write to stdin. 

Having to deal with prompts always hurts scriptability. Of course they can be worked around with something `expect` but thats annoying.

BlackDex:
Well, it might not necessarily allow it, but it makes it possible to do so. And that will allow people to to use it that way even while it is not the safe way.

You can always use the `argon2` tool for scriptability.

buengese:
I was hoping not to have to rely on an external tool. Personally, I’m not a fan of limiting useful features just because users might shoot themselves in the foot with them. However, it’s your call to make.

BlackDex:
I can understand your reasoning too But if you really find that you do not want to use external tools, i would suggest to add the argon2 hashing into ansible then. That would probably be a better option.

I just do not like adding this. To be fair. I have thought about this when i created the hash command.
It just isn't best practice in the cli world. Also tools like `openssl` do not allow this. And i know, other tools like `argon2` it self and `passwd` and probably dozen of other tools allow this in some for all mention to not use it or at least be careful.

I just want to prevent users misusing it. And other tools are available to do the same.

If someone else has an other opinion please share, else I'm leaning towards closing this.

tessus:
I understand both reasonings:

- using some other argon2id provider in ansible or using an external command
- trying to limit what a user can or cannot do

The issue is that certain tools must allow destructive commands. e.g. take the `mkfs.xxx` tools or `fdisk`.
If a user can install vaultwarden, they also have access to commands that can do more harm than exposing a password, e.g. the 2 commands I mentioned above.
Another problem is that users almost never read warnings or documentation. Thus a dev cannot be made responsible for how people use the SW. That's why there are always disclaimers for using SW.

The best option for all might be to hide/guard this behind a flag, e.g. `--i-know-what-i-am-doing` or `--allow-dangerous-stuff` or whatever comes to mind.

I can only speak from my own experience in SW design and dev. Limiting what a user can do is not always the solution. They most likely will find another way to royally screw up then. ;-) 
Jokes aside, I believe a dev should be able to write code in good faith that people use common sense. Otherwise we could be seen as people who think that all users are stupid and rob them of their right to make mistakes. Although my last sentence could be seen as philosphical reasoning in which case, please ignore.

BlackDex:
Closing this as it's not planned to be integrated.

PR #4901, opened 2024-08-27
remove superfluous asterisk

dotlambda:


dani-garcia (commented):
Hmm a long time ago we had the asterisk to point to the `this project is not associated with the Bitwarden project` line, but at some point it got deleted. Should we instead add the asterisk back there?

Here's the old readme for example:
https://github.com/dani-garcia/vaultwarden/blob/a0ae032ea709bee1422a22b9dd022b0a06c33442/README.md

BlackDex:
That line is still there it looks like, but a bit in-between other lines.
It might be good to either move that line a bit more to the top then, and above the labels to let it be more clear.

Not sure if the asterisk is still needed though. There are multiple other locations where we mention Bitwarden or the clients and do not emphasize on the not being a part of..

PR #4903, opened 2024-08-28
Allow enforcing Single Org with pw reset policy

### LINKED ISSUE #4855: Account recovery administration not enforcing Single orginization policy to be enabled.
### Subject of the issue

Account recovery administration is not enforcing "Single orginization" policy to be enabled. You are able to enable the "Account recovery administration" organization policy without the "Single organization" policy being enabled.
### Deployment environment

* vaultwarden version: 1.32.0

* Install method: Docker through tag: vaultwarden/server:1.32.0

* Clients used: 

* Reverse proxy and version:  nginx

* MySQL/MariaDB or PostgreSQL version: 

* Other relevant details:

### Steps to reproduce

Enable the Account recovery administration without the Single Organizaton policy being enabled.
### Expected behaviour

Not being able to enable it.
### Actual behaviour

You can enable it
### Troubleshooting data

image

BlackDex:
Bitwarden only allows the Reset Password policy to be set when the Single Org policy is enabled already. This PR adds a check so that this can be enforced when a config option is enabled.

Since Vaultwarden encouraged to use multiple orgs when groups were not available yet we should not enable this by default now. This might be something to do in the future.

When enabled, it will prevent the Reset Password policy to be enabled if the Single Org policy is not enabled. It will also prevent the Single Org policy to be disabled if the Reset Password policy is enabled.

Fixes #4855

### DIFF
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -1787,9 +1787,7 @@ async fn put_policy(
     // Now that groups are available we can enforce this option when wanted.
     // We put this behind a config option to prevent breaking current installation.
     // Maybe we want to enable this by default in the future, but currently it is disabled by default.
-    if CONFIG.enforce_single_org_with_reset_pw_policy()
-        && (pol_type_enum == OrgPolicyType::ResetPassword || pol_type_enum == OrgPolicyType::SingleOrg)
-    {
+    if CONFIG.enforce_single_org_with_reset_pw_policy() {
         if pol_type_enum == OrgPolicyType::ResetPassword && data.enabled {
             let single_org_policy_enabled =
                 match OrgPolicy::find_by_org_and_type(org_id, OrgPolicyType::SingleOrg, &mut conn).await {
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -1782,6 +1782,40 @@ async fn put_policy(
         None => err!("Invalid or unsupported policy type"),
     };
 
+    // Bitwarden only allows the Reset Password policy when Single Org policy is enabled
+    // Vaultwarden encouraged to use multiple orgs instead of groups because groups were not available in the past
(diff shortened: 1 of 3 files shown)


src/api/core/organizations.rs
```diff
@@ -1782,6 +1782,40 @@ async fn put_policy(
         None => err!("Invalid or unsupported policy type"),
     };
 
+    // Bitwarden only allows the Reset Password policy when Single Org policy is enabled
+    // Vaultwarden encouraged to use multiple orgs instead of groups because groups were not available in the past
+    // Now that groups are available we can enforce this option when wanted.
+    // We put this behind a config option to prevent breaking current installation.
+    // Maybe we want to enable this by default in the future, but currently it is disabled by default.
+    if CONFIG.enforce_single_org_with_reset_pw_policy()
+        && (pol_type_enum == OrgPolicyType::ResetPassword || pol_type_enum == OrgPolicyType::SingleOrg)
```
stefan0xC, line 1791:
```suggestion
```
I think this is a bit superfluous. Do we really need to check for these conditions twice?

I think you could alternatively also get rid of one level of indentation by checking CONFIG.enforce_single_org_with_reset_pw_policy() twice instead.

stefan0xC (changes requested):

src/api/core/organizations.rs
BlackDex: Well, there are other policies to check also.
This does prevent checking the other if's everytime when the config is enabled.

stefan0xC: Are the other ifs that expensive compared to this one? :thinking:

BlackDex: Not that much, there is an extra check done `data.enabled` but that is about it.
I removed it, also as this endpoint is probably not used that much anyway :).

stefan0xC (approved):

dani-garcia (approved):

PR #4906, opened 2024-08-30
Add a CLI feature to backup the SQLite DB

BlackDex:
Many users request to add the sqlite3 binary to the container image. This isn't really ideal as that might bring in other dependencies and will only bloat the image. There main reason is to create a backup of the database.

While there already was a feature within the admin interface to do so (or by using the admin API call), this might not be easy.

This PR adds several ways to generate a backup.
1. By calling the Vaultwarden binary with the `backup` command like:
  - `/vaultwarden backup`
  - `docker exec -it vaultwarden /vaultwarden backup`
2. By sending the USR1 signal to the running process like:
  - `kill -s USR1 $(pidof vaultwarden)`
  - `killall -s USR1 vaultwarden`

This should help users to more easily create backups of there SQLite database.

Also added the Web-Vault version number when using `-v/--version` to the output.

### DIFF
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -384,9 +384,11 @@ pub async fn backup_database(conn: &mut DbConn) -> Result<String, Error> {
 pub fn backup_sqlite_database(conn: &mut diesel::sqlite::SqliteConnection) -> Result<String, Error> {
     use diesel::RunQueryDsl;
     let db_url = CONFIG.database_url();
-    let db_path = std::path::Path::new(&db_url).parent().unwrap().to_string_lossy();
-    let file_date = chrono::Utc::now().format("%Y%m%d_%H%M%S").to_string();
-    let backup_file = format!("{db_path}/db_{file_date}.sqlite3");
+    let db_path = std::path::Path::new(&db_url).parent().unwrap();
+    let backup_file = db_path
+        .join(format!("db_{}.sqlite3", chrono::Utc::now().format("%Y%m%d_%H%M%S")))
+        .to_string_lossy()
+        .into_owned();
     diesel::sql_query(format!("VACUUM INTO '{backup_file}'")).execute(conn)?;
     Ok(backup_file)
 }
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -368,21 +368,27 @@ pub mod models;
 
 /// Creates a back-up of the sqlite database
 /// MySQL/MariaDB and PostgreSQL are not supported.
-pub async fn backup_database(conn: &mut DbConn) -> Result<(), Error> {
+pub async fn backup_database(conn: &mut DbConn) -> Result<String, Error> {
     db_run! {@raw conn:
         postgresql, mysql {
             let _ = conn;
             err!("PostgreSQL and MySQL/MariaDB do not support this backup feature");
         }
         sqlite {
-            use std::path::Path;
-            let db_url = CONFIG.database_url();
-            let db_path = Path::new(&db_url).parent().unwrap().to_string_lossy();
-            let file_date = chrono::Utc::now().format("%Y%m%d_%H%M%S").to_string();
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -560,6 +599,22 @@ async fn launch_rocket(pool: db::DbPool, extra_debug: bool) -> Result<(), Error>
         CONFIG.shutdown();
     });
 
+    #[cfg(unix)]
+    {
+        tokio::spawn(async move {
+            let mut signal_user1 = tokio::signal::unix::signal(SignalKind::user_defined1()).unwrap();
+            loop {
+                // If we need more singles to act upon, we might want to use select! here.
```
dani-garcia, line 607: I think this is what you mean right?
```suggestion
                // If we need more signals to act upon, we might want to use select! here.
```

src/db/mod.rs
```diff
@@ -368,23 +368,29 @@ pub mod models;
...
     }
 }
 
+#[cfg(sqlite)]
+pub fn backup_sqlite_database(conn: &mut diesel::sqlite::SqliteConnection) -> Result<String, Error> {
+    use diesel::RunQueryDsl;
+    let db_url = CONFIG.database_url();
+    let db_path = std::path::Path::new(&db_url).parent().unwrap().to_string_lossy();
+    let file_date = chrono::Utc::now().format("%Y%m%d_%H%M%S").to_string();
+    let backup_file = format!("{db_path}/db_{file_date}.sqlite3");
```
dani-garcia, line 389: I think we should be using path.join() here instead of format!() to add the path separator, something like this (haven't tested it)
```suggestion
    let db_path = std::path::Path::new(&db_url).parent().unwrap();
    let file_date = chrono::Utc::now().format("%Y%m%d_%H%M%S");
    let backup_file = db_path.join(format!("db_{file_date}.sqlite3"));
```

src/main.rs
BlackDex: Unless i would want to sing less 🤣.
I really need to check spelling before i commit haha.

src/db/mod.rs
BlackDex: Ah, yea, that might be a more safe option indeed.
Ill go and test it.

dani-garcia (approved):

git-anish:
Hi @dani-garcia @BlackDex 

Quick question about the new update: it only covers SQLite right? I'm wondering if this would be enough to recover from a complete loss of a Vaultwarden deployment?

Currently, I'm running [ttionya/vaultwarden-backup](https://github.com/ttionya/vaultwarden-backup) to maintain backups in case I need to spin up the server again. It covers

- config.json
- rsa_key*
- Attachments
- Sent Data: sends
- SQLite database

BlackDex:
It only covers the database currently. 
Mainly because people wanted the SQLite binary inside the container. 

So i would suggest to keep using that tool.

git-anish:
> It only covers the database currently. Mainly because people wanted the SQLite binary inside the container.
> 
> So i would suggest to keep using that tool.

Makes complete sense, thank you

gyzerok:
@dani-garcia do you mind updating [backup section](https://github.com/dani-garcia/vaultwarden/wiki/Backing-up-your-vault#sqlite-database-files) in the wiki to mention this command?

If I wouldn't follow release changelogs I'd never know it is there.

BlackDex:
> @dani-garcia do you mind updating [backup section](https://github.com/dani-garcia/vaultwarden/wiki/Backing-up-your-vault#sqlite-database-files) in the wiki to mention this command?
> 
> If I wouldn't follow release changelogs I'd never know it is there.

You may do that your self if you want. The wiki van be adjusted by everybody.

gyzerok:
@BlackDex I've updated it but am surpirsed there is no pre-moderation for such changes. Hopefully my changes are fine. Thank you!

BlackDex:
Looks good. Thanks!

PR #4916, opened 2024-09-03
Actually use Device Type for mails

dfunkt:
- match Bitwarden behaviour
- remove upcase_first, no longer needed and it would also interfere with device types like iOS, macOS

Examples:

Chrome Extension
Before -> New Device Logged In From/Incomplete Two-Step Login From/Device Type: Chrome
After -> New Device Logged In From/Incomplete Two-Step Login From/Device Type: Chrome Extension

Poco F3 (shows up as M2012K11AG)
Before -> New Device Logged In From/Incomplete Two-Step Login From/Device Type: M2012K11AG
After -> New Device Logged In From/Incomplete Two-Step Login From/Device Type: Android

### DIFF
--- a/src/api/identity.rs
+++ b/src/api/identity.rs
@@ -265,7 +265,7 @@ async fn _password_login(
     let twofactor_token = twofactor_auth(&user, &data, &mut device, ip, conn).await?;
 
     if CONFIG.mail_enabled() && new_device {
-        if let Err(e) = mail::send_new_device_logged_in(&user.email, &ip.ip.to_string(), &now, &device.name).await {
+        if let Err(e) = mail::send_new_device_logged_in(&user.email, &ip.ip.to_string(), &now, &device).await {
             error!("Error sending new device email: {:#?}", e);
 
             if CONFIG.require_device_email() {
@@ -421,7 +421,7 @@ async fn _user_api_key_login(
 
     if CONFIG.mail_enabled() && new_device {
         let now = Utc::now().naive_utc();
-        if let Err(e) = mail::send_new_device_logged_in(&user.email, &ip.ip.to_string(), &now, &device.name).await {
+        if let Err(e) = mail::send_new_device_logged_in(&user.email, &ip.ip.to_string(), &now, &device).await {
             error!("Error sending new device email: {:#?}", e);
 
             if CONFIG.require_device_email() {
@@ -535,7 +535,7 @@ async fn twofactor_auth(
         return Ok(None);
     }
 
-    TwoFactorIncomplete::mark_incomplete(&user.uuid, &device.uuid, &device.name, ip, conn).await?;
+    TwoFactorIncomplete::mark_incomplete(&user.uuid, &device.uuid, &device.name, device.atype, ip, conn).await?;
 
     let twofactor_ids: Vec<_> = twofactors.iter().map(|tf| tf.atype).collect();
     let selected_id = data.two_factor_provider.unwrap_or(twofactor_ids[0]); // If we aren't given a two factor provider, assume the first one
--- a/src/db/models/two_factor_incomplete.rs
+++ b/src/db/models/two_factor_incomplete.rs
@@ -13,6 +13,7 @@ db_object! {
         // must complete 2FA login before being added into the devices table.
         pub device_uuid: String,
         pub device_name: String,
+        pub device_type: i32,
         pub login_time: NaiveDateTime,
         pub ip_address: String,
     }
@@ -23,6 +24,7 @@ impl TwoFactorIncomplete {
         user_uuid: &str,
         device_uuid: &str,
         device_name: &str,
+        device_type: i32,
         ip: &ClientIp,
         conn: &mut DbConn,
     ) -> EmptyResult {
@@ -44,6 +46,7 @@ impl TwoFactorIncomplete {
                     twofactor_incomplete::user_uuid.eq(user_uuid),
                     twofactor_incomplete::device_uuid.eq(device_uuid),
                     twofactor_incomplete::device_name.eq(device_name),
+                    twofactor_incomplete::device_type.eq(device_type),
                     twofactor_incomplete::login_time.eq(Utc::now().naive_utc()),
                     twofactor_incomplete::ip_address.eq(ip.ip.to_string()),
                 ))
--- a/src/db/schemas/mysql/schema.rs
+++ b/src/db/schemas/mysql/schema.rs
@@ -169,6 +169,7 @@ table! {
         user_uuid -> Text,
         device_uuid -> Text,
         device_name -> Text,
+        device_type -> Integer,
         login_time -> Timestamp,
         ip_address -> Text,
     }
--- a/src/mail.rs
+++ b/src/mail.rs
@@ -17,7 +17,7 @@ use crate::{
         encode_jwt, generate_delete_claims, generate_emergency_access_invite_claims, generate_invite_claims,
         generate_verify_email_claims,
     },
-    db::models::User,
+    db::models::{Device, DeviceType, User},
     error::Error,
     CONFIG,
 };
@@ -446,9 +446,8 @@ pub async fn send_invite_confirmed(address: &str, org_name: &str) -> EmptyResult
     send_email(address, &subject, body_html, body_text).await
 }
 
-pub async fn send_new_device_logged_in(address: &str, ip: &str, dt: &NaiveDateTime, device: &str) -> EmptyResult {
+pub async fn send_new_device_logged_in(address: &str, ip: &str, dt: &NaiveDateTime, device: &Device) -> EmptyResult {
     use crate::util::upcase_first;
-    let device = upcase_first(device);
 
     let fmt = "%A, %B %_d, %Y at %r %Z";
     let (subject, body_html, body_text) = get_text(
@@ -457,7 +456,8 @@ pub async fn send_new_device_logged_in(address: &str, ip: &str, dt: &NaiveDateTi
             "url": CONFIG.domain(),
             "img_src": CONFIG._smtp_img_src(),
             "ip": ip,
-            "device": device,
+            "device_name": upcase_first(&device.name),
+            "device_type": DeviceType::from_i32(device.atype).to_string(),
             "datetime": crate::util::format_naive_datetime_local(dt, fmt),
         }),
     )?;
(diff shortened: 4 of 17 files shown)


jjlin:
If the device type is Android or iOS, the mobile client should include a device model as the name, so the notification email should ideally include that info for those device types. I'm not sure if any other device types include useful model/name info.

dfunkt:
Let's not forget that the email template says "Device Type", not "Device Type and Name/Model" or anything like that.
Example: https://github.com/dani-garcia/vaultwarden/blob/248e561b3fe6a8172751374df980c6cd43c841d5/src/static/templates/email/new_device_logged_in.html.hbs#L22
Ideally I try to stick as close as possible to Bitwarden's behaviour when implementing these changes, and the emails from them don't specify the device name/model number either.

Nonetheless, let's see what others have to say about this change.

jjlin:
Vaultwarden does various things differently from Bitwarden, as long as they are compatible improvements. For example, Bitwarden does not even have a 2FA incomplete notification AFAIK.

BlackDex:
I agree with @jjlin.
For all the client functionality we should stay as close as possible.
Server-side specific changes should be looked at if it has no impact client side or might alter a way of working.

But for something like emails shouldn't be an issue. 
I actually love it shows we the device name and not only type.
So both would be better. 

I even was thinking about adding basic location detection based upon IP.

dfunkt:
New variant with device name and device type included.
Examples:
Screenshot from 2024-09-04 14-44-43
Screenshot from 2024-09-04 14-44-58

migrations/mysql/2024-09-04-091351_use_device_types/down.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE `twofactor_incomplete` DROP COLUMN `device_atype`;
```
jjlin, line 1:
```suggestion
ALTER TABLE `twofactor_incomplete` DROP COLUMN `device_type`;
```
Same name change for all the other migration files

src/db/models/two_factor_incomplete.rs
```diff
@@ -13,6 +13,7 @@ db_object! {
         // must complete 2FA login before being added into the devices table.
         pub device_uuid: String,
         pub device_name: String,
+        pub device_atype: i32,
```
jjlin, line 16:
```suggestion
        pub device_type: i32,
```

src/db/models/two_factor_incomplete.rs
```diff
@@ -23,6 +24,7 @@ impl TwoFactorIncomplete {
         user_uuid: &str,
         device_uuid: &str,
         device_name: &str,
+        device_atype: &i32,
```
jjlin, line 27:
```suggestion
        device_type: i32,
```

src/db/schemas/mysql/schema.rs
```diff
@@ -169,6 +169,7 @@ table! {
         user_uuid -> Text,
         device_uuid -> Text,
         device_name -> Text,
+        device_atype -> Integer,
```
jjlin, line 172:
```suggestion
        device_type -> Integer,
```
Same comment for the other schema files

src/api/identity.rs
```diff
@@ -535,7 +551,7 @@ async fn twofactor_auth(
         return Ok(None);
     }
 
-    TwoFactorIncomplete::mark_incomplete(&user.uuid, &device.uuid, &device.name, ip, conn).await?;
+    TwoFactorIncomplete::mark_incomplete(&user.uuid, &device.uuid, &device.name, &device.atype, ip, conn).await?;
```
jjlin, line 554:
```suggestion
    TwoFactorIncomplete::mark_incomplete(&user.uuid, &device.uuid, &device.name, device.type, ip, conn).await?;
```

dfunkt:
I chose atype on purpose to match how it was in device.rs:
https://github.com/dani-garcia/vaultwarden/blob/248e561b3fe6a8172751374df980c6cd43c841d5/src/db/models/device.rs#L19
Which one do you prefer @BlackDex? atype or just type?

BlackDex:
`type` is the correct name. The only reason there is an `a` in front is because you can't use that as a variable name.

jjlin:
> I chose atype on purpose to match how it was in device.rs:
> 
> https://github.com/dani-garcia/vaultwarden/blob/248e561b3fe6a8172751374df980c6cd43c841d5/src/db/models/device.rs#L19

This is only called `atype` because you can't have an identifier that's just called `type`. There are many other existing instances of `device_type` in the code.

dfunkt:
Hopefully things are good now

src/api/identity.rs
```diff
@@ -265,7 +265,15 @@ async fn _password_login(
     let twofactor_token = twofactor_auth(&user, &data, &mut device, ip, conn).await?;
 
     if CONFIG.mail_enabled() && new_device {
-        if let Err(e) = mail::send_new_device_logged_in(&user.email, &ip.ip.to_string(), &now, &device.name).await {
+        if let Err(e) = mail::send_new_device_logged_in(
+            &user.email,
+            &ip.ip.to_string(),
+            &now,
+            &device.name,
+            &DeviceType::from_i32(device.atype).to_string(),
```
BlackDex, line 273: I would suggest to just sent `&Device` instead of two different values from the same object.
So `&device` should be enough.

src/api/identity.rs
```diff
@@ -421,7 +429,15 @@ async fn _user_api_key_login(
 
     if CONFIG.mail_enabled() && new_device {
         let now = Utc::now().naive_utc();
-        if let Err(e) = mail::send_new_device_logged_in(&user.email, &ip.ip.to_string(), &now, &device.name).await {
+        if let Err(e) = mail::send_new_device_logged_in(
+            &user.email,
+            &ip.ip.to_string(),
+            &now,
+            &device.name,
+            &DeviceType::from_i32(device.atype).to_string(),
```
BlackDex, line 437: Same here as above of course

BlackDex (changes requested):

dfunkt:
@BlackDex Not sure if this is what you meant, but that's all I could come up with. I'm not particularly familiar with Rust (or programming in general tbh).

src/mail.rs
```diff
@@ -446,9 +446,10 @@ pub async fn send_invite_confirmed(address: &str, org_name: &str) -> EmptyResult
     send_email(address, &subject, body_html, body_text).await
 }
 
-pub async fn send_new_device_logged_in(address: &str, ip: &str, dt: &NaiveDateTime, device: &str) -> EmptyResult {
+pub async fn send_new_device_logged_in(address: &str, ip: &str, dt: &NaiveDateTime, device: &Device) -> EmptyResult {
     use crate::util::upcase_first;
-    let device = upcase_first(device);
+    let device_name = upcase_first(&device.name);
+    let device_type = DeviceType::from_i32(device.atype).to_string();
```
BlackDex, line 452: These do not have to be a variable.
They can be added directly to the `json!()` bellow on lines 461 and 462.

src/mail.rs
```diff
@@ -457,17 +458,24 @@ pub async fn send_new_device_logged_in(address: &str, ip: &str, dt: &NaiveDateTi
...
+pub async fn send_incomplete_2fa_login(
+    address: &str,
+    ip: &str,
+    dt: &NaiveDateTime,
+    device_name: &str,
+    device_type: &str,
+) -> EmptyResult {
     use crate::util::upcase_first;
-    let device = upcase_first(device);
+    let device_name = upcase_first(device_name);
```
BlackDex, line 478: Same here, move this directly into the `json!()` below in line 487.

BlackDex (changes requested):
Yes, that is what i mend indeed.

I just noticed something else too. Not that important, but no need to create new variables for those items actually.

src/mail.rs
dfunkt: Done

src/mail.rs
dfunkt: Done

src/mail.rs
```diff
@@ -457,17 +456,23 @@ pub async fn send_new_device_logged_in(address: &str, ip: &str, dt: &NaiveDateTi
             "url": CONFIG.domain(),
             "img_src": CONFIG._smtp_img_src(),
             "ip": ip,
-            "device": device,
```
dfunkt, line 460: Maybe I should modify this to ```&device.name``` as well in order to keep existing templates compatible?

src/mail.rs
```diff
@@ -476,7 +481,8 @@ pub async fn send_incomplete_2fa_login(address: &str, ip: &str, dt: &NaiveDateTi
...
```
dfunkt, line 479: Same here

src/mail.rs
BlackDex: Well, users will then miss the other variable. So they should now be triggered to adjust the templates.
We could add an extra item in the changelogs for this.

I would leave it as  you have right now from my point of view.

BlackDex (approved):

dani-garcia (approved):

gbe0:
With this merged I get the following on startup:

```
[2024-09-19 10:17:36.721][panic][ERROR] thread 'main' panicked at 'Error running migrations: QueryError(DieselMigrationName { name: "2024-09-04-091351_use_device_type_for_mails", version: MigrationVersion("20240904091351") }, DatabaseError(Unknown, "syntax error at or near \"`\""))': src/db/mod.rs:495
   1: std::panicking::rust_panic_with_hook
   2: std::panicking::begin_panic_handler::{{closure}}
   3: std::sys::backtrace::__rust_end_short_backtrace
   4: rust_begin_unwind
   5: core::panicking::panic_fmt
   7: vaultwarden::db::DbPool::from_config
   8: vaultwarden::main::{{closure}}
   9: vaultwarden::main
  10: std::sys::backtrace::__rust_begin_short_backtrace
  11: main
  12: <unknown>
  13: __libc_start_main
  14: _start
```

My DB is `PostgreSQL 16.4 on x86_64-pc-linux-gnu, compiled by gcc (Debian 12.2.0-14) 12.2.0, 64-bit`.

I rolled back and there are no issues.

Gerardv514:
@dfunkt i just checked this out. I have an iOS device and just tried to login to the web vault using Google chrome iOS app, i closed out on the 2fa page which prompted the 2fa incomplete notification. Device type and device name both say safari, is that normal?

dfunkt:
I think that might be due to Chrome using the browser engine as Safari on iOS. I expect this happens with Bitwarden as well, but I don't have any iOS devices in order to test this theory.

TL;DR: I think this is normal

Gerardv514:
I also attempted the iOS bitwarden app, and device name was iPhone 15,2 and device type was iOS. 

So as you said probably normal due to browser not passing the info along.

PR #4928, opened 2024-09-06
Update Rust version & crates

dfunkt:


Cargo.toml
BlackDex, line 1: Please also update the MSRV `rust-version` to a version which is N-2.
So that would be `1.79.0` in this case.

BlackDex (changes requested):
Thanks, just one remark i see this quickly.

BlackDex (approved):

PR #4937, opened 2024-09-09
Fix collection update from native client

BlackDex:
For some strange reason the native clients send `CollectionIds` instead of `collectionIds` for updating the collection. Added an alias to allow this.

Fixes an issue reported via Matrix.


dani-garcia (approved):

iosmanthus:
Shall we release a new version for it? Since the latest native client with version 2024.8.1(19099) could not connect to the vaultwarden v1.32.0.

BlackDex:
We will, but there still seem to be some users having issue connecting even when using a version with this patch. So before we release a new one, we want to tackle that too.

PR #4939, opened 2024-09-09
feat: Added Elestio as one-click deploy option

kaiwalyakoparkar:
Related discussion #4938


BlackDex:
This is more something for the wiki under the third-party options.
https://github.com/dani-garcia/vaultwarden/wiki/Third-party-packages

We will not promote any specific package on our Readme, especially packages or deployments not supported/build by this project it self.

PR #4957, opened 2024-09-15
Add x.com/twitter.com alias

rhysperry111:
Since twitter.com now redirects to x.com, accounts created on the old domain aren't detected automatically.

I just incremented the "type" number since that's what the rest of the file seemed to be doing, but if that's not correct then happy to change.

### DIFF
--- a/src/static/global_domains.json
+++ b/src/static/global_domains.json
@@ -971,5 +971,13 @@
       "pinterest.se"
     ],
     "excluded": false
+  },
+  {
+    "type": 91,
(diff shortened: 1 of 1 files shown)


BlackDex:
Is this in the Bitwarden repo too?
Else I would suggest to create a PR there.

We keep these files in sync with there repo.

rhysperry111:
Ah - didn't realise that, sorry.

They don't seem to have done anything on their side yet so I'll send a PR their way

PR #4972, opened 2024-09-20
Fix encrypted lastUsedDate

BlackDex:
It appears that some password histories have an encrypted value on the `lastUsedDate` Instead of only checking if it is a string, also check if it is a valid RFC Date/Time String. If not, set it also to epoch 0.


dani-garcia (approved):

BJReplay:
Retested and confirmed working.

Thanks @BlackDex for working with me on this edge case!

BlackDex:
@BJReplay thanks for sharing data. Without it out would not be possible to fix.

PR #4975, opened 2024-09-20
Fix keyword collision in Rust 2024 and add new api/config value

dani-garcia:
`disableUserRegistration` is still not handled by the clients but it should happen soon https://github.com/bitwarden/server/pull/4785/files


BlackDex (approved):
LGTM

PR #4977, opened 2024-09-21
Add extra linting

### LINKED ISSUE #4974: Rust 2024 breaking changes

N/A

### Vaultwarden Build Version

N/A

### Deployment method

Build from source

### Custom deployment method

_No response_

### Reverse Proxy

N/A

Other

### Operating System Version

_No response_

### Clients

Web Vault

### Client Version

_No response_

### Steps To Reproduce

Add `rust-2024-compatibility` to the list of `deny` lints and compile code. One error is caused by some of the macro definitions. For example, `crate::error::make_error` doesn't compile due to "the `expr` fragment specifier will accept more expressions in the 2024 edition".

### Expected Result

Compilation to succeed.

### Actual Result

Compilation failure.

### Logs

_No response_

### Screenshots or Videos

_No response_

### Additional Context

See [the Migration Guide](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/macro-fragment-specifiers.html) for how to fix this.

BlackDex:
Added extra linting for some code styles.
Also added some of the Rust Edition 2024 lints since our MSRV does not yet support the group lint.
Once we reach 1.81.0 as MSRV we can uncomment it and use that lint.

Closes #4974

### DIFF
--- a/src/api/core/accounts.rs
+++ b/src/api/core/accounts.rs
@@ -223,7 +223,7 @@ pub async fn _register(data: Json<RegisterData>, mut conn: DbConn) -> JsonResult
         }
 
         if verified_by_invite && is_email_2fa_required(data.organization_user_id, &mut conn).await {
-            drop(email::activate_email_2fa(&user, &mut conn).await);
+            email::activate_email_2fa(&user, &mut conn).await.ok();
         }
     }
 
@@ -232,7 +232,7 @@ pub async fn _register(data: Json<RegisterData>, mut conn: DbConn) -> JsonResult
     // accept any open emergency access invitations
     if !CONFIG.mail_enabled() && CONFIG.emergency_access_allowed() {
         for mut emergency_invite in EmergencyAccess::find_all_invited_by_grantee_email(&user.email, &mut conn).await {
-            drop(emergency_invite.accept_invite(&user.uuid, &user.email, &mut conn).await);
+            emergency_invite.accept_invite(&user.uuid, &user.email, &mut conn).await.ok();
         }
     }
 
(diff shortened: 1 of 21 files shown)


BlackDex:
Hmm, looks like our MSRV doesn't allow those lints yet.

BlackDex:
Only available since v1.81.0 it looks like.

BlackDex:
i Could comment it out for now and enable it once we are there, and leave the rest as-is.

src/api/core/accounts.rs
```diff
@@ -223,7 +223,7 @@ pub async fn _register(data: Json<RegisterData>, mut conn: DbConn) -> JsonResult
         }
 
         if verified_by_invite && is_email_2fa_required(data.organization_user_id, &mut conn).await {
-            let _ = email::activate_email_2fa(&user, &mut conn).await;
+            drop(email::activate_email_2fa(&user, &mut conn).await);
```
dani-garcia, line 226: Hmm I don't really like the use of drop here, I think this lint is a bit misguided for this use case.

The lint seems to want to make the calling of `Drop` more explicit, but that's not why we're using `let _ =` here, we're using it to indicate that we don't care about the returned `Result` type.

We could also switch to using `.ok()` instead, which I think we also use in other places.

dani-garcia (commented):
I think we can comment that `edition_2024_expr_fragment_specifier` out until we bump the MSRV in two months yeah.

src/api/core/accounts.rs
BlackDex: Yea, if we don't want to traverse the error, the  `.ok()` is good. I don't like `let _` for these cases either, and `.ok()` is more clear then on what we want.

dani-garcia (approved):

PR #5003, opened 2024-09-28
Add update from upstream

AndrewLauu:
Add Docker build
Add Docker images cleanup
Translate email @zituoguan/vaultwarden-lang-zh_CN


BlackDex:
Going to decline this. Probably a wrong button pressed.

BlackDex:
Also, all emails are in English and we currently do not have any translations in this project. 

And the workflows also do not seem to add any extra value

PR #5093, opened 2024-10-16
feat: add root to $PATH as vaultwarden binary is in root

calvin-li-developer:


calvin-li-developer:
I'm not sure how to force push so someone might need to help with that but this should work?

Aoang:
Adding the root directory to the `$PATH` is not the best approach. If necessary, linking to `/usr/local/bin` would be a better solution. However, since `/vaultwarden` is only one character longer than `vaultwarden`, this might not be necessary.

BlackDex:
I'm not sure we want to add this. As currently it's only one char to add extra as @Aoang metnions, And everything currently works this way, and maybe even might break current setups where people might have a different order of there PATH set too somehow.

It's better if we try to fix this in a better way when we will make more breaking changes then adding this.
There for, going to close this right now.

PR #5187, opened 2024-11-12
Support SSH keys on desktop 2024.12

dani-garcia:
Added support for the new SSH item key type in desktop 2024.12+

To test it, you need a desktop build from `bitwarden/clients:main`, and to enable the feature flags on server.

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -615,7 +615,6 @@ pub struct AdminHeaders {
     pub device: Device,
     pub user: User,
     pub org_user_type: UserOrgType,
-    pub client_version: Option<String>,
     pub ip: ClientIp,
 }
 
@@ -625,14 +624,12 @@ impl<'r> FromRequest<'r> for AdminHeaders {
 
     async fn from_request(request: &'r Request<'_>) -> Outcome<Self, Self::Error> {
         let headers = try_outcome!(OrgHeaders::from_request(request).await);
-        let client_version = request.headers().get_one("Bitwarden-Client-Version").map(String::from);
         if headers.org_user_type >= UserOrgType::Admin {
             Outcome::Success(Self {
                 host: headers.host,
                 device: headers.device,
                 user: headers.user,
                 org_user_type: headers.org_user_type,
-                client_version,
                 ip: headers.ip,
             })
         } else {
@@ -898,5 +895,26 @@ impl<'r> FromRequest<'r> for WsAccessTokenHeader {
         Outcome::Success(Self {
             access_token,
         })
+    }
+}
+
+pub struct ClientVersion(pub semver::Version);
+
+#[rocket::async_trait]
+impl<'r> FromRequest<'r> for ClientVersion {
+    type Error = &'static str;
+
+    async fn from_request(request: &'r Request<'_>) -> Outcome<Self, Self::Error> {
+        let headers = request.headers();
+
+        let Some(version) = headers.get_one("Bitwarden-Client-Version") else {
+            err_handler!("No Bitwarden-Client-Version header provided")
+        };
+
+        let Ok(version) = semver::Version::parse(version) else {
+            err_handler!("Invalid Bitwarden-Client-Version header provided")
+        };
+
+        Outcome::Success(ClientVersion(version))
     }
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -811,8 +811,15 @@ fn validate_config(cfg: &ConfigItems) -> Result<(), Error> {
     }
 
     // TODO: deal with deprecated flags so they can be removed from this list, cf. #4263
-    const KNOWN_FLAGS: &[&str] =
-        &["autofill-overlay", "autofill-v2", "browser-fileless-import", "extension-refresh", "fido2-vault-credentials"];
+    const KNOWN_FLAGS: &[&str] = &[
+        "autofill-overlay",
+        "autofill-v2",
+        "browser-fileless-import",
+        "extension-refresh",
+        "fido2-vault-credentials",
+        "ssh-key-vault-item",
+        "ssh-agent",
+    ];
     let configured_flags = parse_experimental_client_feature_flags(&cfg.experimental_client_feature_flags);
     let invalid_flags: Vec<_> = configured_flags.keys().filter(|flag| !KNOWN_FLAGS.contains(&flag.as_str())).collect();
     if !invalid_flags.is_empty() {
--- a/.env.template
+++ b/.env.template
@@ -349,6 +349,8 @@
 ## - "browser-fileless-import": Directly import credentials from other providers without a file.
 ## - "extension-refresh": Temporarily enable the new extension design until general availability (should be used with the beta Chrome extension)
 ## - "fido2-vault-credentials": Enable the use of FIDO2 security keys as second factor.
+## - "ssh-key-vault-item": Enable the creation and use of SSH key vault items. (Needs clients >=2024.12.0)
+## - "ssh-agent": Enable SSH agent support on Desktop. (Needs desktop >=2024.12.0)
 # EXPERIMENTAL_CLIENT_FEATURE_FLAGS=fido2-vault-credentials
 
 ## Require new device emails. When a user logs in an email is required to be sent.
--- a/src/api/core/accounts.rs
+++ b/src/api/core/accounts.rs
@@ -1,3 +1,5 @@
+use std::collections::HashSet;
+
 use crate::db::DbPool;
 use chrono::Utc;
 use rocket::serde::json::Json;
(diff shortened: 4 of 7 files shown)


src/auth.rs
```diff
@@ -900,3 +897,24 @@ impl<'r> FromRequest<'r> for WsAccessTokenHeader {
...
+
+    async fn from_request(request: &'r Request<'_>) -> Outcome<Self, Self::Error> {
+        let headers = request.headers();
+
+        let Some(version) = headers.get_one("Bitwarden-Client-Version") else {
+            err_handler!("No Bitwarden-Client-Version header provided")
+        };
+
+        let Ok(version) = semver::Version::parse(version) else {
+            err_handler!("Invalid Bitwarden-Client-Version header provided")
```
BlackDex, line 915: Should we error here or use some none existing very old version? Like `2018.1.0`? If some clients, like third-party clients do not provide this or even official one's it could break everything.

BlackDex (changes requested):
I have not tested it. But looks good. Just one remark.

src/auth.rs
dani-garcia: On the routes we always pass `Option<ClientVersion>` as parameter, which will be `None` if the `FromRequest` fails. I think it makes more sense for the `FromRequest` to fail, and then the routes can decide if they want to handle that or not.

BlackDex: Ah, ok if it's returning None, it's fine then.

dani-garcia: Yep, kinda cool! 
https://github.com/rwf2/Rocket/blob/master/core/lib/src/request/from_request.rs#L526-L532

src/config.rs
```diff
@@ -811,8 +811,15 @@ fn validate_config(cfg: &ConfigItems) -> Result<(), Error> {
...
     // TODO: deal with deprecated flags so they can be removed from this list, cf. #4263
-    const KNOWN_FLAGS: &[&str] =
-        &["autofill-overlay", "autofill-v2", "browser-fileless-import", "extension-refresh", "fido2-vault-credentials"];
+    const KNOWN_FLAGS: &[&str] = &[
+        "autofill-overlay",
+        "autofill-v2",
+        "browser-fileless-import",
+        "extension-refresh",
+        "fido2-vault-credentials",
+        "ssh-key-vault-item",
```
pquantin, line 820: It would be great to add the new values to the .env.template for documentation purpose

dani-garcia: Good point @pquantin, I always forget! Added in f0e3dcd

BlackDex:
> Nice, so Vaultwarden is only incompatible right now with recent web versions but all other clients fully work?

Once merged and released yes.

quexten:
Note: this might cause problems for key rotation. If the web client does not support parsing and serializing key ciphers, then key rotation might either fail or even end up with undecryptable items (if skipped), so rotation should be blocked.

I'm not sure about what vaultwarden does for validation here, but upstream has rotation validators, that block rotation entirely for cases like this: https://github.com/bitwarden/server/blob/dfbc400520f4c474b28e1279a86b2fae1da052a9/src/Api/Vault/Validators/CipherRotationValidator.cs#L10

BlackDex:
> Note: this might cause problems for key rotation. If the web client does not support parsing and serializing key ciphers, then key rotation might either fail or even end up with undecryptable items (if skipped), so rotation should be blocked.
> 
> I'm not sure about what vaultwarden does for validation here, but upstream has rotation validators, that block rotation entirely for cases like this: https://github.com/bitwarden/server/blob/dfbc400520f4c474b28e1279a86b2fae1da052a9/src/Api/Vault/Validators/CipherRotationValidator.cs#L10

It probably will and i think it will cause issues. 
We currently have some checks, but not the same specific check Bitwarden uses. Which is not that hard to add and would make it more safer.

BlackDex:
We could check if there are any `SSH` type ciphers, if that is the case, cancel/abort the rekey process directly.
But the cipher count would the probably help there too, since if we do not return the `SSH` ciphers, it will be aborted too.
But a better descriptive message in those cases would be better.

BlackDex (changes requested):
As mentioned in the comments already. Before we add this feature, we should add the same kind of checks for the key-rotation to prevent missing keys in the rotation, and make them unavailable anymore, and thus make all SSH Key's unusable.

It might be nice to add a more descriptive warning in case someone is using a newer client then 2024.6.2 to run the key-rotation if there are SSH Ciphers. But a more general error would also be a good start for verifying all ciphers are inculded.

And i think we should do the same for all other rotated items like emergency request, folders etc...

dani-garcia:
Nice, good call about the key rotation validation. I've improved the validation to match what Bitwarden is doing. We now check that we have all the ciphers, folders, sends, etc before doing the rotation

BlackDex (approved):
While i was not able to get a working Desktop v2024.12.0 version this looks to work great.

I tested it with a v2024.11.1 version which was build via Actions. But it was unable to either generate a key or import one from the clipboard.

Besides that, it did saved the ssh-cipher as an item and key-rotation failed with that item in the database as it should to prevent corruption. Without that cipher key-rotation worked fine.

quexten:
> I tested it with a v2024.11.1 version which was build via Actions. But it was unable to either generate a key or import one from the clipboard.

@BlackDex any errors? / what happened when you generated them? Also which build specifically (windows/mac/linux[what package])? If this is still a bug in upstream, I'd like to be able to reproduce it.

During development I've only tested against official server so far, but I don't think the server can make generation/import fail.

> While i was not able to get a working Desktop v2024.12.0 version this looks to work great.
Needs to be a local build at the moment, you can change the version in the package.json file for desktop and it should work.

dani-garcia:
Yeah, I've tested the client against this implementation and at least generation and saving worked fine. I didn't try to use them, but that is indendendent of the server.

If you're having problems building the desktop clients with the updated version, you can also comment out or reduce the version check on the server:
https://github.com/dani-garcia/vaultwarden/blob/2393c3f3c08f451a04643fd9fed5027f491dc12d/src/api/core/ciphers.rs#L121

And then you can use any recent builds from here:
https://github.com/bitwarden/clients/actions/workflows/build-desktop.yml?query=branch%3Amain

BlackDex:
@quexten. I used the AppImage build for Linux.
I'm not behind my computer right now, but it generated a lot of errors in the log/stdout/stderr.

It did this both via the generate and import.

@dani-garcia, i downgraden the version check to v2024.11.0 or greater, so that worked at least.

Not certain which specific build i used, but it was one from that list, just a bit older then the latest there right now.

I also enabled the ssh-agent checkbox in the settings, restarted, but it didn't helped.

I'm using Arch Linux with Gnome 46.x.
I did not looked any further yet though because of time.

BlackDex:
This has nothing to do with Rust. Those keys were generated with OpenSSH tools `ssh-keygen` which generates ed25519 by default. The password-less one worked just fine to try and import, but the one with a password failed.

And since the importing happens in the Desktop client it doesn't even save that, so the server isn't even called, and Vaultwarden just stores this as a blob of data, no parsing is done.

Your version has the exact same problem, but that is to be suspected since it only seems to change the version number, which wasn't an issue for me since i changed the version check in Vaultwarden it self.

BlackDex:
My comments still apply. It has nothing to do with Rust and keys generated using official OpenSSH tools also fail.
So, it probably is a bug in the Desktop which doesn't understand password protected keys, or somehow doesn't show a popup where to provide a password to verify and extract the public-key from it.

I also find it a bit strange the Desktop client doesn't provide an option to generate a password protected ssh-key btw.

quexten:
I'll debug these next week, thank you for the keys.

`clients` does not implement a password prompt UI yet, but the rust part supports `bcrypt-pbkdf + aes256-ctr` encrypted keys, so there should be an error message specifically stating that password protected keys are not supported yet. If they use another way of encrypting, then that is not yet supported in the ssh-key crate.

However, the unencrypted RSA key from @larena1  should work, but it also does not for me. @larena1 what tool were the keys generated with? (The ones used in the unit tests of the importer are generated with `ssh-keygen`).

BlackDex:
Ah, the rust part in the client. I didn't looked there :). That is also a bit confusing but clear now.

quexten:
@larena1 Putty encodes openssh-formatted private keys that are encoded differently to openssh. I don't know whether this is a putty bug or an `ssh-keys` bug, but tracking here: https://github.com/RustCrypto/SSH/issues/316

quexten:
Yeah, there is an internal ticket for pageant support. Unfortunately pageant is seemingly pretty nasty to implement in rust: https://github.com/Eugeny/russh/blob/main/pageant/src/pageant_impl.rs . Not sure what's required exactly there yet.

Svenum:
I have enabled the EXPERIMENTAL_FLAGS, but I dont see the ssh type. What did I wrong?

BlackDex:
Where are you checking this?
It is only visible in the Desktop client as far as i know.

Svenum:
I checked if the options appear in the admon panel and installed the newest desktop client from flathub (stable)

BlackDex:
The desktop version is still v2024.11.x as far as i can see. It's not yet v2024.12.x which is needed for SSH support.

DireMunchkin:
Hello! I also tried out the SSH keys feature with my Vaultwarden. However I wasn't able to get SSH agent to work.

Here's what I did:
1. enabled the feature flags at the server
2. downloaded desktop client 12.0 (AppImage on Linux)
3. made or imported a couple of new SSH key entries
4. enabled SSH-agent setting in the client settings menu
5. started SSH agent, restarted the client

If I understood it correctly I should now be able to see keys saved in Bitwarden using `ssh-add -L` in a separate terminal, and use it for sessions. But the agent doesn't have anything. Did I miss a step in setting it up or is it broken?

DireMunchkin:
OK cool, will wait for the fix upstream then. Thanks for clarifying!

PR #5214, opened 2024-11-20
Fix push not working

BlackDex:
The new native mobile clients seem to use PascalCase for the push payload. Also the date/time could cause issues.

This PR fixes this by formatting the date/time correctly and use PascalCase for the payload key's I now receive cipher updates and login-with-device requests again.

Fixes #5182

### DIFF
--- a/src/api/push.rs
+++ b/src/api/push.rs
@@ -9,6 +9,7 @@ use crate::{
     api::{ApiResult, EmptyResult, UpdateType},
     db::models::{Cipher, Device, Folder, Send, User},
     http_client::make_http_request,
+    util::format_date,
     CONFIG,
 };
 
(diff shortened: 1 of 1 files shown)


dani-garcia (approved):

tessus:
@BlackDex what happens when "old" clients are used? or do they ignore the case of the keys in the payload?

BlackDex:
@tessus, those also use PascalCase too.

tessus:
thx for the info

mada199122:
I still not receive push notification when I try to login with an other device.

BlackDex:
> I still not receive push notification when I try to login with an other device.

Please do not comment on closed PR's that it is not working.
Open a new discussion and provide your support string which you can generate via `/admin/diagnostics`

Because it could be multiple things.
- Wrong Bitwarden id, key or domain (US vs EU)
- Vaultwarden is unable to connect to the internet 
- mobile token is not known

PR #5219, opened 2024-11-21
Add partial role support for manager only using web-vault v2024.12.0

BlackDex:
- Add the custom role which replaces the manager role
- Added mini-details endpoint used by v2024.12.0

These changes try to add the custom role in such a way that it stays compatible with the older manager role. It will convert a manager role into a custom role, and if a manager has `access-all` rights, it will enable the correct custom roles. Upon saving it will convert these back to the old format.

What this does is making sure you are able to revert back to an older version of Vaultwarden without issues. This way we can support newer web-vault's and still be compatible with a previous Vaultwarden version if needed.

In the future this needs to be changed to full role support though.

### DIFF
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -48,6 +48,7 @@ pub fn routes() -> Vec<Route> {
         confirm_invite,
         bulk_confirm_invite,
         accept_invite,
+        get_org_user_mini_details,
         get_user,
         edit_user,
         put_organization_user,
@@ -323,6 +324,13 @@ async fn get_org_collections_details(org_id: &str, headers: ManagerHeadersLoose,
 
     // get all collection memberships for the current organization
     let coll_users = CollectionUser::find_by_organization(org_id, &mut conn).await;
+    // Generate a HashMap to get the correct UserOrgType per user to determine the manage permission
+    // We use the uuid instead of the user_uuid here, since that is what is used in CollectionUser
+    let users_org_type: HashMap<String, i32> = UserOrganization::find_confirmed_by_org(org_id, &mut conn)
+        .await
+        .into_iter()
+        .map(|uo| (uo.uuid, uo.atype))
+        .collect();
 
     // check if current user has full access to the organization (either directly or via any group)
     let has_full_access_to_org = user_org.access_all
@@ -336,11 +344,22 @@ async fn get_org_collections_details(org_id: &str, headers: ManagerHeadersLoose,
             || (CONFIG.org_groups_enabled()
                 && GroupUser::has_access_to_collection_by_member(&col.uuid, &user_org.uuid, &mut conn).await);
 
+        // Not assigned collections should not be returned
+        if !assigned {
+            continue;
+        }
+
         // get the users assigned directly to the given collection
         let users: Vec<Value> = coll_users
             .iter()
             .filter(|collection_user| collection_user.collection_uuid == col.uuid)
-            .map(|collection_user| SelectionReadOnly::to_collection_user_details_read_only(collection_user).to_json())
+            .map(|collection_user| {
+                SelectionReadOnly::to_collection_user_details_read_only(
+                    collection_user,
+                    *users_org_type.get(&collection_user.user_uuid).unwrap_or(&(UserOrgType::User as i32)),
+                )
+                .to_json()
+            })
             .collect();
 
         // get the group details for the given collection
@@ -645,12 +664,24 @@ async fn get_org_collection_detail(
                 Vec::with_capacity(0)
             };
 
+            // Generate a HashMap to get the correct UserOrgType per user to determine the manage permission
+            // We use the uuid instead of the user_uuid here, since that is what is used in CollectionUser
+            let users_org_type: HashMap<String, i32> = UserOrganization::find_confirmed_by_org(org_id, &mut conn)
+                .await
+                .into_iter()
+                .map(|uo| (uo.uuid, uo.atype))
+                .collect();
+
             let users: Vec<Value> =
                 CollectionUser::find_by_collection_swap_user_uuid_with_org_user_uuid(&collection.uuid, &mut conn)
                     .await
                     .iter()
                     .map(|collection_user| {
-                        SelectionReadOnly::to_collection_user_details_read_only(collection_user).to_json()
+                        SelectionReadOnly::to_collection_user_details_read_only(
+                            collection_user,
+                            *users_org_type.get(&collection_user.user_uuid).unwrap_or(&(UserOrgType::User as i32)),
+                        )
+                        .to_json()
                     })
                     .collect();
 
@@ -830,19 +861,36 @@ struct InviteData {
     collections: Option<Vec<CollectionData>>,
     #[serde(default)]
     access_all: bool,
+    #[serde(default)]
+    permissions: HashMap<String, Value>,
 }
 
 #[post("/organizations/<org_id>/users/invite", data = "<data>")]
 async fn send_invite(org_id: &str, data: Json<InviteData>, headers: AdminHeaders, mut conn: DbConn) -> EmptyResult {
-    let data: InviteData = data.into_inner();
+    let mut data: InviteData = data.into_inner();
 
-    let new_type = match UserOrgType::from_str(&data.r#type.into_string()) {
+    // HACK: We need the raw user-type be be sure custom role is selected to determine the access_all permission
+    // The from_str() will convert the custom role type into a manager role type
+    let raw_type = &data.r#type.into_string();
+    // UserOrgType::from_str will convert custom (4) to manager (3)
+    let new_type = match UserOrgType::from_str(raw_type) {
         Some(new_type) => new_type as i32,
         None => err!("Invalid type"),
     };
 
     if new_type != UserOrgType::User && headers.org_user_type != UserOrgType::Owner {
         err!("Only Owners can invite Managers, Admins or Owners")
+    }
+
+    // HACK: This converts the Custom role which has the `Manage all collections` box checked into an access_all flag
+    // Since the parent checkbox is not send to the server we need to check and verify the child checkboxes
+    // If the box is not checked, the user will still be a manager, but not with the access_all permission
+    if raw_type.eq("4")
+        && data.permissions.get("editAnyCollection") == Some(&json!(true))
+        && data.permissions.get("deleteAnyCollection") == Some(&json!(true))
+        && data.permissions.get("createNewCollections") == Some(&json!(true))
+    {
+        data.access_all = true;
     }
 
     for email in data.emails.iter() {
@@ -1254,7 +1302,21 @@ async fn _confirm_invite(
     save_result
 }
 
-#[get("/organizations/<org_id>/users/<org_user_id>?<data..>")]
+#[get("/organizations/<org_id>/users/mini-details", rank = 1)]
+async fn get_org_user_mini_details(org_id: &str, _headers: ManagerHeadersLoose, mut conn: DbConn) -> Json<Value> {
+    let mut users_json = Vec::new();
+    for u in UserOrganization::find_by_org(org_id, &mut conn).await {
+        users_json.push(u.to_json_mini_details(&mut conn).await);
+    }
+
+    Json(json!({
+        "data": users_json,
+        "object": "list",
+        "continuationToken": null,
+    }))
+}
+
+#[get("/organizations/<org_id>/users/<org_user_id>?<data..>", rank = 2)]
 async fn get_user(
     org_id: &str,
     org_user_id: &str,
@@ -1282,6 +1344,8 @@ struct EditUserData {
     groups: Option<Vec<String>>,
     #[serde(default)]
     access_all: bool,
+    #[serde(default)]
+    permissions: HashMap<String, Value>,
 }
 
 #[put("/organizations/<org_id>/users/<org_user_id>", data = "<data>", rank = 1)]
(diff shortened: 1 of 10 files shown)


BlackDex:
This PR needs a web-vault built from this PR https://github.com/dani-garcia/bw_web_builds/pull/186
I'm have yet to fix the CSS, since it now still hides some elements which need to be visible.

But it would be nice if some people are able to test this PR using a web-vault built from the PR linked above.

tessus:
> In the future this needs to be changed to full role support though.

May I suggest to add more comments to the code, where your changes from this PR have to be reverted when full role support is implemented. Unless you you'll keep the compat layer even then.

It is quite possible to forget to remove/change current code/behavior when the "future" changes are not close to this PR. Please ignore, if I am way off.

BlackDex:
@tessus done :)

@stefan0xC I have fixed the icon in the collapsed menu.

BlackDex:
As mentioned before, this PR needs a new web-vault version for which a PR can be found here:
 - https://github.com/dani-garcia/bw_web_builds/pull/186

CI builds of this web-vault can be downloaded here:
 - https://github.com/dani-garcia/bw_web_builds/actions/runs/12013212076 in the artifacts section.

tessus:
If this is the only PR required for the new web vault, is it possible to merge this and then release the new web vault - maybe this weekend?
I've been running it for a few days now and didn't run into any issues. Although I am sure I haven't tested all edge cases and every single functionality.

BlackDex:
I'd rather first fix a native client issue and release a version and after that merge this.

I also think it is better to first release the web-vault and then add that release to this PR.

dfunkt:
Not sure if this belongs here or in the web vault PR, but I seem to be offered the option to log in with a passkey (which obviously doesn't work):
image

BlackDex:
> Not sure if this belongs here or in the web vault PR, but I seem to be offered the option to log in with a passkey (which obviously doesn't work):

Thanks! I fixed it just now.

dfunkt:
Now that I think about it, shouldn't the create account line also be hidden? I have signups disabled.

BlackDex:
> Now that I think about it, shouldn't the create account line also be hidden? I have signups disabled.

Also fixed now :)

BlackDex:
I'm checking to see if we can update to v2024.12.0. There seem to be some minor changes and some endpoints added.

BlackDex:
Code is adjusted to support v2024.12.0 now. Not too much changed.

tessus:
> I'd rather first fix a native client issue and release a version and after that merge this.

The native client issue has been fixed and a new version has been released.

Is it possible to merge this now and release the new web-vault 2024.12.0 (or the other way around as you suggested)?

tessus:
> If not then just try to build and run it yourself

I already do. If you had read my earlier comments, you would have known.

stefan0xC:
@BlackDex the [endpoint `/api/organizations/<org_id>/groups/details`](https://github.com/bitwarden/clients/blob/b7ec74df5aecd0efe8bb4c6433f279d437a135c3/apps/web/src/app/admin-console/organizations/core/services/group/group-api.service.ts#L52-L65) seems to be missing, which appears to break my ability to edit users with `web-v2024.12.0`.
image

BlackDex:
@stefan0xC seems to work for me:
```
[response][INFO] (get_groups_details) GET /api/organizations/<org_id>/groups/details => 200 OK
```

Be sure to pull in the latest commit and are on:
```
git rev-parse HEAD
748a24219ed2b3d74b72ff6c12d9bad7f2fb3cc6
```

stefan0xC:
@BlackDex if the group has `accessAll` it works
```json
{
  "accessAll": true,
  "collections": [],
  "externalId": null,
  "id": "41917ddc-7cf7-47d0-bd7f-7b24d21c2b71",
  "name": "manager group",
  "object": "groupDetails",
  "organizationId": "152bd996-19f9-40ee-b991-e52333a7e719"
}
```

but it fails when the group has permissions for individual collections:
```json
{
  "accessAll": false,
  "collections": [
    {
      "hidePasswords": true,
      "id": "f39e653f-3191-4312-887d-931889be5d20",
      "manage": false,
      "readOnly": false
    },
    {
      "hidePasswords": true,
      "id": "ca4b77ab-c849-47f4-a6f3-c4ebde35d5f0",
      "manage": false,
      "readOnly": true
    }
  ],
  "externalId": null,
  "id": "41917ddc-7cf7-47d0-bd7f-7b24d21c2b71",
  "name": "manager group",
  "object": "groupDetails",
  "organizationId": "152bd996-19f9-40ee-b991-e52333a7e719"
}
```

BlackDex:
@stefan0xC What happens then exactly?
It doesn't work? And is it mixed with with an accessAll?

stefan0xC:
> @stefan0xC What happens then exactly? It doesn't work? And is it mixed with with an accessAll?

The exception happened when I tried to open the Custom user member (which used to be a Manager role without `accessAll`) or the group. But now after changing the group I can't seem to be able to replicate the issue anymore.

src/static/templates/scss/vaultwarden.scss.hbs
```diff
@@ -58,29 +52,63 @@ app-organization-plans > form > bit-section:nth-child(2) {
...
+}
+bit-nav-logo bit-nav-item .bwi-shield {
+  @extend %vw-hide;
+}
 /**** END Static Vaultwarden Changes ****/
 /**** START Dynamic Vaultwarden Changes ****/
 {{#if signup_disabled}}
 /* Hide the register link on the login screen */
-app-frontend-layout > app-login > form > div > div > div > p {
+app-login form p {
```
stefan0xC, line 97: This will also hide the `Logging in as bla@example.com` on the password field.
image

dfunkt: Is this even needed anymore since v2024.12.0 contains commit https://github.com/bitwarden/clients/commit/f5e6fc8ed5a4756d7866ae05118ff795ec68a374?

BlackDex: It should, but somehow this doesn't work.

BlackDex: Fix this now, also hiding the `<hr>` since that is used to divide the text and the button.

stefan0xC (changes requested):
Found a small issue.

src/api/core/organizations.rs
```diff
@@ -323,6 +326,13 @@ async fn get_org_collections_details(org_id: &str, headers: ManagerHeadersLoose,
 
     // get all collection memberships for the current organization
     let coll_users = CollectionUser::find_by_organization(org_id, &mut conn).await;
+    // Generate a HashMap to get the correct UserOrgType per user to determine the manage permission
+    // We use the uuid instead of the user_uuid here, since that is what is used in CollectionUser
+    let users_org_type: HashMap<String, i32> = UserOrganization::find_confirmed_by_org(org_id, &mut conn)
+        .await
+        .into_iter()
+        .map(|uo| (uo.uuid, uo.atype))
```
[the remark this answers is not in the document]
tessus: I've been reading the code a bit and I think more comments would make sense.

e.g. in the atruct `UserOrganization`, what's the `uuid` and `user_uuid` for? Someone can only understand that, if they read and understand the entire code.

And as you have shown here, the usage of these members is apparently not always that clear.

Don't take this the wrong way, I'd add the comments myself, but I doubt I understand all the different abstraction layers and levels in their entirety. I am also not saying that you should add those comments, but that I'd like to see more comments that explain certain connections and layers.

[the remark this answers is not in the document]
BlackDex: I'm almost certain i changed this to have it working at all because it wasn't. I'll have to revisit it.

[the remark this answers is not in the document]
stefan0xC: Well, I did not test if it still works. I just noticed the discrepancy when trying to apply your changes upon my newtype branch (#5320) because it would not even compile with the wrong id.

@tessus I'm not sure if comments would help. As I've shown they can be wrong or misleading. But to answer your question: Generally the `uuid` refers to an entity itself (e.g. the `UserOrganization` relation or whatever is stored in the `users_organizations` table) while it does not with a prefix (e.g. `user_uuid` refers to a `User` entity). I think the database schema is more consistent about this than the rest of the code base.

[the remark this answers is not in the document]
stefan0xC: Here is an visualization of the relevant database relations
output9

Created with [visualize-sqlite](https://github.com/UhhhWaitWhat/visualize-sqlite/) and dot after dropping the other tables from a fresh `db.sqlite3`:
```bash
for t in __diesel_schema_migrations attachments auth_requests ciphers ciphers_collections devices emergency_access event favorites folders folders_ciphers invitations org_policies organization_api_key sends twofactor twofactor_duo_ctx twofactor_incomplete
do 
    sqlite3 data/db.sqlite3 "drop table $t"
    visualize-sqlite data/db.sqlite3  | dot -Tpng > image.png
done
    visualize-sqlite data/db.sqlite3  | dot -Tpng > image.png
```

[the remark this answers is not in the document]
tessus: Ah, ok. Thanks for that. I was wondering why an n:m table would have its own uuid. However looking at the diagram, the naming is rather inconsistent and a bit confusing.

src/api/core/organizations.rs
```diff
@@ -336,11 +346,22 @@ async fn get_org_collections_details(org_id: &str, headers: ManagerHeadersLoose,
...
+
         // get the users assigned directly to the given collection
         let users: Vec<Value> = coll_users
             .iter()
             .filter(|collection_user| collection_user.collection_uuid == col.uuid)
-            .map(|collection_user| SelectionReadOnly::to_collection_user_details_read_only(collection_user).to_json())
+            .map(|collection_user| {
+                SelectionReadOnly::to_collection_user_details_read_only(
+                    collection_user,
+                    *users_org_type.get(&collection_user.user_uuid).unwrap_or(&(UserOrgType::User as i32)),
```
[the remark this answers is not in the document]
stefan0xC: Okay I looked further into it and I think part of what makes it tricky is that we have two functions that retrieve a `CollectionUser` with an `user_uuid` pointing to  `users_organization.uuid` instead of `users.uuid` which is probably why this is so confusing:
https://github.com/dani-garcia/vaultwarden/blob/ed4ad67e732c213beaec78970cdb68e48bee3dc1/src/db/models/collection.rs#L514-L521

https://github.com/dani-garcia/vaultwarden/blob/ed4ad67e732c213beaec78970cdb68e48bee3dc1/src/db/models/collection.rs#L610-L619

src/api/core/organizations.rs
```diff
@@ -645,12 +666,24 @@ async fn get_org_collection_detail(
                 Vec::with_capacity(0)
             };
 
+            // Generate a HashMap to get the correct UserOrgType per user to determine the manage permission
+            // We use the uuid instead of the user_uuid here, since that is what is used in CollectionUser
+            let users_org_type: HashMap<String, i32> = UserOrganization::find_confirmed_by_org(org_id, &mut conn)
+                .await
+                .into_iter()
+                .map(|uo| (uo.uuid, uo.atype))
```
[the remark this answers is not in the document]
stefan0xC: Or not.

src/api/core/organizations.rs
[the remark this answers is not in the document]
BlackDex: @stefan0xC i have renamed the function to also mention the swap, that should for now make it clear.
But this should be revisited i think in the future.

BlackDex:
@stefan0xC & @dfunkt, i have been testing a bit more, but i can't find anything strange my self.
If you have anything which should be solved or issues encountered, please let me know, else will try to merge this soonisch.

stefan0xC (approved):
I don't think I can find anything else. I mean that a Manager that does not have access to all collection will not have access to the admin console is probably expected. One question I would have left is if we should not make this a permanent change  and save the `manage` permission in the CollectionUser / GroupUser relations accordingly? Since that seems a bit weird to just keep deriving it. But I wonder if that's possible without implementing custom permissions as well.

BlackDex:
I'd rather do that while implementing the other rights too.
Because it also need some auth/header rights changes too.

dfunkt (approved):
I've been using this PR for the past few weeks, no further problems to report.

Gerardv514 (changes requested):
to

src/api/core/organizations.rs
```diff
@@ -1303,14 +1369,30 @@ async fn edit_user(
     headers: AdminHeaders,
     mut conn: DbConn,
 ) -> EmptyResult {
-    let data: EditUserData = data.into_inner();
+    let mut data: EditUserData = data.into_inner();
 
-    let Some(new_type) = UserOrgType::from_str(&data.r#type.into_string()) else {
+    // HACK: We need the raw user-type be be sure custom role is selected to determine the access_all permission
```
Gerardv514, line 1374:
```suggestion
    // HACK: We need the raw user-type to be sure custom role is selected to determine the access_all permission
```

Gerardv514 (changes requested):

src/api/core/organizations.rs
BlackDex: thx fixed

src/api/core/organizations.rs
```diff
@@ -845,6 +884,17 @@ async fn send_invite(org_id: &str, data: Json<InviteData>, headers: AdminHeaders
         err!("Only Owners can invite Managers, Admins or Owners")
     }
 
+    // HACK: This converts the Custom role which has the `Manage all collections` box checked into an access_all flag
+    // Since the parent checkbox is not send to the server we need to check and verify the child checkboxes
```
Gerardv514, line 888:
```suggestion
    // Since the parent checkbox is not sent to the server we need to check and verify the child checkboxes
```

src/api/core/organizations.rs
```diff
@@ -1303,14 +1369,30 @@ async fn edit_user(
...
+    let raw_type = &data.r#type.into_string();
+    // UserOrgType::from_str will convert custom (4) to manager (3)
+    let Some(new_type) = UserOrgType::from_str(raw_type) else {
         err!("Invalid type")
     };
 
-    let Some(mut user_to_edit) = UserOrganization::find_by_uuid_and_org(org_user_id, org_id, &mut conn).await else {
-        err!("The specified user isn't member of the organization")
+    // HACK: This converts the Custom role which has the `Manage all collections` box checked into an access_all flag
+    // Since the parent checkbox is not send to the server we need to check and verify the child checkboxes
```
Gerardv514, line 1383:
```suggestion
    // Since the parent checkbox is not sent to the server we need to check and verify the child checkboxes
```

src/db/models/organization.rs
```diff
@@ -608,6 +648,29 @@ impl UserOrganization {
             "object": "organizationUserDetails",
         })
     }
+
+    pub async fn to_json_mini_details(&self, conn: &mut DbConn) -> Value {
+        let user = User::find_by_uuid(&self.user_uuid, conn).await.unwrap();
+
+        // Because BitWarden want the status to be -1 for revoked users we need to catch that here.
```
Gerardv514, line 655:
```suggestion
        // Because BitWarden wants the status to be -1 for revoked users we need to catch that here.
```

Gerardv514 (changes requested):

dani-garcia (approved):

PR #5225, opened 2024-11-23
Adding token in the subject of 2fa emails

frdescam:
This PR Adds the 2FA token in the subject of 2FA emails. This allows user get the code more easily as it will be shown in the notification of most email mobile apps.

Example :

photo_2024-11-23_11-49-43

photo_2024-11-23_11-48-44


BlackDex:
I do not find this very secure, and it's also not best practices to do this in my opinion.

If you want this your self, you can always adjust the templates locally. But I'm going to decline this.

PR #5240, opened 2024-11-27
allow admin token as auth header

farodin91:


BlackDex:
Hmm. I'm not fond of this way to honest. If you want something like this it might be better to let it handle this via the reverse proxy like basic auth or something.

Using a plain text password is not very secure this way.

farodin91:
I found a different Solution.

PR #5248, opened 2024-11-29
Update Rust and crates

BlackDex:
- Updated Rust to v1.83.0
- Updated MSRV to v1.82.0 (Needed for html5gum crate)
- Updated icon fetching code to match new html5gum version
- Updated workflows
- Enabled edition 2024 clippy lints Nightly reports some clippy hints, but that would be too much to change in this PR i think.
- Added crate patches for the fern and yubico crates to update more dependencies
- Fixed an issue with the diesel_logger in combination with the SQLite backup feature

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,14 +67,14 @@ dashmap = "6.1.0"
 
 # Async futures
 futures = "0.3.31"
-tokio = { version = "1.41.1", features = ["rt-multi-thread", "fs", "io-util", "parking_lot", "time", "signal", "net"] }
+tokio = { version = "1.42.0", features = ["rt-multi-thread", "fs", "io-util", "parking_lot", "time", "signal", "net"] }
 
 # A generic serialization/deserialization framework
 serde = { version = "1.0.215", features = ["derive"] }
 serde_json = "1.0.133"
 
 # A safe, extensible ORM and Query builder
-diesel = { version = "2.2.5", features = ["chrono", "r2d2", "numeric"] }
+diesel = { version = "2.2.6", features = ["chrono", "r2d2", "numeric"] }
 diesel_migrations = "2.2.0"
 diesel_logger = { version = "0.4.0", optional = true }
 
@@ -91,7 +91,7 @@ uuid = { version = "1.11.0", features = ["v4"] }
 # Date and time libraries
 chrono = { version = "0.4.38", features = ["clock", "serde"], default-features = false }
 chrono-tz = "0.10.0"
-time = "0.3.36"
+time = "0.3.37"
 
 # Job scheduler
 job_scheduler_ng = "2.0.5"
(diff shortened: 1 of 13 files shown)


BlackDex:
Looks like the Rust v1.83.0 images are not yet available

dfunkt:
Perhaps we could also use commit https://github.com/daboross/fern/commit/3e775ccfafe7d24baee39826d38011981b2e55b5 as an override for fern in order to have `syslog-7` support?

BlackDex:
@dfunkt, yea that might be a good one.
I have updated it, and fixed an issue with the query_logger (diesel-logger) feature too.

Looks like the image is still not available :(.

BlackDex:
We could keep the docker files on v1.82.0, the correct version will be pulled anyway.

BlackDex:
@dani-garcia, we could also patch yubikey-rs which causes several duplicate dependencies to be removed.
Mainly reqwest and all it's dependencies.

It would then be based from my forked repo where the PR is coming from.
```toml
yubico = { git = "https://github.com/BlackDex/yubico-rs", rev = "00df14811f58155c0f02e3ab10f1570ed3e115c6" }
```

Cargo.toml
```diff
@@ -166,6 +166,11 @@ rpassword = "7.3.1"
 # Loading a dynamic CSS Stylesheet
 grass_compiler = { version = "0.13.4", default-features = false }
 
+[patch.crates-io]
+# Patch fern to support syslog v7
+fern = { git = "https://github.com/daboross/fern", rev = "3e775ccfafe7d24baee39826d38011981b2e55b5" }
+yubico = { git = "https://github.com/BlackDex/yubico-rs", rev = "00df14811f58155c0f02e3ab10f1570ed3e115c6" }
```
dfunkt, line 172:
```suggestion

# Patch yubico to remove duplicate dependencies
yubico = { git = "https://github.com/BlackDex/yubico-rs", rev = "00df14811f58155c0f02e3ab10f1570ed3e115c6" }
```

dfunkt (changes requested):

src/main.rs
```diff
@@ -201,25 +201,20 @@ fn parse_args() {
...
-        } else {
-            err_silent!("The database type is not SQLite. Backups only works for SQLite databases")
-        }
-    }
-    #[cfg(not(sqlite))]
-    {
-        err_silent!("The 'sqlite' feature is not enabled. Backups only works for SQLite databases")
+async fn backup_sqlite() -> Result<String, Error> {
+    use crate::db::{backup_database, DbConnType};
+    if DbConnType::from_url(&CONFIG.database_url()).map(|t| t == DbConnType::sqlite).unwrap_or(false) {
```
dani-garcia, line 206: Do we need to check for the database type here? We're already doing that inside `backup_database` as far as I can see.

dani-garcia (approved):
Yeah, patching the dependency for now sounds good, we don't want to be dragging lots of old transitive dependencies if we can avoid it.

PR #5259, opened 2024-12-06
Fix another sync issue with native clients

### LINKED ISSUE #5237: Android bitwarden app stopped having access to vault
* Vaultwarden version: v1.32.5
* OS/Arch: linux/x86_64
* Uses a reverse proxy: true
* Internet access: true
* DNS Check: true
* Server/NTP Time Check: true
* HTTPS Check: false
* Database type: SQLite
* Database version: 3.46.0
* Clients used: 
* Reverse proxy and version: 
* Other relevant information: 

```json
{
  "_duo_akey": null,
  "_enable_duo": true,
  "_enable_email_2fa": false,
  "_enable_smtp": true,
  "_enable_yubico": true,
  "_icon_service_csp": "",
  "_icon_service_url": "",
  "_ip_header_enabled": true,
  "_max_note_size": 10000,
  "_smtp_img_src": "cid:",
  "admin_session_lifetime": 20,
  "admin_token": "***",
  "data_folder": "data",
  "database_conn_init": "",
  "database_max_conns": 10,
  "database_timeout": 30,
  "db_connection_retries": 15,
  "disable_admin_token": false,
  "domain": "****://*********",
  "domain_origin": "****://*********",
  "domain_path": "",
  "domain_set": false,
  "duo_host": null,
  "duo_ikey": null,
  "duo_skey": null,
  "duo_use_iframe": false,
  "email_attempts_limit": 3,
  "email_change_allowed": true,
  "email_expiration_time": 600,
  "email_token_size": 6,
  "enable_db_wal": true,
  "enable_websocket": true,
  "events_days_retain": null,
  "extended_logging": true,
  "helo_name": null,
  "hibp_api_key": null,
  "icon_blacklist_regex": null,
  "icon_cache_negttl": 259200,
  "icon_cache_ttl": 2592000,
  "icon_download_timeout": 10,
  "icon_redirect_code": 302,
  "icon_service": "internal",
  "invitations_allowed": false,
  "ip_header": "X-Real-IP",
  "log_file": "data/vw.log",
  "log_level": "warn",
  "org_attachment_limit": null,
  "org_creation_users": "",
  "org_events_enabled": false,
  "org_groups_enabled": false,
  "push_enabled": false,
  "reload_templates": false,
  "sendmail_command": null,
  "sends_allowed": true,
  "sends_folder": "data/sends",
  "show_password_hint": false,
  "signups_allowed": false,
  "signups_verify": false,
  "smtp_auth_mechanism": null,
  "smtp_debug": false,
  "smtp_embed_images": true,
  "smtp_explicit_tls": null,
  "smtp_from": "",
  "smtp_host": null,
  "smtp_password": null,
  "smtp_port": 587,
  "smtp_security": "starttls",
  "smtp_ssl": null,
  "smtp_timeout": 15,
  "smtp_username": null,
  "tmp_folder": "data/tmp",
  "use_sendmail": false,
  "use_syslog": false,
  "user_send_limit": null,
  "web_vault_enabled": true,
  "yubico_client_id": "'removed'",
  "yubico_secret_key": "***",
  "yubico_server": null
}
```
</details>

### Vaultwarden Build Version

Version 2024.6.2

### Deployment method

Official Container Image

### Custom deployment method

No: Plain docker-compose, caddy, fail2ban

### Reverse Proxy

image: caddy:2

Linux

### Operating System Version

Ubuntu 22.04.5 LTS

### Clients

Android

### Client Version

latest bitwarden

### Steps To Reproduce

1. Open Bitwarden
2. Enter passcode
3. or log out and Enter Masterpassword

### Expected Result

Successful login

### Actual Result

We were unable to process your request. Please try again or contact us.

### Logs

```text
I have not managed enabling logging yet. I am sure that I will need get logging enabled to get some helpful insight.
```

### Screenshots or Videos

I can log into the vault from any other OS, web browser and extension.
This problem existed a few months ago and disappeared  (reboot phone)
This time I signed up for bitwarden beta which has no newer release and also tried uninstall and reinstall

### Additional Context

It seems that I am logged in as I can change settings but there is no vault access.

BlackDex:
The `reprompt` value somehow sometimes has a value of `4`. This isn't a valid value, and doesn't cause issues with other clients, but the native clients are more strict.

This commit fixes this by validating the value before storing and returning.

Fixes #5237

### DIFF
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -511,7 +511,7 @@ pub async fn update_cipher_from_data(
     cipher.fields = data.fields.map(|f| _clean_cipher_data(f).to_string());
     cipher.data = type_data.to_string();
     cipher.password_history = data.password_history.map(|f| f.to_string());
-    cipher.reprompt = data.reprompt;
+    cipher.reprompt = data.reprompt.filter(|r| *r == RepromptType::None as i32 || *r == RepromptType::Password as i32);
 
     cipher.save(conn).await?;
     cipher.move_to_folder(data.folder_id, &headers.user.uuid, conn).await?;
--- a/src/db/models/cipher.rs
+++ b/src/db/models/cipher.rs
@@ -46,10 +46,9 @@ db_object! {
     }
 }
 
-#[allow(dead_code)]
 pub enum RepromptType {
     None = 0,
-    Password = 1, // not currently used in server
+    Password = 1,
 }
 
 /// Local methods
@@ -296,7 +295,7 @@ impl Cipher {
             "creationDate": format_date(&self.created_at),
             "revisionDate": format_date(&self.updated_at),
             "deletedDate": self.deleted_at.map_or(Value::Null, |d| Value::String(format_date(&d))),
-            "reprompt": self.reprompt.unwrap_or(RepromptType::None as i32),
+            "reprompt": self.reprompt.filter(|r| *r == RepromptType::None as i32 || *r == RepromptType::Password as i32).unwrap_or(RepromptType::None as i32),
             "organizationId": self.organization_uuid,
             "key": self.key,
             "attachments": attachments_json,
--- a/src/db/models/mod.rs
+++ b/src/db/models/mod.rs
@@ -18,7 +18,7 @@ mod user;
 
 pub use self::attachment::Attachment;
 pub use self::auth_request::AuthRequest;
-pub use self::cipher::Cipher;
+pub use self::cipher::{Cipher, RepromptType};
 pub use self::collection::{Collection, CollectionCipher, CollectionUser};
 pub use self::device::{Device, DeviceType};
 pub use self::emergency_access::{EmergencyAccess, EmergencyAccessStatus, EmergencyAccessType};


BlackDex:
In theory we should change the column type from Nullable to not Nullable and have all `NULL` values converted to `0`.
But that might cause other issues. Also, since there are more sync issues which we might need to fix on the database side, i think it's better to create some kind of convert/migrate feature to fix all these items in one go in the future once we have identified all possible issues.

jjlin:
> In theory we should change the column type from Nullable to not Nullable and have all `NULL` values converted to `0`. But that might cause other issues. Also, since there are more sync issues which we might need to fix on the database side, i think it's better to create some kind of convert/migrate feature to fix all these items in one go in the future once we have identified all possible issues.

Why would it help to make it not nullable? If the DB value is null, it already gets converted to  `RepromptType::None` AFAICT.

It seems to me it would be cleaner to add `num_derive::FromPrimitive` on `RepromptType` and do a general validity check on the reprompt value, logging any invalid values that are detected. Although I wouldn't be terribly surprised if people are getting invalid values due to disk corruption or something (maybe running off an SD card?) and the issue only manifests now because the mobile clients are stricter.

BlackDex:
I do not understand why to keep something nullable when it shouldn't.

BlackDex:
Since Bitwarden doesn't have it nullable anymore kinda indicates it has a default of 0/None already. Also, as that is the sane default i do not expect it to be changed. Else we could make all lot more Nullable of course. While that might also be beneficial for th database maybe regarding storage, ot doesn't make it easier to query i, it will then always need two compares.

Anyways, it doesn't really matter for now. More for the fixing the database defaults, and fix formats instead of doing that during sync all the time.

Logging invalid values btw doesn't really help here. It might have been broken clients or even non official clients which provided wrong values. It just needs to be fixed.

BlackDex:
Took some look again. But adding `num_derive::FromPrimitive` for this seems a bit of overkill. Also casting from isize(i8) to i64 again to i32 seems a bit of overkill too me, while it probably something which will not have that much of an impact. I think the code as is looks fine too me.

BlackDex:
I also did some more detailed checking by validating https://github.com/bitwarden/android/blob/v2024.11.7/app/src/main/java/com/x8bit/bitwarden/data/vault/datasource/network/model/SyncResponseJson.kt with our code, and i think it is fine.

The only thing i can see which is not the same is the attachment/send size, which we send as a String instead of an Int. But the none-native clients expected a String, and i think it will be converted correctly by the native clients.

dani-garcia (approved):

PR #5260, opened 2024-12-07
refactor: replace static with const for global constants

Integral-Tech:


dani-garcia:
Thanks! Seems like the const support for this was stabilized in Rust 1.83, so we need to wait until we update our minimum supported rust version to 1.83. 

We're on [1.82.0](https://github.com/dani-garcia/vaultwarden/blob/c9860af11ccedfa9b347e22652d75b862eeba391/Cargo.toml#L6) right now, so when 1.84 is released in January this can be merged.

tessus:
@dani-garcia the pipelines require an approval to run

dani-garcia:
Weird, previously Github used to have a big button telling me about the required approval, while now I have to go fishing for it on the Actions page, which is why I didn't notice it this time. They should be running now.

dani-garcia (approved):
Thanks @Integral-Tech this is ready for merging!

PR #5268, opened 2024-12-10
Update crates

dfunkt:
- fixes CVE-2024-12224


BlackDex (approved):
Thanks.

This also removes a duplicate dep on the idna crate and now only one version is used.

PR #5272, opened 2024-12-10
Some Backend Admin fixes and updates

BlackDex:
- Updated datatables
- Added a `X-Robots-Tags` header to prevent indexing
- Modified some layout settings
- Added Websocket check to diagnostics
- Added Security Header checks to diagnostics
- Added Error page response checks to diagnostics
- Modifed support string layout a bit

### DIFF
--- a/src/static/scripts/admin_diagnostics.js
+++ b/src/static/scripts/admin_diagnostics.js
@@ -7,6 +7,8 @@ var timeCheck = false;
 var ntpTimeCheck = false;
 var domainCheck = false;
 var httpsCheck = false;
+var websocketCheck = false;
+var httpResponseCheck = false;
 
 // ================================
 // Date & Time Check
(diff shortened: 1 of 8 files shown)


src/static/scripts/admin_diagnostics.js
```diff
@@ -199,6 +219,162 @@ function checkDns(dns_resolved) {
...
+        return { headers: response.headers, status: response.status, text: await response.text() };
+    } catch (error) {
+        console.error(`Error fetching ${url}: ${error}`);
+        return { error };
+    }
+}
+
+function checkSecurityHeaders(headers, omit) {
+    let securityHeaders = {
+        "x-frame-options": ["SAMEORIGIN"],
```
dani-garcia, line 234: Probably a good idea to add a comment in `util.rs` where we declare these headers mentioning that if they are modified they also need to be updated here.

src/static/scripts/admin_diagnostics.js
```diff
@@ -199,6 +219,162 @@ function checkDns(dns_resolved) {
...
+        }
+    }
+    return messages;
+}
+
+async function checkHttpResponse() {
+    const [apiConfig, webauthnConnector, notFound, notFoundApi, badRequest, unauthorized, forbidden] = await Promise.all([
+        fetchCheckUrl(`${BASE_URL}/api/config`),
+        fetchCheckUrl(`${BASE_URL}/webauthn-connector.html`),
+        fetchCheckUrl(`${BASE_URL}/admin/does-not-exists`),
```
dani-garcia, line 279: Doesn't really matter, but typo
```suggestion
        fetchCheckUrl(`${BASE_URL}/admin/does-not-exist`),
```

dani-garcia (commented):
LGTM, just some small comments

src/static/scripts/admin_diagnostics.js
BlackDex: I mend to do that, but forgot of course.

src/static/scripts/admin_diagnostics.js
BlackDex: Or wasn't it ;)

PR #5281, opened 2024-12-12
feat: mask _smtp_img_src in support string

tessus:
The following item in the support string should be masked as well, because it reveals the domain:

```
"_smtp_img_src": "https://vault.example.com/vw_static/",
```

P.S.: `domain` is masked, so it's kind of strange that it then shows up in `_smtp_img_src`


williamdes (approved):

BlackDex (approved):
Thanks!

PR #5291, opened 2024-12-13
Some refactoring and optimizations

BlackDex:
- Refactored several code to use more modern syntax
- Made some checks a bit more strict
- Updated crates

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -489,9 +489,9 @@ checksum = "ade8366b8bd5ba243f0a58f036cc0ca8a2f069cff1a2351ef1cac6b083e16fc0"
 
 [[package]]
 name = "cc"
-version = "1.2.3"
+version = "1.2.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "27f657647bcff5394bf56c7317665bbf790a137a50eaaa5c6bfbb9e27a518f2d"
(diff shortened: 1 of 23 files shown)


dani-garcia (approved):

gaby:
@dani-garcia @BlackDex This should be submitted as a CVE.

BlackDex:
I'm not sure if we really should report a CVE to be honest.
I'm not against it though, but also not sure if this would really add any benefit either.

It will probably take a long time for it to be requested and granted, since for one i have never done that my self before, and what i hear is that it is a long process.

gaby:
@BlackDex With GitHub it takes 1-2 days. You do it through that URL I posted above. GitHub will handle adding the entry to the registries, assigning a cve #, etc.

BlackDex:
Ok. fine by me

dani-garcia:
For future reference, the advisory lives here, but it's private at the moment while waiting for Github to assign it a CVE:

https://github.com/dani-garcia/vaultwarden/security/advisories/GHSA-g65h-982x-4m5m

I'll update here again once that process is done and the CVE is public.

dani-garcia:
Well that didn't take long at all, advisory and CVE are public at https://github.com/dani-garcia/vaultwarden/security/advisories/GHSA-g65h-982x-4m5m

gaby:
@dani-garcia Thank you 💪

PR #5293, opened 2024-12-14
Allow adding connect-src entries

### LINKED ISSUE #5290: CSP breaking selfhosted addy integration

* Vaultwarden version: v1.32.6
* OS/Arch: linux/x86_64
* Database type: SQLite
* Database version: 3.46.0
* Environment settings overridden!: true
* Uses a reverse proxy: true
* Internet access: true
* DNS Check: true
* Server/NTP Time Check: true
* HTTPS Check: true
* Websocket Check: false
* HTTP Response Checks: true

**Config:**

```json
{
  "_duo_akey": null,
  "_enable_duo": false,
  "_enable_email_2fa": true,
  "_enable_smtp": true,
  "_enable_yubico": true,
  "_icon_service_csp": "",
  "_icon_service_url": "",
  "_ip_header_enabled": true,
  "_max_note_size": 10000,
  "_smtp_img_src": "cid:",
  "admin_session_lifetime": 20,
  "admin_token": "***",
  "data_folder": "data",
  "database_conn_init": "",
  "database_max_conns": 10,
  "database_timeout": 30,
  "db_connection_retries": 15,
  "disable_admin_token": false,
  "domain": "*****://***************",
  "domain_origin": "*****://***************",
  "domain_path": "",
  "domain_set": true,
  "duo_host": null,
  "duo_ikey": null,
  "duo_skey": null,
  "duo_use_iframe": false,
  "email_attempts_limit": 3,
  "email_change_allowed": true,
  "email_expiration_time": 300,
  "email_token_size": 12,
  "enable_db_wal": true,
  "enable_websocket": true,
  "events_days_retain": null,
  "extended_logging": true,
  "helo_name": "***************",
  "hibp_api_key": null,
  "icon_blacklist_regex": null,
  "icon_cache_negttl": 259200,
  "icon_cache_ttl": 2592000,
  "icon_download_timeout": 10,
  "icon_redirect_code": 302,
  "icon_service": "internal",
  "invitations_allowed": true,
  "ip_header": "X-Real-IP",
  "log_file": null,
  "log_level": "info",
  "org_attachment_limit": null,
  "org_creation_users": "",
  "org_events_enabled": false,
  "org_groups_enabled": false,
  "push_enabled": false,
  "reload_templates": false,
  "sendmail_command": null,
  "sends_allowed": true,
  "sends_folder": "data/sends",
  "show_password_hint": false,
  "signups_allowed": false,
  "signups_verify": false,
  "smtp_auth_mechanism": "Plain",
  "smtp_debug": false,
  "smtp_embed_images": true,
  "smtp_explicit_tls": null,
  "smtp_from": "************************",
  "smtp_host": "************",
  "smtp_password": "***",
  "smtp_port": 587,
  "smtp_security": "starttls",
  "smtp_ssl": null,
  "smtp_timeout": 15,
  "smtp_username": "************************",
  "tmp_folder": "data/tmp",
  "use_sendmail": false,
  "use_syslog": false,
  "user_send_limit": null,
  "web_vault_enabled": true,
  "yubico_client_id": null,
  "yubico_secret_key": null,
  "yubico_server": null
}
```

</details>

### Vaultwarden Build Version

1.32.6

### Deployment method

Official Container Image

### Custom deployment method

_No response_

### Reverse Proxy

nginx/1.26.2

Linux

### Operating System Version

_No response_

### Clients

Web Vault

### Client Version

Firefox 131.0.2-1

### Steps To Reproduce

1. Go to Tools
2. Select username generation
3. Pick "Forwarded email alias"
4. Select addy.io as the service
5. Fill in "API access token", "Alias domain" and "Server URL"
6. Click "Regenerate username"

### Expected Result

The alias should be generated

### Actual Result

Fails with 
```
Content-Security-Policy: The page’s settings blocked the loading of a resource (connect-src) at https://your.instance.com/api/v1/aliases because it violates the following directive: “connect-src 'self' https://api.pwnedpasswords.com https://api.2fa.directory https://app.simplelogin.io/api/ https://app.addy.io/api/ https://api.fastmail.com/ https://api.forwardemail.net”
```

### Logs

_No response_

### Screenshots or Videos

_No response_

### Additional Context

_No response_

BlackDex:
Bitwarden allows to use self-hosted forwarded email services. But for this to work you need to add custom URL's to the `connect-src` CSP entry.

This commit allows setting this and checks if the URL starts with `https://` else it will abort loading.

Fixes #5290

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -410,6 +410,14 @@
 ## Multiple values must be separated with a whitespace.
 # ALLOWED_IFRAME_ANCESTORS=
 
+## Allowed connect-src (Know the risks!)
+## https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/connect-src
+## Allows other domains to URLs which can be loaded using script interfaces like the Forwarded email alias feature
+## This adds the configured value to the 'Content-Security-Policy' headers 'connect-src' value.
+## Multiple values must be separated with a whitespace. And only HTTPS values are allowed.
+## Example: "https://my-addy-io.domain.tld https://my-simplelogin.domain.tld"
+# ALLOWED_CONNECT_SRC=""
+
 ## Number of seconds, on average, between login requests from the same IP address before rate limiting kicks in.
 # LOGIN_RATELIMIT_SECONDS=60
 ## Allow a burst of requests of up to this size, while maintaining the average indicated by `LOGIN_RATELIMIT_SECONDS`.
--- a/src/config.rs
+++ b/src/config.rs
@@ -238,6 +238,7 @@ macro_rules! make_config {
                 // Besides Pass, only String types will be masked via _privacy_mask.
                 const PRIVACY_CONFIG: &[&str] = &[
                     "allowed_iframe_ancestors",
+                    "allowed_connect_src",
                     "database_url",
                     "domain_origin",
                     "domain_path",
@@ -610,6 +611,9 @@ make_config! {
         /// Allowed iframe ancestors (Know the risks!) |> Allows other domains to embed the web vault into an iframe, useful for embedding into secure intranets
         allowed_iframe_ancestors: String, true, def,    String::new();
 
+        /// Allowed connect-src (Know the risks!) |> Allows other domains to URLs which can be loaded using script interfaces like the Forwarded email alias feature
+        allowed_connect_src:      String, true, def,    String::new();
+
         /// Seconds between login requests |> Number of seconds, on average, between login and 2FA requests from the same IP address before rate limiting kicks in
         login_ratelimit_seconds:       u64, false, def, 60;
         /// Max burst size for login requests |> Allow a burst of requests of up to this size, while maintaining the average indicated by `login_ratelimit_seconds`. Note that this applies to both the login and the 2FA, so it's recommended to allow a burst size of at least 2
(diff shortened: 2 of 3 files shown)


dani-garcia:
I wonder how Bitwarden handles this, do we have a more strict CSP than they do?

A bit annoying that this is an admin config while the option to configure email forward generator is user level. Would it make sense to allow `*` as a special case to let any domains through (I'm assuming Bitwarden is doing that)?

BlackDex:
We also have addy in our CSP.
But if people are allowed to use custom endpoints, then I'm not sure how they will do this.

I have not checked this my self btw.
Will do that later.

BlackDex:
Looking at the cloud version, that does not support self-hosted forwarded instances.
The self-hosted Bitwarden does, but it also is blocked by default via the CSP.

There is no easy way to fix this in Self-Hosted Bitwarden besides overwriting the nginx config, which is possible, but not easy.

I would not allow `*` as that could cause possible security issues. Since people self-host, i would suggest they know which items are allowed. And we already add the hosted endpoints of those services by default.

dani-garcia (approved):
Nice thanks for the info, LGTM

ligix:
> Since people self-host, i would suggest they know which items are allowed

While this isn't an issue for people who host a small private instance (ex: for their family only) it unnecessarily worsens the CSP for all users. 
A better fix would be scoping the CSP changes to only those users who require it but it'd have marginal benefits and would be annoying to implement.

BlackDex:
> > Since people self-host, i would suggest they know which items are allowed
> 
> While this isn't an issue for people who host a small private instance (ex: for their family only) it unnecessarily worsens the CSP for all users. A better fix would be scoping the CSP changes to only those users who require it but it'd have marginal benefits and would be annoying to implement.

And even impossible since we do not know the user when loading the index page, and thus can not filter on that too change headers per user.

BJReplay:
Nice.  I hadn't noticed that you couldn't use this from the web vault, because the browser extension just works, but it is nice to be able to use it from the web vault, so that you can use the generator while updating logins in bulk from a regular email address to a self-hosted simple login generated email address.

PR #5308, opened 2024-12-18
Fixes fetchWsUrl

DevSplash:
https://github.com/dani-garcia/vaultwarden/discussions/5306

### DIFF
--- a/src/static/scripts/admin_diagnostics.js
+++ b/src/static/scripts/admin_diagnostics.js
@@ -349,7 +349,13 @@ async function checkHttpResponse() {
 async function fetchWsUrl(wsUrl) {
     return new Promise((resolve, reject) => {
         try {
-            const ws = new WebSocket(wsUrl);
+            const url = new URL(wsUrl);
+            if (url.protocol == 'http:') {
+                url.protocol = 'ws:';
+            } else if (url.protocol == 'https:') {
+                url.protocol = 'wss:';
+            }
+            const ws = new WebSocket(url.href);
             ws.onopen = () => {
                 ws.close();
                 resolve(true);


BlackDex:
This is not needed. That is all handled by the WebSocket as described here:
- https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/WebSocket
- https://websockets.spec.whatwg.org/#ref-for-dom-websocket-websocket%E2%91%A0

I know that older browsers might have an issue with this. But as far as i can see, all major browsers support this since half 2024. And to be fair, if someone is still using a browser older that that, they are using an insecure browser and should update a.s.a.p.

BlackDex:
Also, if you run a `console.log(ws)` you will see the `url` starts with `wss://`

DevSplash:
> Also, if you run a `console.log(ws)` you will see the `url` starts with `wss://`

Actually, the URL being passed is using the http(s) protocol, and I think that's the reason of https://github.com/dani-garcia/vaultwarden/discussions/5306 .
image

BlackDex:
```js
async function fetchWsUrl(wsUrl) {
    return new Promise((resolve, reject) => {
        try {
            const ws = new WebSocket(wsUrl);
            console.log(ws);
            ws.onopen = () => {
                ws.close();
                resolve(true);
            };

            ws.onerror = () => {
                reject(false);
            };
        } catch (e) {
            console.error(e)
            reject(false);
        }
    });
}
```

FireFox:
image

Chromium:
image

Gnome Epiphany:
image

As you can see, it works on all the above browsers just fine.

BlackDex:
And what does it do differently then? The URL gets converted to wss when it's https by the browser it self.

Also, we do not even send a message, we only check if we can connect.

BlackDex:
@DevSplash what browser, version and os are you using?

BlackDex:
Also, we can't check all the clients of course from our web diagnostics.
And also, since only the browser and browser components use WebSocket it's probably good to also test the browser.

I'm all for having this work of course. But if new browsers do this automatically as current standards described, I rather have this fail and let users update there browser actually.

But, if the majority of recent browsers do not work with this, then we should adjust this. Bot otherwise I rather not

DevSplash:
> @DevSplash what browser, version and os are you using?

I know that older browsers probably have an issue with this. It seems like the current browser extension still replaces http with ws, which is why the diagnostic page shows an error, but everything works fine in reality (this is what I meant by 'differently'). 
Directly replacing the protocol for now should fix the diagnostic issue. 
Overall, it's not a big deal, since, as you mentioned, updating the browser will make it work properly.

PR #5350, opened 2025-01-05
Prevent new users/members to be stored in db when invite fails

### LINKED ISSUE #5349: New member shows invited when the email was not sent

* Vaultwarden version: v1.32.7
* OS/Arch: linux/x86_64
* Database type: SQLite
* Database version: 3.46.0
* Uses a reverse proxy: false
* Internet access: true
* DNS Check: true
* Server/NTP Time Check: true
* HTTPS Check: false
* Websocket Check: true
* HTTP Response Checks: true

**Config:**

```json
{
  "_duo_akey": null,
  "_enable_duo": true,
  "_enable_email_2fa": true,
  "_enable_smtp": true,
  "_enable_yubico": true,
  "_icon_service_csp": "",
  "_icon_service_url": "",
  "_ip_header_enabled": true,
  "_max_note_size": 10000,
  "_smtp_img_src": "***:",
  "admin_session_lifetime": 20,
  "admin_token": "***",
  "allowed_connect_src": "",
  "data_folder": "data",
  "database_conn_init": "",
  "database_max_conns": 10,
  "database_timeout": 30,
  "db_connection_retries": 15,
  "disable_admin_token": false,
  "domain": "****://*********",
  "domain_origin": "****://*********",
  "domain_path": "",
  "domain_set": true,
  "duo_host": null,
  "duo_ikey": null,
  "duo_skey": null,
  "duo_use_iframe": false,
  "email_attempts_limit": 3,
  "email_change_allowed": true,
  "email_expiration_time": 600,
  "email_token_size": 6,
  "enable_db_wal": true,
  "enable_websocket": true,
  "events_days_retain": null,
  "extended_logging": true,
  "helo_name": null,
  "hibp_api_key": null,
  "icon_blacklist_regex": null,
  "icon_cache_negttl": 259200,
  "icon_cache_ttl": 2592000,
  "icon_download_timeout": 10,
  "icon_redirect_code": 302,
  "icon_service": "internal",
  "invitations_allowed": true,
  "ip_header": "X-Real-IP",
  "log_file": null,
  "log_level": "info",
  "org_attachment_limit": null,
  "org_creation_users": "",
  "org_events_enabled": false,
  "org_groups_enabled": false,
  "push_enabled": false,
  "reload_templates": false,
  "sendmail_command": null,
  "sends_allowed": true,
  "sends_folder": "data/sends",
  "show_password_hint": false,
  "signups_allowed": true,
  "signups_verify": false,
  "smtp_auth_mechanism": null,
  "smtp_debug": false,
  "smtp_embed_images": true,
  "smtp_explicit_tls": null,
  "smtp_from": "****************************",
  "smtp_host": "**************",
  "smtp_password": "***",
  "smtp_port": 465,
  "smtp_security": "force_tls",
  "smtp_ssl": null,
  "smtp_timeout": 15,
  "smtp_username": "****************************",
  "tmp_folder": "data/tmp",
  "use_sendmail": false,
  "use_syslog": false,
  "user_send_limit": null,
  "web_vault_enabled": true,
  "yubico_client_id": null,
  "yubico_secret_key": null,
  "yubico_server": null
}
```

</details>

### Vaultwarden Build Version

1.32.7

### Deployment method

Official Container Image

### Custom deployment method

_No response_

### Reverse Proxy

N/A

Linux

### Operating System Version

Debian

### Clients

Web Vault

### Client Version

_No response_

### Steps To Reproduce

Fresh deployment with for example an incorrect app password for Gmail SMTP, create an org, add a user, it will send an email but give an error because the app password is wrong, but the members page will show the new member as invited when they are not

### Expected Result

The new member is not in the database and not shown as invited

### Actual Result

The member is shown as invited

### Logs

_No response_

### Screenshots or Videos

_No response_

### Additional Context

_No response_

BlackDex:
Currently when a (new) user gets invited as a member to an org, and SMTP is enabled, but sending the invite fails, the user is still created.
They will only not have received a mail, and admins/owners need to re-invite the member again.
Since the dialog window still keeps on-top when this fails, it kinda invites to click try again, but that will fail in mentioning the user is already a member.

To prevent this weird flow, this commit will delete the user, invite and member if sending the mail failed.
This allows the inviter to try again if there was a temporary hiccup for example, or contact the server admin and does not leave stray users/members around.

Fixes #5349

### DIFF
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -954,10 +954,10 @@ async fn send_invite(org_id: &str, data: Json<InviteData>, headers: AdminHeaders
             {
                 // Upon error delete the user, invite and org member records when needed
                 if user_created {
-                    Invitation::take(email, &mut conn).await;
                     user.delete(&mut conn).await?;
+                } else {
+                    new_member.delete(&mut conn).await?;
                 }
-                new_member.delete(&mut conn).await?;
 
                 err!(format!("Error sending invite: {e:?} "));
             };
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -895,6 +895,7 @@ async fn send_invite(org_id: &str, data: Json<InviteData>, headers: AdminHeaders
         data.access_all = true;
     }
 
+    let mut user_created: bool = false;
     for email in data.emails.iter() {
         let mut user_org_status = UserOrgStatus::Invited as i32;
         let user = match User::find_by_mail(email, &mut conn).await {
@@ -908,13 +909,13 @@ async fn send_invite(org_id: &str, data: Json<InviteData>, headers: AdminHeaders
                 }
 
                 if !CONFIG.mail_enabled() {
-                    let invitation = Invitation::new(email);
-                    invitation.save(&mut conn).await?;
+                    Invitation::new(email).save(&mut conn).await?;
                 }
 
-                let mut user = User::new(email.clone());
-                user.save(&mut conn).await?;
-                user
+                let mut new_user = User::new(email.clone());
+                new_user.save(&mut conn).await?;
+                user_created = true;
+                new_user
             }
             Some(user) => {
                 if UserOrganization::find_by_user_and_org(&user.uuid, org_id, &mut conn).await.is_some() {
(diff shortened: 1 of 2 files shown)


stefan0xC:
What about the CollectionUser records? Wouldn't they need to be deferred as well?

BlackDex:
> What about the CollectionUser records? Wouldn't they need to be deferred as well?

Ah yes that needs to be moved to.

BlackDex:
I also wonder if this is the right way to go btw.
Now i defer saving the objects, but that could also mean, that after sending an email, storing the objects might fail, which i think is far worse then not being able to sent out an email.

I think it's better to save the objects, and upon mail failure, delete the items.

The only thing i would then put after sending an email is adding the user to the groups and collections.

BlackDex:
Ok, I made some changes. Instead of deferring, I now delete the records if sending the mail fails.
While this will cause some extra db queries to be executed, I do not think this will happen that much, and still keeps things more as they were.

I did moved creating the groups and collection records to after sending the mail, as those need the other records anyway.

src/api/core/organizations.rs
```diff
@@ -929,11 +930,49 @@ async fn send_invite(org_id: &str, data: Json<InviteData>, headers: AdminHeaders
...
+                Some(headers.user.email.clone()),
+            )
+            .await
+            {
+                // Upon error delete the user, invite and org member records when needed
+                if user_created {
+                    Invitation::take(email, &mut conn).await;
+                    user.delete(&mut conn).await?;
+                }
+                new_member.delete(&mut conn).await?;
```
stefan0xC, line 960: I would either call `new_member.delete()` before the `if` statement or only in an `else` block because `user.delete()` should also delete the `UserOrganization` and even if it works it feels wrong to call `update_uuid_revision` after deleting the `User`  https://github.com/dani-garcia/vaultwarden/blob/ef4bff09eb4414117cd8ea25c809988872e2016b/src/db/models/organization.rs#L708-L709

stefan0xC: Oh, and Invitation::take() is also called by user.delete, so this probably can be simplified to just 
```suggestion
                if user_created {
                    user.delete(&mut conn).await?;
                } else {
                    new_member.delete(&mut conn).await?;
                }
```

BlackDex: Uch, I should get some more sleep

BlackDex: Besides all that, Invites are also only created if smtp is **not** enabled 🤣 , so that would have resulted in a useless query anyway. While it still is done upon deleting the user, it's probably the better option.

stefan0xC (approved):
Looks good to me. Even if it does not prevent sending mails to non-existent hosts.

Not sure if that is something that we can fix because that's something that happens after the mail has been successfully received by the MTA
```
an 07 12:38:23 mx postfix/lmtp[2296669]: 8573C4078F: to=<stefan+mailinvitetest@example.com>, relay=127.0.0.1[127.0.0.1]:10024, delay=5.1, delays=0.28/0.1/0.02/4.7, dsn=2.0.0, status=sent (250 2.0.0 from MTA(smtp:[127.0.0.1]:10025): status=sent (250 2.0.0 from MTA(smtp:[127.0.0.1]:10025): 250 2.0.0 Ok: queued as 5703C40709
an 07 12:38:23 mx postfix/qmgr[1824492]: 8573C4078F: removed
an 07 12:38:23 mx postfix/smtp[2296676]: 5703C40709: to=<stefan+mailinvitetest@example.com>, relay=none, delay=0.36, delays=0.09/0.25/0.02/0, dsn=5.1.0, status=bounced (Domain example.com does not accept mail (nullMX))
```

BlackDex:
That will be difficult indeed, also, sometimes MTA's use graylogging, which could cause these kind of messages.
Thanks for testing and reviewing!

PR #5378, opened 2025-01-10
Add Attestations for containers and artifacts

BlackDex:
This commit will add attestation actions to sign the containers and binaries which can be verified via the gh cli. https://cli.github.com/manual/gh_attestation_verify

The binaries from both Alpine and Debian based images are extracted and attested so that you can verify the binaries of all the containers.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,11 +27,16 @@ jobs:
         if: ${{ github.ref_type == 'branch' }}
 
   docker-build:
+    permissions:
+      packages: write
+      contents: read
+      attestations: write
+      id-token: write
     runs-on: ubuntu-24.04
     timeout-minutes: 120
     needs: skip_check
     if: ${{ needs.skip_check.outputs.should_skip != 'true' && github.repository == 'dani-garcia/vaultwarden' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -27,11 +27,16 @@ jobs:
         if: ${{ github.ref_type == 'branch' }}
 
   docker-build:
+    permissions:
+      packages: write
+      contents: read
```
dani-garcia, line 32: Do we need the `packages` and `contents` permissions? The attestation action doesn't include these two:
https://github.com/actions/attest-build-provenance?tab=readme-ov-file#usage

----
Also if possible, can we move the attestation to a separate job so that `docker-build` doesn't have any unnecessary permissions? Ideally something like this:

```yaml
  docker-build:
    steps:
     .....

  attest-images:
    permissions:
      ..permissions here..
    steps:
      ....
```

This would make the workflow more secure as there won't be a chance of anything malicious from the build steps getting write permissions.

We're building inside a docker image, so we should be safe in general, but extra layers of security won't hurt

BlackDex: Ill look into the permissions a bit better and more in depth a later time. For now i think those are correct.

PR #5438, opened 2025-01-24
Org fixes

BlackDex:
### Security fixes for admin and sendmail

Because the Vaultwarden Admin Backend endpoints did not validated the Content-Type during a request, it was possible to update settings via CSRF. But, this was only possible if there was no `ADMIN_TOKEN` set at all. To make sure these environments are also safe I added the needed content-type checks at the functions.
This could cause some users who have scripts which uses cURL for example to adjust there commands to provide the correct headers.

By using a crafted favicon and having access to the Admin Backend an attacker could run custom commands on the host/container where Vaultwarden is running on. The main issue here is that we allowed the sendmail binary name/path to be updated via the Admin Backend. ~~To mitigate this we removed this configuration item and only then `sendmail` binary as a name can be used.
This could cause some issues where the `sendmail` binary is not in the `$PATH` and thus not able to be started. In these cases the admins should make sure `$PATH` is set correctly or create a custom shell script or symlink at a location which is in the `$PATH`.~~

Added an extra security header and adjusted the CSP to be more strict by setting `default-src` to `none` and added the needed missing specific policies.

Also created a general email validation function which does some more checking to catch invalid email address not found by the email_address crate.

### Fix security issue with organizationId validation

Because of a invalid check/validation of the OrganizationId which most of the time is located in the path but sometimes provided as a URL Parameter, the parameter overruled the path ID during the Guard checks.
This resulted in someone being able to execute commands as an Admin or Owner of the OrganizationId fetched from the parameter, but the API endpoints then used the OrganizationId located in the path instead.

This PR fixes the extraction of the OrganizationId in the Guard and also added some extra validations of this OrgId in several functions.

Also added an extra `OrgMemberHeaders` which can be used to only allow access to organization endpoints which should only be accessible by members of that org.

### Update server version in config endpoint

Updated the server version reported to the clients to `2025.1.0`.
This should make Vaultwarden future proof for the newer clients released by Bitwarden.

### Fix and adjust build workflow

The build workflow had an issue with some `if` checks.
For one they had two `$` signs, and it is not recommended to use `always()` since canceling a workflow does not cancel those calls.
Using `!cancelled()` is the preferred way.

### Update crates

Updated crates to there latest version

---

### Edit Allow sendmail binary config again.

Revert a previous change which removed the sendmail to be configurable.
We now set the config to be read-only, and omit all read-only values from being stored during a save action from the admin interface.


tessus:
Wouldn't it also be a solution to make the `sendmail_command` config var read-only? Changing the config file requires root.

Although in most cases sendmail is in the PATH anyway.

BlackDex:
> Wouldn't it also be a solution to make the `sendmail_command` config var read-only? Changing the config file requires root.
> 
> Although in most cases sendmail is in the PATH anyway.

Well, it basically isn't read-only in a hard way. You can still modify the contents and POST it.
The only reason we have them read-only right now is to prevent editing (easily) in the UI, since those values need a restart.
There might be a way now for us to reload the config on-the-fly, or maybe even warn users that some settings require a restart before they will be active. But that is not yet a feature added.

dani-garcia (approved):

nomandera:
It appears that something in this commit has broken valid `SMTP_FROM` addresses if the domain contains `-` character.

It is beyond me to work our why but reverting to `vaultwarden/server:1.32.7-alpine` from `vaultwarden/server:1.30.0-alpine` fixes the issue.

Replacing the `SMTP_FROM` with a different domain without `-` also fixes the issue.

BlackDex:
> It appears that something in this commit has broken valid `SMTP_FROM` addresses if the domain contains `-` character.
> 
> It is beyond me to work our why but reverting to `vaultwarden/server:1.32.7-alpine` from `vaultwarden/server:1.30.0-alpine` fixes the issue.
> 
> Replacing the `SMTP_FROM` with a different domain without `-` also fixes the issue.

Already fixed in #5442

nomandera:
Update on my post for Version 1.33.1. Using an email without a `-` in the TLD works fine. 

Real domain name redacted but form is the same. If I am OT for this chat let me know.

```
2025-02-04T15:37:55.150704726Z /--------------------------------------------------------------------\
2025-02-04T15:37:55.150792136Z |                        Starting Vaultwarden                        |
2025-02-04T15:37:55.150815558Z |                           Version 1.33.1                           |
2025-02-04T15:37:55.150831682Z |--------------------------------------------------------------------|
2025-02-04T15:37:55.150845553Z | This is an *unofficial* Bitwarden implementation, DO NOT use the   |
2025-02-04T15:37:55.150859139Z | official channels to report bugs/features, regardless of client.   |
2025-02-04T15:37:55.150875391Z | Send usage/configuration questions or feature requests to:         |
2025-02-04T15:37:55.150890648Z |   https://github.com/dani-garcia/vaultwarden/discussions or        |
2025-02-04T15:37:55.150946084Z |   https://vaultwarden.discourse.group/                             |
2025-02-04T15:37:55.150967742Z | Report suspected bugs/issues in the software itself at:            |
2025-02-04T15:37:55.150983699Z |   https://github.com/dani-garcia/vaultwarden/issues/new            |
2025-02-04T15:37:55.150999100Z \--------------------------------------------------------------------/
2025-02-04T15:37:55.151015165Z
2025-02-04T15:37:55.152266736Z [INFO] Using saved config from `data/config.json` for configuration.
2025-02-04T15:37:55.152292579Z
2025-02-04T15:37:55.152497888Z Error loading config:
2025-02-04T15:37:55.152540281Z   SMTP_FROM '<noreply@x-xxx.org>' is not a valid email address
```

BlackDex:
I do not think it is an issue with the `-`, but more the `<` and `>` signs.

BlackDex:
I can use `noreply@x-domain.tld` and `noreply@x-y-domain.tld` just fine.
But, using `<noreply@x-domain.tld>` or `<noreply@x-y-domain.tld>` breaks.

While not totally wrong to use `<` and `>` to enclose a mail address, it is not the format we want.

nomandera:
Testing all variations here now but it looks like you are spot on. Will update this post when I am done or with any useful info if I find any. Thanks for taking the time to replicate this.

PR #5469, opened 2025-01-27
Make sure the icons are displayed correctly in desktop clients

### LINKED ISSUE #5452: No Icons in Desktop Clients with Vaultwarden 1.33.0

* Vaultwarden version: v1.33.0
* Web-vault version: v2025.1.1
* OS/Arch: linux/x86_64
* Database type: MySQL
* Database version: 8.0.41
* Uses a reverse proxy: true
* IP Header check: true (x-forwarded-for)
* Internet access: true
* DNS Check: true
* Server/NTP Time Check: true
* HTTPS Check: true
* Websocket Check: true
* HTTP Response Checks: true

**Config:**

```json
{
  "_duo_akey": null,
  "_enable_duo": false,
  "_enable_email_2fa": false,
  "_enable_smtp": true,
  "_enable_yubico": true,
  "_icon_service_csp": "",
  "_icon_service_url": "",
  "_ip_header_enabled": true,
  "_max_note_size": 10000,
  "_smtp_img_src": "***:",
  "admin_session_lifetime": 20,
  "admin_token": "***",
  "allowed_connect_src": "",
  "data_folder": "data",
  "database_conn_init": "",
  "database_max_conns": 10,
  "database_timeout": 30,
  "database_url": "*****://*******************************************************************",
  "db_connection_retries": 15,
  "disable_admin_token": false,
  "domain_path": "",
  "domain_set": true,
  "duo_host": "api-bc8d4d38.duosecurity.com",
  "duo_ikey": "DICFPHD0UJEWUVKIKAVM",
  "duo_skey": "***",
  "duo_use_iframe": false,
  "email_attempts_limit": 3,
  "email_change_allowed": true,
  "email_expiration_time": 600,
  "email_token_size": 6,
  "enable_db_wal": true,
  "enable_websocket": true,
  "events_days_retain": null,
  "extended_logging": true,
  "helo_name": null,
  "hibp_api_key": null,
  "icon_blacklist_regex": null,
  "icon_cache_negttl": 259200,
  "icon_cache_ttl": 2592000,
  "icon_download_timeout": 10,
  "icon_redirect_code": 302,
  "icon_service": "internal",
  "invitation_org_name": "Miller Family Password Vault",
  "invitations_allowed": true,
  "ip_header": "x-forwarded-for",
  "log_level": "info",
  "org_attachment_limit": null,
  "org_creation_users": "",
  "org_events_enabled": false,
  "org_groups_enabled": false,
  "password_iterations": 1000000,
  "push_enabled": true,
  "reload_templates": false,
  "sendmail_command": null,
  "sends_allowed": true,
  "sends_folder": "data/sends",
  "show_password_hint": false,
  "signups_allowed": false,
  "signups_verify": true,
  "smtp_auth_mechanism": "off",
  "smtp_debug": false,
  "smtp_embed_images": true,
  "smtp_explicit_tls": null,
  "smtp_from": "******************",
  "smtp_from_name": "MillerNet Password Vault",
  "smtp_host": "*****************",
  "smtp_password": null,
  "smtp_port": 587,
  "smtp_security": "off",
  "smtp_ssl": null,
  "smtp_timeout": 15,
  "smtp_username": null,
  "tmp_folder": "data/tmp",
  "use_sendmail": false,
  "use_syslog": false,
  "user_send_limit": null,
  "web_vault_enabled": true,
  "yubico_client_id": "102714",
  "yubico_secret_key": "***",
  "yubico_server": null
}
```

</details>

### Vaultwarden Build Version

1.33.0

### Deployment method

Official Container Image

### Custom deployment method

_No response_

### Reverse Proxy

Caddy v2.9.1

NAS/SAN

### Operating System Version

macOS 12.7.6

### Clients

Desktop

### Client Version

2025.1.3

### Steps To Reproduce

1. Open Bitwarden client
2. View any or all entries in the vault
3. File -> Sync
4. Still no icons

### Expected Result

Icons for websites.

### Actual Result

a mixture of the default no icon and a broken image.

### Logs

```text

```

### Screenshots or Videos

Image

Reverting to 1.32.7 restores the icons.

### Additional Context

_No response_

WinLinux1028:
Fixes #5452

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -55,7 +55,10 @@ impl Fairing for AppHeaders {
         res.set_raw_header("Referrer-Policy", "same-origin");
         res.set_raw_header("X-Content-Type-Options", "nosniff");
         res.set_raw_header("X-Robots-Tag", "noindex, nofollow");
-        res.set_raw_header("Cross-Origin-Resource-Policy", "same-origin");
+
+        if !res.headers().get_one("Content-Type").is_some_and(|v| v.starts_with("image/")) {
+            res.set_raw_header("Cross-Origin-Resource-Policy", "same-origin");
+        }
 
         // Obsolete in modern browsers, unsafe (XS-Leak), and largely replaced by CSP
         res.set_raw_header("X-XSS-Protection", "0");


src/util.rs
```diff
@@ -55,7 +55,9 @@ impl Fairing for AppHeaders {
         res.set_raw_header("Referrer-Policy", "same-origin");
         res.set_raw_header("X-Content-Type-Options", "nosniff");
         res.set_raw_header("X-Robots-Tag", "noindex, nofollow");
-        res.set_raw_header("Cross-Origin-Resource-Policy", "same-origin");
+        if !(req_uri_path.starts_with("/icons/") && req_uri_path.ends_with("/icon.png")) {
```
BlackDex, line 58: I think we only need to check if it ends with `/icon.png`.
The other check is not really needed and might even not work if someone uses a sub-path to host Vaultwarden on.

BlackDex (changes requested):

src/util.rs
```diff
@@ -55,7 +55,9 @@ impl Fairing for AppHeaders {
...
+            res.set_raw_header("Cross-Origin-Resource-Policy", "same-origin");
+        }
```
tessus, line 60: I'd go with the same process as above (might even be slightly more performant):

```suggestion
        res.set_raw_header("Cross-Origin-Resource-Policy", "same-origin");
        if req_uri_path.ends_with("/icon.png") {
            res.remove_header("Cross-Origin-Resource-Policy");
        }
```

src/util.rs
WinLinux1028: I was adding this check because I thought there might be other paths that end in `icon.png`.
I deleted it as it was unnecessary.

src/util.rs
BlackDex: How would this be more performant?
You first always set the header, then always check if it ends on a specific string.
While you can also negate the check if it always ends on a specific string and then just set the header.

jjlin: Why would it be more performant to set a header and then potentially remove it? Seems like it would be simplest just to do
```
if !req_uri_path.ends_with("/icon.png") {
    res.set_raw_header("Cross-Origin-Resource-Policy", "same-origin");
}
```

src/util.rs
```diff
@@ -57,6 +57,10 @@ impl Fairing for AppHeaders {
         res.set_raw_header("X-Robots-Tag", "noindex, nofollow");
         res.set_raw_header("Cross-Origin-Resource-Policy", "same-origin");
 
+        if req_uri_path.ends_with("/icon.png") {
+            res.remove_header("Cross-Origin-Resource-Policy");
+        }
+
```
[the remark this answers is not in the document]
WinLinux1028: There seems to be a conflict of opinion.
@tessus How do you think about this?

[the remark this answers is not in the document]
tessus: @BlackDex in the4 code above the headers are removed when a condition is met. The question is whether most of the calls are calls that end in icon.png or not.
If other calls are more likely, removing the header is better for performance

src/util.rs
tessus: branch prediction. but maybe not anymore. maybe the compilers are much better these days. I am probably too old....

tessus: also, I have noticed that in some projects clippy will complain because clippy prefers matches over not matches. But maybe that's not the case here either

src/util.rs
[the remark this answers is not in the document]
BlackDex: If you are referring to the `notifications/hub` checks, those do an early return.

It doesn't matter if most of the calls end with `/icon.png` or not, the check is done always not matter if you put a `!` in-front of it or not.

In your example there is always 1 call extra, the removing of a just added header, which to me is illogical.
Only set it when needed, and not set it, and then afterwards say, ow wait, i do not need it anyway, lets do an other function call to delete it again.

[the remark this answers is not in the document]
tessus: It's your decision of course. I am way past my bed time, so I am heading to bed now.

src/util.rs
WinLinux1028:
```Rust
        if let Some(path_stripped) = req_uri_path.as_str().strip_prefix(CONFIG.domain_path()) {
            if !(path_stripped.starts_with("/icons/") && path_stripped.ends_with("/icon.png")) {
                res.set_raw_header("Cross-Origin-Resource-Policy", "same-origin");
            }
        }
```
How about doing it like this?

src/util.rs
BlackDex: The clippy lint gets triggered if there would have been a function `not_ends_with()` which makes it more clear because missing the `!` in front is easier to do, and a function with that name is more descriptive.

And, doesn't branch prediction work the same way for the exclusion? If the CPU can predict when to not call it, that is fine too ;)

src/util.rs
BlackDex: That is overly complicated, and not needed, as there only is one valid path which ends on `/icon.png`

So, only checking if the path doesn't end on it will be fine.

BlackDex (dismissed):

src/util.rs
```diff
@@ -55,7 +55,9 @@ impl Fairing for AppHeaders {
         res.set_raw_header("Referrer-Policy", "same-origin");
         res.set_raw_header("X-Content-Type-Options", "nosniff");
         res.set_raw_header("X-Robots-Tag", "noindex, nofollow");
-        res.set_raw_header("Cross-Origin-Resource-Policy", "same-origin");
+        if !req_uri_path.ends_with("/icon.png") {
+            res.set_raw_header("Cross-Origin-Resource-Policy", "same-origin");
+        }
```
stefan0xC, line 60: Isn't something like that also needed for "mail-github.png" and "logo-gray.png" in case `SMTP_EMBED_IMAGES=false` (so they can be fetched by mail clients)? Something like
```suggestion
        if !(req_uri_path.ends_with("/icon.png")
             || (!CONFIG.smtp_embed_images
                && (req_uri_path.ends_with("/logo-gray.png")
                || req_uri_path.ends_with("/mail-github.png"))) {
            res.set_raw_header("Cross-Origin-Resource-Policy", "same-origin");
        }
```

BlackDex: Hmmm, i have not tested that yet to be fair.
But then only excluding those files will probably still cause issues, since you could server custom image files too, not just those we set our self.

BlackDex: I think there is a better way.

WinLinux1028: After checking the email I received from Vaultwarden, I found that the two files are under `/vw_static/`.
How about doing it like this?

```Rust
if let Some(path_stripped) = req_uri_path.strip_prefix(&*CONFIG.domain_path()) {
    if !(path_stripped.starts_with("/icons/") || path_stripped.starts_with("/vw_static/") {
        res.set_raw_header("Cross-Origin-Resource-Policy", "same-origin");
    }
}
```

stefan0xC: Sounds good. I mean, if we want to tackle this in a separate PR this would be fine for me too.

stefan0xC (approved):

src/util.rs
BlackDex: We can use `res.content_type()` and check if it is an image :)

WinLinux1028: Isn't it too much to not apply `Cross-Origin-Resource-Policy` to all images?

src/util.rs
```diff
@@ -55,7 +55,12 @@ impl Fairing for AppHeaders {
         res.set_raw_header("Referrer-Policy", "same-origin");
         res.set_raw_header("X-Content-Type-Options", "nosniff");
         res.set_raw_header("X-Robots-Tag", "noindex, nofollow");
-        res.set_raw_header("Cross-Origin-Resource-Policy", "same-origin");
+
+        if let Some(path_stripped) = req_uri_path.strip_prefix(&*CONFIG.domain_path()) {
+            if !(path_stripped.starts_with("/icons/") || path_stripped.starts_with("/vw_static/")) {
+                res.set_raw_header("Cross-Origin-Resource-Policy", "same-origin");
+            }
+        }
```
stefan0xC, line 63: I think I'd also prefer @BlackDex suggestion. I.e, something like:
```suggestion
        if !res.is_png() {
            res.set_raw_header("Cross-Origin-Resource-Policy", "same-origin");
        }
```

src/util.rs
BlackDex: I think we could use `res.headers().get_one("Content-Type").is_some_and(|v| v.starts_with("image"))` which should solve it.

Well, what if someone uses a modified image from the web-vault somewhere? Like browsers might use it for the favicon of the domain for example?

It's not that the images really need protection from being fetched in my opinion.
But more to prevent other scripts or sites from being able to abuse the API endpoints.

src/util.rs
WinLinux1028: What about non-PNG images?
In my opinion, it's too much to not apply Cross-Origin-Resource-Policy to all images.

BlackDex:
```suggestion
        if !res.headers().get_one("Content-Type").is_some_and(|v| v.starts_with("image")) {
            res.set_raw_header("Cross-Origin-Resource-Policy", "same-origin");
        }
```

stefan0xC: I don't think that we can have it both ways without complicating the this further. I think allowing only `PNG` should be enough. (If you don't agree, I've also posted an suggestion above https://github.com/dani-garcia/vaultwarden/pull/5469#discussion_r1930244657 which would only allow the bare minimum.)

WinLinux1028: In the future, we may use WebP or AVIF, so [this](https://github.com/dani-garcia/vaultwarden/pull/5469#discussion_r1930318145) method is a good choice if you do not want to assign Cross-Origin-Resource-Policy to all images.

BlackDex: That isn't possible. The content type is based upon the image type by it's headers. And thus not only `PNG` types are returned.

https://github.com/dani-garcia/vaultwarden/blob/2903a3a13ad06a09805e8842511b4b715a686d8e/src/api/icons.rs#L582-L592

BlackDex: And:
https://github.com/dani-garcia/vaultwarden/blob/2903a3a13ad06a09805e8842511b4b715a686d8e/src/api/icons.rs#L113-L119

BlackDex (approved):

dani-garcia (approved):

Livefour2day:
Icon issue in Chrome extension and desktop still present after latest update.

BlackDex:
> Missing icons issue in Chrome extension and desktop still present after latest update.

Are you using icon redirects?

Livefour2day:
> > Missing icons issue in Chrome extension and desktop still present after latest update.
> 
> Are you using icon redirects?

Yes is this the issue it has been working fine for years

BlackDex:
> > > Missing icons issue in Chrome extension and desktop still present after latest update.
> > 
> > 
> > Are you using icon redirects?
> 
> Yes is this the issue it has been working fine for years

This is fixed via #5536

PR #5490, opened 2025-01-28
Add passkey support to Vaultwarden

p4rzl:
Add support for saving and using passkeys, and importing them via .json.

* **src/api/core/ciphers.rs**
  - Add `Passkey` type to `CipherData` struct.
  - Update `update_cipher_from_data` function to handle passkeys.
  - Modify `post_ciphers_import` function to import passkeys.

* **src/db/models/cipher.rs**
  - Add `Passkey` type to `Cipher` struct.
  - Update `type_data_json` handling to include passkeys.
  - Add validation for passkey entries.

* **Database Migrations**
  - Add SQL statements to add passkey fields to MySQL, PostgreSQL, and SQLite schemas.

* **src/api/admin.rs**
  - Add endpoints for managing passkeys: `get_passkeys`, `get_passkey`, `create_passkey`, `update_passkey`, `delete_passkey`.

### DIFF
--- a/migrations/mysql/2025-01-09-172300_add_passkey_support/up.sql
+++ b/migrations/mysql/2025-01-09-172300_add_passkey_support/up.sql
@@ -1,0 +1,3 @@
+ALTER TABLE ciphers
+ADD COLUMN passkey_id TEXT,
+ADD COLUMN passkey_public_key TEXT;
--- a/migrations/postgresql/2025-01-09-172300_add_passkey_support/up.sql
+++ b/migrations/postgresql/2025-01-09-172300_add_passkey_support/up.sql
@@ -1,0 +1,3 @@
+ALTER TABLE ciphers
+ADD COLUMN passkey_id TEXT,
+ADD COLUMN passkey_public_key TEXT;
--- a/migrations/sqlite/2025-01-09-172300_add_passkey_support/up.sql
+++ b/migrations/sqlite/2025-01-09-172300_add_passkey_support/up.sql
@@ -1,0 +1,3 @@
+ALTER TABLE ciphers
+ADD COLUMN passkey_id TEXT,
+ADD COLUMN passkey_public_key TEXT;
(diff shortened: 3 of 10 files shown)


BlackDex:
Do  you have a link to the client side code? I'm currently unable to find this CipherType.
Testing this without a working client will be rather hard.

BlackDex:
Also, As far as i know, Passkeys already work just fine, they are stored within the Login type.
But there might be a change upcoming of course.

p4rzl:
honestly, I tried to use them several times but they didn't work in any way.

BlackDex:
So, how do you think this is going to work then? Since none of the Bitwarden clients are going to support this?

p4rzl:
At first I had only thought about the web version, to integrate better, now I'm trying to fix the code making everything compatible.

BlackDex:
If this is for logging in with Passkeys, then this is not the right way to store them though.
Storage of Logins/2FA has nothing to do with Ciphers.

And if you are planning on using a custom client other then Bitwarden, then I'm afraid we are not going to support it like this.

Main reason also is, that if Bitwarden is going to add a new CipherType, which then probably gets the followup number 6, that will cause issues.

If you can provide more details on what you are trying to do here, that would be helpful.

p4rzl:
I'm making this conversation a draft so I can reopen it after I've finished editing.

BlackDex:
If you would try to explain what you are doing, and how the bitwarden clients would interact with it, that could safe you a lot of work for nothing.

tessus:
@p4rzl I think you might not understand what the issue is with your PR. The cypher type and the API endpoints are used by the clients (web, mobile, cli). But if none of the clients make use of the changes you are coding in this PR, the code is not used and thus useless.

However, vw does not have the endpoints for logging in (into vaultwarden itself) with a passkey. Is this what you are trying to do? (blackdex also asked this in an earlier reply)

PR #5491, opened 2025-01-28
Fix Duo Field Names for Web Client

ratiner:
Looks like the web client expects different fields for duo integration.
The changes were introduced in the following commit:
https://github.com/bitwarden/clients/commit/41e1d9155850581be85d11e4b74088aa022f8381#diff-2be3b52c1ad01eac329688bca2644e0aa1f819167ed8a9d52ed57f72e18b8aa8L4
https://github.com/bitwarden/clients/commit/41e1d9155850581be85d11e4b74088aa022f8381#diff-8813d04cfcaad734e34e034e8a85eeb4a528d55aeb52b5c88e6bcbfced9fc260L13

Web client sends different fields to the API (and also receives wrong ones):
`clientSecret` instead of `secretKey` and `clientId` instead of `integrationKey`

I believe, this PR can address this.

When saving:
image

When loading (Client Id and Client Secret is empty):
image

### DIFF
--- a/src/api/core/two_factor/duo.rs
+++ b/src/api/core/two_factor/duo.rs
@@ -26,8 +26,8 @@ pub fn routes() -> Vec<Route> {
 #[derive(Serialize, Deserialize)]
 struct DuoData {
     host: String, // Duo API hostname
-    ik: String,   // integration key
-    sk: String,   // secret key
+    ik: String,   // client id
+    sk: String,   // client secret
 }
 
 impl DuoData {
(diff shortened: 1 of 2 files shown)


BDvirus (approved):

Gerardv514:
Seems similar to what I had brought up back in October

https://github.com/dani-garcia/vaultwarden/discussions/5135

ratiner:
> Seems similar to what I had brought up back in October. The admin panel should get renamed as well as I mentioned.
> 
> #5135

Thank you. I've added a commit which renames the fields in admin page as well.

BlackDex (approved):

BlackDex:
Thanks!

PR #5537, opened 2025-02-03
Update workflows and enhance security

BlackDex:
This commit updates the workflow files and also fixes some security issues which were reported by using zizmor https://github.com/woodruffw/zizmor

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,4 +1,5 @@
 name: Build
+permissions: {}
 
 on:
   push:
@@ -13,6 +14,7 @@ on:
       - "diesel.toml"
       - "docker/Dockerfile.j2"
       - "docker/DockerSettings.yaml"
+
   pull_request:
     paths:
       - ".github/workflows/build.yml"
@@ -28,13 +30,17 @@ on:
 
 jobs:
   build:
+    name: Build and Test ${{ matrix.channel }}
+    permissions:
+      actions: write
+      contents: read
     # We use Ubuntu 22.04 here because this matches the library versions used within the Debian docker containers
     runs-on: ubuntu-22.04
     timeout-minutes: 120
     # Make warnings errors, this is to prevent warnings slipping through.
     # This is done globally to prevent rebuilds when the RUSTFLAGS env variable changes.
     env:
-      RUSTFLAGS: "-D warnings"
+      RUSTFLAGS: "-Dwarnings"
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -28,34 +29,37 @@ on:
```
dfunkt, line 29:
```suggestion
permissions: {}

```
https://woodruffw.github.io/zizmor/audits/#remediation_2
```
warning[excessive-permissions]: overly broad permissions
   --> .github/workflows/build.yml:1:1
    |
  1 | / name: Build
  2 | |
...   |
220 | |           echo "### :tada: Checks Passed!" >> "${GITHUB_STEP_SUMMARY}"
221 | |           echo "" >> "${GITHUB_STEP_SUMMARY}"
    | |______________________________________________- default permissions used due to no permissions: block
    |
    = note: audit confidence → Medium
```

.github/workflows/hadolint.yml
```diff
@@ -8,14 +8,11 @@ on: [
 jobs:
```
dfunkt, line 8:
```suggestion
permissions: {}

jobs:
```

.github/workflows/release.yml
```diff
@@ -6,13 +6,16 @@ on:
       - main
 
     tags:
-      - '*'
+      # https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet
+      - '[1-2].[0-9]+.[0-9]+'
 
 jobs:
```
dfunkt, line 12:
```suggestion
permissions: {}

jobs:
```

.github/workflows/releasecache-cleanup.yml
```diff
@@ -13,6 +13,8 @@ name: Cleanup
 jobs:
```
dfunkt, line 13:
```suggestion
permissions: {}

jobs:
```

dfunkt (changes requested):
Haven't had the chance to test these changes yet, just a quick surface look.

.github/workflows/build.yml
BlackDex: I think you are using `-p` here right?
Those are not really an issue i think, see https://github.com/woodruffw/zizmor/issues/472

BlackDex:
@dfunkt, i think those items are not correct.
The GitHub documentation also states that you can either provide a global permission, or a per job setting.
But maybe adding a default of not allowing anything is a good option though, to prevent more broader access if someone adds a new job without adding permissions.

.github/workflows/build.yml
dfunkt: Yeah, forgot to mention that.

dfunkt:
You can disregard my previous suggestions, completly forgot that I was checking the files with ```--pedantic``` enabled.

BlackDex:
> You can disregard my previous suggestions, completly forgot that I was checking the files with `--pedantic` enabled.

Well, you made me thinking, and having some added security by setting no permissions at all for the whole workflow by default, and then setting it per job is probably a nice way of enforcing it everywhere.

In that case, when forks still have a read-write token by default that is also a bit more hard to exploit there.
Also converted a few other items, and now i get a result of 0 items wrong, also checked with a GitHub Token for some added online checks.

BlackDex:
I also run these workflows through https://rhysd.github.io/actionlint/ and they are also clean there.

dfunkt (approved):

dani-garcia (approved):

dimejo:
@BlackDex:
In case you want to further tighten permissions you can run [GitHubSecurityLab/actions-permissions](https://github.com/GitHubSecurityLab/actions-permissions) to scan and print minimally required permissions per job.

PR #5620, opened 2025-02-20
DKIM implementations

DorianCoding:
Hello,

I would like to implement DKIM on Rust emails, so SPF and DKIM can be fulfilled and allows better transmission of mails.

Thanks.


BlackDex:
@tessus, sendmail is nothing more then a client in our case. It's how the sendmail is configured which might make it differently

DorianCoding:
Hello,

Yes I know but some person might use DKIM on their own without their server and it would be wise to allow people to do it if they want to, because the implementation is minimal as you can see. It might not be useful for everyone, I fully agree and if the server already manages SPF/DKIM and so on, there is no need to implement anything more but it might not be the case and people should be able to implement it.

src/mail.rs
```diff
@@ -641,7 +643,60 @@ async fn send_with_selected_transport(email: Message) -> EmptyResult {
...
     let smtp_from = &CONFIG.smtp_from();
-
+    let dkim = match (CONFIG.dkim_signature(), CONFIG.dkim_infos()) {
+        (Some(sig), Some(infos)) => {
+            let config = {
+                let algo = match CONFIG.dkim_algo() {
+                    Some(true) => DkimSigningAlgorithm::Rsa,
+                    _ => DkimSigningAlgorithm::Ed25519,
+                };
+                let mut key = String::with_capacity(4096);
```
williamdes, line 653: does it assume all keys are 4096 ?

DorianCoding: Hello, the maximum size of RSA keys is AFAIK 4096 bits so 512 bytes I forgot the conversion and 64 bytes for ed25519. So we can separate both to reduce memory allocation.

DorianCoding: It should work whatever the key size. The capacity is just to avoid multiple allocations by estimating the expected maximum size of the key in the file.

DorianCoding:
@williamdes Sorry, I've forgot I can implement `fs::read_to_string` which automatically manages the reading and the allocation, it would be way easier. What do you think?

williamdes (approved):
Nice

DorianCoding:
Weird the changes I have made to put Rust 2024 has been pushed. Please just push changes to mail if you proceed. Sorry and thanks.

src/mail.rs
```diff
@@ -661,12 +709,14 @@ async fn send_email(address: &str, subject: &str, body_html: String, body_text:
...
-    let email = Message::builder()
+    let mut email = Message::builder()
         .message_id(Some(format!("<{}@{}>", crate::util::get_uuid(), smtp_from.split('@').collect::<Vec<&str>>()[1])))
         .to(Mailbox::new(None, Address::from_str(address)?))
         .from(Mailbox::new(Some(CONFIG.smtp_from_name()), Address::from_str(smtp_from)?))
         .subject(subject)
         .multipart(body)?;
-
+    if let Some(sig) = dkim {
+        dkim_sign(&mut email, &sig);
```
stefan0xC:
```suggestion
    if let Some(dkim_config) = dkim {
        dkim_sign(&mut email, &dkim_config);
```

src/config.rs
```diff
@@ -704,6 +704,12 @@ make_config! {
         smtp_from_name:                String, true,   def,     "Vaultwarden".to_string();
         /// Username
         smtp_username:                 String, true,   option;
+        /// Dkim signature (type:privatekey). Private must be base64-encoded ed key or PKCS#1 format RSA key.
+        dkim_signature:                String, true,   option;
+        /// Dkim algo (true if RSA else ed25519)
+        dkim_algo:                     bool,   true,   option;
+        /// Dkim infos (selector:domain)
+        dkim_infos:                    String, true,   option;
```
stefan0xC: Why not let users configure `selector` and `domain` separately?

src/config.rs
```diff
@@ -704,6 +704,12 @@ make_config! {
         smtp_from_name:                String, true,   def,     "Vaultwarden".to_string();
         /// Username
         smtp_username:                 String, true,   option;
+        /// Dkim signature (type:privatekey). Private must be base64-encoded ed key or PKCS#1 format RSA key.
+        dkim_signature:                String, true,   option;
```
stefan0xC: What does `type:privatekey` mean? Nothing indicates that this should be a path to a file. If you want to configure it via a file you should be able to configure a `DKIM_SIGNATURE_FILE`.

src/mail.rs
```diff
@@ -641,7 +643,53 @@ async fn send_with_selected_transport(email: Message) -> EmptyResult {
...
+            let config = {
+                let algo = match CONFIG.dkim_algo() {
+                    Some(true) => DkimSigningAlgorithm::Rsa,
+                    _ => DkimSigningAlgorithm::Ed25519,
+                };
+                let sig = match fs::read_to_string(sig) {
+                    Err(e) => {
+                        debug!("Cannot read DKIM file. Err is {:?}", e);
+                            None
+                    },
```
stefan0xC: We also already have a mechanism for loading config values from files so this should not be necessary.

src/mail.rs
```diff
@@ -641,7 +643,53 @@ async fn send_with_selected_transport(email: Message) -> EmptyResult {
 
 async fn send_email(address: &str, subject: &str, body_html: String, body_text: String) -> EmptyResult {
     let smtp_from = &CONFIG.smtp_from();
-
+    let dkim = match (CONFIG.dkim_signature(), CONFIG.dkim_infos()) {
```
stefan0xC: The validation should be done at startup (in `config.rs`) instead of when sending a mail.

Cargo.toml
```diff
@@ -4,8 +4,8 @@ workspace = { members = ["macros"] }
 name = "vaultwarden"
 version = "1.0.0"
 authors = ["Daniel García <dani-garcia@users.noreply.github.com>"]
-edition = "2021"
-rust-version = "1.83.0"
+edition = "2024"
+rust-version = "1.85.0"
```
stefan0xC: > Weird the changes I have made to put Rust 2024 has been pushed. Please just push changes to mail if you proceed. Sorry and thanks.

Why don't you remove those changes from your PR if you don't want them in it?

src/config.rs
```diff
@@ -704,6 +704,12 @@ make_config! {
...
+        /// Dkim algo (true if RSA else ed25519)
+        dkim_algo:                     bool,   true,   option;
```
stefan0xC: Do we have to support both methods? Also the naming is wrong if this is a toggle to use an RSA key. (I would call it `dkim_use_rsa` instead (and say that the default is `ed25519` otherwise).

src/config.rs
stefan0xC: Also this should probably called `dkim_signing_key` instead.

stefan0xC (changes requested):
I'm not sure I agree that this should be added but I added some comments. The way to specify the signing key should be better documented (and configuration options should also be added to the `.env.template`).

stefan0xC:
Also you should probably run `cargo fmt` and `cargo clippy` otherwise the automatic checks (that have not run yet) will fail.

rustfmt.toml
```diff
@@ -1,4 +1,4 @@
-edition = "2021"
+edition = "2024"
```
BlackDex, line 1: This should not be changed!

rust-toolchain.toml
BlackDex, line 1: This file should not be edited, and you probably want to rebase on our current main branch.

Cargo.toml
BlackDex: Besides that, it this PR needs a rebase on main, and the edition should not be changed!

BlackDex (changes requested):

tessus:
I am still a bit fuzzy on the why and how. 

> but some person might use DKIM on their own without their server and it would be wise to allow people to do it if they want to

How do people use DKIM on their own?

DKIM requires the DNS record to be updated with TXT entries. Do you expect the vw admin to update the mail server's DNS record? Or do you plan to send emails via a random mail server, but let people think they are coming from vaultwarden and add the DKIM TXT entries to the vw DNS record?

DorianCoding:
Hello,

Thanks everyone for your review, I have redo my changes in light of the new version. I've taken into account your comments.

> How do people use DKIM on their own?
> 
> DKIM requires the DNS record to be updated with TXT entries. Do you expect the vw admin to update the mail server's DNS record? Or do you plan to send emails via a random mail server, but let people think they are coming from vaultwarden and add the DKIM TXT entries to the vw DNS record?

TXT is the public key. You just need vaultwarden to get the private key and sign. However, depending on the MTA, it might be useless or wrong. However if the MTA does not implement it and does not change header, DKIM implementation by the client would work. It is for me really important as some people might need it in their configuration and email sending by Vaultwarden should pass filters to avoid issues and ensure security.

By default, From/Subject/To/Date are signed. Those are headers that are unlikely to be changed by default MTA.

tessus:
> TXT is the public key. You just need vaultwarden to get the private key and sign. However, depending on the MTA, it might be useless or wrong. However if the MTA does not implement it and does not change header, DKIM implementation by the client would work. It is for me really important as some people might need it in their configuration and email sending by Vaultwarden should pass filters to avoid issues and ensure security.

My thought process went into another direction and maybe I did express myself poorly. Let me try to unpack this.

You can derive the public key from the private key, but not the other way around. This means the person who setups vw to use DKIM, must be the person who has access to the private keys AND must add the public keys to the DNS record.
You can't send emails from your home network, because providers (at least in North America and Europe) block outgoing mails. And even if you could, dynamic IP ranges are included in various DSNBL lists, which means the receiving email server will most likely discard them.
Using a public mail server like google/yahoo/protonmail/whatever is fragile, because they sign the mail with DKIM. In theory this should be fine according to the DKIM specs, but in reality is not so straightforward (double signing can be a mess). Apart from that some mailservers add a footer (similar to what mailing lists sometimes do), which makes your locally signed mail invalid.
Other mail servers use 7bit encoding. This alone screws up the signature.

This means you either have to be in control (the admin) of the mail server or know exactly how that mail server behaves (no mail body or header changes, transfer encoding, ...). 

If it is the former, why not setup DKIM on the mail server? Even if you have never done it before, it takes less than 15 minutes to do so. My question here is: what is the gain? 15 minutes? Let me remind you, the vw admin has to change the DNS record to add the public keys. So the only gain is not having to add dkim to the mail server - 15 minutes.

If it is the latter, don't you think the group of people who have all this type of info is rather small? How many people would actually be able to use your PR? All I am saying is that this use case is very specific and the target audience probably small. 

Please don’t take this the wrong way, I do not want to belittle your PR or your use case. I just do not understand why it would warrant adding something that only a few people are able to use.
But it is also very possible that I misunderstood your use case.
Maybe people have access to mail servers that do not sign email, use the right transfer encoding, don't change body and headers, follow all specs to the letter, ...
My comment was not meant to deter anyone from merging your work (which I couldn't, even if I wanted to, since my opinion does not matter at all). It's about me. I am the one, who does not understand your use case, which is why I asked for the why and how.
I am curious about the gain and reusability. That's all.

> By default, From/Subject/To/Date are signed. Those are headers that are unlikely to be changed by default MTA.

This is unfortunately not always the case. e.g. mail servers sometimes use the mail server's timestamp, instead of just forwarding the date: header. As mentioned before, if you were to use your local MTA to send email, you would have to use a smart host to submit the email via an SMTP relay. Once again, you have to know exactly how all the parts in the chain behave.

DorianCoding:
Hello,

I agree with nearly all your comments.

> If it is the former, why not setup DKIM on the mail server? Even if you have never done it before, it takes less than 15 minutes to do so.

If you relay your mail like you say on a server that you don't control. Maybe that server don't do DKIM. Maybe you don't want this server to sign for you. Maybe it is not free to implement DKIM... Even this, if you use a relay that sends your email, you or this provider needs DNS access, unless you don't send mails from your domain.

> My question here is: what is the gain? 15 minutes? Let me remind you, the vw admin has to change the DNS record to add the public keys. So the only gain is not having to add dkim to the mail server - 15 minutes.

The gain is where the person cannot or does not want to implement DKIM on the MTA but has domain control and can know headers signed (the one I've sent to you) are not altered by their MTA.

> If it is the latter, don't you think the group of people who have all this type of info is rather small? How many people would actually be able to use your PR? All I am saying is that this use case is very specific and the target audience probably small.
> Please don’t take this the wrong way, I do not want to belittle your PR or your use case. I just do not understand why it would warrant adding something that only a few people are able to use.


I don't take the wrong way at all and your comments are legitimate. The target audience might be small. This implementation is not optimal but it is a work-around for specific people and deliverability of emails sent by Vaultwarden is crucial for many reasons. As such, it is just an addition of a feature from a crate that already exists in mail. If necessary, we could also implement it as a feature. In conclusion, even if DKIM support should not be done from the client side, it has some applications for some people and the minimum changes required to achieve this should give us reasons to implement it.

tessus:
> The gain is where the person cannot or does not want to implement DKIM on the MTA but has domain control and can know headers signed (the one I've sent to you) are not altered by their MTA.

Ok, this makes sense, at least the "cannot implement" part.

For me it was rather a given that people who had server control (for vw) and control over DNS, also had control over a mail server. But I guess it's possible tha this might not be the case. Weird, but legitimately ok.

My question is answered. Thank you.

As for the implementation I think the current approach is fine. It's not a huge change, the binary size wouldn't increase by a lot, and there are no performace implications, thus a feature is most likely an overkill.

src/mail.rs
```diff
@@ -689,7 +692,44 @@ async fn send_with_selected_transport(email: Message) -> EmptyResult {
...
+                    Err(e) => {
+                        return Err(format!("Cannot read DKIM file. Err is {:?}", e));
+                    }
+                    Ok(key) => match DkimSigningKey::new(&key, algo) {
+                        Ok(d) => d,
+                        Err(e) => {
+                            return Err(format!("Cannot read DKIM file. Err is {:?}", e));
+                        }
+                    },
+                };
```
stefan0xC: Like I said reading the signing key like this should not be necessary if we use the built-in `_file` configuration method. That way we can assume that `CONFIG.dkim_signing_key` has the correct value.

Also the second error message is wrong; it should say that the dkim signing key is invalid not that the file cannot be read.

src/main.rs
```diff
@@ -90,7 +90,10 @@ async fn main() -> Result<(), Error> {
     schedule_jobs(pool.clone());
     db::models::TwoFactor::migrate_u2f_to_webauthn(&mut pool.get().await.unwrap()).await.unwrap();
     db::models::TwoFactor::migrate_credential_to_passkey(&mut pool.get().await.unwrap()).await.unwrap();
-
+    if let Err(e) = mail::check_dkim() {
+        error!("{}", e);
+        exit(1);
+    }
```
stefan0xC: Please look in `src/config.rs` for how configuration values are validated. If you only validate it on startup this will miss configurations changes via the admin panel.

stefan0xC (changes requested):
Thanks for cleaning up your PR a bit. However, I think that you have not implemented the validation in the correct place and also you have not addressed my other concerns at all, so I still think this PR needs to be reworked a bit before it should be merged.

Besides my earlier suggestions of renaming the signing key, splitting dkim_infos, using the `_file` method and documenting the new options to the `.env.template` I would also urge you to look at other configuration options and differentiate between the name and the description of the configuration option, which you can by adding `|>`.

Also instead of having one function that does two things I think that the function for getting the DKIM configuration for signing can be simplified further if you split out the validation as the `check_dkim()` function currently a bit messy and does too many things at once.

src/config.rs
stefan0xC: If you do that you can simplify https://github.com/dani-garcia/vaultwarden/blob/3f40fc49c04fab9f739dc269dfaf8dc0c0fd2720/src/mail.rs#L714-L719

Personally while I do think that we should keep the number of configuration options low I don't think that this should come at the expense of additional complexity. And since the configured value is not used as is anywhere else as far as I can tell (and  the only  benefit of having both options in one seems to be that we can save an additional configuration option.

src/config.rs
stefan0xC: I think if you rename it to DKIM Signing Key it becomes a lot clearer what this is for. The description (separated by `|>`) can than be explicitly about the format requirements of the value.

BlackDex:
I'm not sure if @DorianCoding still want's this implemented. There are some comments from @stefan0xC which need to be addressed. While I'm not seeing a huge amount of people who want this, I'm not really against it, though I still think it should be done by a mail server though.

DorianCoding:
Hello,

I have rebased the commits, made some corrections about the config. The only comment I cannot get is the _file method in the config (https://github.com/dani-garcia/vaultwarden/pull/5620#discussion_r2724880877)

If everything looks fine, I will check that it works. And yes I agree it is a very specific issue but because of all the discussions done and the low overload, I guess it can be implemented.

Thanks.

stefan0xC:
@DorianCoding 
> The only comment I cannot get is the _file method in the config (https://github.com/dani-garcia/vaultwarden/pull/5620#discussion_r2724880877)

There is a built-in function that automatically reads the environment variable from a file (if you have appended `_FILE` to the environment variable name):
https://github.com/dani-garcia/vaultwarden/blob/7cf0c5d67eb81c8b4f2e86b5c8d030bb330faa28/src/util.rs#L396-L410

So instead of requiring users to specify the key via a file path (which can be error prone) you should just let users define the value as we do with every other configuration option and users that want to read it from a file can set a path with `DKIM_SIGNATURE_FILE` as mentioned in https://github.com/dani-garcia/vaultwarden/pull/5620#discussion_r2262201777

stefan0xC:
Also I'm not sure why you are ignoring my comments about the names of the variables and configuration options? It's called a [signing key](https://docs.rs/lettre/0.11.21/lettre/message/dkim/struct.DkimSigningKey.html) and not signature and likewise it's a [`dkim config`](https://docs.rs/lettre/0.11.21/lettre/message/dkim/struct.DkimConfig.html) and not a `sig`. And I still think that you can simplify the `check_dkim()` function further (because the temporary `config` tuple should not be really needed either.).

src/mail.rs
```diff
@@ -703,7 +702,28 @@ async fn send_with_selected_transport(email: Message) -> EmptyResult {
         }
     }
 }
-
+pub fn check_dkim() -> Result<Option<DkimConfig>, String> {
+    match (get_env::<String>("dkim_privatekey"), CONFIG.dkim_infos()) {
```
stefan0xC: If you look at the code base this is not quite how that should be used.

```suggestion
    match (CONFIG.dkim_privatekey(), CONFIG.dkim_infos()) {
```

Neustradamus:
@DorianCoding: This feature is very important, thanks for your work!

src/mail.rs
```diff
@@ -703,7 +702,28 @@ async fn send_with_selected_transport(email: Message) -> EmptyResult {
...
+        (Some(pk), Some(infos)) => {
+                let algo = if CONFIG.dkim_use_rsa() {DkimSigningAlgorithm::Rsa } else { DkimSigningAlgorithm::Ed25519 };
+                let (selector, domain, privatekey) = match (DkimSigningKey::new(pk.as_str(), algo), infos.split(':').collect::<Vec<&str>>()) {
+                    (Ok(sig), split2) if split2.len() == 2 => {
+                        let (selector, domain, sig) =
+                            (String::from(*split2.first().unwrap()), String::from(*split2.last().unwrap()), sig);
+                        (selector, domain, sig)
+                    }
+                    _ => {
+                        return Err("DKIM issue, invalid domain, selector.".to_string());
```
stefan0xC: This error message is unhelpful. Why don't you validate the signing key separately from the other two values? Also why do you insist on keeping the selector and domain with a custom format?

stefan0xC (changes requested):
@DorianCoding You still have not addressed all of my suggestions:

- [ ] let users configure and validate dkim domain and selector separately https://github.com/dani-garcia/vaultwarden/pull/5620#discussion_r2262197166 
- [ ] name the config option `dkim_signing_key` https://github.com/dani-garcia/vaultwarden/pull/5620#discussion_r2262254897
- [ ] improve the documentation for the dkim signing key https://github.com/dani-garcia/vaultwarden/pull/5620#discussion_r2262201777
- [ ] rename the `sig` variable to something less confusing like `dkim_config`https://github.com/dani-garcia/vaultwarden/pull/5620#discussion_r2262194170 
- [ ] document the config options also in `.env.template` https://github.com/dani-garcia/vaultwarden/pull/5620#pullrequestreview-3099791910 
- [ ] run `cargo fmt` and `cargo clippy` https://github.com/dani-garcia/vaultwarden/pull/5620#issuecomment-3167005253

Can you please communicate where you need further assistance or why you don't want to implement them so this becomes less frustrating for me as a reviewer?

DorianCoding:
> @DorianCoding You still have not addressed all of my suggestions:
> 
>     * []  let users configure and validate dkim domain and selector separately [#5620 (comment)](https://github.com/dani-garcia/vaultwarden/pull/5620#discussion_r2262197166)
>
>     * []  name the config option `dkim_signing_key` [#5620 (comment)](https://github.com/dani-garcia/vaultwarden/pull/5620#discussion_r2262254897)
> 
>     * [ ]  improve the documentation for the dkim signing key [#5620 (comment)](https://github.com/dani-garcia/vaultwarden/pull/5620#discussion_r2262201777)
> 
>     * []  rename the `sig` variable to something less confusing like `dkim_config`[#5620 (comment)](https://github.com/dani-garcia/vaultwarden/pull/5620#discussion_r2262194170)
> 
>     * []  document the config options also in `.env.template` [#5620 (review)](https://github.com/dani-garcia/vaultwarden/pull/5620#pullrequestreview-3099791910)
> 
>     * []  run `cargo fmt` and `cargo clippy` [#5620 (comment)](https://github.com/dani-garcia/vaultwarden/pull/5620#issuecomment-3167005253)
> 
> 
> Can you please communicate where you need further assistance or why you don't want to implement them so this becomes less frustrating for me as a reviewer?

I'm sorry you are feeling frustated. I have understand the comments and don't need help. I have made the changes you have asked.

* [x]  let users configure and validate dkim domain and selector separately [#5620 (comment)](https://github.com/dani-garcia/vaultwarden/pull/5620#discussion_r2262197166)
* [x]  name the config option `dkim_signing_key` [#5620 (comment)](https://github.com/dani-garcia/vaultwarden/pull/5620#discussion_r2262254897)
* [x]  improve the documentation for the dkim signing key [#5620 (comment)](https://github.com/dani-garcia/vaultwarden/pull/5620#discussion_r2262201777)
 * [x]  rename the `sig` variable to something less confusing like `dkim_config`[#5620 (comment)](https://github.com/dani-garcia/vaultwarden/pull/5620#discussion_r2262194170)
* [x]  document the config options also in `.env.template` [#5620 (review)](https://github.com/dani-garcia/vaultwarden/pull/5620#pullrequestreview-3099791910)
* [x]  run `cargo fmt` and `cargo clippy` [#5620 (comment)](https://github.com/dani-garcia/vaultwarden/pull/5620#issuecomment-3167005253)

DorianCoding:
I have put examples to show the needed format required by ```lettre``` crate. But if you don't want them we can remove them. DKIM itself is already tested in this crate and there is no need to re-test.

PR #5626, opened 2025-02-21
Abstract persistent files through Apache OpenDAL

txase:
This PR represents the first set of attempting to incorporate the changes from the AWS Serverless POC in #5591 and contains two commits:
1. Refactor the codebase to access persistent files through [Apache OpenDAL](https://opendal.apache.org/)
2. Add support for an optional AWS S3 OpenDAL backend behind the new `s3` feature flag

> I decided to put these two commits into one PR to make it easier for me to get the new abstractions right, and to demonstrate what it takes to add an additional OpenDAL backend. I'm happy to split these into separate sequential PRs if that's preferred.

The first commit looks quite large at first glance, but many changes simply make all file accesses asynchronous and fallible by returning `Result` types. I think the key changes worth reviewing are:
* src/auth.rs: Slightly refactored `initialize_keys()` as needed for OpenDAL access of private key file
* src/config.rs:
    * `CONFIG` is still a synchronous `Lazy` type, but we have to calculate it from async methods. I shoehorned a tiny tokio async thread that runs to completion to calculate the value. The alternative of patching every use of `CONFIG` to be async, for no runtime benefit, seemed not worth doing.
    * See the new `opendal_operator_for_path()` and the abstracted `CONFIG.opendal_operator_for_path_type()` methods for the core of how operators are managed for various paths.
* src/util.rs: Has a new function `save_temp_file()` that abstracts the saving of [TempFiles](https://api.rocket.rs/v0.5/rocket/fs/enum.TempFile) that Rocket creates when files are uploaded
* Wherever persistent data files are used, we now generate an [OpenDAL Operator](https://docs.rs/opendal/latest/opendal/struct.Operator.html) to access them. This includes config.json, sends, attachments, and icons.

The second commit is much smaller and more straightforward. The only thing worth pointing out is that OpenDAL uses [reqsign](https://docs.rs/reqsign/latest/reqsign/) under the covers to configure AWS credentials. However, AWS SDK configs have repeatedly been extended for better credential generation. For example, I use AWS Identity Center (aka AWS SSO) to generate temporary access tokens in my dev environment. reqsign doesn't support AWS SSO configs, but it has an escape hatch I utilized to load credentials. In the escape hatch I load the official AWS SDK config and credential generation crates to generate credentials. The one annoying part of the escape hatch is that reqsign's [AwsCredentialLoad](https://docs.rs/reqsign/latest/reqsign/trait.AwsCredentialLoad.html) trait uses `anyhow::Result`, so we have to pull in anyhow just for this escape hatch :(.

## Trying it out
These changes should be a behavioral no-op for existing use cases. The one minor change is the attachments, icon_cache, and sends folders aren't created at startup as OpenDAL FS service creates them when the first Operator is instantiated for each.

To try out the new S3 changes:
1. Build with the `s3` feature turned on
2. Configure an AWS profile (this implementation honors all standard env vars like `AWS_PROFILE` and `AWS_REGION` along with standard AWS configs like `~/.aws/config`)
3. Set the following config values:
    * `DATA_FOLDER` -> `s3://<bucket in the matching AWS region>[/<optional path prefix>]`
    * `TMP_FOLDER` -> `data/tmp` (or your preference, but must be set to a local path)
    * `TEMPLATES_FOLDER` -> `data/templates` (or your preference, but must be set to a local path)
    * `DATABASE_URL` -> `data/db.sqlite3` (or your preference, but must be set to a valid value)


williamdes:
> Set the following config values: 

can they all be S3 ?
even  DATABASE_URL ?

txase:
No, they can’t all be abstracted through opendal, nor should they be. You don’t want your entire sqlite db to be streamed back and forth to s3 on every record change, for example. One could also imagine tmp files could go through opendal, but it doesn’t make sense to do so even if it’s possible.

williamdes:
> No, they can’t all be abstracted through opendal, nor should they be. You don’t want your entire sqlite db to be streamed back and forth to s3 on every record change, for example. One could also imagine tmp files could go through opendal, but it doesn’t make sense to do so even if it’s possible.

Okay, but TEMPLATES_FOLDER can be in s3 ?
What are they ?

txase:
I believe they are for customizing email and/or admin page html using handlebars. I had two thoughts on them:
* They could be considered data, but they are probably more like assets one would include in an installation. It doesn’t seem like they should be considered user data to me that should be propagated through opendal.
* The folder location is passed to a handlebars library function. Maybe there’s a way to customize the handlebars library behavior to operate through opendal rather than direct file access, but due to the previous point it doesn’t seem worthwhile.

txase:
I fixed all the GHA check issues so they pass. No material changes were made.

BlackDex:
> I fixed all the GHA check issues so they pass. No material changes were made.

Might want to install `pre-commit`, it will help on catching those before you push 😀.

src/util.rs
```diff
@@ -82,6 +82,12 @@ impl Fairing for AppHeaders {
             // 2FA/MFA Site check: api.2fa.directory
             // # Mail Relay: https://bitwarden.com/blog/add-privacy-and-security-using-email-aliases-with-bitwarden/
             // app.simplelogin.io, app.addy.io, api.fastmail.com, quack.duckduckgo.com
+
+            #[cfg(s3)]
+            let s3_connect_src = "https://*.amazonaws.com";
```
williamdes, line 87: this should be configurable, not all S3 use AWS
there is MinIO or Garage as examples

txase: We could remove this and instruct users to add their service's specific url via `ALLOWED_CONNECT_SRC`?

williamdes: yes, less specific code could be better

txase:
Just wanted to check in here. Is the CSP concern the only thing that needs to be addressed? Or do we need further reviews by anyone?

williamdes:
> Just wanted to check in here. Is the CSP concern the only thing that needs to be addressed? Or do we need further reviews by anyone?

> We could remove this and instruct users to add their service's specific url via ALLOWED_CONNECT_SRC?

I agree (https://github.com/dani-garcia/vaultwarden/pull/5626#discussion_r1966193714)

Is there a missing commit ?

txase:
> > Just wanted to check in here. Is the CSP concern the only thing that needs to be addressed? Or do we need further reviews by anyone?
> 
> > We could remove this and instruct users to add their service's specific url via ALLOWED_CONNECT_SRC?
> 
> I agree ([#5626 (comment)](https://github.com/dani-garcia/vaultwarden/pull/5626#discussion_r1966193714))
> 
> Is there a missing commit ?

I have now removed the S3 CSP change and rebased to address conflicts in Cargo.lock. I'm not 100% sure whether you were saying that's the only thing that needed to be addressed, but if so it should be good to go now :).

williamdes (approved):

williamdes:
Thanks, for me ready for review from other users that know better

txase:
@dani-garcia @BlackDex Do you have an estimate of when you can review this? I don't want it to bit-rot :).

BlackDex:
I'm sorry for the silence, but unfortunately i didn't had much time lately. I hope ill have some time soon.

src/config.rs
```diff
@@ -1128,11 +1154,95 @@ fn smtp_convert_deprecated_ssl_options(smtp_ssl: Option<bool>, smtp_explicit_tls
...
+
+    let operator = if path.starts_with("s3://") {
+        #[cfg(not(s3))]
+        return Err(opendal::Error::new(opendal::ErrorKind::ConfigInvalid, "S3 support is not enabled").into());
+
+        #[cfg(s3)]
+        opendal_s3_operator_for_path(path)?
+    } else {
+        let builder = opendal::services::Fs::default().root(path);
+        opendal::Operator::new(builder).map_err(Into::<Error>::into)?.finish()
```
dani-garcia, line 1177: I don't think the map_err is needed here either
```suggestion
        opendal::Operator::new(builder)?.finish()
```

src/config.rs
```diff
@@ -1128,11 +1154,95 @@ fn smtp_convert_deprecated_ssl_options(smtp_ssl: Option<bool>, smtp_explicit_tls
...
+
+    let bucket = url.host_str().ok_or_else(|| format!("Missing Bucket name in data folder S3 URL {path:?}"))?;
+
+    let builder = opendal::services::S3::default()
+        .customized_credential_load(Box::new(OPEN_DAL_S3_CREDENTIAL_LOADER))
+        .bucket(bucket)
+        .root(url.path())
+        .default_storage_class("INTELLIGENT_TIERING");
+
+    Ok(opendal::Operator::new(builder).map_err(Into::<Error>::into)?.finish())
```
dani-garcia, line 1230: I don't think the map_err is needed here either
```suggestion
    Ok(opendal::Operator::new(builder)?.finish())
```

txase:
> Some comments based on an initial review, looks pretty good but I'll need to test it a bit more

Thanks for the review! Your comments look good, I’ll work on addressing them early next week.

src/util.rs
```diff
@@ -816,6 +816,28 @@ pub fn is_global(ip: std::net::IpAddr) -> bool {
...
+    use tokio::io::AsyncReadExt as _;
+
+    let operator = CONFIG.opendal_operator_for_path_type(path_type)?;
+
+    let mut read_stream = temp_file.open().await?;
+    let mut buf = Vec::with_capacity(temp_file.len() as usize);
+    read_stream.read_to_end(&mut buf).await?;
+    operator.write(path, buf).await?;
+    Ok(())
+}
```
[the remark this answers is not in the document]
txase: Oh! This is exactly what I tried and failed to find. Thanks!

src/db/models/attachment.rs
```diff
@@ -104,26 +115,26 @@ impl Attachment {
...
-                Err(e) => Err(e.into()),
-                _ => Ok(()),
+            .map(|_| ())
+            .map_res("Error deleting attachment")
+        }}?;
+
+        let operator = CONFIG.opendal_operator_for_path_type(PathType::Attachments)?;
+        let file_path = self.get_file_path();
+
+        if let Err(e) = operator.delete_iter([file_path.clone()]).await {
```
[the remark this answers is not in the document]
txase: 🤦 I had first tried using the [`remove()`](https://docs.rs/opendal/latest/opendal/struct.Operator.html#method.remove) method. That is now deprecated with a message advising the use of `delete_iter()` instead. I didn't realize there was a simple `delete()` method I could be using. Thanks for catching this!

txase:
@dani-garcia I rebased this PR, then added a commit on top that includes improvements from your comments so you can review the changes more easily. It's now ready for review again. I'm also happy to rebase these changes into the original commits if you prefer.

There was one small change that I had forgotten to make earlier, and that was to switch the S3 operator config to use "virtual-host style" urls. S3 has deprecated the path-style urls that OpenDAL uses by default, and this makes CSP controls easier to implement, too. This is an innocuous one-line change to call `.enable_virtual_host_style()`, but I figured I should call it out to prevent confusion :).

txase:
@dani-garcia Friendly poke :). Do you know when you can review these updates? Thanks!

txase:
@BlackDex I've rebased this PR, and updated the dependencies to the latest versions since this PR has been open. I'd rather not have to rebase it again and again, thankfully this time the only conflicting changes were in the Cargo.toml and Cargo.lock files :). Hopefully you can review this soon?

txase:
@BlackDex @dani-garcia Hey folks, pinging again to see if we can get this across the finish line? Thanks!

dani-garcia (commented):
Sorry for the delay, I've had some time to test this and it's working great. I gave the code a second review and left some comments with some small simplifications we can do, but besides that it looks good to me. 

We should make sure to document which config path options are local only and which ones are local or s3, so there's no confusion. That should be done both in the `.env.template` file and in [the field comments](https://github.com/dani-garcia/vaultwarden/blob/main/src/config.rs#L376) in the config struct, as those are displayed in the admin page.

Once those are done and conflicts are resolved this should be ready for merging. 

After that we should consider creating a wiki page with setup guides, or something like that.

src/api/core/sends.rs
```diff
@@ -406,19 +404,29 @@ async fn post_send_file_v2_data(
...
-    let file_path = folder_path.join(file_id);
+    let file_path = format!("{send_id}/{file_id}");
 
-    // Check if the file already exists, if that is the case do not overwrite it
-    if tokio::fs::metadata(&file_path).await.is_ok() {
-        err!("Send file has already been uploaded.", format!("File {file_path:?} already exists"))
-    }
+    save_temp_file(PathType::Sends, &file_path, data.data, false).await.map_err(|e| {
+        let was_file_exists_error = e
+            .source()
```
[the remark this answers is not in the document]
txase: This was a case where I attempted to keep the original functionality, but it did seem unnecessary. I'm glad we can simplify this.

src/auth.rs
```diff
@@ -40,37 +38,44 @@ static JWT_REGISTER_VERIFY_ISSUER: Lazy<String> = Lazy::new(|| format!("{}|regis
...
+            operator.write(&rsa_key_filename, priv_key_buffer).await?;
+            info!("Private key '{}' created correctly", CONFIG.private_rsa_key());
+            Err(Error::new(ErrorKind::NotFound, "Private key created, forcing attempt to read it again"))
+        }
     }
 
-    let (priv_key, priv_key_buffer) = read_key(true).or_else(|_| read_key(false))?;
+    let (priv_key, priv_key_buffer) = match read_key(true).await {
+        Ok(key) => key,
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => read_key(false).await?,
```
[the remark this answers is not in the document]
txase: Another case where I attempted to keep the exact same behavior because I didn't know of potential side-effects if I changed it, but everything is way simpler by inlining the functionality. I'm glad you have/remember enough context so we can feel good about simplifying it :).

txase:
I updated the PR:
* Rebased onto the latest main branch
* Implemented latest PR comments (d418b812c686618daae904864cf00af1f7af2bdd)
* Added some comment docs in .env.template (dcee1df6d0cdc5b570db7d7f02b50da4dc237af4)

I did not modify config doc comments in config.rs because those comments:
* Only show up in the alt text of the setting names, so they are quite buried
* Would all be in settings that are in the Read-Only section, where doc messages are not helpful in the same way they are in .env.template

This is ready for review again, thanks @dani-garcia!

dani-garcia (approved):
LGTM! Thanks @txase, appreciate the patience.

This is ready to merge now, we had to merge some changes first to make the new release that went out today, so there are some conflicts. I was going to resolve them myself but I don't have permission.

If you can get them resolved or give me permission we will merge your changes.

BlackDex:
@txase, i was looking at dependencies, and i saw that rustls is brought in with S3 enabled, the binary gets a lot bigger, around ~9MiB's bigger. Here is an output of cargo-bloat.

```
no-s3:    57.1% 100.0%  23.5MiB .text section size, the file size is 41.2MiB
w/o-def:  57.6% 100.0%  26.3MiB .text section size, the file size is 45.6MiB
original: 58.7% 100.0%   29.4MiB .text section size, the file size is 50.1MiB
```

I looked at the default features, and saw their might be some optimizations there.
While it still increases the binary with ~4MiB's it's better then the ~9MiB's with the current default features enabled.
It also removes a lot of duplicate crates which have different versions currently, which is also nice.

Since I'm not 100% sure if this would be break stuff, but i think not, i wanted to ask if you could check and test.

```diff
diff --git a/Cargo.toml b/Cargo.toml
index a0e8a61b..9389a88e 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -179,11 +179,11 @@ rpassword = "7.4.0"
 grass_compiler = { version = "0.13.4", default-features = false }
 
 # File are accessed through Apache OpenDAL
-opendal = { version = "0.53.3", features = ["services-fs"] }
+opendal = { version = "0.53.3", features = ["services-fs"], default-features = false}
 
 # For retrieving AWS credentials, including temporary SSO credentials
 anyhow = { version = "1.0.98", optional = true }
-aws-config = { version = "1.6.3", features = ["behavior-version-latest"], optional = true }
+aws-config = { version = "1.6.3", features = ["behavior-version-latest", "rt-tokio", "credentials-process", "sso"], default-features = false, optional = true }
 aws-credential-types = { version = "1.2.3", optional = true }
 reqsign = { version = "0.16.3", optional = true }

```

Let me know if this is possible or not 🙂 , or if there are more features which can be removed of course.

Xuanwo:
I'm willing to upstream SSO support in `reqsign` so that we can remove `aws-config` entirely. Would you like to create a seperate issue instead and ping me there?

txase:
@BlackDex It's not quite so simple to remove the http client feature from the AWS SDK. It compiles, but immediate fails at runtime. _Some_ kind of HTTP client must be provided.

I decided to try building an adapter for the reqwest client already available for other uses in the code base. I was able to get it working in #5917, which both continues to work for me and demonstrates a lot of dropped dependencies in Cargo.lock. Feel free to review and merge that PR if you like what you see :).

txase:
> I'm willing to upstream SSO support in `reqsign` so that we can remove `aws-config` entirely. Would you like to create a seperate issue instead and ping me there?

@Xuanwo I wouldn't recommend this approach. AWS improves their SDK continuously over time, including in config and credential management. You'll just continue to chase them as they add more features and secure mechanisms to manage credentials and configuration. Instead, I suggest making it possible to provide an [aws_config::SdkConfig](https://docs.rs/aws-config/latest/aws_config/struct.SdkConfig.html) into OpenDAL. Then, you can get out of having to keep up with AWS changes and focus on the core parts of OpenDAL that provide more value :).

BlackDex:
@txase , i didn't mentioned removing the http client. That also isn't what is shown in my patch example.

What the case is here is that by default some of the new crates were using rustls, and then via reqwest. By disabling the rustls part, reqwest will use native-tls again, same as the rest of Vaultwarden. And, that brings down the size, and compile time.

So, it would still be using all the code as it already was, only different features.

txase:
@BlackDex This proposed change in the Cargo.toml file:

```diff
-aws-config = { version = "1.6.3", features = ["behavior-version-latest"], optional = true }
+aws-config = { version = "1.6.3", features = ["behavior-version-latest", "rt-tokio", "credentials-process", "sso"], default-features = false, optional = true }
```

leaves all the [aws-config default features](https://docs.rs/crate/aws-config/latest/features#default) enabled except for `default-https-client`. Without `default-https-client` enabled, and without any additional https client provided (like in #5917), the AWS SDK panics at runtime:

```
thread '<unnamed>' panicked at /.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/aws-smithy-runtime-1.8.3/src/client/orchestrator/operation.rs:408:13:
a http_client is required. Enable the `default-https-client` crate feature or configure an HTTP client to fix this.
```

If we want to remove the AWS SDK built-in rustls-based https client, we need to provide a client ourselves. That's what #5917 does.

MOZGIII:
Note that the wiki doesn't mention S3 still - so this feature is implemented but not documented

BlackDex:
> Note that the wiki doesn't mention S3 still - so this feature is implemented but not documented

It also is not build by default, and i think we will not do this by default for now at least.
And you are free to update the wiki to add this.

MOZGIII:
> And you are free to update the wiki to add this.

Yes, but don't expect me to. :)

I feel like this is something that be "maintained" properly by the core team first - as in, acknowledged by taking ownership of this whole ordeal. This also connects the feature not being enabled by default - why not? I assume there is some hesitation because this feature has not yet been properly emancipated.

Also, did anyone look into the OpenDAL in terms of audit?

BlackDex:
Enabling a feature which adds a lot of dependencies and increases the size of the binary for a feature almost nobody uses (as far as we know) is a bit too much for a default.

To be fair, we have enough to do already, any help for maintaining the wiki would be very nice.

MOZGIII:
> Enabling a feature which adds a lot of dependencies and increases the size of the binary for a feature almost nobody uses (as far as we know) is a bit too much for a default.

Well, I'd use this feature if it was readily available. I won't bother with making custom builds. Let's create a separate issue to discuss now we can bring this to the users?

> To be fair, we have enough to do already, any help for maintaining the wiki would be very nice.

I understand you very well here :D That said, personally I'd rather help by writing the code than docs. If I have time for it.

txase:
This is the first part of a multi-stage effort outlined in #5591. While it is possible to document how to use just this one s3 part, every time we merge another aws service implementation the same users will likely want to use them, too, and they would have to do extra effort to keep up with the progress. In the end, my goal is to make it easy to deploy via a CloudFormation template, but I don’t have the bandwidth to maintain docs and a cfn template through these changes.

I suggest we instead focus on getting the other aws functionality merged (eg #5910), then we can focus on making it easy for users to adopt. Until then, I think it should remain in a state where people can use it as bleeding edge functionality, eg via guidance in this PR summary.

MOZGIII:
Well, it is reasonable.

For context: have a similar goal of ease of deployment, however I don't care about AWS - only S3 (and mostly non-AWS flavors of it); my plan is to deploy to Kubernetes. It is not vendor-locked, and more people have access to a k8s cluster somewhere than to an AWS accounts nowadays. There are a lot of helm charts out there already - but none have S3 support...

So, for myself, no further work in the #5591 pipeline would be useful. But I'll just wait for this work to get to completion - the main blocker for me is the lack of the official s3-enabled docker images.

fxzxmicah:
Why is it only for AWS instead of the general S3 standard? Otherwise, my hostility towards this feature would be a bit less.
Another source of my hostility towards this feature is that there are quite a few tools that can work with Vaultwarden to achieve similar things. So, do we really need to implement it again in Vaultwarden?
Moreover, based on this development trend, I can't even imagine how many different backends there will be in the future, turning it into another Rclone (not to belittle Rclone).
Finally, I really can't imagine what use this feature has (or rather, its usefulness is too limited), because after all, it can't upload all files to AWS.

EDIT: I just read #5591 and understood what you're trying to do. It also answers some of the questions I raised above. However, I don't even think these kind of abilities specialized for a specific cloud platform should be merged into the upstream. Instead, it should exist as a separate project.

EDIT: If AWS has donated to this project, please ignore all my remarks.

7heMech:
I see the benefit in modularity and S3 support, if it could be deployed to Cloudflare via WASM and R2 as storage backend would be dope, but that's probably way off, I also do get not increasing binary size pointlessly.

PR #5673, opened 2025-03-07
On member invite and edit access_all is not sent anymore

Timshel:
With client `bw_web_v2025.1.1` the `access_all` field is not sent anymore.

As such when selecting `Owner` or `Admin` the user end-up with no permissions.

### DIFF
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -997,8 +997,6 @@ struct InviteData {
     r#type: NumberOrString,
     collections: Option<Vec<CollectionData>>,
     #[serde(default)]
-    access_all: bool,
-    #[serde(default)]
     permissions: HashMap<String, Value>,
 }
 
@@ -1012,7 +1010,7 @@ async fn send_invite(
     if org_id != headers.org_id {
         err!("Organization not found", "Organization id's do not match");
     }
-    let mut data: InviteData = data.into_inner();
+    let data: InviteData = data.into_inner();
 
     // HACK: We need the raw user-type to be sure custom role is selected to determine the access_all permission
     // The from_str() will convert the custom role type into a manager role type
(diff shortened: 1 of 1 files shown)


Timshel:
Sorry for the broken PR, was testing on top of the SSO branch and missed that the cherry-pick was not building on top of `main` :(.

src/api/core/organizations.rs
```diff
@@ -1030,13 +1028,12 @@ async fn send_invite(
...
-    if raw_type.eq("4")
-        && data.permissions.get("editAnyCollection") == Some(&json!(true))
-        && data.permissions.get("deleteAnyCollection") == Some(&json!(true))
-        && data.permissions.get("createNewCollections") == Some(&json!(true))
-    {
-        data.access_all = true;
-    }
+    let access_all = new_type == MembershipType::Owner
+        || new_type == MembershipType::Admin
+        || (raw_type.eq("4")
```
iDaoU, line 1033: Why do we still want to have this `raw_type` check when we are able to parse it into `new_type`?

Also looks like if `raw_type` cannot be parsed to `new_type`, it throw an error? So that should always be possible?

Timshel: It's not something I changed but as mentioned:

> Membership::from_str will convert custom (4) to manager (3)

Timshel:
I wonder if a migration should be added to correct the `access_all` value depending on the `atype` otherwise post patch admin will have to save each affected admin/owner user role again to fix it.

Timshel:
@BlackDex I have seen some mention of this PR in another issue: https://github.com/dani-garcia/vaultwarden/issues/5711#issuecomment-2804231808

I decided not to answer over there since it concerns only the issue:

> Once his account was back he had user permission without any collections and once I gave him admin again he still did not get any of the collections. 

You mention:

> But in the end, the access_all feature should be removed, as this is just not how Bitwarden works anymore and support for this way of working will be difficult in the future with all the separate rolls and collection/group permissions they are using now.

While I agree that this is just a quick fix and more change are probably needed, I'm not aware of any issue which would prevent merging this to ensure that when saving a `admin` and `owner` roles access to all current and futures collections is granted.

src/api/core/organizations.rs
```diff
@@ -1030,13 +1028,12 @@ async fn send_invite(
...
     // If the box is not checked, the user will still be a manager, but not with the access_all permission
-    if raw_type.eq("4")
-        && data.permissions.get("editAnyCollection") == Some(&json!(true))
-        && data.permissions.get("deleteAnyCollection") == Some(&json!(true))
-        && data.permissions.get("createNewCollections") == Some(&json!(true))
-    {
-        data.access_all = true;
-    }
+    let access_all = new_type == MembershipType::Owner
+        || new_type == MembershipType::Admin
```
BlackDex, line 1032: this can be converted to:
```rust
let access_all = new_type >= MembershipType::Admin
```

src/api/core/organizations.rs
```diff
@@ -1565,13 +1559,12 @@ async fn edit_member(
...
```
BlackDex, line 1563: this can be converted to:
```rust
let access_all = new_type >= MembershipType::Admin
```

BlackDex (changes requested):
I think this is a good change to add, to preserve the previous way of working.
Thanks!!

Just a minor adjustment to make it more inline with the rest of the code.

src/api/core/organizations.rs
Timshel: Unless I'm mistaken `MembershipType::Owner == 0` and `MembershipType::Admin == 1` so the check should be `new_type <= MembershipType::Admin` ?

BlackDex (dismissed):

PR #5696, opened 2025-03-15
cockroachdb support

haras-unicorn:
`2020-08-02-025025_add_favorites_table` migration patch applied from [support for cockroachdb discussion](https://github.com/dani-garcia/vaultwarden/discussions/2689).
The reason the migration had to be split is because [create table and insert fail in the same transaction in cockroachdb](https://github.com/cockroachdb/cockroach/issues/14548).

Specified `INT4` instead of `INTEGER` datatype in all postgresql migrations because [postgresql uses 32-bit integers by default](https://www.postgresql.org/docs/current/datatype-numeric.html) while [cockroachdb uses 64-bit integers by default](https://www.cockroachlabs.com/docs/stable/int).

Tested on version 1.33.0 and can confirm migrations work on my machine for what its worth.

I was able to make a user account and import all of my keepassxc stuff without a hitch.
For me this functionality is enough so I'll open this up for review but I'd still greatly appreciate other people testing because I only did basic tests to confirm functionality that I'm interested in works.


Timshel:
Hey,

Not a maintainer but have you tested adding your patch on an existing PG database ?

Because unless I'm mistaken from the point of view of existing deployment you are adding a new migration `2020-08-02-025026_add_favorites_table_fix` which I believe will fail when run.

Last time I tested the fact that some recent migration were run will not prevent running a migration with an old date prefix (Otherwise I will have troubles with my own PR ^^).

haras-unicorn:
good point - i didnt know that it runs migrations like that
usually migration order matters so i assumed it just runs missing migrations from the latest ran and not the ones before the latest
do you have a suggestion how to make it work?
maybe in the same migration add a COMMIT and BEGIN right after table creation? idk if that would work with diesel tho

Timshel:
From a quick search Diesel run each migration in a transaction.
PG and cockroachdb appears to support nested transaction, but it will probably still trigger the bug.

Alternative could be to recreate the `favorite` column if missing at the start of the new migration to handle the case of existing deployment.

haras-unicorn:
maybe in the migration you can commit the transaction started by Diesel after the table creation then start another one ? - this was my original solution but i was worried it wasnt gonna work so i took the patch from the discussion
ill try that - if it works in cockroachdb would you be willing to test in postgres?

Timshel:
> if it works in cockroachdb would you be willing to test in postgres?

Not very hard to spin-up PG :sweat_smile:, but yes should have some time to test it

BlackDex:
I have tried this, using the latest CockroachDB version available.
But as all my previous tests, it fails. Either it's complaining about Transactions not being active, which is false, since migrations are always run within a transaction. Trying to add `BEGIN` and `COMMIT` on places where needed helped a bit, but it just refuses to fully finish.

Since this has been asked multiple times already i did gave it a try again.
But as long as Diesel is not going to support CockroachDB, or CockroachDB isn't going to support all/full PostgreSQL i think this will just not happen.

Too much risks and check which can go wrong using this. While PostgreSQL just works, and just starts.

Therefor, going to close this PR, as I'm hesitant to imply supporting this database.

PR #5709, opened 2025-03-19
Update Rust, Crates and other deps

BlackDex:
- Updated Rust to v1.85.1
- Updated crates and fixed breaking changes
- Updated datatables js
- Updated GitHub Actions

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -70,13 +70,13 @@ jobs:
 
     steps:
       - name: Initialize QEMU binfmt support
-        uses: docker/setup-qemu-action@53851d14592bedcffcf25ea515637cff71ef929a # v3.3.0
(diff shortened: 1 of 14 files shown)


.github/workflows/release.yml
```diff
@@ -70,13 +70,13 @@ jobs:
 
     steps:
       - name: Initialize QEMU binfmt support
-        uses: docker/setup-qemu-action@53851d14592bedcffcf25ea515637cff71ef929a # v3.3.0
+        uses: docker/setup-qemu-action@53851d14592bedcffcf25ea515637cff71ef929a # v3.6.0
```
dani-garcia, line 73: This one wasn't updated but the version number was, accidental?

BlackDex: Yes, the replacement failed i guess

BlackDex: Fixed!

dani-garcia (approved):

PR #5712, opened 2025-03-19
Fix invited user registration without SMTP

Timshel:
Perform the user check only if verification email is used.

It was blocking the registration of invited user when email was disabled.

Edit: I wonder if it would not make sense to just always send the email ? 
It does create an unintuitive flow if the user already exists since it will fail later on, but currently it means that if the user has deleted the email he has no way to verify his email anymore.

### DIFF
--- a/src/api/identity.rs
+++ b/src/api/identity.rs
@@ -744,8 +744,13 @@ async fn register_verification_email(
 
     let should_send_mail = CONFIG.mail_enabled() && CONFIG.signups_verify();
 
-    if User::find_by_mail(&data.email, &mut conn).await.is_some() {
-        if should_send_mail {
+    let token_claims =
+        crate::auth::generate_register_verify_claims(data.email.clone(), data.name.clone(), should_send_mail);
+    let token = crate::auth::encode_jwt(&token_claims);
+
+    if should_send_mail {
+        let user = User::find_by_mail(&data.email, &mut conn).await;
+        if user.filter(|u| u.private_key.is_some()).is_some() {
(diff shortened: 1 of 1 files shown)


src/api/identity.rs
```diff
@@ -744,26 +744,25 @@ async fn register_verification_email(
...
-    }
-
     let token_claims =
         crate::auth::generate_register_verify_claims(data.email.clone(), data.name.clone(), should_send_mail);
     let token = crate::auth::encode_jwt(&token_claims);
 
     if should_send_mail {
-        mail::send_register_verify_email(&data.email, &token).await?;
+        if User::find_by_mail(&data.email, &mut conn).await.is_some() {
+            if should_send_mail {
```
dani-garcia, line 753: This `if should_send_mail` is always true, as we're already checking it two lines before

dani-garcia (commented):
LGTM, just one duplicate if with the move.

We can consider changing the email sending from `user.is_none()` to `user.is_none() || user.password_hash.is_none()`, which should allow resending  registration emails to users that are invited but not registered yet. Is that what you're thinking about?

Timshel:
> just one duplicate if with the move.

Sorry about that :(.

For the resending was thinking of sending it in all cases but your suggestion is way better since it will prevent abuse.

Hope you don't mind used the `private_key` as an indicator that the user is initialized (A leftover of when I made a VW version which was not storing the MP when using OIDC).

dani-garcia (approved):
LGTM, yeah checking for private key works as well

PR #5719, opened 2025-03-24
Update email.rs

setsecurity:
Add ip_src on logs when email 2fa token is invalid or not available
Changes for fail2ban purposes

### DIFF
--- a/src/api/core/two_factor/email.rs
+++ b/src/api/core/two_factor/email.rs
@@ -204,7 +204,7 @@ pub async fn validate_email_code_str(user_id: &UserId, token: &str, data: &str, 
         .map_res("Two factor not found")?;
     let Some(issued_token) = &email_data.last_token else {
         err!(
-            "No token available",
+            format!("No token available! Server time: {} IP: {}", current_time.format("%F %T UTC"), ip.ip),
             ErrorEvent {
                 event: EventType::UserFailedLogIn2fa
             }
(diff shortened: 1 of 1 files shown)


dani-garcia:
Adding the IP address to that error message sounds reasonable, but I don't think we need to add the current time as well (also current_time is not defined in that scope, so it wouldn't work either way).

tessus:
Isn't date and time automatically added in the log? Thus it isn't needed anyway.

e.g. in my vaultwarden log the entries look like this:

```
[2025-04-04 14:46:08.362][start][INFO] Rocket has launched from http://0.0.0.0:4040
[2025-04-04 14:46:43.802][request][INFO] GET /api/config
[2025-04-04 14:46:43.802][response][INFO] (config) GET /api/config => 200 OK
```

BlackDex:
The date/time is already part of the mail headers.

setsecurity (commented):
removed current_time

tessus:
@dani-garcia can you please run the pipelines? this PR seems to be ready...

tessus:
thanks @BlackDex

tessus:
@setsecurity if you give me access to your branch I can fix the error. Otherwise, you need to add the ip to the call of the function, but don't add the entire header, just the ip.

setsecurity (commented):
ok :)

PR #5743, opened 2025-04-04
Fix mysqlclient-sys building

BlackDex:
Because of some issues with mysqlclient we need to use buildtime bindgen. This also needed some extra environment variables to point the bindgen to the correct files and correct version.

Also update some other crates.


docker/Dockerfile.debian
```diff
@@ -105,8 +105,15 @@ RUN source /env-cargo && \
         echo "export CARGO_TARGET_$(echo "${CARGO_TARGET}" | tr '[:lower:]' '[:upper:]' | tr - _)_LINKER=/usr/bin/$(xx-info)-gcc" >> /env-cargo && \
         echo "export PKG_CONFIG=/usr/bin/$(xx-info)-pkg-config" >> /env-cargo && \
         echo "export CROSS_COMPILE=1" >> /env-cargo && \
+        # Export build env's for OpenSSL
         echo "export OPENSSL_INCLUDE_DIR=/usr/include/$(xx-info)" >> /env-cargo && \
         echo "export OPENSSL_LIB_DIR=/usr/lib/$(xx-info)" >> /env-cargo ; \
+        # Export build env's for mysqlclient-sys (mariadb)
+        # The version is linked to the mariadb-connector-c a.k.a libmariadb-dev version
+        # See: https://mariadb.com/kb/en/about-mariadb-connector-c/
+        echo "export MYSQLCLIENT_VERSION=3.3.14" >> /env-cargo ; \
```
dani-garcia, line 114: Would this need to be updated as the `libmariadb-dev` package in our image gets new versions? If so can we extract the version number using `dpkg` or `dpkg-query`?

BlackDex: No, not really. I tried to look for that, but the main point is that the current mysqlclient-sys doesn't support the version number located anywhere inside any of the files in a sensible way. And, if that version doesn't match anything currently matchable by mysqlclient-sys, building will break again.

BlackDex: Also, Debian will probably only update this in there new **Trixie** version and in those cases we probably need to adjust a bit more then only `bookworm` to `trixie`.

And, i reported an issue regarding the version check which is invalid here: https://github.com/sgrif/mysqlclient-sys/issues/65

dani-garcia (approved):

PR #5694, opened 2025-03-15
Add AnonAddy/SimpleLogin self host feature flag

PseudoResonance:
This adds the feature flags for bitwarden/android/pull/4711 and bitwarden/android/pull/4715. These only apply to Android at the moment unfortunately, but hopefully will come to iOS soon.

PR #5802, opened 2025-04-25
Fix minimum Android version for self-host email alias feature flags

PseudoResonance:
This is very minor, but I just noticed I made a typo in the minimum Android client versions for the AnonAddy/Addy.io and SimpleLogin self hosted feature flags from #5694.

They require a minimum of [2025.3.0](https://github.com/bitwarden/android/releases/tag/untagged-4731eaadac73f3dfbbb8), not 2025.2.0!

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -354,8 +354,8 @@
 ## - "inline-menu-positioning-improvements": Enable the use of inline menu password generator and identity suggestions in the browser extension.
 ## - "ssh-key-vault-item": Enable the creation and use of SSH key vault items. (Needs clients >=2024.12.0)
 ## - "ssh-agent": Enable SSH agent support on Desktop. (Needs desktop >=2024.12.0)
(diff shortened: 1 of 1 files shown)


PseudoResonance:
Apparently these should be [enabled by default in 2025.4.0 though](https://github.com/bitwarden/android/issues/4308#issuecomment-2835414397), although still not sure about iOS status.

BlackDex (approved):

seang96:
I am on 2025.4.0 and still don't see it so I assume this is still preventing it even though it is enabled by default?

PseudoResonance:
> I am on 2025.4.0 and still don't see it so I assume this is still preventing it even though it is enabled by default?

This is just a documentation fix for a PR that's already been accepted earlier (#5694). I thought they were planning to enable it by default on 2025.4.0, but I'm not entirely sure how that works or if it has truly happened. Regardless, you can just add the flags to the Vaultwarden config and it should work.

PR #5832, opened 2025-05-05
Toggle providers using class

Timshel:
Starting with `web-vault` release `v2025.4.0` it will become possible to toggle the 2fa providers using a class. 
Cf https://github.com/vaultwarden/vw_web_builds/pull/15

Without this change the selector will start breaking with  `v2025.4.1` since the element switched to `bit-item` (cf [166cb75](https://github.com/vaultwarden/vw_web_builds/commit/166cb75401f4578a8d809186df3ca9e1134c0113)).

### DIFF
--- a/src/static/templates/scss/vaultwarden.scss.hbs
+++ b/src/static/templates/scss/vaultwarden.scss.hbs
@@ -112,14 +112,14 @@ app-login form div + div + div + div + hr + p {
 
 {{#unless mail_enabled}}
 /* Hide `Email` 2FA if mail is not enabled */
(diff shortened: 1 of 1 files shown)


BlackDex (approved):

stefan0xC (approved):
As stated in the referenced PR the [provider types](https://github.com/bitwarden/clients/blob/web-v2025.5.0/libs/common/src/auth/enums/two-factor-provider-type.ts) seem to be statically assigned, so this should should be added once #5860 has been merged.

dani-garcia (approved):

PR #5871, opened 2025-05-20
web-client now request email 2fa

Timshel:
With version `v2025.5.0` it appears that the web client is now calling the endpoint `api/two-factor/send-email-login` when an email 2FA is necessary. 

This result in sending two emails with verification codes (the first one becoming immediately invalid).

Alternative could be to check the client version, but I have not check if future extension or mobile app will have the same logic.

### DIFF
--- a/src/api/identity.rs
+++ b/src/api/identity.rs
@@ -17,7 +17,7 @@ use crate::{
         push::register_push_device,
         ApiResult, EmptyResult, JsonResult,
     },
-    auth::{generate_organization_api_key_login_claims, ClientHeaders, ClientIp},
+    auth::{generate_organization_api_key_login_claims, ClientHeaders, ClientIp, ClientVersion},
     db::{models::*, DbConn},
     error::MapResult,
     mail, util, CONFIG,
@@ -28,7 +28,12 @@ pub fn routes() -> Vec<Route> {
 }
 
 #[post("/connect/token", data = "<data>")]
-async fn login(data: Form<ConnectData>, client_header: ClientHeaders, mut conn: DbConn) -> JsonResult {
+async fn login(
+    data: Form<ConnectData>,
+    client_header: ClientHeaders,
+    client_version: Option<ClientVersion>,
+    mut conn: DbConn,
+) -> JsonResult {
     let data: ConnectData = data.into_inner();
 
     let mut user_id: Option<UserId> = None;
@@ -48,7 +53,7 @@ async fn login(data: Form<ConnectData>, client_header: ClientHeaders, mut conn: 
             _check_is_some(&data.device_name, "device_name cannot be blank")?;
             _check_is_some(&data.device_type, "device_type cannot be blank")?;
 
-            _password_login(data, &mut user_id, &mut conn, &client_header.ip).await
+            _password_login(data, &mut user_id, &mut conn, &client_header.ip, &client_version).await
         }
         "client_credentials" => {
             _check_is_some(&data.client_id, "client_id cannot be blank")?;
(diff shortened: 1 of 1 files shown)


BlackDex:
Yea i noticed this yesterday during testing, didn't looked into it further.

src/db/models/device.rs
```diff
@@ -139,6 +139,20 @@ impl Device {
     pub fn is_cli(&self) -> bool {
         matches!(DeviceType::from_i32(self.atype), DeviceType::WindowsCLI | DeviceType::MacOsCLI | DeviceType::LinuxCLI)
     }
+
+    pub fn is_browser(&self) -> bool {
```
stefan0xC, line 143: Given that Bitwarden calls the browser extension [browser](https://github.com/bitwarden/clients/releases/tag/browser-v2025.5.0) I think we should rename this to `is_web` instead.

src/api/identity.rs
```diff
@@ -689,8 +693,9 @@ async fn _json_err_twofactor(
                     err!("No twofactor email registered")
                 };
 
-                // Send email immediately if email is the only 2FA option
-                if providers.len() == 1 {
+                // Send email immediately if email is the only 2FA option.
+                // If the device is a browser the client will make the request to send the token.
+                if providers.len() == 1 && !device.is_browser() {
```
stefan0xC, line 698: As far as I've tested it this also happens in the Browser extension (using chromium and `v2025.5.0`). I've not yet tested the CLI version (`v2025.5.0` is not yet released) but given that the clients share the same code base this most likely will also happen in the CLI version.

stefan0xC (changes requested):
Given that this also happens in other clients (at least the browser extension) which we have no control over could we do a `ClientVersion` check if they are >= `v2025.5.0`?

BlackDex:
I btw didn't noticed this at Bitwarden it self for some reason.
Ill have to check again.

src/api/identity.rs
Timshel: Ok will check, initially took the easiest approach :).

stefan0xC:
> I btw didn't noticed this at Bitwarden it self for some reason. Ill have to check again.

The Bitwarden server does not seem to send the `TwoFactorProvider` and `TwoFactorProvider2` fields in the response anymore, so maybe that's why?
```json
{"error":"device_error","error_description":"New device verification required","ErrorModel":{"Message":"new device verification required","Object":"error"}}
```

Timshel:
@stefan0xC switching the response change the flow, the token is not sent twice anymore but the following `/identity/connect/token` query is done with :

```url
scope=api%20offline_access
&client_id=web
&deviceType=10
&deviceIdentifier=4fb8e7f7-e260-4c48-959f-21901f4b7fb8
&deviceName=firefox
&grant_type=password
&username=test%40yopmail.com
&password=vOsxJYu1r5XTtHZo9rbQQfC1%2B9ckVZZ9lF%2Fy16MuByY%3D
&newDeviceOtp=408381
```

vs 

```url
scope=api%20offline_access
&client_id=web
&deviceType=10
&deviceIdentifier=4fb8e7f7-e260-4c48-959f-21901f4b7fb8
&deviceName=firefox
&twoFactorToken=721596
&twoFactorProvider=1
&twoFactorRemember=0
&grant_type=password
&username=test%40yopmail.com
&password=vOsxJYu1r5XTtHZo9rbQQfC1%2B9ckVZZ9lF%2Fy16MuByY%3D
&captchaResponse=

```

I'll the client code directly

Timshel:
So changing the response trigger the device verification flow:

vaultwarden_device-verification

The nice part is the screen now has the ability to resend the code if needed (`api/accounts/resend-new-device-otp`).
But I tested with two authenticators activated, and it's not possible to select which one to use and appears to rely only on the email.

For now, I'll look at checking the version to send or not the email.

Timshel:
I decided not to filter against the mobile applications, hopefully the same logic will apply and if not they have the ability to request to send a new code (at least on android did not check IOS).

PR #5798, opened 2025-04-23
Sync with Upstream

BlackDex:
Synced the API responses of Vaultwarden to match more closely the current Bitwarden version (Mainly v2025.4.x and some v2025.5.x). This solved several issues which are currently open, but also some outstanding tasks.

- Updated several Bitwarden source links to current versions
- Prevent deprecated experimental client feature flags from being returned, but also from causing errors during starting.
- Fixed several issues with the Push Notifications and Android client registrations (And possibly also iOS)
- Added or removed several response key/values to match Bitwarden
- Send an email if someone tries to change their email address to an address which already has an account (Fixes #5630)
- Added a `Small` error-kind which is needed for the org-import to show the correct error message
- Allow members with `can manage` access to a collection to import personal/non-collection backups/exports into the allowed collection. Fixes #5592
- Added remove collection functionality with Bulk Collection Editing. Fixes #5737

By fixing several responses like which users are a member of a collection or group some issues seem to have been resolved since I'm unable to reproduce them anymore. This is either thanks to those changes or some web-vault updates or both.
- Fixes #5590
- Fixes #5609

PR #5874, opened 2025-05-21
Remove old client version check

Timshel:
While adding the version check for email 2fa found this old check which is probably not needed anymore.


BlackDex:
I have this also in my #5798 PR.

PR #5917, opened 2025-06-01
Use existing reqwest client for AWS S3 requests

txase:
This removes a lot of duplicate client dependency bloat for roughly equivalent functionality.

Based on suggestion by @BlackDex in https://github.com/dani-garcia/vaultwarden/pull/5626#issuecomment-2926701148.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -182,7 +182,7 @@ dependencies = [
  "futures-lite",
  "parking",
  "polling",
- "rustix 1.0.7",
+ "rustix",
  "slab",
  "tracing",
  "windows-sys 0.59.0",
@@ -214,7 +214,7 @@ dependencies = [
  "cfg-if",
  "event-listener 5.4.0",
  "futures-lite",
- "rustix 1.0.7",
+ "rustix",
  "tracing",
 ]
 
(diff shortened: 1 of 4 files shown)


txase:
I tested the updates from @BlackDex's merge just now. They still work without issue. 👍

dani-garcia (approved):

tessus:
Just as an info, since this PR was supposed to make the binary smaller:

| Version         | Description     | Bytes      |
|-----------------|-----------------|------------|
| 1.34.1-a545636e | without this PR | 39,574,288 |
| 1.34.1-3b48e6e9 | with this PR    | 39,568,456 |

Build command: `cargo +stable build --features sqlite,mysql,postgresql --release`

stable is rust 1.88.0 (6b00bc388 2025-06-23)

P.S.: It seems that s3 is not a default feature, thus this comment is obsolete. My bad.

BlackDex:
Well, not only size, but also compile time and it also brought a lot of outdated crates for some reason. Duplicate crates with different versions is also not very useful.

txase:
The effects of this PR are mostly mooted if you don’t enable the s3 feature. The compile size should be much less after this PR with the feature enabled.

PR #5929, opened 2025-06-04
Add support for the "Login with passkey" option

zUnixorn:
Hello,

This PR implements the endpoints needed for passkey login into the Bitwarden webclient.

I marked this as a draft for now, since I'd like to first get some feedback on this PR before I start pursuing this further or if this is even a wanted feature for Vaultwarden.

## Working Features
* Login with passkey without using it for encryption
* Login with passkey while using it for encryption
* Adding a new passkey
* Listing all registered passkeys
* Deleting a passkey

## Testing this Branch
To test this branch, the fronted needs to be build with [this](https://github.com/vaultwarden/vw_web_builds/commit/08498dd2059d10f1eed7dd6082619686f4118bd7) line appended, to allow for passkey registration.

## Open TODOs
- [ ] Cleanup the code
	- [ ] there is a bunch of duplicate code between _login_password() and _login_webauthn()
	- [ ] the webauthn stuff could be moved into its own file / module
	- [ ] Resolve the TODOs
- [ ] Figure out how to handle the intermidate webauthn state:
	- [ ] Saved in the database (I think that's what's already done in the 2FA webauthn), but this probably would only work for registration, not login
	- [ ] Passed around as an encrypted string in the `token` field (I think this is what bitwarden does with [this C# Api](https://learn.microsoft.com/en-us/aspnet/core/security/data-protection/introduction?view=aspnetcore-9.0))
- [ ] Decide what to do about the unimplemented [endpoints](https://github.com/bitwarden/server/blob/v2025.5.3/src/Api/Auth/Controllers/WebAuthnController.cs) that are implemented by upstream (they don't seem to be used currently by the webclient)
- [ ] Figure out why the frontend doesn't work in firefox for passkeys


BlackDex:
First, thanks for contributing to Vaultwarden!

Second, if adding this, it might also be the best to update webauthn to the latest version. The only issue here is, that this needs some form of migration for the currently configured 2FA objects, which would become invalid. Luckily the webauthn crate has the old struct still available for this reason in their latest version, but it might be complex.

We already have a migration which is done from an even older format we used before. Though, I'm not sure if we would still need to support that as it has been a long long time ago already, and admins should have installed a newer version already for a long time now. (What do you think @dani-garcia ? )

The main reason for this is that there are several fixes and updates to better support passkeys via webauthn in the latest version.

tessus:
> I'm not sure if we would still need to support that

I don't think it is necessary, unless the migration is not a one-go approach, but only happens when an item with one or more 2FA objects is accessed (which would be a serious problem anyway).

Admins can migrate to the latest version of vw with the "old" migration path and then to the most recent vw version. Such migration paths are vey common: wikimedia, database engines, operating systems, ....

> The only issue here is, that this needs some form of migration for the currently configured 2FA objects, which would become invalid.

Hasn't Stefan started a branch for upgrading webauthn at one point?

BlackDex:
> I don't think it is necessary, unless the migration is not a one-go approach, but only happens when an item with one or more 2FA objects is accessed (which would be a serious problem anyway).

The current migration check/step is done during startup.
While i think most people should be running a version for which this step is ready done long time ago, it is still something to take into account, and skip items which are that old, and exclude those from the 2fa, and upon detection of such an old, unsupported version, maybe enable email 2fa when that was the only 2fa feature enabled. Simply to prevent no 2fa at all anymore.

That would make it much easier to migrate, and maybe even act upon invalid or unsupported webauthn objects.

Try to convert from old, if that also doesn't work, then disable.

tessus:
> The current migration check/step is done during startup.

In that case, it is very easy.

Rmove the old migration path from the code. Upon startup start transaction. Migrate to new webauthn objects. If it encounters an "old" object that cannot be migrated, roll back transaction and show the following message:

`You have to upgrade to vaultwarden 1.34.1 first.`

P.S.: In the release notes and maybe even in the `.env.template`, you can mention that if one was still using vaultawrden < x.y.z, they'd have to do an intermediary upgrade.

zUnixorn:
Just to be sure I understood you both correctly, there would be 3 versions for 2FA Webauthn Registration Structs:
* The original [U2FRegistration](https://github.com/dani-garcia/vaultwarden/blob/1.34.1/src/api/core/two_factor/webauthn.rs#L38)
* The current [WebauthnRegistration](https://github.com/dani-garcia/vaultwarden/blob/1.34.1/src/api/core/two_factor/webauthn.rs#L88) that stores a [Credential](https://docs.rs/webauthn-rs/0.3.2/webauthn_rs/proto/struct.Credential.html)
* A future WebauthnRegistration that stores a [Passkey](https://docs.rs/webauthn-rs/latest/webauthn_rs/prelude/struct.Passkey.html) instead of a [Credential](https://docs.rs/webauthn-rs/0.3.2/webauthn_rs/proto/struct.Credential.html) (since `Credential` is private in webauthn-rs 0.5 and only [exposed](https://docs.rs/webauthn-rs/latest/src/webauthn_rs/interface.rs.html#57-59) via the `Passkey` struct)

The [current migration at the start](https://github.com/dani-garcia/vaultwarden/blob/1.34.1/src/main.rs#L91) is from `U2FRegistration` to `WebauthnRegistration`.
When updating to webauthn-rs 0.5, the migration at the start should instead migrate from `WebauthnRegistration` with a `Credential` to a `WebauthnRegistration` version with a `Passkey`.
If it encounters an error because the Database still contains `U2FRegistration` instances, it should exit with a help message that the tells the user to upgrade to an intermediate version first, correct?

If that is the case, I think this should be done in other PR first, should it?

BlackDex:
> When updating to webauthn-rs 0.5, the migration at the start should instead migrate from WebauthnRegistration with a Credential to a WebauthnRegistration version with a Passkey.
If it encounters an error because the Database still contains U2FRegistration instances, it should exit with a help message that the tells the user to upgrade to an intermediate version first, correct?

Kinda, the only thing I'm not sure about is what to do with the old `U2FRegistration` objects, which in my opinion shouldn't exists anymore since all clients currently do not support older server versions anymore. Unless users do not use any of these clients and still use older clients of course.

So... Do we want to support migration from `U2FRegistration` to the new `Passkey` if they exists if that is possible?
And thus migrate both `U2FRegistration` and `WebauthnRegistration` to `Passkey`, or, skip `U2FRegistration` migration at all now?

zUnixorn:
Another thing I just noticed is that the webauthn-rs 0.5 crate doesn't seem support resident keys [without also forcing attestation](https://docs.rs/webauthn-rs/latest/webauthn_rs/struct.Webauthn.html#method.start_attested_resident_key_registration). In the 0.3 version there was the option to implement [get_require_resident_key()](https://docs.rs/webauthn-rs/0.3.2/src/webauthn_rs/core.rs.html#1038-1040) to ensures that a resident key is created on the security key of the user. This wasn't needed for 2FA but since the Bitwarden frontend decides to use Discoverable Credentials (That's why you don't need to supply a username when logging in with a Passkey), "Login with Passkey" can only be implemented with that feature to stay compatible with the Frontend. 

The maintainers of webauthn-rs seem to have a strong opinion on residential keys (see [here](https://github.com/kanidm/webauthn-rs/issues/372#issuecomment-1769515576) and [here](https://github.com/kanidm/webauthn-rs/issues/378#issuecomment-2247078907)). Since there was a way to do this without attestation in the past, maybe this could be copied and done outside of the webauthn crate. 

Alternatively I could try if the Frontend cares if registrations request includes attestation, if it doesn't, then the passkey support could still be implemented with the 0.5 release and the `resident-key-support` feature but would then be limited to attested Passkeys only. And if Bitwarden decides to port that feature to other Platforms this could break as well, if it would even work in the first place.

zUnixorn:
Ok, after looking through the webauthn crate, I found [webauthn_rs_core](https://crates.io/crates/webauthn-rs-core), this should solve this problem as it exposes the internally used webauthn builders, so that could be used for the resident key case. My previous concern shouldn't matter therefore.

lukasj98:
When will it be available in testing?

ishcarr:
I would love to have this feature. Thank you all for trying.

DenuxPlays:
Hey,

First of all thank you for your great work.

I have a question.
Would this feature in general allow me to also use a physical passkey to log into my vaultwaren account?
Currently I have to type out my long password if I want to log into devices I never logged in before and I want to replace it my simply using a physical key.
Would that be possible with this pr?

Again thank you everyone for your work and great holidays :)

gh10ltz:
@DenuxPlays yes it would but it doesn't look like this will be merged any time soon. The author has not contributed to the PR in a long time so pretty safe to assume this is abandoned.

DenuxPlays:
Oh nice that's great to hear.

I know but a lot of PRs are open for a while and are not dead.
I know it from myself sometimes things come up and you don't have the time to finish it.
I am thankful for everything people do for free.
And if at least someone can always pick up here and finish it just like with the whole SSO stuff.

zUnixorn:
> Would this feature in general allow me to also use a physical passkey to log into my vaultwaren account?
> Currently I have to type out my long password if I want to log into devices I never logged in before and I want to replace it my simply using a physical key.
> Would that be possible with this pr?

Yes, that's my use-case for this as well (Yubikey in my case). But since the official frontends only support it in the web-version, this feature would therefore also only work on the website for now.

zUnixorn:
> @DenuxPlays yes it would but it doesn't look like this will be merged any time soon. The author has not contributed to the PR in a long time so pretty safe to assume this is abandoned.

I didn't forget about this PR. I still miss this feature too, just hadn't had the time to port it to the new webauthn crate version. But I still plan to do that when I find some time again.

BJReplay:
With the recently announced support for logging into the browser with a passkey (https://bitwarden.com/blog/log-into-bitwarden-with-a-passkey/) this would be a great feature for feature parity.

That said, I host Vaultwarden for my family, and the feature that they all love is login with biometrics (i.e. Windows PIN / fingerprint reader) working in browser with desktop installed, logged in once, and set up properly - which I have set up for all of them, and is working perfectly in the 1.35.x release series - so I'm the only user who would use the passkey login with my Yubikeys :)

Watching this with interest, but rust is well beyond my technical capabilities to contribute.

zR-JB:
Just a question:

Is this login option dependent on the prf extension to webauthn?

richardwonka:
Adding to the general love expressed towards everyone working on this, I'd like to suggest increasing the priority here - 
Login with passkeys is a use case that is becoming more and more frequent, normalised, essentially expected behaviour. 

ETA: I don't know if it is possible at all to add resources to this? I do not have the qualification, otherwise I'd be on it already.

Thank you all.

stefan0xC:
> ETA: I don't know if it is possible at all to add resources to this? I do not have the qualification, otherwise I'd be on it already.

Not really. I mean there also seems to be a competing PR right now #6820 that has to be dealt with (i.e. reviewed or asked to contribute to this PR instead, or if this one should be closed in favor of the other one, etc) which also requires additional time and energy from the maintainers and other contributors.

NathanHuisman:
The rotate account endpoint (/accounts/key-management/rotate-user-account-keys) also needs to be updated. I have a local branch where I implemented webAuthnPrfOptions for sync and properly implement prfStatus. LMK if you need any help

blaggacao:
@zUnixorn Could you short-cut maintainer assessment effort and reclaim against or defer to #6820 so that maintainers have a viable opportunity to commit to and prioritize an implementation path?  &mdash; cause users are screeming 😉

zaid-marji:
Hi @zUnixorn — cross-posting from #6820. I've done a thorough review-and-fix pass on the passkey-login work and I'm offering to drive the single consolidated PR the maintainers asked for.

To be upfront about where things stand: #6820 is essentially your #5929 carried forward — ported to webauthn-rs 0.5, a current base, and all three DB backends — so the consolidation would build on #6820, and #5929 would be closed in its favour. Your work is the foundation of all of it, and you'd be credited as `Co-authored-by:`. If you'd rather stay involved or weigh in on the approach, you're very welcome to.

The full proposal is in #6820 — I'd genuinely value your input there. Thanks for getting this feature off the ground.

zaid-marji:
Cross-posting: the consolidated passkey-login PR is now open at #7297. It
carries your original work forward (you're credited via `Co-authored-by:`) and
resolves the open TODOs from this thread — the intermediate-state handling now
lives in a DB-backed single-use challenge table, the code is split into its own
module, and the duplicate login/password paths are gone. The feature is
complete in #7297, so I'd suggest the maintainers close #5929 in favor of it.

JitteryDoodle:
> Cross-posting: the consolidated passkey-login PR is now open at #7297. It carries your original work forward (you're credited via `Co-authored-by:`) and resolves the open TODOs from this thread — the intermediate-state handling now lives in a DB-backed single-use challenge table, the code is split into its own module, and the duplicate login/password paths are gone. The feature is complete in #7297, so I'd suggest the maintainers close #5929 in favor of it.

I see that your PR was closed by yourself.

Mostly just stating this for anyone reading it in the future: #7297 was generated by Claude and is a huge diff. It put the bulk of the work on the actual maintainers of Vaultwarden and @zaid-marji confessed that they weren't very experienced in rust development. This PR (#5929) is still the better path, even though it was stated otherwise by the quoted author.

computingchap:
Hi @zUnixorn — thank you for this PR, it's an excellent foundation and clearly the furthest-along implementation of login-with-passkeys for Vaultwarden.

I'd like to help carry this work forward to a mergeable state, since the branch has drifted a fair way behind main (notably the webauthn-rs 0.5.x upgrade in #5934, the SSO merge, and the unified Diesel schema). I've started porting your implementation onto current main in my fork, keeping your schema/API design: webauthn-rs 0.5 discoverable-credential APIs, DB-persisted single-use challenge state instead of the in-memory maps, migrations for all three database backends, key-rotation handling for the PRF-wrapped keys, and prfStatus reporting.

Before going further: are you happy for this branch to be adopted and continued? I'd keep your commits/credit in the branch history, and would equally welcome doing this collaboratively on your branch if you prefer. Happy to share the WIP diff for comparison first.

computingchap:
Quick follow-up — I posted the above before catching the newer consolidated PR #7297 and #7370 further down the thread, apologies for the noise. Given those exist, the last thing this feature needs is another parallel branch. I'll redirect my effort into building and testing the open PRs against current main and real clients, and posting reproducible findings — hopefully that makes review easier rather than harder. If either author wants an extra pair of hands, happy to help.

PR #5934, opened 2025-06-06
Update webauthn-rs to 0.5.x

zUnixorn:
Hello,

since there was a discussion in #5929 about upgrading webauthn-rs first, I tried to do this in this PR. 

I tested it a few times and didn't have any problems on my machine. It also converts from the old webauthn credential to the new webauthn passkey. Currently this is done by trying to deserialize into the old credential, but this could also be checked via a new field in the DB or some other way. 

Since I didn't remove the U2F migration for now, I could also test if that still works in combination with the new migration. Because U2F was used a long time ago, what version could I use, that uses U2F, to create the credentials, before updating to this branch for testing the migration?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -120,7 +120,10 @@ totp-lite = "2.0.1"
 yubico = { package = "yubico_ng", version = "0.13.0", features = ["online-tokio"], default-features = false }
 
 # WebAuthn libraries
-webauthn-rs = "0.3.2"
+# danger-allow-state-serialisation is needed to save the state in the db
+# danger-credential-internals is needed to support U2F to Webauthn migration
+webauthn-rs = { version = "0.5", features = ["danger-allow-state-serialisation", "danger-credential-internals"] }
+webauthn-rs-proto = "0.5"
 
 # Handling of URL's for WebAuthn and favicons
 url = "2.5.4"
--- a/src/api/core/two_factor/webauthn.rs
+++ b/src/api/core/two_factor/webauthn.rs
@@ -133,12 +133,15 @@ async fn generate_webauthn_challenge(data: Json<PasswordOrOtpData>, headers: Hea
         Some(registrations),
     )?;
 
-    // TODO is there a nicer way to do this?
     // this is done since `start_passkey_registration()` always sets this to `Required` which shouldn't be needed for 2FA
-    challenge.public_key.authenticator_selection = challenge.public_key.authenticator_selection.map(|mut a| {
-        a.user_verification = UserVerificationPolicy::Discouraged_DO_NOT_USE;
-        a
-    });
+    challenge.public_key.extensions = None;
+    if let Some(asc) = challenge.public_key.authenticator_selection.as_mut() {
+        asc.user_verification = UserVerificationPolicy::Discouraged_DO_NOT_USE;
+    }
+
+    let mut state = serde_json::to_value(&state)?;
+    state["rs"]["policy"] = Value::String("discouraged".to_string());
+    state["rs"]["extensions"].as_object_mut().unwrap().clear();
 
     let type_ = TwoFactorType::WebauthnRegisterChallenge;
     TwoFactor::new(user.uuid.clone(), type_, serde_json::to_string(&state)?).save(&mut conn).await?;
@@ -368,10 +371,12 @@ pub async fn generate_webauthn_login(user_id: &UserId, webauthn: Webauthn2FaConf
 
     // Generate a challenge based on the credentials
     let (mut response, state) = webauthn.start_passkey_authentication(&creds)?;
+
+    // Modify to discourage user verification
+    let mut state = serde_json::to_value(&state)?;
+    state["ast"]["policy"] = Value::String("discouraged".to_string());
+    state["ast"]["appid"] = Value::String(format!("{}/app-id.json", &CONFIG.domain()));
     response.public_key.user_verification = UserVerificationPolicy::Discouraged_DO_NOT_USE;
-    
-    // TODO does the appid extension matter? As far as I understand, this was only put into the authentication state anyway
-    // let ext = RequestAuthenticationExtensions::builder().appid(format!("{}/app-id.json", &CONFIG.domain())).build();
 
     // Save the challenge state for later validation
     TwoFactor::new(user_id.clone(), TwoFactorType::WebauthnLoginChallenge, serde_json::to_string(&state)?)
--- a/src/api/core/two_factor/webauthn.rs
+++ b/src/api/core/two_factor/webauthn.rs
@@ -1,9 +1,14 @@
+use std::str::FromStr;
+use std::sync::Arc;
+use once_cell::sync::Lazy;
 use rocket::serde::json::Json;
 use rocket::Route;
 use serde_json::Value;
 use url::Url;
-use webauthn_rs::{base64_data::Base64UrlSafeData, proto::*, AuthenticationState, RegistrationState, Webauthn};
-
+use uuid::Uuid;
+use webauthn_rs::{Webauthn, WebauthnBuilder};
+use webauthn_rs::prelude::{Base64UrlSafeData, Passkey, PasskeyAuthentication, PasskeyRegistration};
+use webauthn_rs_proto::{AuthenticationExtensionsClientOutputs, AuthenticatorAssertionResponseRaw, AuthenticatorAttestationResponseRaw, PublicKeyCredential, RegisterPublicKeyCredential, RegistrationExtensionsClientOutputs, UserVerificationPolicy};
 use crate::{
     api::{
         core::{log_user_event, two_factor::_generate_recover_code},
@@ -18,6 +23,26 @@ use crate::{
     util::NumberOrString,
     CONFIG,
 };
+
+pub static WEBAUTHN_2FA_CONFIG: Lazy<Arc<Webauthn>> = Lazy::new(|| {
+    let domain = CONFIG.domain();
+    let domain_origin = CONFIG.domain_origin();
+    let rp_id = Url::parse(&domain).map(|u| u.domain().map(str::to_owned)).ok().flatten().unwrap_or_default();
+    let rp_origin = Url::parse(&domain_origin).unwrap();
+
+    let webauthn = WebauthnBuilder::new(
+        &rp_id,
+        &rp_origin,
+    ).expect("Creating WebauthnBuilder failed")
+        .rp_name(&domain);
+
+    // TODO check what happened to get_require_uv_consistency()
+
+    // TODO check if there is a better way to handle these errors (would they instantly through or only when used?)
+    Arc::new(webauthn.build().expect("Building Webauthn failed"))
+});
+
+pub type Webauthn2FaConfig<'a> = &'a rocket::State<Arc<Webauthn>>;
 
 pub fn routes() -> Vec<Route> {
     routes![get_webauthn, generate_webauthn_challenge, activate_webauthn, activate_webauthn_put, delete_webauthn,]
@@ -45,52 +70,14 @@ pub struct U2FRegistration {
     pub migrated: Option<bool>,
 }
 
-struct WebauthnConfig {
-    url: String,
-    origin: Url,
-    rpid: String,
-}
-
-impl WebauthnConfig {
-    fn load() -> Webauthn<Self> {
-        let domain = CONFIG.domain();
-        let domain_origin = CONFIG.domain_origin();
-        Webauthn::new(Self {
-            rpid: Url::parse(&domain).map(|u| u.domain().map(str::to_owned)).ok().flatten().unwrap_or_default(),
-            url: domain,
-            origin: Url::parse(&domain_origin).unwrap(),
-        })
-    }
-}
-
-impl webauthn_rs::WebauthnConfig for WebauthnConfig {
-    fn get_relying_party_name(&self) -> &str {
-        &self.url
-    }
-
-    fn get_origin(&self) -> &Url {
-        &self.origin
-    }
-
-    fn get_relying_party_id(&self) -> &str {
-        &self.rpid
-    }
-
-    /// We have WebAuthn configured to discourage user verification
-    /// if we leave this enabled, it will cause verification issues when a keys send UV=1.
-    /// Upstream (the library they use) ignores this when set to discouraged, so we should too.
-    fn get_require_uv_consistency(&self) -> bool {
-        false
-    }
-}
-
 #[derive(Debug, Serialize, Deserialize)]
 pub struct WebauthnRegistration {
     pub id: i32,
     pub name: String,
     pub migrated: bool,
 
-    pub credential: Credential,
+    // TODO should this be renamed or just stay this way
+    pub credential: Passkey,
 }
 
 impl WebauthnRegistration {
@@ -125,7 +112,7 @@ async fn get_webauthn(data: Json<PasswordOrOtpData>, headers: Headers, mut conn:
 }
 
 #[post("/two-factor/get-webauthn-challenge", data = "<data>")]
-async fn generate_webauthn_challenge(data: Json<PasswordOrOtpData>, headers: Headers, mut conn: DbConn) -> JsonResult {
+async fn generate_webauthn_challenge(data: Json<PasswordOrOtpData>, headers: Headers, webauthn: Webauthn2FaConfig<'_>, mut conn: DbConn) -> JsonResult {
     let data: PasswordOrOtpData = data.into_inner();
     let user = headers.user;
 
@@ -135,17 +122,23 @@ async fn generate_webauthn_challenge(data: Json<PasswordOrOtpData>, headers: Hea
         .await?
         .1
         .into_iter()
-        .map(|r| r.credential.cred_id) // We return the credentialIds to the clients to avoid double registering
+        .map(|r| r.credential.cred_id().to_owned()) // We return the credentialIds to the clients to avoid double registering
         .collect();
 
-    let (challenge, state) = WebauthnConfig::load().generate_challenge_register_options(
-        user.uuid.as_bytes().to_vec(),
-        user.email,
-        user.name,
+    // TODO handle errors
+    let (mut challenge, state) = webauthn.start_passkey_registration(
+        Uuid::from_str(&*user.uuid).unwrap(),
+        &user.email,
+        &user.name,
         Some(registrations),
-        None,
-        None,
     )?;
+
+    // TODO is there a nicer way to do this?
+    // this is done since `start_passkey_registration()` always sets this to `Required` which shouldn't be needed for 2FA
+    challenge.public_key.authenticator_selection = challenge.public_key.authenticator_selection.map(|mut a| {
+        a.user_verification = UserVerificationPolicy::Discouraged_DO_NOT_USE;
+        a
+    });
 
     let type_ = TwoFactorType::WebauthnRegisterChallenge;
     TwoFactor::new(user.uuid.clone(), type_, serde_json::to_string(&state)?).save(&mut conn).await?;
@@ -193,8 +186,10 @@ impl From<RegisterPublicKeyCredentialCopy> for RegisterPublicKeyCredential {
             response: AuthenticatorAttestationResponseRaw {
                 attestation_object: r.response.attestation_object,
                 client_data_json: r.response.client_data_json,
+                transports: None,
             },
             type_: r.r#type,
+            extensions: RegistrationExtensionsClientOutputs::default(),
         }
     }
 }
@@ -205,7 +200,7 @@ pub struct PublicKeyCredentialCopy {
     pub id: String,
     pub raw_id: Base64UrlSafeData,
     pub response: AuthenticatorAssertionResponseRawCopy,
-    pub extensions: Option<AuthenticationExtensionsClientOutputs>,
+    pub extensions: AuthenticationExtensionsClientOutputs,
     pub r#type: String,
 }
 
@@ -238,7 +233,7 @@ impl From<PublicKeyCredentialCopy> for PublicKeyCredential {
 }
 
 #[post("/two-factor/webauthn", data = "<data>")]
-async fn activate_webauthn(data: Json<EnableWebauthnData>, headers: Headers, mut conn: DbConn) -> JsonResult {
+async fn activate_webauthn(data: Json<EnableWebauthnData>, headers: Headers, webauthn: Webauthn2FaConfig<'_>, mut conn: DbConn) -> JsonResult {
     let data: EnableWebauthnData = data.into_inner();
     let mut user = headers.user;
 
@@ -253,7 +248,7 @@ async fn activate_webauthn(data: Json<EnableWebauthnData>, headers: Headers, mut
     let type_ = TwoFactorType::WebauthnRegisterChallenge as i32;
     let state = match TwoFactor::find_by_user_and_type(&user.uuid, type_, &mut conn).await {
         Some(tf) => {
-            let state: RegistrationState = serde_json::from_str(&tf.data)?;
+            let state: PasskeyRegistration = serde_json::from_str(&tf.data)?;
             tf.delete(&mut conn).await?;
             state
         }
@@ -261,8 +256,8 @@ async fn activate_webauthn(data: Json<EnableWebauthnData>, headers: Headers, mut
     };
 
     // Verify the credentials with the saved state
-    let (credential, _data) =
-        WebauthnConfig::load().register_credential(&data.device_response.into(), &state, |_| Ok(false))?;
+    let credential = webauthn
+        .finish_passkey_registration(&data.device_response.into(), &state)?;
 
     let mut registrations: Vec<_> = get_webauthn_registrations(&user.uuid, &mut conn).await?.1;
     // TODO: Check for repeated ID's
@@ -291,8 +286,8 @@ async fn activate_webauthn(data: Json<EnableWebauthnData>, headers: Headers, mut
 }
 
 #[put("/two-factor/webauthn", data = "<data>")]
-async fn activate_webauthn_put(data: Json<EnableWebauthnData>, headers: Headers, conn: DbConn) -> JsonResult {
-    activate_webauthn(data, headers, conn).await
+async fn activate_webauthn_put(data: Json<EnableWebauthnData>, headers: Headers, webauthn: Webauthn2FaConfig<'_>, conn: DbConn) -> JsonResult {
+    activate_webauthn(data, headers, webauthn, conn).await
 }
 
 #[derive(Debug, Deserialize)]
@@ -335,7 +330,7 @@ async fn delete_webauthn(data: Json<DeleteU2FData>, headers: Headers, mut conn: 
             Err(_) => err!("Error parsing U2F data"),
         };
 
-        data.retain(|r| r.reg.key_handle != removed_item.credential.cred_id);
+        data.retain(|r| r.reg.key_handle != removed_item.credential.cred_id().as_slice());
         let new_data_str = serde_json::to_string(&data)?;
 
         u2f.data = new_data_str;
@@ -362,9 +357,9 @@ pub async fn get_webauthn_registrations(
     }
 }
 
-pub async fn generate_webauthn_login(user_id: &UserId, conn: &mut DbConn) -> JsonResult {
+pub async fn generate_webauthn_login(user_id: &UserId, webauthn: Webauthn2FaConfig<'_>, conn: &mut DbConn) -> JsonResult {
     // Load saved credentials
-    let creds: Vec<Credential> =
+    let creds: Vec<_> =
         get_webauthn_registrations(user_id, conn).await?.1.into_iter().map(|r| r.credential).collect();
 
     if creds.is_empty() {
@@ -372,8 +367,11 @@ pub async fn generate_webauthn_login(user_id: &UserId, conn: &mut DbConn) -> Jso
     }
 
     // Generate a challenge based on the credentials
-    let ext = RequestAuthenticationExtensions::builder().appid(format!("{}/app-id.json", &CONFIG.domain())).build();
-    let (response, state) = WebauthnConfig::load().generate_challenge_authenticate_options(creds, Some(ext))?;
+    let (mut response, state) = webauthn.start_passkey_authentication(&creds)?;
+    response.public_key.user_verification = UserVerificationPolicy::Discouraged_DO_NOT_USE;
+    
+    // TODO does the appid extension matter? As far as I understand, this was only put into the authentication state anyway
+    // let ext = RequestAuthenticationExtensions::builder().appid(format!("{}/app-id.json", &CONFIG.domain())).build();
 
     // Save the challenge state for later validation
     TwoFactor::new(user_id.clone(), TwoFactorType::WebauthnLoginChallenge, serde_json::to_string(&state)?)
@@ -384,11 +382,11 @@ pub async fn generate_webauthn_login(user_id: &UserId, conn: &mut DbConn) -> Jso
     Ok(Json(serde_json::to_value(response.public_key)?))
 }
 
-pub async fn validate_webauthn_login(user_id: &UserId, response: &str, conn: &mut DbConn) -> EmptyResult {
+pub async fn validate_webauthn_login(user_id: &UserId, response: &str, webauthn: Webauthn2FaConfig<'_>, conn: &mut DbConn) -> EmptyResult {
     let type_ = TwoFactorType::WebauthnLoginChallenge as i32;
     let state = match TwoFactor::find_by_user_and_type(user_id, type_, conn).await {
         Some(tf) => {
-            let state: AuthenticationState = serde_json::from_str(&tf.data)?;
+            let state: PasskeyAuthentication = serde_json::from_str(&tf.data)?;
             tf.delete(conn).await?;
             state
         }
@@ -405,13 +403,11 @@ pub async fn validate_webauthn_login(user_id: &UserId, response: &str, conn: &mu
 
     let mut registrations = get_webauthn_registrations(user_id, conn).await?.1;
 
-    // If the credential we received is migrated from U2F, enable the U2F compatibility
-    //let use_u2f = registrations.iter().any(|r| r.migrated && r.credential.cred_id == rsp.raw_id.0);
-    let (cred_id, auth_data) = WebauthnConfig::load().authenticate_credential(&rsp, &state)?;
+    let authentication_result = webauthn.finish_passkey_authentication(&rsp, &state)?;
 
     for reg in &mut registrations {
-        if &reg.credential.cred_id == cred_id {
-            reg.credential.counter = auth_data.counter;
+        if reg.credential.cred_id() == authentication_result.cred_id() && authentication_result.needs_update() {
+            reg.credential.update_credential(&authentication_result);
 
             TwoFactor::new(user_id.clone(), TwoFactorType::Webauthn, serde_json::to_string(&registrations)?)
                 .save(conn)
--- a/src/db/models/two_factor.rs
+++ b/src/db/models/two_factor.rs
@@ -3,6 +3,7 @@ use webauthn_rs::prelude::{Credential, ParsedAttestation};
 use webauthn_rs_proto::{AttestationFormat, RegisteredExtensions};
 use super::UserId;
 use crate::{api::EmptyResult, db::DbConn, error::MapResult};
+use crate::api::core::two_factor::webauthn::WebauthnRegistration;
 
 db_object! {
     #[derive(Identifiable, Queryable, Insertable, AsChangeset)]
@@ -45,7 +46,27 @@ mod webauthn_0_3 {
     use webauthn_rs::prelude::ParsedAttestation;
     use webauthn_rs_proto::{AttestationFormat, RegisteredExtensions};
 
+    #[derive(Deserialize)]
+    pub struct WebauthnRegistration {
+        pub id: i32,
+        pub name: String,
+        pub migrated: bool,
+        pub credential: Credential,
+    }
+
+    impl From<WebauthnRegistration> for crate::api::core::two_factor::webauthn::WebauthnRegistration {
+        fn from(value: WebauthnRegistration) -> Self {
+            Self {
+                id: value.id,
+                name: value.name,
+                migrated: value.migrated,
+                credential: webauthn_rs::prelude::Credential::from(value.credential).into(),
+            }
+        }
+    }
+
     // Copied from https://docs.rs/webauthn-rs/0.3.2/src/webauthn_rs/proto.rs.html#316-339
+    #[derive(Deserialize)]
     pub struct Credential {
         pub cred_id: Vec<u8>,
         pub cred: COSEKey,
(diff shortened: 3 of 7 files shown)


src/db/models/two_factor.rs
```diff
@@ -40,6 +42,127 @@ pub enum TwoFactorType {
...
+        }
+    }
+
+    // Copied from https://docs.rs/webauthn-rs/0.3.2/src/webauthn_rs/proto.rs.html#260-278
+    #[allow(non_camel_case_types)]
+    #[derive(Deserialize)]
+    pub enum COSEKeyType {
+        EC_OKP,
+        EC_EC2(COSEEC2Key),
+        RSA(COSERSAKey),
```
dfunkt:
```suggestion
        Rsa(COSERSAKey),
```

src/db/models/two_factor.rs
```diff
@@ -40,6 +42,127 @@ pub enum TwoFactorType {
...
+    }
+
+    impl From<COSEKeyType> for webauthn_rs::prelude::COSEKeyType {
+        fn from(value: COSEKeyType) -> Self {
+            match value {
+                COSEKeyType::EC_OKP => panic!(), // TODO what to do here
+                COSEKeyType::EC_EC2(a) => Self::EC_EC2(a.into()),
+                COSEKeyType::RSA(a) => Self::RSA(a.into()),
```
dfunkt:
```suggestion
                COSEKeyType::Rsa(a) => Self::RSA(a.into()),
```

src/api/core/two_factor/webauthn.rs
```diff
@@ -135,18 +122,27 @@ async fn generate_webauthn_challenge(data: Json<PasswordOrOtpData>, headers: Hea
...
+        .map(|r| r.credential.cred_id().to_owned()) // We return the credentialIds to the clients to avoid double registering
         .collect();
 
-    let (challenge, state) = WebauthnConfig::load().generate_challenge_register_options(
-        user.uuid.as_bytes().to_vec(),
-        user.email,
-        user.name,
+    // TODO handle errors
+    let (mut challenge, state) = webauthn.start_passkey_registration(
+        Uuid::from_str(&*user.uuid).unwrap(),
```
dfunkt:
```suggestion
        Uuid::from_str(&user.uuid).unwrap(),
```

dfunkt (changes requested):
Got the following clippy errors when trying to build:
```
error: deref which would be done by auto-deref
   --> src/api/core/two_factor/webauthn.rs:133:24
    |
133 |         Uuid::from_str(&*user.uuid).unwrap(),
    |                        ^^^^^^^^^^^ help: try: `&user.uuid`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#explicit_auto_deref
    = note: `-D clippy::explicit-auto-deref` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::explicit_auto_deref)]`
error: name `RSA` contains a capitalized acronym
   --> src/db/models/two_factor.rs:118:9
    |
118 |         RSA(COSERSAKey),
    |         ^^^ help: consider making the acronym lowercase, except the initial letter: `Rsa`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#upper_case_acronyms
    = note: `-D clippy::upper-case-acronyms` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::upper_case_acronyms)]`
error: could not compile `vaultwarden` (bin "vaultwarden") due to 2 previous errors
```

```
error[E0599]: no variant or associated item named `RSA` found for enum `db::models::two_factor::webauthn_0_3::COSEKeyType` in the current scope
   --> src/db/models/two_factor.rs:126:30
    |
115 |     pub enum COSEKeyType {
    |     -------------------- variant or associated item `RSA` not found for this enum
...
126 |                 COSEKeyType::RSA(a) => Self::RSA(a.into()),
    |                              ^^^ variant or associated item not found in `COSEKeyType`
    |
help: there is a variant with a similar name
    |
126 -                 COSEKeyType::RSA(a) => Self::RSA(a.into()),
126 +                 COSEKeyType::Rsa(a) => Self::RSA(a.into()),
    |

For more information about this error, try `rustc --explain E0599`.
error: could not compile `vaultwarden` (bin "vaultwarden") due to 1 previous error
```

zUnixorn:
Hi thanks for checking, I just fixed the clippy warnings.

src/db/models/two_factor.rs
```diff
@@ -40,6 +42,128 @@ pub enum TwoFactorType {
...
+    pub enum COSEKeyType {
+        EC_OKP,
+        EC_EC2(COSEEC2Key),
+        RSA(COSERSAKey),
+    }
+
+    impl From<COSEKeyType> for webauthn_rs::prelude::COSEKeyType {
+        fn from(value: COSEKeyType) -> Self {
+            match value {
+                COSEKeyType::EC_OKP => panic!(), // TODO what to do here
```
zUnixorn: I just added a panic here for now, since [new](https://docs.rs/webauthn-rs/latest/webauthn_rs/prelude/enum.COSEKeyType.html) Version contains more data than the [old](https://docs.rs/webauthn-rs/0.3.2/webauthn_rs/proto/enum.COSEKeyType.html) type, but I am not sure what the right thing to do here would be, or if this was even used in the past.

zUnixorn: After looking through the webauthn-rs 0.3 code, I think COSEKeyType::EC_OKP was only a placeholder and wasn't constructed anywhere in the webauthn crate at that time. [This](https://docs.rs/webauthn-rs/0.3.2/src/webauthn_rs/crypto.rs.html#414) is where COSEKeyType eventually gets constructed when [register_credential()](https://docs.rs/webauthn-rs/0.3.2/webauthn_rs/core/struct.Webauthn.html#method.register_credential) gets called, which never constructs `COSEKeyType::EC_OKP` but throws an error instead.

zUnixorn:
Ok, I resolved the TODOs I still had in the code so this can now be converted to a regular PR I think.

tessus:
@dani-garcia @BlackDex can you start the workflows?

tessus:
@zUnixorn unfortunately the pipeline has clippy and format issues. I can fix them and push a commit, if you give me access to your branch.

zUnixorn:
Thanks for taking a look and offering your help! Seems like I didn't use the right clippy config locally. I don't know how to give you access to my branch specifically, so I just invited you as a collaborator to my fork for now, that should work too, should it?

tessus:
Yep, should work. I'll see in a few minutes. If not, I'll just open a PR in your repo.

tessus:
Yep, worked. But I had to rebase to solve the conflict. Now we need @dani-garcia or @BlackDex to restart the workflows again.

tessus:
Thanks, Daniel!

dfunkt:
There might be some openssl/pkg-config related build issues with this PR applied, specifically when cross compiling. Haven't figured out the fix yet:
https://github.com/dfunkt/vaultwarden/actions/runs/15995090259/job/45116585187#step:14:4765
```
#62 [linux/amd64->arm64 build 11/13] RUN source /env-cargo &&     cargo build --features sqlite,mysql,postgresql --profile "release" --target="${CARGO_TARGET}" &&     find . -not -path "./target*" -delete
#62 86.86    Compiling openssl v0.10.73
#62 87.09    Compiling zeroize v1.8.1
#62 87.59    Compiling try-lock v0.2.5
#62 87.61    Compiling tower-service v0.3.3
#62 87.74    Compiling darling_macro v0.20.11
#62 87.80    Compiling want v0.3.1
#62 87.81    Compiling rustls-pki-types v1.12.0
#62 88.33 warning: openssl-sys@0.9.109: In file included from /usr/include/openssl/opensslv.h:109,
#62 88.33 warning: openssl-sys@0.9.109:                  from build/expando.c:1:
#62 88.33 warning: openssl-sys@0.9.109: /usr/include/openssl/macros.h:14:10: fatal error: openssl/opensslconf.h: No such file or directory
#62 88.33 warning: openssl-sys@0.9.109:    14 | #include <openssl/opensslconf.h>
#62 88.33 warning: openssl-sys@0.9.109:       |          ^~~~~~~~~~~~~~~~~~~~~~~
#62 88.33 warning: openssl-sys@0.9.109: compilation terminated.
#62 88.33 error: failed to run custom build command for `openssl-sys v0.9.109`
#62 88.33 
#62 88.33 Caused by:
#62 88.33   process didn't exit successfully: `/app/target/release/build/openssl-sys-fbf17aecef99cef3/build-script-main` (exit status: 101)
#62 88.33   --- stdout
#62 88.33   cargo:rustc-check-cfg=cfg(osslconf, values("OPENSSL_NO_OCB", "OPENSSL_NO_SM4", "OPENSSL_NO_SEED", "OPENSSL_NO_CHACHA", "OPENSSL_NO_CAST", "OPENSSL_NO_IDEA", "OPENSSL_NO_CAMELLIA", "OPENSSL_NO_RC4", "OPENSSL_NO_BF", "OPENSSL_NO_PSK", "OPENSSL_NO_DEPRECATED_3_0", "OPENSSL_NO_SCRYPT", "OPENSSL_NO_SM3", "OPENSSL_NO_RMD160", "OPENSSL_NO_EC2M", "OPENSSL_NO_OCSP", "OPENSSL_NO_CMS", "OPENSSL_NO_COMP", "OPENSSL_NO_SOCK", "OPENSSL_NO_STDIO", "OPENSSL_NO_EC", "OPENSSL_NO_SSL3_METHOD", "OPENSSL_NO_KRB5", "OPENSSL_NO_TLSEXT", "OPENSSL_NO_SRP", "OPENSSL_NO_RFC3779", "OPENSSL_NO_SHA", "OPENSSL_NO_NEXTPROTONEG", "OPENSSL_NO_ENGINE", "OPENSSL_NO_BUF_FREELISTS", "OPENSSL_NO_RC2"))
#62 88.33   cargo:rustc-check-cfg=cfg(openssl)
#62 88.33   cargo:rustc-check-cfg=cfg(libressl)
#62 88.33   cargo:rustc-check-cfg=cfg(boringssl)
#62 88.33   cargo:rustc-check-cfg=cfg(awslc)
#62 88.33   cargo:rustc-check-cfg=cfg(libressl250)
#62 88.33   cargo:rustc-check-cfg=cfg(libressl251)
#62 88.33   cargo:rustc-check-cfg=cfg(libressl252)
#62 88.33   cargo:rustc-check-cfg=cfg(libressl261)
#62 88.33   cargo:rustc-check-cfg=cfg(libressl270)
#62 88.33   cargo:rustc-check-cfg=cfg(libressl271)
#62 88.33   cargo:rustc-check-cfg=cfg(libressl273)
#62 88.33   cargo:rustc-check-cfg=cfg(libressl280)
#62 88.33   cargo:rustc-check-cfg=cfg(libressl281)
#62 88.33   cargo:rustc-check-cfg=cfg(libressl291)
#62 88.33   cargo:rustc-check-cfg=cfg(libressl310)
#62 88.33   cargo:rustc-check-cfg=cfg(libressl321)
#62 88.33   cargo:rustc-check-cfg=cfg(libressl332)
#62 88.33   cargo:rustc-check-cfg=cfg(libressl340)
#62 88.33   cargo:rustc-check-cfg=cfg(libressl350)
#62 88.33   cargo:rustc-check-cfg=cfg(libressl360)
#62 88.33   cargo:rustc-check-cfg=cfg(libressl361)
#62 88.33   cargo:rustc-check-cfg=cfg(libressl370)
#62 88.33   cargo:rustc-check-cfg=cfg(libressl380)
#62 88.33   cargo:rustc-check-cfg=cfg(libressl381)
#62 88.33   cargo:rustc-check-cfg=cfg(libressl382)
#62 88.33   cargo:rustc-check-cfg=cfg(libressl390)
#62 88.33   cargo:rustc-check-cfg=cfg(libressl400)
#62 88.33   cargo:rustc-check-cfg=cfg(libressl410)
#62 88.33   cargo:rustc-check-cfg=cfg(ossl101)
#62 88.33   cargo:rustc-check-cfg=cfg(ossl102)
#62 88.33   cargo:rustc-check-cfg=cfg(ossl102f)
#62 88.33   cargo:rustc-check-cfg=cfg(ossl102h)
#62 88.33   cargo:rustc-check-cfg=cfg(ossl110)
#62 88.33   cargo:rustc-check-cfg=cfg(ossl110f)
#62 88.33   cargo:rustc-check-cfg=cfg(ossl110g)
#62 88.33   cargo:rustc-check-cfg=cfg(ossl110h)
#62 88.33   cargo:rustc-check-cfg=cfg(ossl111)
#62 88.33   cargo:rustc-check-cfg=cfg(ossl111b)
#62 88.33   cargo:rustc-check-cfg=cfg(ossl111c)
#62 88.33   cargo:rustc-check-cfg=cfg(ossl111d)
#62 88.33   cargo:rustc-check-cfg=cfg(ossl300)
#62 88.33   cargo:rustc-check-cfg=cfg(ossl310)
#62 88.33   cargo:rustc-check-cfg=cfg(ossl320)
#62 88.33   cargo:rustc-check-cfg=cfg(ossl330)
#62 88.33   cargo:rustc-check-cfg=cfg(ossl340)
#62 88.33   cargo:rerun-if-env-changed=X86_64_UNKNOWN_LINUX_GNU_OPENSSL_LIB_DIR
#62 88.33   X86_64_UNKNOWN_LINUX_GNU_OPENSSL_LIB_DIR unset
#62 88.33   cargo:rerun-if-env-changed=OPENSSL_LIB_DIR
#62 88.33   OPENSSL_LIB_DIR unset
#62 88.33   cargo:rerun-if-env-changed=X86_64_UNKNOWN_LINUX_GNU_OPENSSL_INCLUDE_DIR
#62 88.33   X86_64_UNKNOWN_LINUX_GNU_OPENSSL_INCLUDE_DIR unset
#62 88.33   cargo:rerun-if-env-changed=OPENSSL_INCLUDE_DIR
#62 88.33   OPENSSL_INCLUDE_DIR unset
#62 88.33   cargo:rerun-if-env-changed=X86_64_UNKNOWN_LINUX_GNU_OPENSSL_DIR
#62 88.33   X86_64_UNKNOWN_LINUX_GNU_OPENSSL_DIR unset
#62 88.33   cargo:rerun-if-env-changed=OPENSSL_DIR
#62 88.33   OPENSSL_DIR unset
#62 88.33   cargo:rerun-if-env-changed=OPENSSL_NO_PKG_CONFIG
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_x86_64-unknown-linux-gnu
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_x86_64_unknown_linux_gnu
#62 88.33   cargo:rerun-if-env-changed=HOST_PKG_CONFIG
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG
#62 88.33   cargo:rerun-if-env-changed=OPENSSL_STATIC
#62 88.33   cargo:rerun-if-env-changed=OPENSSL_DYNAMIC
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_ALL_STATIC
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_ALL_DYNAMIC
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_PATH_x86_64-unknown-linux-gnu
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_PATH_x86_64_unknown_linux_gnu
#62 88.33   cargo:rerun-if-env-changed=HOST_PKG_CONFIG_PATH
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_PATH
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR_x86_64-unknown-linux-gnu
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR_x86_64_unknown_linux_gnu
#62 88.33   cargo:rerun-if-env-changed=HOST_PKG_CONFIG_LIBDIR
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR_x86_64-unknown-linux-gnu
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR_x86_64_unknown_linux_gnu
#62 88.33   cargo:rerun-if-env-changed=HOST_PKG_CONFIG_SYSROOT_DIR
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR
#62 88.33   cargo:rerun-if-env-changed=SYSROOT
#62 88.33   cargo:rerun-if-env-changed=OPENSSL_STATIC
#62 88.33   cargo:rerun-if-env-changed=OPENSSL_DYNAMIC
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_ALL_STATIC
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_ALL_DYNAMIC
#62 88.33   cargo:rustc-link-search=native=/usr/lib/aarch64-linux-gnu
#62 88.33   cargo:rustc-link-lib=ssl
#62 88.33   cargo:rustc-link-lib=crypto
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_x86_64-unknown-linux-gnu
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_x86_64_unknown_linux_gnu
#62 88.33   cargo:rerun-if-env-changed=HOST_PKG_CONFIG
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG
#62 88.33   cargo:rerun-if-env-changed=OPENSSL_STATIC
#62 88.33   cargo:rerun-if-env-changed=OPENSSL_DYNAMIC
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_ALL_STATIC
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_ALL_DYNAMIC
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_PATH_x86_64-unknown-linux-gnu
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_PATH_x86_64_unknown_linux_gnu
#62 88.33   cargo:rerun-if-env-changed=HOST_PKG_CONFIG_PATH
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_PATH
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR_x86_64-unknown-linux-gnu
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR_x86_64_unknown_linux_gnu
#62 88.33   cargo:rerun-if-env-changed=HOST_PKG_CONFIG_LIBDIR
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR_x86_64-unknown-linux-gnu
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR_x86_64_unknown_linux_gnu
#62 88.33   cargo:rerun-if-env-changed=HOST_PKG_CONFIG_SYSROOT_DIR
#62 88.33   cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR
#62 88.33   cargo:rerun-if-changed=build/expando.c
#62 88.33   OPT_LEVEL = Some(0)
#62 88.33   OUT_DIR = Some(/app/target/release/build/openssl-sys-dffc0174b3e8c280/out)
#62 88.33   TARGET = Some(x86_64-unknown-linux-gnu)
#62 88.33   HOST = Some(x86_64-unknown-linux-gnu)
#62 88.33   cargo:rerun-if-env-changed=CC_x86_64-unknown-linux-gnu
#62 88.33   CC_x86_64-unknown-linux-gnu = None
#62 88.33   cargo:rerun-if-env-changed=CC_x86_64_unknown_linux_gnu
#62 88.33   CC_x86_64_unknown_linux_gnu = None
#62 88.33   cargo:rerun-if-env-changed=HOST_CC
#62 88.33   HOST_CC = None
#62 88.33   cargo:rerun-if-env-changed=CC
#62 88.33   CC = None
#62 88.33   cargo:rerun-if-env-changed=CC_ENABLE_DEBUG_OUTPUT
#62 88.33   RUSTC_WRAPPER = None
#62 88.33   cargo:rerun-if-env-changed=CRATE_CC_NO_DEFAULTS
#62 88.33   CRATE_CC_NO_DEFAULTS = None
#62 88.33   DEBUG = Some(false)
#62 88.33   CARGO_CFG_TARGET_FEATURE = Some(fxsr,sse,sse2)
#62 88.33   cargo:rerun-if-env-changed=CFLAGS
#62 88.33   CFLAGS = None
#62 88.33   cargo:rerun-if-env-changed=HOST_CFLAGS
#62 88.33   HOST_CFLAGS = None
#62 88.33   cargo:rerun-if-env-changed=CFLAGS_x86_64_unknown_linux_gnu
#62 88.33   CFLAGS_x86_64_unknown_linux_gnu = None
#62 88.33   cargo:rerun-if-env-changed=CFLAGS_x86_64-unknown-linux-gnu
#62 88.33   CFLAGS_x86_64-unknown-linux-gnu = None
#62 88.33   CARGO_ENCODED_RUSTFLAGS = Some()
#62 88.33   cargo:warning=In file included from /usr/include/openssl/opensslv.h:109,
#62 88.33   cargo:warning=                 from build/expando.c:1:
#62 88.33   cargo:warning=/usr/include/openssl/macros.h:14:10: fatal error: openssl/opensslconf.h: No such file or directory
#62 88.33   cargo:warning=   14 | #include <openssl/opensslconf.h>
#62 88.33   cargo:warning=      |          ^~~~~~~~~~~~~~~~~~~~~~~
#62 88.33   cargo:warning=compilation terminated.
#62 88.33 
#62 88.33   --- stderr
#62 88.33 
#62 88.33   thread 'main' panicked at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/openssl-sys-0.9.109/build/main.rs:315:13:
#62 88.33 
#62 88.33   Header expansion error:
#62 88.33   Error { kind: ToolExecError, message: "command did not execute successfully (status code exit status: 1): LC_ALL=\"C\" \"cc\" \"-O0\" \"-ffunction-sections\" \"-fdata-sections\" \"-fPIC\" \"-m64\" \"-I\" \"/usr/include\" \"-Wall\" \"-Wextra\" \"-E\" \"build/expando.c\"" }
#62 88.33 
#62 88.33   Failed to find OpenSSL development headers.
#62 88.33 
#62 88.33   You can try fixing this setting the `OPENSSL_DIR` environment variable
#62 88.33   pointing to your OpenSSL installation or installing OpenSSL headers package
#62 88.33   specific to your distribution:
#62 88.33 
#62 88.33       # On Ubuntu
#62 88.33       sudo apt-get install pkg-config libssl-dev
#62 88.33       # On Arch Linux
#62 88.33       sudo pacman -S pkgconf openssl
#62 88.33       # On Fedora
#62 88.33       sudo dnf install pkgconf perl-FindBin perl-IPC-Cmd openssl-devel
#62 88.33       # On Alpine Linux
#62 88.33       apk add pkgconf openssl-dev
#62 88.33 
#62 88.33   See rust-openssl documentation for more information:
#62 88.33 
#62 88.33       https://docs.rs/openssl
#62 88.33 
#62 88.33   note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
#62 88.33 warning: build failed, waiting for other jobs to finish...
#62 ERROR: process "/bin/bash -o pipefail -c source /env-cargo &&     cargo build --features ${DB} --profile \"${CARGO_PROFILE}\" --target=\"${CARGO_TARGET}\" &&     find . -not -path \"./target*\" -delete" did not complete successfully: exit code: 101
```

BlackDex:
@dfunkt it happens on both Alpine and Debian it looks like.
Something seems to override some CC settings somewhere, it must be coming for the new webauth crate since all other crates are up-to-date already, and do not cause issues.

BlackDex:
It seems to break on there basic github CI too if I try to cross-build that.
I have opened an issue there https://github.com/kanidm/webauthn-rs/issues/498, maybe someone there knows something.

tessus:
I don't understand. The CI pipelines were successful. Or are you telling me they are for show? I have compiled vw on this PR branch locally (Linux and macOS) as well. No issues. I even rebased on main yesterday.

P.S.: If there is an issue with this PR, additional workflows have to be added. Otherwise the pipelines are useless.

P.P.S.: I hope this didn't sound harsh, but why do we have pipelines, if we can't trust them to deploy (merge to master).

BlackDex:
@tessus is building for arm64, armv7 or armv6 that seems to fall.

tessus:
@BlackDex in that case wouldn't it make sense to add this to the pipeline as a target matrix? Or is failing those platforms not a reason not to merge?

P.S.: Thanks for the info.

BlackDex:
Well it would prevent a merge. But adding that will also make the process a whole lot longer to validate. It has been on my mind though. But building for every arch will take time. But maybe using debug builds for those would be enough.

BlackDex:
My PR has been merged and that should fix cross-compiling.
Once v0.5.2 has been released this should work.

Cargo.toml
```diff
@@ -120,7 +120,10 @@ totp-lite = "2.0.1"
 yubico = { package = "yubico_ng", version = "0.13.0", features = ["online-tokio"], default-features = false }
 
 # WebAuthn libraries
-webauthn-rs = "0.3.2"
+# danger-allow-state-serialisation is needed to save the state in the db
+# danger-credential-internals is needed to support U2F to Webauthn migration
+webauthn-rs = { version = "0.5", features = ["danger-allow-state-serialisation", "danger-credential-internals"] }
+webauthn-rs-proto = "0.5"
```
dfunkt:
```suggestion
webauthn-rs = { version = "0.5.2", features = ["danger-allow-state-serialisation", "danger-credential-internals"] }
webauthn-rs-proto = "0.5.2"
```
`webauthn-rs` 0.5.2 has now been released, needed to fix the compilation issues.

src/api/core/two_factor/webauthn.rs
```diff
@@ -18,6 +12,38 @@ use crate::{
     util::NumberOrString,
     CONFIG,
 };
+use once_cell::sync::Lazy;
+use rocket::serde::json::Json;
+use rocket::Route;
+use serde_json::Value;
+use std::str::FromStr;
+use std::sync::Arc;
```
dfunkt:
```suggestion
use std::sync::{Arc, LazyLock};
```

src/api/core/two_factor/webauthn.rs
```diff
@@ -18,6 +12,38 @@ use crate::{
     util::NumberOrString,
     CONFIG,
 };
+use once_cell::sync::Lazy;
```
dfunkt:
```suggestion
```
Why not use LazyLock, seeing that's now available in the standard library?

src/api/core/two_factor/webauthn.rs
```diff
@@ -18,6 +12,38 @@ use crate::{
...
+use uuid::Uuid;
+use webauthn_rs::prelude::{Base64UrlSafeData, Passkey, PasskeyAuthentication, PasskeyRegistration};
+use webauthn_rs::{Webauthn, WebauthnBuilder};
+use webauthn_rs_proto::{
+    AuthenticationExtensionsClientOutputs, AuthenticatorAssertionResponseRaw, AuthenticatorAttestationResponseRaw,
+    PublicKeyCredential, RegisterPublicKeyCredential, RegistrationExtensionsClientOutputs,
+    RequestAuthenticationExtensions, UserVerificationPolicy,
+};
+
+pub static WEBAUTHN_2FA_CONFIG: Lazy<Arc<Webauthn>> = Lazy::new(|| {
```
dfunkt:
```suggestion
pub static WEBAUTHN_2FA_CONFIG: LazyLock<Arc<Webauthn>> = LazyLock::new(|| {
```

src/api/core/two_factor/webauthn.rs
BlackDex: Not everything once_cell offers is available in std. Not sure from the top of my head which differences there were specifically.

dfunkt: Most of them are though, even `OnceCell:wait` has been stabilized in 1.86.0 as `OnceLock::wait`. `LazyLock` is pretty much the standard library equivalent of once_cell's `Lazy`.

For that matter, I've swapped all instances of `Lazy` and `OnceCell` with `LazyLock` and `OnceLock` in my builds and got rid  of the `once_cell` dependency altogether.

dfunkt:
The `Cargo.lock` file needs to be updated, it currently points to the 0.5.1 version of the crates.

tessus:
@dani-garcia I think this one needs one last workflow approval. With 0.5.2 building for arm64, armv7 and armv6 has been solved. The requested changes have also been implemented.

tessus:
This looks good to me. Is there still something missing? Merging this will pave the way for #5929

BlackDex:
> This looks good to me. Is there still something missing? Merging this will pave the way for #5929

Looks ok, but have you tested it? I still need to test this, like going from a very old version, to the current to this version and using this version directly etc... for which i did not had time yet to fully test.
I might have time today to test it.

stefan0xC:
I only tested from current version to this PR with a FIDO U2F key which works but I had no time for further testing.

src/db/models/two_factor.rs
```diff
@@ -40,6 +42,130 @@ pub enum TwoFactorType {
     ProtectedActions = 2000,
 }
 
+mod webauthn_0_3 {
```
dani-garcia: This is only by the migration code in `migrate_credential_to_passkey`, can we move this whole module to the bottom of the file so it's closer to the function that uses it? I think that would make this file simpler to read if we don't have old and new formats intermixed.

src/api/core/two_factor/webauthn.rs
```diff
@@ -135,18 +126,26 @@ async fn generate_webauthn_challenge(data: Json<PasswordOrOtpData>, headers: Hea
...
         .into_iter()
-        .map(|r| r.credential.cred_id) // We return the credentialIds to the clients to avoid double registering
+        .map(|r| r.credential.cred_id().to_owned()) // We return the credentialIds to the clients to avoid double registering
         .collect();
 
-    let (challenge, state) = WebauthnConfig::load().generate_challenge_register_options(
-        user.uuid.as_bytes().to_vec(),
-        user.email,
-        user.name,
+    let (mut challenge, state) = webauthn.start_passkey_registration(
```
dani-garcia: I haven't tested this, but would it be easier to use `webauthn.start_securitykey_registration` here, and add `.danger_set_user_presence_only_security_keys(true)` to `WebauthnBuilder`?

I think with those settings we won't need to manually modify the challenge to remove the UV and modify the serialized value.

dani-garcia (commented):
Some initial review, I have some thoughts on the use of `start_passkey_registration` and `start_passkey_authentication`.  

The `webauthn-rs` crate provides what seem like more appropiate functions for the use of webauthn as 2FA only, with `start_securitykey_registration` and `start_securitykey_authentication`. 

 I haven't tested these, but looking at the code it seems like they handle UV and extensions closer to the way we want, as long as we call `.danger_set_user_presence_only_security_keys(true)` on the `WebAuthnBuilder`. If possible I'd rather use those and avoid modifying challenges/states/responses manually, which feels very error prone.

BlackDex:
I also noticed an issue with reverting back to a version without this PR, it mentions:
`[CAUSE] Error("missing field `type_`", line: 1, column: 672)`

I think we should try to prevent causing issues like this if someone for some reason reverted to a previous version, or, tired testing and then went back to latest.

tessus:
I only tested from the current version to this one. 

I'm not using docker/podman, so testing old rust binaries can be tricky (even compiling older code with new toolchains can be problematic). Many times I've experienced old rust binaries not running on newer systems and vice versa. Usually it complains about the libc. I can setup a VM later tonight and try to test an old binary. Can you please tell me which version that might be?

A while back I suggested the following strategy https://github.com/dani-garcia/vaultwarden/pull/5929#issuecomment-2941354080 but nobody told me what they thought about it.
(Unless it was because of being a really stupid idea, in which case I still would like to know that and why. ;-))
I understand the wish to support a lot of migration paths (I am similar in that regard), but at one point I usually concede that certain paths just make the code overly complicated. There is nothing shameful in requiring people to migrate to an intermediate version, especially when the original vesion is really old.

I am trying to setup a test migration `very old (still need to know the version) -> this PR` this evening.

BlackDex:
Functionally it seems to work ok. I tested Firefox, Chromium, Android (Vivaldi). Migration from previous version to new version seems to be fine. Only reverting to an older version of Vaultwarden will break a users login. Not sure if that is a big issue, but we should probably mention this in the changelog/release-notes once merged.

I do not know if there is an easy way to have the new json work with the old code. But that probably means some weird exceptions and checks which might also not be what we want.

BlackDex:
> I am trying to setup a test migration `very old (still need to know the version) -> this PR` this evening.

Version v1.21.0 was the last version which used U2F before we switched to WebauthN and did the migration code.
But... You would also need a browser which still supports U2F, since that has been removed in favor of WebauthN.

I tested it with Firefox v91.9 ESR, and i was able to create a U2F Key and login with it.
But, starting the current `testing` tagged version did seem to migrate the key, but didn't worked.
Also, WebauthN doesn't work on v91.9 ESR either, so there are probably a lot of security related items which might cause weird conflicts etc... Same as changing a domain for example, that isn't support for WebauthN. At least, not of Firefox, since it doesn't support related domains.

Also, since v1.21.0 is more than 3 years old, i do hope most people are migrated already and used their 2FA key again and should have noticed that it either works, or not, and have resolved it.

> A while back I suggested the following strategy https://github.com/dani-garcia/vaultwarden/pull/5929#issuecomment-2941354080 but nobody told me what they thought about it.

Since i noticed the previous migration probably doesn't work that well anymore anyways, i think it should be fine as-is.

tessus:
> You would also need a browser which still supports U2F,

Ha, I hoped to not have to install a desktop env on the test VM. 

> But, starting the current testing tagged version did seem to migrate the key, but didn't work

In this case the migration from very old to testing has already an issue, which makes any further discussion about migrating from very old (`<= 1.21.0`) pretty much moot. Unless I am missing something.

But as you said, people would have to use very old browsers, which is a lot less likely than using an old server component. For that reason, it's probably safe to say that such a migration is not feasible.

I guess I won't have to setup a test VM after all...

dani-garcia:
I've tried to migrate this to use the `*_securitykey_*` methods as mentioned above, and it seems to work correctly: https://github.com/dani-garcia/vaultwarden/tree/use-securitykey-methods This removes most of the manual modifications to state and responses, the only exception is the appid change to support U2F compat mode, but as we have been talking, that code is untested and likely to not work, so we might just delete it.

About reverting the migration, we're changing the format so that might be hard to do. I guess we could release a version that can revert from 0.5->0.3 and then a new version that moves over to 0.5, but that seems quite complicated and fragile. We could change the webauthn struct to be something like:
```rust
enum WebAuthnRegistration {
V03(webauthn_0_3::Registration),
V05(webauthn_0_5::Registration),
```
Then it would be very clear which version the data is, and we could do the conversion on the fly as needed? But that's a much bigger lift than this PR.

tessus:
> that code is untested and likely to not work, so we might just delete it.

I agree on that!

> About reverting the migration, we're changing the format so that might be hard to do. I guess we could release a version that can revert from 0.5->0.3 and then a new version that moves over to 0.5, but that seems quite complicated and fragile

I agree here as well. What would be a legitimate use case for installing an older version after having used a newer version?
Usually this only happens, if prod is down. And even then it is often required to actually restore a backup.
e.g. schema changes in the db backend can be a nightmare to rollback, especially, if new data was inserted which would be lost after changing back to the old schema. I have dealt with a myriad of mitigation strategies and such code bases are fragile and become a maintenance burden. Many SW projects state that a rollback is only possible by restoring a backup (which also means that new data will be lost).
(Please note that the schema change was just an example.)
Even if it is not convenient from an admin and user point of view, it beats testing hundreds of edge cases. And in a real life rollback it might still fail, because an edge case was missing. 

Don't get me wrong, there are areas where rollbacks w/o data loss are absolutely necessary, but in such cases a rollback strategy is its own project (and making sure that the process works takes longer than developing a new release).
But I don't think that this is necessary in this situation.

BlackDex:
Well. It seems to work ok as-is, and i tested an official YubiKey 5c, SoftU2F and Android it self (Not Windows Hello though).
And i also tested the version from @dani-garcia with all these items. And also works.

So, with all that in mind. I think merging this (once it's rebased again, and the suggestions from @dani-garcia) should be fine. And with the final release we should just mention this clearly in the release notes.

Also, if needed, a sys admin could just restore the previous records of only that table/records to restore the previous format if really needed. And since we do always recommend to create backups, that should be fine i think.

zUnixorn:
Ok, so now the branch was rebased by @BlackDex and the suggestions from @dani-garcia were applied, I also cleaned up the webauthn_0_3 module part. Was there something else I missed? Should the app_id part also be removed or stay like it is? Another question I still have is: Should the current way of checking if there are any old credentials to be migrated be changed or is it fine as it currently is?

dani-garcia:
I think we can leave it for now, and we'll do a separate PR removing all FIDO2 backwards compatibility code at a later time.

dani-garcia (approved):
Had to resolve some conflicts with the SSO PR, but looks ready to merge now, thanks!

Timshel:
Hey,
I believe this change bring issues when the `DOMAIN` is set with an  `ip`.
Setting `DOMAIN=http://127.0.0.1:8000` will results in a server start failure with:

```log
[2025-08-09 12:56:25.242][webauthn_rs][ERROR] rp_id is not an effective_domain of rp_origin
[2025-08-09 12:56:25.242][panic][ERROR] thread 'main' panicked at 'Creating WebauthnBuilder failed: Configuration': src/api/core/two_factor/webauthn.rs:41
```

Switching to `DOMAIN=http://localhost:8000` allow bypassing the issue, but I often set my external ip to be able to test mobile applications.

dani-garcia:
Hmm the `127.0.0.1`/`localhost` sounds annoying, but I wasn't aware that an external IP would work here, I thought webauthn required a secure context, (so https or localhost).

When testing mobile what I do is have a domain point to the internal network IP, I figured mobile would also require HTTPS for those.

Timshel:
With `Firefox` it's possible to use `dom.securecontext.allowlist` to set additional secure host. 
Edit: Had no issue in the android application with an http+ip server host.

With this it was possible to fully test without the need to setup https or a proxy.

It's not that setting a reverse proxy with nginx is complicated but was just easier ^^.

stefan0xC:
Maybe we could fix that by disabling webauthn support if the `domain` is an IP?

Timshel:
I wonder if a domain is truly needed or if only this check is blocking : https://docs.rs/webauthn-rs/latest/src/webauthn_rs/lib.rs.html#293
Might try to test with a custom build.

Blacks-Army:
This PR seems to break Software Passkeys... See issue #6154

BlackDex:
> This PR seems to break Software Passkeys... See issue #6154

Well, it doesn't break all software passkeys, as some still work. Like Android, SoftU2F, and even Bitwarden's own worked for me.

Blacks-Army:
> > This PR seems to break Software Passkeys... See issue #6154
> 
> 
> 
> Well, it doesn't break all software passkeys, as some still work. Like Android, SoftU2F, and even Bitwarden's own worked for me.

Hm Okay weird.. For me it did not work with Bitwarden's

BlackDex:
> > > This PR seems to break Software Passkeys... See issue #6154
> > 
> > 
> > Well, it doesn't break all software passkeys, as some still work. Like Android, SoftU2F, and even Bitwarden's own worked for me.
> 
> Hm Okay weird.. For me it did not work with Bitwarden's

I'll check it again tomorrow.

stefan0xC:
> I wonder if a domain is truly needed or if only this check is blocking : https://docs.rs/webauthn-rs/latest/src/webauthn_rs/lib.rs.html#293 Might try to test with a custom build.

I think it is according to https://www.w3.org/TR/webauthn-2/#relying-party-identifier - so that's why I've decided to hide it if we don't support it, cf. #6160

PR #5993, opened 2025-06-27
Misc Updates and favicon fixes

BlackDex:
- Updated crates
- Switched to rustls instead of native-tls Some dependency were already using rustls by default or without option. By removing native-tls we also have just one way of working here.

Updated favicon fetching which now is able to fetch more icons.
- Use rustls instead of native-tls This seems to work better, probably because of tls sniffing
- Use different user-agent and added several other headers
- Added SVG support. SVG Images will be sanitized first before stored or presented. Also, a special CSP for images will be sent to prevent scripts etc.. from SVG images.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -128,9 +128,9 @@ dependencies = [
 
 [[package]]
 name = "async-compression"
-version = "0.4.24"
+version = "0.4.25"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d615619615a650c571269c00dca41db04b9210037fa76ed8239f70404ab56985"
+checksum = "40f6024f3f856663b45fd0c9b6f2024034a702f453549449e0d84a305900dad4"
 dependencies = [
  "brotli",
  "flate2",
@@ -138,6 +138,8 @@ dependencies = [
  "memchr",
  "pin-project-lite",
  "tokio",
+ "zstd",
+ "zstd-safe",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


BlackDex:
> Nice, do we still benefit from #5917 on top of this?

That depends. Using our built-in reqwest client has the dns-bind-attack prevention in it, and maybe other items we share.
So i think it will.

txase:
With the move to rustls, I might have pushed back on the value of trying to get the AWS SDK to use a custom reqwest client from the vaultwarden codebase if I hadn't already done it :). But the #5917 PR exists, everything still works, and I can't think of any downsides, so I advocate that we should merge it in.

(Also, I'm happy to see the move to rustls overall. I was close to some AWS Rust SDK engineers who provided some valid sounding high-level reasons for why they changed the SDK default from ring to rustls. I wish I had more I could share, but I wasn't actually part of that effort and don't know the details.)

BlackDex:
> 
> (Also, I'm happy to see the move to rustls overall. I was close to some AWS Rust SDK engineers who provided some valid sounding high-level reasons for why they changed the SDK default from ring to rustls. I wish I had more I could share, but I wasn't actually part of that effort and don't know the details.)

Well, we still use `ring` via `rustls` though, not sure if everything already supports `aws-lc` in a sane way.
But it is at least a start going into the good direction. We would have to check which items still use ring, and if there is an alternative feature for those items.

BlackDex:
> With the move to rustls, I might have pushed back on the value of trying to get the AWS SDK to use a custom reqwest client from the vaultwarden codebase if I hadn't already done it :). But the #5917 PR exists, everything still works, and I can't think of any downsides, so I advocate that we should merge it in.

Well, the value too me is the custom dns-bind-attack stuff @dani-garcia built for one.
And maybe so other stuff we globally configure.

PR #5998, opened 2025-06-30
chore: fix some minor issues in the comments

mountdisk:
remove redundant words in comment

### DIFF
--- a/src/api/icons.rs
+++ b/src/api/icons.rs
@@ -338,7 +338,7 @@ struct IconUrlResult {
 
 /// Returns a IconUrlResult which holds a Vector IconList and a string which holds the referer.
 /// There will always two items within the iconlist which holds http(s)://domain.tld/favicon.ico.
-/// This does not mean that that location does exists, but it is the default location browser use.
+/// This does not mean that location exists, but (it) is the default location the browser uses.
 ///
 /// # Argument
 /// * `domain` - A string which holds the domain with extension.


src/static/scripts/datatables.js
dani-garcia, line 1: This file is vendored from the datatables project, so you should remove this change from here and send a PR there instead:
https://github.com/DataTables/DataTablesSrc/blob/0f3fdb5703230a876d5edda79a62be0658de9aee/js/core/core.draw.js#L893

tessus: additionally there are a few other issues: running sentences, wrong grammar, ...

this sentence alone has 2 errors:

```
This does not mean that location does exists, but it is the default location browser use.
```

removing the second  `that` is valid, but could also be replaced by a `this`.

However, the rest still has 2 errors:

It should be either:

```
This does not mean that location does exist, but (it) is the default location the browser uses.
```

or

```
This does not mean that location exists, but (it) is the default location the browser uses.
```

And this was just one sentence. You'd have to spend some time to rewrite those comments....

mountdisk: Thanks for your review !

Modified.

tessus: I think you misunderstood. You have to open the PR in the repository for DataTables.
Any changes to this file in this repo might be overwritten in the future. It makes zero sense to change it in this repo.

mountdisk: Oh, I apologize for my misunderstanding. I have already submitted a PR to the upstream repository. https://github.com/DataTables/DataTablesSrc/pull/351

Thank you for your patient guidance.

tessus: No worries, nothing to apologize for. I do understand your desire to change faulty comments. I also like to have neat comments.
In this case it unfortunately makes no sense to do so in this repo. Stuff like that happened to me too, when I changed auto-generated files... ;-)

BlackDex (approved):

PR #6017, opened 2025-07-03
fix: resolve group permission conflicts with multiple groups

### LINKED ISSUE #5590: Inconsistent permissions shown on clients for members with multiple groups

* Vaultwarden version: v1.33.2
* Web-vault version: v2025.1.1
* OS/Arch: linux/x86_64
* Database type: SQLite
* Database version: 3.48.0
* Uses a reverse proxy: true
* Internet access: true
* DNS Check: true
* Server/NTP Time Check: true
* HTTPS Check: true
* Websocket Check: true
* HTTP Response Checks: true

**Config:**

```json
{
  "_duo_akey": null,
  "_enable_duo": true,
  "_enable_email_2fa": false,
  "_enable_smtp": true,
  "_enable_yubico": true,
  "_icon_service_csp": "",
  "_icon_service_url": "",
  "_ip_header_enabled": true,
  "_max_note_size": 10000,
  "_smtp_img_src": "***:",
  "admin_session_lifetime": 20,
  "admin_token": "***",
  "allowed_connect_src": "",
  "data_folder": "data",
  "database_conn_init": "",
  "database_max_conns": 10,
  "database_timeout": 30,
  "db_connection_retries": 15,
  "disable_admin_token": false,
  "domain": "*****://****************",
  "domain_origin": "*****://****************",
  "domain_path": "",
  "domain_set": true,
  "duo_host": null,
  "duo_ikey": null,
  "duo_skey": null,
  "duo_use_iframe": false,
  "email_attempts_limit": 3,
  "email_change_allowed": true,
  "email_expiration_time": 600,
  "email_token_size": 6,
  "enable_db_wal": true,
  "enable_websocket": true,
  "events_days_retain": null,
  "extended_logging": true,
  "helo_name": null,
  "hibp_api_key": null,
  "icon_blacklist_regex": null,
  "icon_cache_negttl": 259200,
  "icon_cache_ttl": 2592000,
  "icon_download_timeout": 10,
  "icon_redirect_code": 302,
  "icon_service": "internal",
  "invitations_allowed": true,
  "ip_header": "X-Real-IP",
  "log_file": null,
  "log_level": "info",
  "org_attachment_limit": null,
  "org_creation_users": "",
  "org_events_enabled": false,
  "org_groups_enabled": true,
  "push_enabled": false,
  "reload_templates": false,
  "sendmail_command": null,
  "sends_allowed": true,
  "sends_folder": "data/sends",
  "show_password_hint": false,
  "signups_allowed": true,
  "signups_verify": false,
  "smtp_auth_mechanism": null,
  "smtp_debug": false,
  "smtp_embed_images": true,
  "smtp_explicit_tls": null,
  "smtp_from": "",
  "smtp_host": null,
  "smtp_password": null,
  "smtp_port": 587,
  "smtp_security": "starttls",
  "smtp_ssl": null,
  "smtp_timeout": 15,
  "smtp_username": null,
  "tmp_folder": "data/tmp",
  "use_sendmail": false,
  "use_syslog": false,
  "user_send_limit": null,
  "web_vault_enabled": true,
  "yubico_client_id": null,
  "yubico_secret_key": null,
  "yubico_server": null
}
```

</details>

### Vaultwarden Build Version

v1.33.2

### Deployment method

Official Container Image

### Custom deployment method

_No response_

### Reverse Proxy

Traefik 2.11.10

Linux

### Operating System Version

Kubernetes

### Clients

Web Vault

### Client Version

Web-vault: v2025.1.1, Desktop 2025.2.0, Browser 2024.12.4 (Firefox)

### Steps To Reproduce

1. Create UserA, GroupA, GroupB, CollectionA and ItemA assigned to CollectionA.
2. Give GroupA readonly permissions for CollectionA.
3. Give GroupB readwrite permissions for CollectionA.
4. Add UserA to both groups. (A Member with the user role)
5. UserA has readwrite access to ItemA.
6. In the Admin Console, open the edit dialog for GroupA and click save without changing anything.
7. After a client reload, clients for UserA now only show readonly access to Items in CollectionA. (grayed out fields, etc)

### Expected Result

User permissions are consistent regardless which member group was updated/changed most recent.

### Actual Result

The most recent updated member group determines the visible permission on the clients.

### Logs

```text

```

### Screenshots or Videos

_No response_

### Additional Context

Although the permissions are shown as readonly, clients like the bw cli tool have readwrite access to items in CollectionA.

Also, opening ItemA in edit mode on the browser extension (no fields are modifiable) and clicking save (without making changes), ItemA suddenly appears are readwrite able in the UI of all clients (Webvault, Desktop, browser extension).

Is this a bug in Vaultwarden or on all the Bitwarden clients?

DasCanard:
When a user belonged to multiple groups with different permissions for the same collection, only the permissions from one group were applied instead of combining them properly. This caused users to see incorrect access levels when initially viewing collection items.

The fix combines permissions from all user groups by taking the most permissive settings:
- read_only: false if ANY group allows write access
- hide_passwords: false if ANY group allows password viewing
- manage: true if ANY group allows management

This ensures users immediately see the correct permissions when opening collection entries, matching the behavior after editing and saving.

resolves #5590


src/api/core/ciphers.rs
```diff
@@ -1924,11 +1924,24 @@ impl CipherSyncData {
...
+                        existing.hide_passwords = existing.hide_passwords && cg.hide_passwords; // false if ANY group allows password view
+                        existing.manage = existing.manage || cg.manage; // true if ANY group allows manage
+                    }
+                    None => {
+                        // First group for this collection
+                        combined_permissions.insert(cg.collections_uuid.clone(), cg);
+                    }
+                }
+            }
+            combined_permissions
```
BlackDex: We might be able to use `fold` here and generate the combined permissions directly instead of using a `match`.

```suggestion
            CollectionGroup::find_by_user(user_id, conn)
                .await
                .into_iter()
                .fold(HashMap::new(), |mut combined_permissions, cg| {
                    combined_permissions.entry(cg.collections_uuid.clone())
                        .and_modify(|existing| {
                            // Combine permissions: take the most permissive settings.
                            existing.read_only &= cg.read_only; // false if ANY group allows write
                            existing.hide_passwords &= cg.hide_passwords; // false if ANY group allows password view
                            existing.manage |= cg.manage; // true if ANY group allows manage
                        })
                        .or_insert(cg);
                    combined_permissions
                })
```

I quickly tested this, and that seems to do the trick for me as well.
But the logic stays the same, so that was a nice one :).

DasCanard: Great idea, thanks!

BlackDex (approved):

BlackDex:
Nice! Thanks!

stefan0xC:
I'm not sure if those permissions are correct or if this is still incomplete because I think we also do some permission crunching here:
https://github.com/dani-garcia/vaultwarden/blob/3b48e6e90386bcee3e8a2b99ec2472cba979adf1/src/db/models/collection.rs#L88-L126

E.g. when I am a member of a group with manage permissions but I have also direct view access like this:
image

that means I will only have View permission. (This might be intentional but I'm not sure it is.)

---

Also as a User I'll get logged out even if I have the `manage` permission

https://github.com/user-attachments/assets/1d69f93d-4c07-4684-ab2b-94626887308e

because of the `ManagerHeaderLoose` check on at least these routes:
https://github.com/dani-garcia/vaultwarden/blob/3b48e6e90386bcee3e8a2b99ec2472cba979adf1/src/api/core/organizations.rs#L350-L353
https://github.com/dani-garcia/vaultwarden/blob/3b48e6e90386bcee3e8a2b99ec2472cba979adf1/src/api/core/organizations.rs#L1499-L1502
https://github.com/dani-garcia/vaultwarden/blob/3b48e6e90386bcee3e8a2b99ec2472cba979adf1/src/api/core/organizations.rs#L2662-L2663

(This might be unrelated to this PR but I noticed it when testing.)

BlackDex:
Hmm, might have to do some further testing. I didn't encountered this. But might have skipped this specific test.

DasCanard:
> Also as a User I'll get logged out even if I have the manage permission

Yes also noticed this one a while ago.

BlackDex:
> Also as a User I'll get logged out even if I have the `manage` permission

@stefan0xC, that is because we do not support `can manage` for anyone else but a manager or higher. Users are not allowed to. So, we probably should also check which access level they have, and based upon that down-scale to edit instead of manage for those users.

I noticed that Bitwarden returns the uuid of the collection twice, and doesn't merge it, but the permissions are only returned just once.

stefan0xC:
> @stefan0xC, that is because we do not support `can manage` for anyone else but a manager or higher. Users are not allowed to. So, we probably should also check which access level they have, and based upon that down-scale to edit instead of manage for those users.
> 

But the clients allows setting the manage permission so shouldn't we support that then? Until we do, could we return `cu.manage && m.atype != MemberShiptype::User` instead of https://github.com/dani-garcia/vaultwarden/blob/3b48e6e90386bcee3e8a2b99ec2472cba979adf1/src/db/models/collection.rs#L100 since this is an explicit option now? (I mean we could also keep the `is_manager && !cu.read_only && !cu.hide_passwords` part but I kinda think that this is kinda wrong...)

BlackDex:
> > @stefan0xC, that is because we do not support `can manage` for anyone else but a manager or higher. Users are not allowed to. So, we probably should also check which access level they have, and based upon that down-scale to edit instead of manage for those users.
> 
> But the clients allows setting the manage permission so shouldn't we support that then? Until we do, could we return `cu.manage && m.atype != MemberShiptype::User` instead of
> 
> https://github.com/dani-garcia/vaultwarden/blob/3b48e6e90386bcee3e8a2b99ec2472cba979adf1/src/db/models/collection.rs#L100
> 
> since this is an explicit option now? (I mean we could also keep the `is_manager && !cu.read_only && !cu.hide_passwords` part but I kinda think that this is kinda wrong...)

Not supporting the `can manage` is of course already a deviation of the Bitwarden way. And since we would need better rbac for that to allow users also, i think a quick fix to at least allow edit in those situations would help for now. Unless someone fixes all of it of course. Including all the custom rights etc...

But since there are currently so many places where we check this, converting manage to edit for normal users is a good way i think.

I didn't had time(and spirit) yet to dive deep into this topic. But i would love to fix it all hehe.

BlackDex:
Did a quick look. I think the folding part is correct.
I'm only indeed not sure if the direct assigned permission on a collection should take the lead above group permissions.
Because that is what currently is done at the code @stefan0xC linked to. Which is where the group assigned permissions are ignored if there are direct assigned permissions.

So, having the correct group assigned permissions crunched already is good, we just need to do a bit more later on i think.

I have not yet looked into the other rights issue with the manage/edit part yet and how to solve that.

BlackDex:
Looks like direct assigned permissions do overrule group permissions also on Bitwarden, so that is probably ok than.

stefan0xC:
So this can be merged?

BlackDex:
> So this can be merged?

I'm checking the `can manage` part right now which you also mentioned.

BlackDex:
@DasCanard, regarding the `can manage` issues and logging out when a user has `can manage` rights, there needs to be some added changes to this PR.

```diff
--- a/src/db/models/collection.rs
+++ b/src/db/models/collection.rs
@@ -97,13 +97,13 @@ impl Collection {
                         (
                             cu.read_only,
                             cu.hide_passwords,
-                            cu.manage || (is_manager && !cu.read_only && !cu.hide_passwords),
+                            is_manager && (cu.manage || (!cu.read_only && !cu.hide_passwords)),
                         )
                     } else if let Some(cg) = cipher_sync_data.user_collections_groups.get(&self.uuid) {
                         (
                             cg.read_only,
                             cg.hide_passwords,
-                            cg.manage || (is_manager && !cg.read_only && !cg.hide_passwords),
+                            is_manager && (cg.manage || (!cg.read_only && !cg.hide_passwords)),
                         )
                     } else {
                         (false, false, false)
```

I have not yet checked the part which isn't using the `cipher_sync_data`, because that currently doesn't use the edited code. But, I'm also not sure if that is used at all anywhere though.

BlackDex:
Ok, that part returns invalid data since it doesn't take the membership role into account.
I think it can be solved by setting it like this.

```diff
--- a/src/db/models/collection.rs
+++ b/src/db/models/collection.rs
@@ -114,7 +116,7 @@ impl Collection {
         } else {
             match Membership::find_confirmed_by_user_and_org(user_uuid, &self.org_uuid, conn).await {
                 Some(m) if m.has_full_access() => (false, false, m.atype >= MembershipType::Manager),
-                Some(_) if self.is_manageable_by_user(user_uuid, conn).await => (false, false, true),
+                Some(m) if m.atype == MembershipType::Manager && self.is_manageable_by_user(user_uuid, conn).await => (false, false, true),
                 Some(m) => {
                     let is_manager = m.atype == MembershipType::Manager;
                     let read_only = !self.is_writable_by_user(user_uuid, conn).await;
```

With those changes i think it should work as expected.

DasCanard:
@BlackDex @stefan0xC Thanks for the detailed analysis!

Implemented the suggested changes to restrict `manage` permissions to managers only. This prevents users from getting logged out when they have manage permissions while maintaining proper role hierarchy.

The fix ensures only `MembershipType::Manager` and higher can receive `manage: true` in collection responses.

BlackDex:
@DasCanard Could you also rebase on the current main branch? That would help testing and merging :)

BlackDex:
@stefan0xC would be good if you can check to, i might have missed something (again ;) )

BlackDex (approved):

stefan0xC:
> @stefan0xC would be good if you can check to, i might have missed something (again ;) )

The collection management issue seems to work fine. As a User I can't edit a collection anymore.

The only problem I now see is somewhere else: if we say that a directly assigned collection permission should take precedence then this should also be taken into account when we calculate the permissions for the ciphers. But that is calculated a bit differently as described here:
https://github.com/dani-garcia/vaultwarden/blob/ce70cd2cf4f1faf84ce32266bcc65ca9482a5514/src/db/models/cipher.rs#L635-L643

So the question would be if this comment is correct or if we are doing something wrong and we should calculate it the same way for both collections and ciphers?

To reproduce the inconsistency give a User direct access with `View items, hidden passwords` permission and add them to a group that has `Manage collection` permission to that collection. The items (but not the collection itself) will be editable by the user and also the password will be viewable. (Or as a Manager the collection will not be editable because of the precedence rule but the items are.)

BlackDex:
Ok... @DasCanard  I did some more testing, and i think if you apply this test it finally should be solved!
We currently merge the personal and group permissions which we shouldn't.
And a boolean AND was done instead of an OR on the manage bool.

```diff
diff --git a/src/db/models/cipher.rs b/src/db/models/cipher.rs
index 5e8971c872..f8d60505f4 100644
--- a/src/db/models/cipher.rs
+++ b/src/db/models/cipher.rs
@@ -609,22 +609,23 @@ impl Cipher {
             let mut rows: Vec<(bool, bool, bool)> = Vec::new();
             if let Some(collections) = cipher_sync_data.cipher_collections.get(&self.uuid) {
                 for collection in collections {
-                    //User permissions
+                    // User permissions
                     if let Some(cu) = cipher_sync_data.user_collections.get(collection) {
                         rows.push((cu.read_only, cu.hide_passwords, cu.manage));
-                    }
-
-                    //Group permissions
-                    if let Some(cg) = cipher_sync_data.user_collections_groups.get(collection) {
+                    // Group permissions
+                    } else if let Some(cg) = cipher_sync_data.user_collections_groups.get(collection) {
                         rows.push((cg.read_only, cg.hide_passwords, cg.manage));
                     }
                 }
             }
             rows
         } else {
-            let mut access_flags = self.get_user_collections_access_flags(user_uuid, conn).await;
-            access_flags.append(&mut self.get_group_collections_access_flags(user_uuid, conn).await);
-            access_flags
+            let user_permissions = self.get_user_collections_access_flags(user_uuid, conn).await;
+            if !user_permissions.is_empty() {
+                user_permissions
+            } else {
+                self.get_group_collections_access_flags(user_uuid, conn).await
+            }
         };
 
         if rows.is_empty() {
@@ -633,6 +634,9 @@ impl Cipher {
         }
 
         // A cipher can be in multiple collections with inconsistent access flags.
+        // Also, user permission overrule group permissions
+        // and only user permissions are returned by the code above.
+        //
         // For example, a cipher could be in one collection where the user has
         // read-only access, but also in another collection where the user has
         // read/write access. For a flag to be in effect for a cipher, upstream
@@ -641,13 +645,15 @@ impl Cipher {
         // and `hide_passwords` columns. This could ideally be done as part of the
         // query, but Diesel doesn't support a min() or bool_and() function on
         // booleans and this behavior isn't portable anyway.
+        //
+        // The only exception is for the `manage` flag, that needs a boolean OR!
         let mut read_only = true;
         let mut hide_passwords = true;
         let mut manage = false;
         for (ro, hp, mn) in rows.iter() {
             read_only &= ro;
             hide_passwords &= hp;
-            manage &= mn;
+            manage |= mn;
         }
 
         Some((read_only, hide_passwords, manage))
```

BlackDex:
@stefan0xC, would be great if you can test again. I have looked for any other place where these rights are used, and i think all should be covered.

BlackDex (approved):

DasCanard:
Thank you everyone for the great teamwork on this!

BlackDex:
> Looks good to me. Just to confirm: If an item is in multiple collection one of which the user has only read access to directly and the other they have also write access to via a group then the write access wins (which I think is the correct behavior).

Yes, i think so too.

BlackDex:
Ok.. lets wait with merging until RedHat has solved there Quay issues: https://status.redhat.com/incidents/rvtmg00ycbjh

PR #6028, opened 2025-07-04
docs(readme): add reference to data backup documentation

rudra-singh1:
this pr adds a new data backup documentation in the "resources" folder, detailing backup and restore processes for vaultwarden, including automated scripts for backups and cleanup. a reference to this documentation is added under the "Usage" section in the main README.md. addresses partially the database migration point mentioned in: https://github.com/dani-garcia/vaultwarden/issues/246 . i believe in the future this documentation can be updated to support those different databases, but for now, a basic data backup walk thru is important

why this pr Is needed: vaultwarden users need reliable backup solutions to prevent data loss, enable disaster recovery, and support migrations. many community networks have asked my team how we add backup capabilities to vaultwarden. i thought it'd be easier to put this information out here instead, for ease of access to everyone.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,8 @@ A nearly complete implementation of the Bitwarden Client API is provided, includ
 > [!TIP]
 >**For more detailed examples on how to install, use and configure Vaultwarden you can check our [Wiki](https://github.com/dani-garcia/vaultwarden/wiki).**
 
+For information on how to back up your Vaultwarden data, refer to the [Data Backup Documentation](./resources/data-backup.md).
+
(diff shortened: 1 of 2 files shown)


tessus:
I like the fact that you wrote documentation, but I see an issue with this. Your doc is limited to docker containers. What about native installs and databases other than sqlite?

Don't worry, I am not asking you to write it. (I know how to backup my vaultwarden and create database backups.) I am just saying this info is better added to the project's wiki, if it isn't there already.

I am writing this because I believe somebody else created a similar PR in the past, but was rejected.

P.S.: But let's see what the devs say...

BlackDex:
We have the wiki for that which already contains several articles regarding backups. Maybe they could be updated, but I'd rather have it listed there instead of some random file in the source tree.

And we already have a reference to the wiki  in the readme, so that should be enough I think. Would be a bit tedious to keep updating both the wiki and the readme everytime a change is done or article added.

Please leave documentation in the wiki and adjust those where needed.

rudra-singh1:
Thanks. I just updated the wiki with that information

PR #6056, opened 2025-07-12
Update crates, workflow and issue template

BlackDex:
- Updated all the crates, which probably fixes #5959
- Updated all the workflows and tested it with zizmor Also added zizmor as a workflow it self.
- Updated the issue template to better mention to search first.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -47,7 +47,7 @@ jobs:
     # Start a local docker registry to extract the compiled binaries to upload as artifacts and attest them
     services:
       registry:
-        image: registry:2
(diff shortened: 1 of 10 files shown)


.github/workflows/release.yml
```diff
@@ -47,7 +47,7 @@ jobs:
     # Start a local docker registry to extract the compiled binaries to upload as artifacts and attest them
     services:
       registry:
-        image: registry:2
+        image: registry:sha256@1fc7de654f2ac1247f0b67e8a459e273b0993be7d2beda1f3f56fbf1001ed3e7 # v3.0.0 - https://hub.docker.com/_/registry/tags
```
stefan0xC, line 50: @dani-garcia there was a small typo here:
```suggestion
        image: registry@sha256:1fc7de654f2ac1247f0b67e8a459e273b0993be7d2beda1f3f56fbf1001ed3e7 # v3.0.0 - https://hub.docker.com/_/registry/tags
```

which should be fixed by #6058

PR #6057, opened 2025-07-12
Update release.yml

dani-garcia:
Seems like docker can't use the hash references, and it fails to build: https://github.com/dani-garcia/vaultwarden/actions/runs/16242780267/job/45861396226

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -47,7 +47,7 @@ jobs:
     # Start a local docker registry to extract the compiled binaries to upload as artifacts and attest them
     services:
       registry:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -47,7 +47,7 @@ jobs:
     # Start a local docker registry to extract the compiled binaries to upload as artifacts and attest them
     services:
       registry:
-        image: registry:sha256@1fc7de654f2ac1247f0b67e8a459e273b0993be7d2beda1f3f56fbf1001ed3e7 # v3.0.0 - https://hub.docker.com/_/registry/tags
```
stefan0xC, line 50: that was caused by a typo because [sha256:f970a8a224075e6f0b99365d215812352a5eb84c719c786a774079db62ea4194](https://hub.docker.com/layers/library/registry/3.0.0/images/sha256-f970a8a224075e6f0b99365d215812352a5eb84c719c786a774079db62ea4194) exists.

PR #6127, opened 2025-08-03
OpenDAL S3 parameter support

### LINKED ISSUE #6112: S3 default storage class breaks Minio compatibility
This line:

https://github.com/dani-garcia/vaultwarden/blob/a0198d8d7cadc666643a4e32a2a8d4217038ea3a/src/config.rs#L1243

Causes problems if using [Minio](https://github.com/minio) as the S3 endpoint:
```
[REQUEST s3.PutObject] [2025-07-28T10:34:52.211] [Client IP: redacted]
PUT /rsa_key.pem
Proto: HTTP/1.1
Host: [redacted]
X-Amz-Content-Sha256: UNSIGNED-PAYLOAD
X-Forwarded-For: [redacted]
X-Forwarded-Proto: https
Accept: */*
Authorization: AWS4-HMAC-SHA256 Credential=vaultwarden/20250728/us-east-1/s3/aws4_request, SignedHeaders=content-length;host;x-amz-content-sha256;x-amz-date;x-amz-storage-class, Signature=4d257b5c8807a7cddf1c2c49749af1f96dce936a0cdf11f96ed363fce5323f81
X-Amz-Date: 20250728T173452Z
X-Amz-Storage-Class: INTELLIGENT_TIERING
X-Real-Ip: [redacted]
Accept-Encoding: gzip, br, zstd, deflate
Content-Length: 1675
<BLOB>
[RESPONSE] [2025-07-28T10:34:52.211] [ Duration 58µs TTFB 50.045µs ↑ 153 B  ↓ 338 B ]
400 Bad Request
Content-Length: 338
Content-Type: application/xml
Strict-Transport-Security: max-age=31536000; includeSubDomains
Vary: Origin,Accept-Encoding
X-Amz-Id-2: 37b62f02de360c6949659e1d2fb72852efb5d74e5dcbc7d4e665920a1c72a29e
X-Ratelimit-Limit: 20413
X-Xss-Protection: 1; mode=block
Server: MinIO
X-Amz-Request-Id: 18567AC52E4FC471
X-Content-Type-Options: nosniff
X-Ratelimit-Remaining: 20413
Accept-Ranges: bytes
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>InvalidStorageClass</Code><Message>Invalid storage class.</Message><Key>rsa_key.pem</Key><BucketName>vaultwarden</BucketName><Resource>/rsa_key.pem</Resource><RequestId>18567AC52E4FC471</RequestId><HostId>37b62f02de360c6949659e1d2fb72852efb5d74e5dcbc7d4e665920a1c72a29e</HostId></Error>
```
Can that storage class configuration be made user-configurable or even optional, please?

txase:
This change allows s3:// data paths to include additional parameters as querystrings. The default parameters remain optimized for the official AWS S3 service, but any OpenDAL S3 parameters can be set. For example, to override the default `true` value for `enable_virtual_host_style` you can set the data config to:

s3://my-bucket?enable_virtual_host_style=false

Boolean parameters can be set to `true` by simply adding the parameter:

s3://my-bucket?enable_versioning

Fixes #6112

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,17 +9,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
-name = "aes"
-version = "0.8.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
-dependencies = [
- "cfg-if",
- "cipher",
- "cpufeatures 0.2.17",
-]
-
-[[package]]
 name = "ahash"
 version = "0.8.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -671,17 +660,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "backon"
-version = "1.6.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cffb0e931875b666fc4fcb20fee52e9bbd1ef836fd9e9e04ec21555f9f85f7ef"
-dependencies = [
- "fastrand",
- "gloo-timers",
- "tokio",
-]
-
-[[package]]
 name = "base16ct"
 version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -776,15 +754,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cdd35008169921d80bc60d3d0ab416eecb028c4cd653352907921d95084790be"
 dependencies = [
  "hybrid-array",
-]
-
-[[package]]
-name = "block-padding"
-version = "0.3.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a8894febbff9f758034a5b8e12d87918f56dfc64a8e1fe757d65e29041538d93"
-dependencies = [
- "generic-array",
 ]
 
 [[package]]
@@ -893,15 +862,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ade8366b8bd5ba243f0a58f036cc0ca8a2f069cff1a2351ef1cac6b083e16fc0"
 
 [[package]]
-name = "cbc"
-version = "0.1.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "26b52a9543ae338f279b96b0b9fed9c8093744685043739079ce85cd58f289a6"
-dependencies = [
- "cipher",
-]
-
-[[package]]
 name = "cc"
 version = "1.2.61"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -918,12 +878,6 @@ name = "cfg-if"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
-
-[[package]]
-name = "cfg_aliases"
-version = "0.2.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
 
 [[package]]
 name = "chacha20"
@@ -958,16 +912,6 @@ checksum = "a6139a8597ed92cf816dfb33f5dd6cf0bb93a6adc938f11039f371bc5bcd26c3"
 dependencies = [
  "chrono",
  "phf 0.12.1",
-]
-
-[[package]]
-name = "cipher"
-version = "0.4.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "773f3b9af64447d2ce9850330c473515014aa235e6a783b02db81ff39e4a3dad"
-dependencies = [
- "crypto-common 0.1.6",
- "inout",
 ]
 
 [[package]]
@@ -2358,15 +2302,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "home"
-version = "0.5.12"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cc627f471c528ff0c4a49e1d5e60450c8f6461dd6d10ba9dcd3a61d3dff7728d"
-dependencies = [
- "windows-sys 0.61.2",
-]
-
-[[package]]
 name = "hostname"
 version = "0.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2516,11 +2451,9 @@ dependencies = [
  "hyper 1.9.0",
  "hyper-util",
  "rustls 0.23.40",
- "rustls-native-certs",
  "tokio",
  "tokio-rustls 0.26.4",
  "tower-service",
- "webpki-roots",
 ]
 
 [[package]]
@@ -2717,16 +2650,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c8fae54786f62fb2918dcfae3d568594e50eb9b5c25bf04371af6fe7516452fb"
 
 [[package]]
-name = "inout"
-version = "0.1.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "879f10e63c20629ecabbb64a8010319738c66a5cd0c29b02d63d272b03751d01"
-dependencies = [
- "block-padding",
- "generic-array",
-]
-
-[[package]]
 name = "ipconfig"
 version = "0.3.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2798,10 +2721,12 @@ checksum = "f00b5dbd620d61dfdcb6007c9c1f6054ebd75319f163d886a9055cec1155073d"
 dependencies = [
  "jiff-static",
  "jiff-tzdb-platform",
+ "js-sys",
  "log",
  "portable-atomic",
  "portable-atomic-util",
  "serde_core",
+ "wasm-bindgen",
  "windows-sys 0.61.2",
 ]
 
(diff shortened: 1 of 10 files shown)


ProbstDJakob:
When this has been merged, it would be great if the official container images would be build with S3 support enabled.

tessus:
@BlackDex since you are on a roll today merging PRs, this seems like a low hanging fruit... ;-)

kaiyou:
Thank you so much for this much expected feature.

I have tested this for a couple days and deployed it to production today. It works really well and also enables compatibility with Garage, yet another S3 cluster implementation.

I have feedback about the admin experience though: `ICON_CACHE_FOLDER`, `SENDS_FOLDER` and `ATTACHMENTS_FOLDER`, when not explicitly provided, are automatically derived from the more general `DATA_FOLDER` in `config.rs`. One might either set all of these explicitly, e.g. to store data in separate buckets, or simply set `DATA_FOLDER` to an S3 URL and let OpenDAL deal with derived subpaths.

However, when `DATA_FOLDER`  contains a query string, it cannot be concaternated with `/attachments`, `/sends` or `/icons_cache` anymore. For instance : 

```
s3://mybucket?endpoint=my-service.tld&region=myregion
```

... leads to the following attachment folder URL :

```
s3://mybucket?endpoint=my-service.tld&region=myregion/attachments
```

The subpath is now part of the region in the query string, which just fails (due to `/` being forbidden in region names and sigv4 scope construction). I suspect other situations might lead to more subtle failure modes, where requests are handled properly but data is directed to the bucket root instead of the intended subpath.

It would be really nice if the subpath was properly inserted before the query string. This either requires some more logic in `config.rs`, parsing and manipulating the OpenDAL URL, or a separate configuration variable for the query string. Otherwise, a note in `.env.template` would help avoid some failure modes.

BlackDex:
Thanks @kaiyou for testing! This really helps us.

dani-garcia:
Looking through the changelog on the latest OpenDAL version, it looks like `Operator::from_uri` is now available:
https://github.com/apache/opendal/issues/5445 
https://docs.rs/opendal/latest/opendal/struct.Operator.html#method.from_uri

I'm wondering if that would also cover this usecase, potentially in a generic way that would make it easier to enable more backends in the future?

JoooostB:
Kinda waiting on this one to be merged, any estimates on when this can be expected? Thanks!

txase:
I think there have simply been some timing issues on getting this ready for review and reviewers having the bandwidth to review it. Maybe we can solve for that...

@BlackDex @dani-garcia: I can rebase this quickly, using OpenDAL's `Operator::from_uri()` method and taking a look at fixing the issue @kaiyou found. When will you have bandwidth to review this PR? I can aim to have it rebased and ready for you on your timeline if that helps.

BlackDex:
@txase i think a new release is coming soon, and a few PR's will probably be merged before this release. So, probably it's best after that release i think?

ProbstDJakob:
For people waiting on this PR, merging it before the release would be welcome, but if you want a "working" release (no huge changes) before this PR has been merged, that would be understandable. In that case, will a release also be created not long after this PR has been merged (and will this feature be enabled in the container images)?

txase:
> @txase i think a new release is coming soon, and a few PR's will probably be merged before this release. So, probably it's best after that release i think?

@BlackDex I saw some releases go by this past week. I'll be busy next week, but maybe can get to this the week of April 27th. Would that work well for you to slot in reviews to get this rebased, tested, and merged?

txase:
I spent time rebasing and figuring out how to address @kaiyou's issue and incorporate the from_uri() config method that OpenDAL now provides. The result is this new commit. It is a bit larger due to the following:

- OpenDAL is updated to the latest version
- The latest version of OpenDAL has a different manner of setting up custom AWS request signing (we use the official AWS SDK for signing config instead of OpenDAL's implementation because it tends to lag behind AWS SDK functionality like newer SSO auth options)
  - That said, this new version is nicer in that it doesn't require adding the `anyhow` crate anymore and uses rustls for TLS
- It's a non-trivial change to solve @kaiyou's issue where params and path prefixes are mixed
  - To solve for this, I created a new `storage` module with an `s3`-specific submodule to abstract away the details

This is now ready for review again!

BlackDex:
@txase, thanks for looking at this again.

I have one main concern and that is the pull-in of Reqwest v0.13 besides the current version we use v0.12.
This also brings-in aws-lc, which we want to prevent, which is why we wanted to use a custom http client in the first place if i remember correctly.

I think the only crate holding us back currently is `openidconnect`, which by default uses the outh2 reqwest feature.

We could also see if we can disable reqwest completely for openidconnect, and implement a custom http client, since it looks like that is supported.

Besides that, we probably do want reqwest to still use rustls for now and not switch to aws-lc with this PR.

txase:
Ahh right, I'm remembering you've got specific encryption dependencies you want to maintain. I can look to rework this some more. This is what I got from your comment, please let me know if I have this right:

- Find a way to continue using the existing Reqwest dependency, which is currently v0.12
- Do not pull in aws-lc, even if as a dependency of Rustls

It sounds like you were mentioning `openidconnect` just to say that it's a reason vaultwarden still is on v0.12 of reqwest, but if I find a way to make this work on Reqwest v0.12 then that is not an issue for this PR?

BlackDex:
Well, i think we are probably better of implementing a custom http-client wrapper for openidconnect.
That way we can use our custom implementation including Hickory specifics etc...

It will also ensure it will break a future new release of reqwest.
I'm not sure if Opendal has something similar though? Using any http-client, just as long as it requests and responds in the correct way?

I'm currently looking into to that, to remove the reqwest feature from openidconnect, since that seems to be a possibility.

The main thing for this PR then would be, update reqwest also with this PR (And yubico-ng, since the latest version supports v0.13 too). Ensure reqwest keeps using rustls, and not bring in aws-lc (Might change in the future if we want to switch).

So, for your PR do not mind openidconnect, and that it brings in a duplicate reqwest version i think, unless you really have time to spare haha.

txase:
I'm trying to parse your comments to understand how I should attempt to update this PR. I'm sorry that I'm only partially following along, and I know you are rightly focused on making sure the dependencies for this project stay clean.

Can you help me out by stating what specifically you'd like to see in this branch? Is it just to get rid of the aws-lc dependency, and that it's ok to use Reqwest 0.13? Would you like me to attempt removing the reqwest feature from openidconnect? I'm fairly open to whatever you're looking for, I'm just not sure what that is right now.

Thanks!

dani-garcia:
I think what he's trying to say is that we're okay with reqwest 0.13 even if it introduces duplication, as we're going to have to update to it eventually. But we're trying to avoid introducing another crypto library like aws-lc, so we can probably do something like this (not tested):

```diff
diff --git a/Cargo.toml b/Cargo.toml
index 019eaf3..52fe17c 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ vendored_openssl = ["openssl/vendored"]
 # Enable MiMalloc memory allocator to replace the default malloc
 # This can improve performance for Alpine builds
 enable_mimalloc = ["dep:mimalloc"]
-s3 = ["opendal/services-s3", "opendal/reqwest-rustls-tls", "dep:aws-config", "dep:aws-credential-types", "dep:aws-smithy-runtime-api", "dep:http", "dep:reqsign-aws-v4", "dep:reqsign-core"]
+s3 = ["opendal/services-s3", "dep:opendal-reqwest", "dep:rustls", "dep:aws-config", "dep:aws-credential-types", "dep:aws-smithy-runtime-api", "dep:http", "dep:reqsign-aws-v4", "dep:reqsign-core"]
 
 # OIDC specific features
 oidc-accept-rfc3339-timestamps = ["openidconnect/accept-rfc3339-timestamps"]
@@ -198,6 +198,16 @@ grass_compiler = { version = "0.13.4", default-features = false }
 # File are accessed through Apache OpenDAL
 opendal = { version = "0.56.0", features = ["services-fs"], default-features = false }
 
+# Reference reqwest 0.13 so we can enable `rustls-no-provider`
+opendal-reqwest = { package = "reqwest", version = "0.13", features = ["rustls-no-provider"], default-features = false, optional = true }
+# Needed to install the provider for reqwest 0.13
+rustls = { version = "0.23", features = ["ring"], default-features = false, optional = true }
+
 # For retrieving AWS credentials, including temporary SSO credentials
 aws-config = { version = "1.8.16", features = ["behavior-version-latest", "rt-tokio", "credentials-process", "sso"], default-features = false, optional = true }
 aws-credential-types = { version = "1.2.14", optional = true }
diff --git a/src/main.rs b/src/main.rs
index 96aa55b..31ab2ee 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -71,6 +71,11 @@ pub use util::is_running_in_container;
 
 #[rocket::main]
 async fn main() -> Result<(), Error> {
+    #[cfg(feature = "s3")]
+    let _ = rustls::crypto::ring::default_provider().install_default();
+
     parse_args();
     launch_info();
```

I believe that should be enough to remove the aws-lc dep.

BlackDex:
Kinda, but we can drop reqwest from openidconnect by using a custom `AsyncHttpClient`, that will use our own http_client implementation.

With that dropped, we can just switch fully to v0.13, and make sure it uses rustls. That way, no need for `opendal-reqwest` as an extra package.

Sorry for being a bit unclear earlier hehe.

txase:
Ok, I understand now. I'm going to rework this PR to have two commits:

- Upgrade vaultwarden's reqwest dependency to 0.13, including through a custom integration for `openidconnect` using existing vaultwarden reqwest http client code
- Refactored s3-params commit without pulling in aws-lc

Thanks for the clarifications!

txase:
I was able to address all the concerns (I hope!). There are now two commits.

The first one upgrades all usage of reqwest to 0.13 while still using ring instead of aws-lc for all crypto. This had a handful of knock-on implementation code changes (e.g. upgrading opendal, which then required a code change to move off a deprecated function). It also changes the dependency lock file a good amount, but on net it removes much more than it adds.

The second commit remains the original s3-params work.

BlackDex (approved):
Nice, seems to work like a charm, and a lot of dependencies are gone/removed!

dani-garcia (approved):
Looks good to me too, thanks!

PR #6141, opened 2025-08-06
Perform same checks when setting kdf

Timshel:
In the SSO PR it was [recommended](https://github.com/dani-garcia/vaultwarden/pull/3899#discussion_r2254247199) to check the kdf settings in `post_set_password`.

Since it seemed logical to apply it to the register endpoint decided to create this PR too.

From what I can see the `Option` for `kdf` and `kdf_iterations` was not necessary:
 - Old [register.request.ts](https://github.com/bitwarden/clients/blob/web-v2025.1.1/libs/common/src/models/request/register.request.ts)
 - [register-finish.request](https://github.com/bitwarden/clients/blob/main/libs/common/src/auth/models/request/registration/register-finish.request.ts)

### DIFF
--- a/src/api/core/accounts.rs
+++ b/src/api/core/accounts.rs
@@ -68,13 +68,20 @@ pub fn routes() -> Vec<rocket::Route> {
 
 #[derive(Debug, Deserialize)]
 #[serde(rename_all = "camelCase")]
+pub struct KDFData {
+    kdf: i32,
+    kdf_iterations: i32,
+    kdf_memory: Option<i32>,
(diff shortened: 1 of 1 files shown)


dani-garcia (approved):
LGTM, thanks!

Yeah as far as I know both kdf and iterations should always be set, for both pbkdf and argon so that seems correct to me.

BlackDex:
In the past it was option because it either wasn't send by default. and/or, we wanted to be backwards compatible with older clients.

Nice to have some tech debt gone.

PR #6151, opened 2025-08-08
Fix several more multi select push issues

BlackDex:
There were some more items which would still overload the push endpoint. This PR fixes the remaining items (I hope).

I also encountered a missing endpoint for restoring multiple ciphers from the trash via the admin console.

Overall, we could improve a lot of these items in a different way. Like bundle all SQL Queries etc... But that takes more time, and this fixes overloading the Bitwarden push servers, and speeds up these specific actions.

### DIFF
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -78,6 +78,7 @@ pub fn routes() -> Vec<Route> {
         restore_cipher_put,
         restore_cipher_put_admin,
         restore_cipher_selected,
+        restore_cipher_selected_admin,
         delete_all,
         move_cipher_selected,
         move_cipher_selected_put,
@@ -318,7 +319,7 @@ async fn post_ciphers_create(
     // or otherwise), we can just ignore this field entirely.
     data.cipher.last_known_revision_date = None;
 
-    share_cipher_by_uuid(&cipher.uuid, data, &headers, &mut conn, &nt).await
+    share_cipher_by_uuid(&cipher.uuid, data, &headers, &mut conn, &nt, None).await
 }
 
 /// Called when creating a new user-owned cipher.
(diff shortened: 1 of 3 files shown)


tessus:
I have a question about the comment on improving the queries. I understand that abstraction layers and adding SQL to the code is convenient.
However, I worked in database performance and development for almost 2 decades (until around 2013) and I was wondering why projects don't use (more) stored procedures.
Is it because it is too much work and fragile in terms of compat between the db backends?

It would certainly make the code a lot less convoluted. Although I also understand that vw's db schema is rather simple and not much might be gained by using SPs.

This is not an attempt to change how vw handles data, I am really just curious.

BlackDex:
The problem with this project is mainly the huge differences between all supported database engines.

Also, views are not fully supported in Diesel if I'm correct, and that also doesn't really help for some items.

And, our current database is, in my opinion to much normalized without those features.

tessus:
Thanks for the reply. Ok, makes sense, although I talked about stored procedures not views. But I suspect diesel might not support SPs either.

When using SPs, an abstraction layer becomes irrelevant. You only call the SPs with input parameters and get predefined columns returned.

e.g. currently you have functions that call a few sql queries and then return the respective result. all these sql queries (from a function) can be packed into a stored procedure. additionally sql queries have client/server communication. the sql in a SP runs exclusively on the db server and only the final result set is returned.

But yes, your reply answers my question. Thanks again.

src/api/core/ciphers.rs
```diff
@@ -1555,35 +1575,47 @@ async fn move_cipher_selected(
...
-            err!("Cipher doesn't exist")
-        };
-
-        if !cipher.is_accessible_to_user(&user_id, &mut conn).await {
-            err!("Cipher is not accessible by user")
+    let cipher_count = data.ids.len();
+    let mut single_cipher: Option<Cipher> = None;
+
+    // TODO: Convert this to use a single query (or at least less) to update all items
+    // Find all ciphers a user has access too, all others will be ignored
```
dani-garcia:
```suggestion
    // Find all ciphers a user has access to, all others will be ignored
```

dani-garcia (approved):

PR #6158, opened 2025-08-09
Allow SSO role mapping to add admin cookie

Timshel:
Add the ability to parse the `IDToken` and User Information endpoint response from an OIDC Provider to retrieve a `role` to grant access to Vaultwarden admin console. Support two roles: `admin` or `user`.

This feature is controlled by the following configuration:

- `SSO_ROLES_ENABLED: control if the mapping is done, default is `false`.
- `SSO_ROLES_DEFAULT_TO_USER`: do not block login in case of missing role, default is `true`.
- `SSO_ROLES_TOKEN_PATH=/resource_access/${SSO_CLIENT_ID}/roles`: path to read roles in the ID Token or User Information response.

The role claim parsing is done using generic since the goal is to reuse it to parse Organization role claims (The idea is to have `admin`/`user` for Vaultwarden admin page access. And `OrgOwner`, `OrgAdmin`, `OrgManager` and `OrgUser` for user membership).

~~Replaced the `FromRequest for Secure` since it seemed simpler to just check if the `DOMAIN` start with `https`.~~

### DIFF
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -38,7 +38,10 @@ use crate::{
 };
 
 pub fn routes() -> Vec<Route> {
-    if !CONFIG.disable_admin_token() && !CONFIG.is_admin_token_set() {
+    if !(CONFIG.disable_admin_token()
+        || CONFIG.is_admin_token_set()
+        || (CONFIG.sso_enabled() && CONFIG.sso_roles_enabled()))
+    {
         return routes![admin_disabled];
     }
 
@@ -166,6 +169,7 @@ fn render_admin_login(msg: Option<&str>, redirect: Option<&str>) -> ApiResult<Ht
     let json = json!({
         "page_content": "admin/login",
         "error": msg,
+        "sso_only": CONFIG.sso_enabled() && CONFIG.sso_roles_enabled(),
         "redirect": redirect,
         "urlpath": CONFIG.domain_path()
     });
@@ -179,6 +183,24 @@ fn render_admin_login(msg: Option<&str>, redirect: Option<&str>) -> ApiResult<Ht
 struct LoginForm {
     token: String,
     redirect: Option<String>,
+}
+
+pub fn add_admin_cookie(cookies: &CookieJar<'_>, is_secure: bool) {
+    let claims = generate_admin_claims();
+    let jwt = encode_jwt(&claims);
+
+    let cookie = Cookie::build((COOKIE_NAME, jwt))
+        .path(admin_path())
+        .max_age(time::Duration::minutes(CONFIG.admin_session_lifetime()))
+        .same_site(SameSite::Strict)
+        .http_only(true)
+        .secure(is_secure);
+
+    cookies.add(cookie);
+}
+
+pub fn remove_admin_cookie(cookies: &CookieJar<'_>) {
+    cookies.remove(Cookie::build(COOKIE_NAME).path(admin_path()));
 }
 
 #[post("/", format = "application/x-www-form-urlencoded", data = "<data>")]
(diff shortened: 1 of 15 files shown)


BlackDex:
I'm not sure if we would want this to be honest.
This would allow everyone with the admin role to access the backend admin, which might be valid or not.
While it's still of course the server admin who needs to configure it, it does add extra complexity to the admin login.

But, i do see the ease of it though. But It's more a question of, do we really want this.

Also, replacing the `FromRequest for Secure` with just a simple `https` check isn't valid.
While i would recommend everyone to access the `/admin` interface on the same domain and also use `https`, that simply isn't the case in all situations.

One could configure `DOMAIN=https://vaultwarden.domain.tld`, but access the admin interface via `http://vwadmin.internal.local` or `http://192.168.0.2`. Which, in the PR code would cause issues and the cookie to be invalid.

tessus:
> But, i do see the ease of it though. But It's more a question of, do we really want this.

Well, that's for the admin to decide, which is why there is a config option. Just my 2 cents. ;-)

Timshel:
> that simply isn't the case in all situations.

Ok will revert this and pass the parameter around should not be too much of an issue.

> with the admin role to access the backend admin, which might be valid or not.

A more custom role could be used like `VaultAdmin` / `VaultUser` (or anything else) to prevent involuntary grant.

As for wanting this :

- As you mentioned its ease of use
- In the context of SSO this allows to revoke access without having to roll admin password.
- It allows restricting access to a subset of users (it was a feature request).
- In itself it's not much code, much of the PR is some boilerplate and code that will be reused to parse the IDToken and User information (to allow for Organization role and groups mapping).

dani-garcia:
My worry here would be that we've had security issues due to improper access control in the past because of how complex the rules are around organizations/users/groups, etc. Thankfully most data is E2E encrypted so the worst that those can lead to is data loss, but the admin page has a lot more controls that an admin can touch. So in that sense, the unintentional benefit of the admin page having such a simple access control means that it's much less likely that we'll get it wrong.

Also is the strategy here that the admin first has to log in to the web vault (which stores the admin cookie), and then they can manually go to the admin page? Because I feel like that's going to be a pretty obscure and hard to discover behavior.

Timshel:
> having such a simple access control

Well it's simple once correctly configured, first an admin need to activate the feature then grant the correct role (which will be independent/different from the Organization roles).
As opposed to just sharing a password which can be more easily misused.

> Because I feel like that's going to be a pretty obscure and hard to discover behavior.

In my mind only one or two people will be granted the Admin access, so discovery was not really a priority.
Edit: and if you go to the admin page it will tell you to log.

rizlas:
Hi, we discussed this FR with @Timshel 2years ago and you can find it here https://github.com/Timshel/vaultwarden/issues/10.

As an org admin I would like to manage authn/authz from my sso provider instead of doing it per application.

My2c

ArcticLampyrid:
Frankly speaking, for a password management tool, it should first follow some security practices, and full SSO capability is very important.
We can keep the codebase simple, but real-world security problems are never simple—they require complex strategies.
In my view, minimizing the use of passwords as much as possible is a safer implementation, and this can be achieved through SSO. (The master password is different; that is a encryption key, not a password.)

src/api/admin.rs
```diff
@@ -31,8 +31,12 @@ use crate::{
     CONFIG, VERSION,
 };
 
+#[allow(clippy::nonminimal_bool)]
```
BlackDex: This isn't needed if you change the if statement.

src/api/admin.rs
```diff
@@ -31,8 +31,12 @@ use crate::{
...
 pub fn routes() -> Vec<Route> {
-    if !CONFIG.disable_admin_token() && !CONFIG.is_admin_token_set() {
+    if !CONFIG.disable_admin_token()
+        && !CONFIG.is_admin_token_set()
+        && !(CONFIG.sso_enabled() && CONFIG.sso_roles_enabled())
```
BlackDex: This could be simplified to this:

```suggestion
if !(disable_admin_token || is_admin_token_set || sso_enabled && sso_roles_enabled)
```

BlackDex (changes requested):
I have not tested the specific SSO feature for the admin interface. but as far as i can tell it looks ok, other then a change to remove the clippy allow.

Timshel:
Updated the docker `Keycloak` image to `trixie`, make the whole setup simpler :).

e64462:
This seems to have stalled so I'll just +1 that I currently use this feature on Timshel's fork without issue. I'm still hoping it can make it into vaultwarden at some point.

HomoSapiens:
+1
Still on Timshel's fork code, because of that

tim-schneider:
+1

I love using Vaultwarden personally, but when it comes to deploying it in larger environments, this PR would be extremely important. As a system administrator, it is always reassuring to have full RBAC in place everywhere.

Honestly, I'm not sure how using SSO without this PR makes much sense. Yes, it introduces some external dependencies, which may slightly reduce the security of the vault itself, but that is the trade-off for enabling it and eliminating manual permission management.

It only affects setups where this feature is enabled, so it does not change anything for any other use case.

PR #6164, opened 2025-08-10
Switch to GHA's concurrency control

dfunkt:
- removes the need to use a 3rd party action


dfunkt:
Need to do some further testing.

BlackDex:
@dfunkt, it now started to build containers in my fork. It didn't do that before.

dfunkt:
Whoops, somehow it didn't even occur to me that others may not have container building enabled. This can be solved by adding back the `if: ${{ github.repository == 'dani-garcia/vaultwarden' }}` to the `docker-build` job.
Should I make a PR with the change?

EDIT: Made one anyway -> https://github.com/dani-garcia/vaultwarden/pull/6227

PR #6166, opened 2025-08-10
Make database connection pool dynamic

Samoth69:
Hello,

I've notice that Vaultwarden always keep his database pool fully connected to the database. Since my instance of Vaultwarden has a light load this resulted in 10 database connections being almost always in idle.

This PR fix this by using `min_idle` and `idle_timeout` options of the database pool manager. This allow the app to keep only 2 sessions at minimum while allowing the pool to grow when there is more load.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -636,8 +636,14 @@ make_config! {
         /// Timeout when acquiring database connection
         database_timeout:       u64,    false,  def,    30;
 
-        /// Database connection pool size
+        /// Timeout in seconds before idle connections to the database are closed
+        database_idle_timeout:  u64,    false, def,     60;
+
+        /// Database connection max pool size
         database_max_conns:     u32,    false,  def,    10;
+
+        /// Database connection min pool size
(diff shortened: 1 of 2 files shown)


src/config.rs
```diff
@@ -639,9 +639,15 @@ make_config! {
         /// Timeout when acquiring database connection
         database_timeout:       u64,    false,  def,    30;
 
-        /// Database connection pool size
+        /// Timeout in seconds before idle connections to the database are closed
+        database_idle_timeout:  u64,    false, def,     60;
```
dani-garcia:
```suggestion
        database_idle_timeout:  u64,    false, def,     600;
```

The default value of this in `r2d2` is 10 minutes (600 seconds), so changing it to be now 60 seems unnecessary here. We can let it be configurable but follow the default provided to us.

https://docs.rs/r2d2/latest/r2d2/struct.Builder.html#method.idle_timeout

Samoth69: No problem, the original value was arbitrary.

PR #6187, opened 2025-08-15
Enable block-browser-injections-by-domain feature

spicychickensauce:
I tried to follow https://bitwarden.com/help/blocking-uris/, but the option is not visible in my browser extension.
It's an important feature, as the extension messes up some sites.

I tracked it down to the `block-browser-injections-by-domain` feature flag.

It seems to be enabled in the official client, so it makes sense to also enable here.

Btw. it would have been much easier if the `experimental_client_feature_flags` would not be checked against a hardcoded list of `KNOWN_FLAGS`.

Would it be possible to remove that check? Would have saved me a couple of hours...

### DIFF
--- a/src/api/core/mod.rs
+++ b/src/api/core/mod.rs
@@ -212,6 +212,7 @@ fn config() -> Json<Value> {
     feature_states.insert("unauth-ui-refresh".to_string(), true);
     feature_states.insert("enable-pm-flight-recorder".to_string(), true);
     feature_states.insert("mobile-error-reporting".to_string(), true);
+    feature_states.insert("block-browser-injections-by-domain".to_string(), true);
 
     Json(json!({
         // Note: The clients use this version to handle backwards compatibility concerns


dani-garcia:
Hey thanks for the PR! This feature flag was removed from the clients last week, so we can probably wait until next client's release rather than merge this now for us to delete it in a month: https://github.com/bitwarden/clients/pull/16008

About the hardcoded `KNOWN_FLAGS` list, I agree it's unfortunate specially for flags like this that are entirely client-side, but other flags modify the way that the clients communicate with the server, and so they are dangerous to enable if the server doesn't support the corresponding APIs, but the clients think it does.

Sadly we don't have a way to differentiate between client-only flags and client-server flags, so the current strategy was to just have a limited set that is allowed, and for the rest to just wait until the clients support the features without flags.

I'm not really sure what's the best way to go here, I don't really want people to go toggle random flags without considering what they involve, but the current system is too restrictive. Maybe we can just add a new option called `DANGEROUS_ALLOW_UNKNOWN_FEATURE_FLAGS` that documents that enabling unknown flags can be problematic and point users to do a backup first?

spicychickensauce:
Alright, I guess in that case I'll just wait for that to be released :+1:

> other flags modify the way that the clients communicate with the server

Ok, I see how those could cause issues,  so it makes sense to limit it in some ways.

> Maybe we can just add a new option called DANGEROUS_ALLOW_UNKNOWN_FEATURE_FLAGS

I think that would be a great compromise :+1: 
It would allow adventurous users to experiment, and it clearly communicates that if something goes wrong, it's on the user.

Issue #6163, opened 2025-08-10
panic when SSO_MASTER_PASSWORD_POLICY is disabled


harshavmb:
### Prerequisites

- [x] I have searched the existing **Closed _AND_ Open** [Issues](https://github.com/dani-garcia/vaultwarden/issues?q=is%3Aissue%20) **_AND_** [Discussions](https://github.com/dani-garcia/vaultwarden/discussions?discussions_q=)
- [x] I have searched and read the [documentation](https://github.com/dani-garcia/vaultwarden/wiki/)

### Your environment (Generated via diagnostics page)

* Vaultwarden version: v1.34.3-2a5489a4
* Web-vault version: v2025.7.0
* OS/Arch: linux/aarch64
* Running within a container: true (Base: Debian)
* Database type: MySQL
* Database version: 10.11.13-MariaDB-0ubuntu0.24.04.1-log
* Uses config.json: true
* Uses a reverse proxy: true
* IP Header check: false (X-Forwarded-For)
* Internet access: true
* Internet access via a proxy: false
* DNS Check: true
* Browser/Server Time Check: true
* Server/NTP Time Check: true
* Domain Configuration Check: true
* HTTPS Check: true
* Websocket Check: true
* HTTP Response Checks: true

### Config & Details (Generated via diagnostics page)

<details><summary>Show Config & Details</summary>

**Environment settings which are overridden:** DOMAIN, ADMIN_TOKEN, SSO_ENABLED, SSO_ONLY, SSO_CLIENT_ID, SSO_CLIENT_SECRET, SSO_AUTHORITY, SSO_SCOPES, SSO_AUDIENCE_TRUSTED

**Config:**

```json
{
  "_duo_akey": null,
  "_enable_duo": true,
  "_enable_email_2fa": false,
  "_enable_smtp": true,
  "_enable_yubico": true,
  "_icon_service_csp": "",
  "_icon_service_url": "",
  "_ip_header_enabled": true,
  "_max_note_size": 10000,
  "_smtp_img_src": "***:",
  "admin_ratelimit_max_burst": 3,
  "admin_ratelimit_seconds": 300,
  "admin_session_lifetime": 20,
  "admin_token": "***",
  "allowed_connect_src": "",
  "allowed_iframe_ancestors": "",
  "attachments_folder": "data/attachments",
  "auth_request_purge_schedule": "30 * * * * *",
  "authenticator_disable_time_drift": false,
  "data_folder": "data",
[...]

PR #6193, opened 2025-08-17
#6163: Fix panic when SSO_MASTER_PASSWORD_POLICY is non-json

harshavmb:
Dear Maintainers,

Thanks for this project. I replaced gmail drafts (as my secret store lol), Google authenticator, Apple Notes & several others with this vaultwarden & authenticator services.

Regarding this PR, it's a proposal to [#6163](https://github.com/dani-garcia/vaultwarden/issues/6163). I am novice when it comes go `rust` but I think I got the problem right. I hope this will address this problem.

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -111,7 +111,14 @@ async fn master_password_policy(user: &User, conn: &DbConn) -> Value {
             }
         }))
     } else if let Some(policy_str) = CONFIG.sso_master_password_policy().filter(|_| CONFIG.sso_enabled()) {
-        serde_json::from_str(&policy_str).unwrap_or(json!({}))
+        // Use a match statement to handle Ok and Err results gracefully.
+        match serde_json::from_str(&policy_str) {
+            Ok(val) => val,
+            Err(_) => {
+                warn!("Failed to parse sso_master_password_policy configuration string: {}", policy_str);
+                json!({})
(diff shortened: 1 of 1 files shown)


BlackDex:
Not to judge, but this looks like AI generated code. But also, it doesn't change anything, except for adding a warn message.

It doesn't resolve the panic, as that doesn't happen here, since at this code part no panic can occur.

harshavmb:
Hi @BlackDex ,

Thanks for your review. It's indeed AI generated code. I should have done a bit more work on this. I spent a bit more time on this now, could [this](https://github.com/dani-garcia/vaultwarden/blob/main/src/api/mod.rs#L102) be the cause when `master_password_policies` is empty, it returns `None` & that causes `json!(None)` to panic during conversion?

> From reduce() fn in iterator.rs file
> /// Reduces the elements to a single one, by repeatedly applying a reducing
>     /// operation.
>     ///
>     /// If the iterator is empty, returns [`None`]; otherwise, returns the
>     /// result of the reduction.
>     ///

Also, to add I am not a fulltime SDE, so my coding skills are rudimentary but I would like to do improve. Please excuse me if this analysis is also incorrect.

tessus:
fixed in https://github.com/dani-garcia/vaultwarden/pull/6233

This can be closed.

PR #6205, opened 2025-08-20
Improve sso auth flow

Timshel:
This PR make multiple modifications to the authentication flow:

- stop wrapping the returned values from the authorization (`code` or `error`) in a JWT token to pass through Bitwarden redirection. Instead, it stores them in database and use the `state` to find them again. 
The length of the JWT Token might have been causing issue with the desktop client and Chrome on Windows.
- stop using a memory cache to store the User information in case a 2FA flow is triggered, use the database that we were reading anyway.
- Validate the Bitwarden clients PKCE code challenge to ensure that the client exchanging the `code` is the one which initiated the request. Cannot pass this challenge to the providers since `PkceCodeChallenge` cannot be instantiated with arbitrary values; but it does mean it's checked even if the provider does not support PKCE.

-------

Documentation which could be added to the wiki:

### On `SSO_PKCE`

Only control sending a PKCE challenge generated by the server to the provider. This is more a compatibility feature than a security one since it's separate from the challenge generated by the client.
Even when disabled the Bitwarden clients will send a challenge which will be verified before exchanging the code with the provider (It would be preferable to pass-it to the provider, but due to limitations with the underlying library it's not possible at the moment).


src/db/models/sso_auth.rs
```diff
@@ -0,0 +1,238 @@
...
+        code: OIDCCode,
+    },
+    Error {
+        error: String,
+        error_description: Option<String>,
+    },
+}
+
+#[cfg(sqlite)]
+impl ToSql<Text, diesel::sqlite::Sqlite> for OIDCCodeWrapper {
```
Timshel: The different `ToSql` probably can be improved but wanted some feedbacks before spending more times.

src/sso.rs
```diff
@@ -215,78 +218,52 @@ impl OIDCIdentifier {
...
     }
 
-    let nonce = match SsoNonce::find(&state, conn).await {
-        None => err!(format!("Invalid state cannot retrieve nonce")),
-        Some(nonce) => nonce,
+    let verifier = openidconnect::PkceCodeVerifier::new(client_verifier.into());
+    let challenge = openidconnect::PkceCodeChallenge::from_code_verifier_sha256(&verifier);
+    if challenge.as_str() != String::from(sso_auth.client_challenge.clone()) {
+        err!(format!("PKCE client challenge failed"))
+        // Might need to notify admin ? how ?
```
Timshel: It might be a sign that someone is trying to redeem an intercepted `code`.
What would you recommend ?

src/db/models/sso_auth.rs
```diff
@@ -0,0 +1,238 @@
...
+    #[derive(Identifiable, Queryable, Insertable, AsChangeset)]
+    #[diesel(table_name = sso_auth)]
+    #[diesel(primary_key(state))]
+    pub struct SsoAuth {
+        pub state: OIDCState,
+        pub client_challenge: OIDCCodeChallenge,
+        pub nonce: String,
+        pub verifier: Option<String>,
+        pub redirect_uri: String,
+        pub code_response: Option<OIDCCodeWrapper>,
```
Timshel: Chose to persist the `OIDCCodeWrapper` and `OIDCAuthenticatedUser` either as `Text` or `JSON` instead of mapping all columns since it makes it easier to handle the fact that they are Optional with optional fields.

Could be written as separate table, but I did not think it was worth it for such short-lived data (After the auth expiration of 10 min everything should be deleted anyway).

src/sso.rs
ArcticLampyrid: Is there a similar kind of notification now, for example, sending an alert to the admin when a password is being brute-forcing? If not, then I think it’s unnecessary here as well.

src/db/models/sso_auth.rs
BlackDex, line 1: I find it a bit much to create a new sql-type here.
This brings a lot of boiler-plate code which seems to be different per database type.
The problem with this is that issues could be overlooked quickly, especially if the implementation is different per type.

It might be better to just store the data in text, and add a `get` or `set` kinda function for the functions/locations where you need the value of the `OIDCCodeWrapper` it self, but keep it as a default text type.
And, maybe using the `DieselNewType` derive might help here as well?

I want to prevent future issues when something gets overlooked on a specific database type.
I would actually want most DB Queries to even not need any knowledge of which database engine is running, and custom type implements with different code isn't going to help with that.

BlackDex (commented):
Except for my comment on the `sso_auth` file i think it looks ok.

src/db/models/sso_auth.rs
Timshel: Could switch to using `Text` for all databases, this should simplify the To/FromSql.

For the `DieselNewType`  is it https://github.com/quodlibetor/diesel-derive-newtype ? If I understand correctly it's only for wrapper type ?

BlackDex: Yes, and only useful if you want to use a custom name instead of just having it as `Text`

Timshel: Switched to using `Text` in all databases.
Added the `impl_FromToSqlText` macro to for the `ToSql<Text, DB>` and `FromSql<Text, DB>`.

And rewrote the `SsoAuth.save` to use upsert for all db.

sandervandegeijn:
We are going to do the pentest on the SSO-functionality soon. Will this be merged in the coming weeks? @Timshel suggested this PR should be tested with the pentest.

BlackDex (approved):

sandervandegeijn:
Thanks!

PR #6222, opened 2025-08-25
Fix: SSO Numeric Normalization to recover from preventable SSO failures

krparajuli:
In testing SSO with Microsoft Entra, the SSO fails due to Microsoft Entra supplying time valued fields particularly related to expiry of tokens consist of strings rather than integers.

This causes current SSO flow to error out.

Screenshot 2025-08-19 at 11 39 36

This commit attempts recovery by trying to parse the strings value of relevant fields to integers.

### DIFF
--- a/src/sso_client.rs
+++ b/src/sso_client.rs
@@ -1,4 +1,5 @@
 use regex::Regex;
+use serde_json::Value;
 use std::borrow::Cow;
 use std::time::Duration;
 use url::Url;
@@ -15,7 +16,6 @@ use crate::{
     sso::{OIDCCode, OIDCState},
     CONFIG,
 };
-
 static CLIENT_CACHE_KEY: Lazy<String> = Lazy::new(|| "sso-client".to_string());
 static CLIENT_CACHE: Lazy<Cache<String, Client>> = Lazy::new(|| {
     Cache::builder().max_capacity(1).time_to_live(Duration::from_secs(CONFIG.sso_client_cache_expiration())).build()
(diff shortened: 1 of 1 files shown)


krparajuli:
Required due to poor vendor implementation of OAuth OIDC

krparajuli:
Tagging @Timshel since this is an improvement on recently merged SSO work

Timshel:
Honestly, I probably would not merge it since as you mentioned it's not following the spec (https://openid.net/specs/openid-connect-core-1_0.html#JSONSerialization)

At this point Entra could have its own fork since it's divergent on multiple point (boolean are often serialized as string too but there is a compile time setting `accept-string-booleans` for it in [openidconnect-rs](https://github.com/ramosbugs/openidconnect-rs)).
I won't go too much on what I think of Open Source project having to support the likes of Microsoft not following a spec ...

To be more constructive I would recommend adding support for this in `openidconnect` in similar way that what was done for [accept-string-booleans](https://github.com/ramosbugs/openidconnect-rs/blob/b639b5d39eac6903238867aeb2b29326502e6b26/src/helpers.rs#L140).

Then maintainers could decide to toggle both by default or add a custom build for Entra ...

krparajuli:
@Timshel I wholeheartedly agree with sticking to the specs and not having to support the bs put out by big companies. I also agree with not going down the rabbit hole of having to support out of spec implementations.

However, for the project to flourish, it would be good to have it be able to go places it could not go before. Thus, I will try to maintain a fork of the project to the best of my abilities and support Entra related issues.

Timshel:
> However, for the project to flourish

I'm not a maintainer of the project so can't speak for them on the subject :sweat_smile: you should wait for their feedback.

But I do believe adding support for the parsing in `openidconnect` will be more maintainable and might be useful to others using the lib :).

dani-garcia:
I agree with @Timshel here, there are a lot of out of spec openid servers here and we can't afford to fix the quirks for every single one of them as part of the Vaultwarden project.

If you can submit a PR to the `openidconnect` crate that adds a new feature like the existing `accept-rfc3339-timestamps` or `accept-string-booleans`, we can then enable those features here.

krparajuli:
Thanks for the feedback. Closing this PR. :)

PR #6229, opened 2025-08-25
refactor(config): update template, add validation

tessus:
This change is a follow up to #6166

- add new options to `.env.template`
- add validation for new config option values


BlackDex (approved):
Thanks! LGTM

PR #6233, opened 2025-08-26
Fix panic around sso_master_password_policy

Timshel:
Should fix https://github.com/dani-garcia/vaultwarden/issues/6163 which happened when the config was a valid `serde_json::Value` but not an `Object`.


dani-garcia (approved):
LGTM, thanks

PR #6234, opened 2025-08-26
Update crates, gha and web-vault

BlackDex:
- Update crates to the latest version (Some are yanked and downgraded)
- Update GHA's
- Update web-vault to v2025.8.0

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,8 +34,7 @@ jobs:
     permissions:
       actions: write
       contents: read
-    # We use Ubuntu 22.04 here because this matches the library versions used within the Debian docker containers
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     timeout-minutes: 120
     # Make warnings errors, this is to prevent warnings slipping through.
     # This is done globally to prevent rebuilds when the RUSTFLAGS env variable changes.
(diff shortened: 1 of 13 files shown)


reneleonhardt:
Why not use dependabot to automatically run CI pipelines and then just decide when that update should be merged?

BlackDex:
> Why not use dependabot to automatically run CI pipelines and then just decide when that update should be merged?

Because some, not all, might need some more love and attention or might even become obsolete. 

This way we can keep a bit more control on it.
I my self tend to check what has changed with those crates even if we do not use them directly.

reneleonhardt:
dependabot literally includes the release notes of all dependencies which provide release notes, so you don't even have to search for the changes manually most of the time.
Control is to decide when everything is green to merge or to wait a bit longer.

BlackDex:
As far as i know it only checks direct dependencies, items in `Cargo.toml` and not the lock file. And also doesn't report all changes of the indirect dependencies, but i might be wrong.

reneleonhardt:
What's the harm in trying to get a better understanding?
It creates pull requests for you to review and merge, Cargo.lock is updated too:
https://deepwiki.com/dependabot/dependabot-core/3.8-rust-(cargo)

BlackDex:
> What's the harm in trying to get a better understanding? It creates pull requests for you to review and merge, Cargo.lock is updated too: https://deepwiki.com/dependabot/dependabot-core/3.8-rust-(cargo)

What is the harm in doing it manually? It's not that we do not update them or something. There is a pending PR with database changes which contains all updated crates.

williamdes:
> > Why not use dependabot to automatically run CI pipelines and then just decide when that update should be merged?
> 
> Because some, not all, might need some more love and attention or might even become obsolete.
> 
> This way we can keep a bit more control on it. I my self tend to check what has changed with those crates even if we do not use them directly.

Same, bots do not expose the real diff in the code of each crate. That's a problem, ChangeLogs are not explicit. Manual checking is good

PR #6240, opened 2025-08-27
Fix 2fa recovery endpoint

### LINKED ISSUE #6203: 2FA Recovery Code no longer working
I've enabled two 2FA mechanisms: mail and authenticator app.
Following the [upstream docs](https://bitwarden.com/help/two-step-recovery-code/#use-your-recovery-code), I'm trying to recover using the recovery code which I've just copied from <https://mydomain.example.localhost/#/settings/security/two-factor> (alongside the user name and master password, of course).

I do not get an error in the web UI, but I'm just redirected to the main login view; the network pane reads as follows:

image

The log shows a single line:
> vaultwarden-5d964b5d76-zqmsr [2025-08-19 22:49:21.283][vaultwarden::api::identity][ERROR] Invalid two factor provider

As per the discussion [here](https://github.com/dani-garcia/vaultwarden/discussions/6200#discussioncomment-14161185), we could identify that upstream has refactored their recovery code handling and added a new type.

_Originally posted by @shadow1runner in https://github.com/dani-garcia/vaultwarden/discussions/6200_

BlackDex:
The newer web-vaults handle the 2fa recovery code differently. This commit fixes this by adding this new flow.

Fixes #6200
Fixes #6203

### DIFF
--- a/src/api/identity.rs
+++ b/src/api/identity.rs
@@ -267,7 +267,7 @@ async fn _sso_login(
         }
         Some((mut user, sso_user)) => {
             let mut device = get_device(&data, conn, &user).await?;
-            let twofactor_token = twofactor_auth(&user, &data, &mut device, ip, client_version, conn).await?;
+            let twofactor_token = twofactor_auth(&mut user, &data, &mut device, ip, client_version, conn).await?;
 
             if user.private_key.is_none() {
                 // User was invited a stub was created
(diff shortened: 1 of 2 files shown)


dani-garcia (approved):
Nice LGTM. We should check if all the clients have been updated to the new endpoint, so we can delete the older one in a couple of releases.

tessus:
Is there also legacy code that can be removed in the future?

In such a case something similar to bw's approach might be useful:

```
[Obsolete("To be removed when the feature flag pm-17128-recovery-code-login is removed PM-18175.")]
```

I remember that there were often comments in other PRs that code can be removed at a later point, but I never saw any PRs that removed any code. Or maybe I just missed them. Sorry for being this inquisitive, but I have seen projects where old code was never removed and made new development extremely complicated (and introduced actually a few security issues). 

P.S.: Sorry, Daniel. I only saw your comment after I posted mine.

BlackDex:
> Nice LGTM. We should check if all the clients have been updated to the new endpoint, so we can delete the older one in a couple of releases.

Yea, i think there are a few others we might be able to delete too.
Some tech debt we need to cleanup.

PR #6247, opened 2025-08-28
Fix around singleorg policy

Timshel:
Should fix:

- Prevent a user already in an org to join an organization with the single org policy activated
- When activating the single org policy revoke the membership (as mentioned in the warning) instead of deleting.
- Allow to restore an invitation

Moved all the logic back to the `check_user_allowed` instead of having to duplicate it from all call site, might make sense to move it out, inside a mod.rs maybe ?

Add the Admin and Invited membership test in `check_user_allowed`, means that all membership modification need to be done before calling the  check.

### DIFF
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -23,7 +23,7 @@ use crate::{
         backup_sqlite, get_sql_server_version,
         models::{
             Attachment, Cipher, Collection, Device, Event, EventType, Group, Invitation, Membership, MembershipId,
-            MembershipType, OrgPolicy, OrgPolicyErr, Organization, OrganizationId, SsoUser, TwoFactor, User, UserId,
+            MembershipType, OrgPolicy, Organization, OrganizationId, SsoUser, TwoFactor, User, UserId,
         },
         DbConn, DbConnType, ACTIVE_DB_TYPE,
     },
@@ -556,23 +556,9 @@ async fn update_membership_type(data: Json<MembershipTypeData>, token: AdminToke
         }
     }
 
+    member_to_edit.atype = new_type;
     // This check is also done at api::organizations::{accept_invite, _confirm_invite, _activate_member, edit_member}, update_membership_type
-    // It returns different error messages per function.
-    if new_type < MembershipType::Admin {
-        match OrgPolicy::is_user_allowed(&member_to_edit.user_uuid, &member_to_edit.org_uuid, true, &conn).await {
-            Ok(_) => {}
-            Err(OrgPolicyErr::TwoFactorMissing) => {
-                if CONFIG.email_2fa_auto_fallback() {
-                    two_factor::email::find_and_activate_email_2fa(&member_to_edit.user_uuid, &conn).await?;
-                } else {
-                    err!("You cannot modify this user to this type because they have not setup 2FA");
-                }
-            }
-            Err(OrgPolicyErr::SingleOrgEnforced) => {
-                err!("You cannot modify this user to this type because it is a member of an organization which forbids it");
-            }
-        }
-    }
+    OrgPolicy::check_user_allowed(&member_to_edit, "modify", &conn).await?;
 
     log_event(
         EventType::OrganizationUserUpdated as i32,
@@ -585,7 +571,6 @@ async fn update_membership_type(data: Json<MembershipTypeData>, token: AdminToke
     )
     .await;
 
-    member_to_edit.atype = new_type;
     member_to_edit.save(&conn).await
 }
 
--- a/src/api/core/mod.rs
+++ b/src/api/core/mod.rs
@@ -53,7 +53,7 @@ use crate::{
     api::{EmptyResult, JsonResult, Notify, UpdateType},
     auth::Headers,
     db::{
-        models::{Membership, MembershipStatus, MembershipType, OrgPolicy, OrgPolicyErr, Organization, User},
+        models::{Membership, MembershipStatus, OrgPolicy, Organization, User},
         DbConn,
     },
     error::Error,
@@ -269,26 +269,11 @@ async fn accept_org_invite(
         err!("User already accepted the invitation");
     }
 
-    // This check is also done at accept_invite, _confirm_invite, _activate_member, edit_member, admin::update_membership_type
-    // It returns different error messages per function.
-    if member.atype < MembershipType::Admin {
-        match OrgPolicy::is_user_allowed(&member.user_uuid, &member.org_uuid, false, conn).await {
-            Ok(_) => {}
-            Err(OrgPolicyErr::TwoFactorMissing) => {
-                if crate::CONFIG.email_2fa_auto_fallback() {
-                    two_factor::email::activate_email_2fa(user, conn).await?;
-                } else {
-                    err!("You cannot join this organization until you enable two-step login on your user account");
-                }
-            }
-            Err(OrgPolicyErr::SingleOrgEnforced) => {
-                err!("You cannot join this organization because you are a member of an organization which forbids it");
-            }
-        }
-    }
-
     member.status = MembershipStatus::Accepted as i32;
     member.reset_password_key = reset_password_key;
+
+    // This check is also done at accept_invite, _confirm_invite, _activate_member, edit_member, admin::update_membership_type
+    OrgPolicy::check_user_allowed(&member, "join", conn).await?;
 
     member.save(conn).await?;
 
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -15,7 +15,7 @@ use crate::{
         models::{
             Cipher, CipherId, Collection, CollectionCipher, CollectionGroup, CollectionId, CollectionUser, EventType,
             Group, GroupId, GroupUser, Invitation, Membership, MembershipId, MembershipStatus, MembershipType,
-            OrgPolicy, OrgPolicyErr, OrgPolicyType, Organization, OrganizationApiKey, OrganizationId, User, UserId,
+            OrgPolicy, OrgPolicyType, Organization, OrganizationApiKey, OrganizationId, User, UserId,
         },
         DbConn,
     },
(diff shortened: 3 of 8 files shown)


src/db/models/org_policy.rs
```diff
@@ -285,31 +279,35 @@ impl OrgPolicy {
...
-        org_uuid: &OrganizationId,
-        exclude_current_org: bool,
-        conn: &mut DbConn,
-    ) -> OrgPolicyResult {
-        // Enforce TwoFactor/TwoStep login
-        if TwoFactor::find_by_user(user_uuid, conn).await.is_empty() {
-            match Self::find_by_org_and_type(org_uuid, OrgPolicyType::TwoFactorAuthentication, conn).await {
-                Some(p) if p.enabled => {
-                    return Err(OrgPolicyErr::TwoFactorMissing);
+    pub async fn check_user_allowed(m: &Membership, action: &str, conn: &mut DbConn) -> EmptyResult {
```
stefan0xC: I'd rename this function because it has side effects
```suggestion
    pub async fn enforce_membership_policies(m: &Membership, action: &str, conn: &mut DbConn) -> EmptyResult {
```

src/db/models/org_policy.rs
```diff
@@ -285,31 +279,35 @@ impl OrgPolicy {
...
+        if m.atype < MembershipType::Admin && m.status > (MembershipStatus::Invited as i32) {
+            // Enforce TwoFactor/TwoStep login
+            if let Some(p) = Self::find_by_org_and_type(&m.org_uuid, OrgPolicyType::TwoFactorAuthentication, conn).await
+            {
+                if p.enabled && TwoFactor::find_by_user(&m.user_uuid, conn).await.is_empty() {
+                    if CONFIG.email_2fa_auto_fallback() {
+                        two_factor::email::find_and_activate_email_2fa(&m.user_uuid, conn).await?;
+                    } else {
+                        err!(format!("Cannot {} because 2FA is required (membership {})", action, m.uuid));
```
stefan0xC, line 286:
```suggestion
                        err!(format!("Cannot {} member {} because 2FA is required", action, m.uuid));
```

src/db/models/org_policy.rs
```diff
@@ -285,31 +279,35 @@ impl OrgPolicy {
...
                 }
-                _ => {}
-            };
-        }
+            }
+
+            // Check if the user is part of another Orgnization with SingleOrg activated
+            if Self::is_applicable_to_user(&m.user_uuid, OrgPolicyType::SingleOrg, Some(&m.org_uuid), conn).await {
+                err!(format!(
+                    "Cannot {} because another organization policy forbids it (membership {})",
```
stefan0xC, line 294:
```suggestion
                    "Cannot {} member {} because another organization's SingleOrg policy forbids it",
```

src/db/models/org_policy.rs
```diff
@@ -285,31 +279,35 @@ impl OrgPolicy {
...
-        };
-        if Self::is_applicable_to_user(user_uuid, OrgPolicyType::SingleOrg, exclude_org, conn).await {
-            return Err(OrgPolicyErr::SingleOrgEnforced);
+            if let Some(p) = Self::find_by_org_and_type(&m.org_uuid, OrgPolicyType::SingleOrg, conn).await {
+                if p.enabled
+                    && Membership::count_accepted_and_confirmed_by_user(&m.user_uuid, &m.org_uuid, conn).await > 0
+                {
+                    err!(format!("Cannot {} because the organization policy forbids being part of other organization (membership {})", action, m.uuid));
+                }
+            }
```
stefan0xC, line 305: I'm not sure I understand why we check for the SingleOrg policy twice? Does this differ from the call to `is_applicable_to_user` above?

src/api/admin.rs
```diff
@@ -544,23 +544,9 @@ async fn update_membership_type(data: Json<MembershipTypeData>, token: AdminToke
...
-                } else {
-                    err!("You cannot modify this user to this type because they have not setup 2FA");
-                }
-            }
-            Err(OrgPolicyErr::SingleOrgEnforced) => {
-                err!("You cannot modify this user to this type because it is a member of an organization which forbids it");
-            }
-        }
-    }
+    OrgPolicy::check_user_allowed(&member_to_edit, "modify", &mut conn).await?;
```
stefan0xC:
```suggestion
    OrgPolicy::check_user_allowed(&member_to_edit, "change type of", &mut conn).await?;
```

src/api/core/mod.rs
```diff
@@ -271,27 +271,12 @@ async fn accept_org_invite(
...
-                err!("You cannot join this organization because you are a member of an organization which forbids it");
-            }
-        }
-    }
-
     member.status = MembershipStatus::Accepted as i32;
     member.reset_password_key = reset_password_key;
 
+    // This check is also done at accept_invite, _confirm_invite, _activate_member, edit_member, admin::update_membership_type
+    OrgPolicy::check_user_allowed(&member, "join", conn).await?;
```
stefan0xC, line 276:
```suggestion
    OrgPolicy::check_user_allowed(&member, "accept invitation of", conn).await?;
```

src/static/templates/email/send_single_org_removed_from_org.hbs
```diff
@@ -1,4 +1,4 @@
-You have been removed from {{{org_name}}}
+Your access to {{{org_name}}} has been revoked
 <!---------------->
-Your user account has been removed from the *{{org_name}}* organization because you are a part of another organization. The {{org_name}} organization has enabled a policy that prevents users from being a part of multiple organizations. Before you can re-join this organization you need to leave all other organizations or join with a different account.
+Your acess to the *{{org_name}}* organization has been revoked because you are a part of another organization. The {{org_name}} organization has enabled a policy that prevents users from being a part of multiple organizations. Before your access can be restored you need to leave all other organizations or join with a different account.
```
stefan0xC:
```suggestion
Your access to the *{{org_name}}* organization has been revoked because you are a part of another organization. The {{org_name}} organization has enabled a policy that prevents users from being a part of multiple organizations. Before your access can be restored you need to leave all other organizations or join with a different account.
```

src/static/templates/email/send_single_org_removed_from_org.html.hbs
```diff
@@ -1,10 +1,10 @@
...
-         Your user account has been removed from the <b style="margin: 0; font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif; box-sizing: border-box; font-size: 16px; color: #333; line-height: 25px; -webkit-font-smoothing: antialiased; -webkit-text-size-adjust: none;">{{org_name}}</b> organization because you are a part of another organization. The {{org_name}} organization has enabled a policy that prevents users from being a part of multiple organizations. Before you can re-join this organization you need to leave all other organizations or join with a different account.
+         Your acess to the <b style="margin: 0; font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif; box-sizing: border-box; font-size: 16px; color: #333; line-height: 25px; -webkit-font-smoothing: antialiased; -webkit-text-size-adjust: none;">{{org_name}}</b> organization has been revoked because you are a part of another organization. The {{org_name}} organization has enabled a policy that prevents users from being a part of multiple organizations. Before your access can be restored you need to leave all other organizations or join with a different account.
```
stefan0xC:
```suggestion
         Your access to the <b style="margin: 0; font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif; box-sizing: border-box; font-size: 16px; color: #333; line-height: 25px; -webkit-font-smoothing: antialiased; -webkit-text-size-adjust: none;">{{org_name}}</b> organization has been revoked because you are a part of another organization. The {{org_name}} organization has enabled a policy that prevents users from being a part of multiple organizations. Before your access can be restored you need to leave all other organizations or join with a different account.
```

src/api/core/organizations.rs
```diff
@@ -2163,14 +2134,15 @@ async fn put_policy(
 
     // When enabling the SingleOrg policy, remove this org's members that are members of other orgs
```
stefan0xC, line 2126:
```suggestion
    // When enabling the SingleOrg policy, revoke the memberships of users that are members of other organizations.
```

stefan0xC (changes requested):
Thanks for the refactoring. I have a few suggestions.

src/db/models/org_policy.rs
Timshel: `is_applicable_to_user` check if another Org with `SingleOrg` policy exist (the current org is excluded); it completely ignores orgs without this policy.

This make the second check needed to ensure that if the membership org has `SingleOrg` activated then we need to ensure the user is not already part of another org.

src/api/core/mod.rs
```diff
@@ -271,27 +271,12 @@ async fn accept_org_invite(
...
-            }
-        }
-    }
-
     member.status = MembershipStatus::Accepted as i32;
     member.reset_password_key = reset_password_key;
 
+    // This check is also done at accept_invite, _confirm_invite, _activate_member, edit_member, admin::update_membership_type
+    OrgPolicy::enforce_membership_policies(&member, "accept invitation of", conn).await?;
+
```
BlackDex, line 277: You have moved this check after already changing some `member` settings.
For example, the `status` is now changed here to `Accepted` while it might have been something else. Even though here a check is done at the start of the function which should prevent this, it still isn't sound from my point of view to change the values before doing a check on those same values.

So, i would suggest to move this check before changing those values to ensure they will only be changed if the check passes.

src/api/core/organizations.rs
```diff
@@ -1466,27 +1466,12 @@ async fn _confirm_invite(
...
-            }
-        }
-    }
-
     member_to_confirm.status = MembershipStatus::Confirmed as i32;
     member_to_confirm.akey = key.to_string();
 
+    // This check is also done at accept_invite, _confirm_invite, _activate_member, edit_member, admin::update_membership_type
+    OrgPolicy::enforce_membership_policies(&member_to_confirm, "confirm", conn).await?;
+
```
BlackDex, line 1471: Same situation here, do not modify attributes before validating.

src/api/core/organizations.rs
```diff
@@ -1640,27 +1625,13 @@ async fn edit_member(
...
-        }
-    }
-
     member_to_edit.access_all = access_all;
     member_to_edit.atype = new_type as i32;
 
+    // This check is also done at accept_invite, _confirm_invite, _activate_member, edit_member, admin::update_membership_type
+    // We need to perform the check after changing the type since `admin` is exempt.
+    OrgPolicy::enforce_membership_policies(&member_to_edit, "modify", &mut conn).await?;
+
```
BlackDex, line 1625: Same situation here, do not modify attributes before validating.

src/api/core/organizations.rs
```diff
@@ -2627,25 +2600,10 @@ async fn _restore_member(
...
-                    Err(OrgPolicyErr::SingleOrgEnforced) => {
-                        err!("You cannot restore this user because they are a member of an organization which forbids it");
-                    }
-                }
-            }
-
             member.restore();
+            // This check is also done at accept_invite, _confirm_invite, _activate_member, edit_member, admin::update_membership_type
+            // This check need to be done after restoring to work with the correct status
+            OrgPolicy::enforce_membership_policies(&member, "restore", conn).await?;
```
BlackDex: Same here, you now restore the member already even though the next check might fail.
While the `restore()` function currently doesn't save the user and saving is done after this check, modifying that attribute already is not sound.

BlackDex (changes requested):
I have not yet checked everything in this PR, but i have a few remarks.

src/api/core/mod.rs
Timshel: The status has some impact on the check itself, so it needs to be changed before checking.
The goal is to verify if the membership modifications are valid or blocked by the org policies.

BlackDex: No, it actually shouldn't impact, since the first check done is if the status is equal to `Invited`. Then `enforce_membership_policies` should probably check for `>=` instead of `>`.

Timshel: > should probably check for >= instead of >

I don't believe 2FA is enforced on invitation.
So it becomes enforced only when `Accepted`.

The point of the PR is that the policies check are complicated and to have only one function to make them.
If the checks need to be done before the membership is modified it's impossible to share the code.

BlackDex (approved):
Ok... I Think it looks ok.
@danig, if you can take a look as well :)

Issue #6179, opened 2025-08-12
SSO login/registration kinda buged from yesterday to today


Joly0:
### Prerequisites

- [x] I have searched the existing **Closed _AND_ Open** [Issues](https://github.com/dani-garcia/vaultwarden/issues?q=is%3Aissue%20) **_AND_** [Discussions](https://github.com/dani-garcia/vaultwarden/discussions?discussions_q=)
- [x] I have searched and read the [documentation](https://github.com/dani-garcia/vaultwarden/wiki/)

### Your environment (Generated via diagnostics page)

* Vaultwarden version: v1.34.3-8e7eeab2
* Web-vault version: v2025.7.0
* OS/Arch: linux/x86_64
* Running within a container: true (Base: Debian)
* Database type: SQLite
* Database version: 3.50.2
* Uses config.json: true
* Uses a reverse proxy: true
* IP Header check: true (X-Real-IP)
* Internet access: true
* Internet access via a proxy: false
* DNS Check: true
* Browser/Server Time Check: true
* Server/NTP Time Check: true
* Domain Configuration Check: true
* HTTPS Check: true
* Websocket Check: true
* HTTP Response Checks: true

### Config & Details (Generated via diagnostics page)

<details><summary>Show Config & Details</summary>

**Environment settings which are overridden:** DOMAIN, ADMIN_TOKEN

**Config:**

```json
{
  "_duo_akey": null,
  "_enable_duo": true,
  "_enable_email_2fa": false,
  "_enable_smtp": true,
  "_enable_yubico": true,
  "_icon_service_csp": "",
  "_icon_service_url": "",
  "_ip_header_enabled": true,
  "_max_note_size": 10000,
  "_smtp_img_src": "***:",
  "admin_ratelimit_max_burst": 3,
  "admin_ratelimit_seconds": 300,
  "admin_session_lifetime": 20,
  "admin_token": "***",
  "allowed_connect_src": "",
  "allowed_iframe_ancestors": "",
  "attachments_folder": "data/attachments",
  "auth_request_purge_schedule": "30 * * * * *",
  "authenticator_disable_time_drift": false,
  "data_folder": "data",
  "database_conn_init": "",
  "database_max_conns": 10,
  "database_timeout": 30,
  "database_url": "***************",
  "db_connection_retries": 15,
  "disable_2fa_remember": false,
  "disable_admin_token": false,
  "disable_icon_download": false,
  "domain": "*****://************************",
[...]

PR #6262, opened 2025-09-01
Fix `sso_user` dropped on `User::save`

Timshel:
Should fix https://github.com/dani-garcia/vaultwarden/issues/6179#issuecomment-3242088962

The issue is that if there is no `ForeignKeyViolation` a `User::save` first delete the record with `sqlite` and `mysql`.

Since the `sso_user` table was added with `ON UPDATE CASCADE ON DELETE CASCADE`, the violation was not triggered and the entry was deleted.

I thought changing the `save` logic was better than changing the cascade logic since :

- Especially for users `update` is the happy path (as opposed to the violation being triggered in most cases).
- It does not prevent using cascade
- Less of a trap (it was missed during the review and testing, running with `SSO_SIGNUPS_MATCH_EMAIL` the association was just restored).

Made the change only for `User` but it probably would apply to most table.

### DIFF
--- a/src/db/models/user.rs
+++ b/src/db/models/user.rs
@@ -284,21 +284,22 @@ impl User {
 
         db_run! {conn:
             sqlite, mysql {
-                match diesel::replace_into(users::table)
-                    .values(UserDb::to_db(self))
+                let value = UserDb::to_db(self);
+                // Don't use replace_into() since it wants to delete the record first.
+                match diesel::update(users::table)
+                    .filter(users::uuid.eq(&self.uuid))
+                    .set(&value)
                     .execute(conn)
                 {
-                    Ok(_) => Ok(()),
-                    // Record already exists and causes a Foreign Key Violation because replace_into() wants to delete the record first.
-                    Err(diesel::result::Error::DatabaseError(diesel::result::DatabaseErrorKind::ForeignKeyViolation, _)) => {
-                        diesel::update(users::table)
-                            .filter(users::uuid.eq(&self.uuid))
-                            .set(UserDb::to_db(self))
+                    Ok(1) => Ok(()),
+                    Ok(_) => {
+                        diesel::insert_into(users::table)
+                            .values(value)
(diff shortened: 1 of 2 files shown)


stefan0xC:
Could we change the logic to be more like the postgres method? According to the diesel documentation for [`IncompleteOnConflict::do_update`](https://docs.rs/diesel/2.2.12/diesel/upsert/struct.IncompleteOnConflict.html#method.do_update) postgres and sqlite could be called the same and only mariadb/mysql requires `diesel::dsl::DuplicatedKeys` for the upsert functionality? 

Support for this was added in [diesel v2.1.0](https://github.com/diesel-rs/diesel/releases/tag/v2.1.0). 

edit: fix link to documentation

Timshel:
It works :tada:

src/db/models/user.rs
```diff
@@ -283,24 +283,25 @@ impl User {
...
-            sqlite, mysql {
-                match diesel::replace_into(users::table)
-                    .values(UserDb::to_db(self))
+            mysql {
+                let value = UserDb::to_db(self);
+                // Don't use replace_into() since it wants to delete the record first.
+                match diesel::update(users::table)
+                    .filter(users::uuid.eq(&self.uuid))
+                    .set(&value)
                     .execute(conn)
```
stefan0xC: I thought something more like this should be possible now:

```suggestion
                diesel::insert_into(users::table)
                    .values(&value)
                    .on_conflict(diesel::dsl::DuplicatedKeys)
                    .do_update()
                    .set(&value)
                    .execute(conn)
                    .map_res("Error saving user")
```

and also removing  https://github.com/dani-garcia/vaultwarden/blob/9b7a2018962a2dc1faf78565246d9e16e5662ee9/src/db/models/user.rs#L293-L302 (which GitHub won't let me suggest nicely)...

Timshel: O yeah sorry :sweat:

PR #6263, opened 2025-09-01
Change OIDC dummy identifier

Timshel:
The main use of the fake identifier is to return a value in `Organizations::get_auto_enroll_status` to allow bypassing the screen which would otherwise ask for one.

Though that the current value was a bit too simple and run the risk of someone creating an Organization with the same name.

### DIFF
--- a/docker/DockerSettings.yaml
+++ b/docker/DockerSettings.yaml
@@ -1,6 +1,6 @@
 ---
-vault_version: "v2025.8.0"
-vault_image_digest: "sha256:41c2b51c87882248f405d5a0ab37210d2672a312ec5d4f3b9afcdbbe8eb9d57d"
+vault_version: "v2025.9.1"
+vault_image_digest: "sha256:15a126ca967cd2efc4c9625fec49f0b972a3f7d7d81d7770bb0a2502d5e4b8a4"
 # Cross Compile Docker Helper Scripts v1.6.1
 # We use the linux/amd64 platform shell scripts since there is no difference between the different platform scripts
 # https://github.com/tonistiigi/xx | https://hub.docker.com/r/tonistiigi/xx/tags
--- a/docker/Dockerfile.alpine
+++ b/docker/Dockerfile.alpine
@@ -19,15 +19,15 @@
 # - From https://hub.docker.com/r/vaultwarden/web-vault/tags,
 #   click the tag name to view the digest of the image it currently points to.
 # - From the command line:
-#     $ docker pull docker.io/vaultwarden/web-vault:v2025.8.0
-#     $ docker image inspect --format "{{.RepoDigests}}" docker.io/vaultwarden/web-vault:v2025.8.0
-#     [docker.io/vaultwarden/web-vault@sha256:41c2b51c87882248f405d5a0ab37210d2672a312ec5d4f3b9afcdbbe8eb9d57d]
+#     $ docker pull docker.io/vaultwarden/web-vault:v2025.9.1
+#     $ docker image inspect --format "{{.RepoDigests}}" docker.io/vaultwarden/web-vault:v2025.9.1
+#     [docker.io/vaultwarden/web-vault@sha256:15a126ca967cd2efc4c9625fec49f0b972a3f7d7d81d7770bb0a2502d5e4b8a4]
 #
 # - Conversely, to get the tag name from the digest:
-#     $ docker image inspect --format "{{.RepoTags}}" docker.io/vaultwarden/web-vault@sha256:41c2b51c87882248f405d5a0ab37210d2672a312ec5d4f3b9afcdbbe8eb9d57d
-#     [docker.io/vaultwarden/web-vault:v2025.8.0]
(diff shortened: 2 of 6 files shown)


src/sso.rs
```diff
@@ -19,7 +19,7 @@ use crate::{
     CONFIG,
 };
 
-pub static FAKE_IDENTIFIER: &str = "Vaultwarden";
+pub static FAKE_IDENTIFIER: &str = "DUMMY_IDENTIFIER_FOR_OIDC";
```
tessus:
```suggestion
pub static FAKE_IDENTIFIER: &str = "VW_DUMMY_IDENTIFIER_FOR_OIDC";
```

I know it is not necessary, but I usually like to use project specific namespaces for stuff like that. But if you are that concerned, a uuidv4 should also do it. ;-)

Timshel: I added the prefix, I prefer something which make it clear that it's a dummy value as opposed to a more opaque uuid.

stefan0xC:
I think this should be fine. But while looking at where this fake identifier is used I noticed that this comment 
https://github.com/dani-garcia/vaultwarden/blob/a2ad1dc7c3d28834749d4b14206838d795236c27/src/api/core/organizations.rs#L342 is misleading because that function does not exist. Could you change `get_org_domain_sso_details` to `get_org_domain_sso_verified` (or rename the function)? At least I think that this function is meant:
https://github.com/dani-garcia/vaultwarden/blob/a2ad1dc7c3d28834749d4b14206838d795236c27/src/api/core/organizations.rs#L974

stefan0xC:
Also I think that the usage of `Organization::find_by_name` (if it's not the fake identifier) https://github.com/dani-garcia/vaultwarden/blob/a2ad1dc7c3d28834749d4b14206838d795236c27/src/api/core/accounts.rs#L369-L370 and https://github.com/dani-garcia/vaultwarden/blob/a2ad1dc7c3d28834749d4b14206838d795236c27/src/api/core/organizations.rs#L346-L353 seems a bit flaky to me, because there is no guarantee that the set name of an organization is unique or the first result is the correct organization.
https://github.com/dani-garcia/vaultwarden/blob/a2ad1dc7c3d28834749d4b14206838d795236c27/src/db/models/organization.rs#L395-L402

So this should probably also be improved? Bitwarden allows to set a [unique identifier for SSO](https://bitwarden.com/help/configure-sso-oidc/) - so maybe we should do something similar instead? (I mean if we only support one organization for SSO, maybe it would be better to let users create an organization before configuring SSO and have them set the UUID of that organization explicitly so that it can be used as an unique identifier?)

edit: and if no organization is needed than the code should be changed as well to reflect that, so that no organization is returned? (I'm not actually sure how this is/was intended.)

Timshel:
> I mean if we only support one organization for SSO

The SSO config is global so no organization is required. Since some flow require it the dummy value is used as a fallback.

> if no organization is needed than the code should be changed as well to reflect that, so that no organization is returned

While not needed if an Organization associated with the user is found I believe it's better to return it since then it's own org policy can be used (password strength ...).

> seems a bit flaky to me, because there is no guarantee that the set name of an organization is unique

I agree, but at the same time having users set a value for each org is a pain. \
You mention the `uuid` I think it should work since with the current `web-vault` the value should never be displayed (I believe it was not the case at some time and why I decided to go with the name).

stefan0xC:
> While not needed if an Organization associated with the user is found I believe it's better to return it since then it's own org policy can be used (password strength ...).

That's what I don't get because this could introduce unwanted side effects (if it's possible that for users in multiple organizations different organizations could be returned), I'd rather we have an organization required or to never return an organization to eliminate any issues regarding that. (If you can't join multiple organizations when using SSO then consider my point moot of course.)

> having users set a value for each org is a pain

I understand. My idea would be to have the admin user create an organization and to use the UUID of that organization (also instead of the dummy identifier), so it's properly defined which organization is responsible for policies etc. I don't think the user should then ever have to enter that value (until we add proper support for multiple organizations and setting an unique identifier for each organization). But yeah, I have no idea if this makes sense in practice.

Timshel:
> if it's possible that for users in multiple organizations different organizations could be returned

That's kind of the point of [get_org_domain_sso_verified](https://github.com/dani-garcia/vaultwarden/blob/main/src/api/core/organizations.rs#L977) which return all the organization from a user. 
At the moment it appears to use only the first one, but it's not something I can change.

> I'd rather we have an organization required

The feature was added with the goal of not to having this requirement.


> My idea would be to have the admin user create an organization and to use the UUID of that organization

This would then need to be done for all Organization (And depending on the setup all users can create Orgs).


I believe some improvement could be done to order of the orgs returned in `get_org_domain_sso_verified` but outside of it, when multiple org policy might conflict it's probably better to do something like a merge as is done for the [master_password_policy](https://github.com/dani-garcia/vaultwarden/blob/a2ad1dc7c3d28834749d4b14206838d795236c27/src/api/mod.rs#L89).

PR #6279, opened 2025-09-08
Use Diesels MultiConnections Derive

BlackDex:
With this PR we remove almost all custom macro's to create the multiple database type code. This is now handled by Diesel it self.

This removed the need of the following functions/macro's:
 - `db_object!`
 - `::to_db`
 - `.from_db()`

It is also possible to just use one schema instead of multiple per type.

Also done:
 - Refactored the SQLite backup function
 - Some formatting of queries so every call is one a separate line, this looks a bit better
 - Declare `conn` as mut inside each `db_run!` instead of having to declare it as `mut` in functions or calls
 - Added an `ACTIVE_DB_TYPE` static which holds the currently active database type
 - Removed `diesel_logger` crate and use Diesel's `set_default_instrumentation()`
   If you want debug queries you can now simply change the log level of `vaultwarden::db::query_logger`
 - Use PostgreSQL v17 in the Alpine images to match the Debian Trixie version
 - Optimized the Workflows since `diesel_logger` isn't needed anymore

And on the extra plus-side, this lowers the compile-time and binary size too.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,15 +3,6 @@
 version = 4
 
 [[package]]
-name = "addr2line"
-version = "0.25.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1b5d307320b3181d6d7954e663bd7c774a838b8220fe0593c86d9fb09f498b4b"
-dependencies = [
- "gimli",
-]
-
-[[package]]
 name = "adler2"
 version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -449,9 +440,9 @@ dependencies = [
 
 [[package]]
 name = "aws-sdk-ssooidc"
-version = "1.88.0"
+version = "1.89.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9a68d675582afea0e94d38b6ca9c5aaae4ca14f1d36faa6edb19b42e687e70d7"
+checksum = "695dc67bb861ccb8426c9129b91c30e266a0e3d85650cafdf62fcca14c8fd338"
 dependencies = [
  "aws-credential-types",
  "aws-runtime",
@@ -661,28 +652,13 @@ dependencies = [
 
 [[package]]
 name = "backon"
-version = "1.5.2"
+version = "1.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "592277618714fbcecda9a02ba7a8781f319d26532a88553bbacc77ba5d2b3a8d"
+checksum = "cffb0e931875b666fc4fcb20fee52e9bbd1ef836fd9e9e04ec21555f9f85f7ef"
 dependencies = [
  "fastrand",
  "gloo-timers",
  "tokio",
-]
-
-[[package]]
-name = "backtrace"
-version = "0.3.76"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bb531853791a215d7c62a30daf0dde835f381ab5de4589cfe7c649d2cbe92bd6"
-dependencies = [
- "addr2line",
- "cfg-if",
- "libc",
- "miniz_oxide",
- "object",
- "rustc-demangle",
- "windows-link 0.2.1",
 ]
 
 [[package]]
@@ -721,9 +697,9 @@ checksum = "55248b47b0caf0546f7988906588779981c43bb1bc9d0c44087278f80cdb44ba"
 
 [[package]]
 name = "base64urlsafedata"
-version = "0.5.2"
+version = "0.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e5913e643e4dfb43d5908e9e6f1386f8e0dfde086ecef124a6450c6195d89160"
+checksum = "215ee31f8a88f588c349ce2d20108b2ed96089b96b9c2b03775dc35dd72938e8"
 dependencies = [
  "base64 0.21.7",
  "pastey",
@@ -732,9 +708,9 @@ dependencies = [
 
 [[package]]
 name = "bigdecimal"
-version = "0.4.8"
+version = "0.4.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1a22f228ab7a1b23027ccc6c350b72868017af7ea8356fbdf19f8d991c690013"
+checksum = "560f42649de9fa436b73517378a147ec21f6c997a546581df4b4b31677828934"
 dependencies = [
  "autocfg",
  "libm",
@@ -751,9 +727,9 @@ checksum = "383d29d513d8764dcdc42ea295d979eb99c3c9f00607b3692cf68a431f7dca72"
 
 [[package]]
 name = "bitflags"
-version = "2.9.4"
+version = "2.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2261d10cca569e4643e526d8dc2e62e433cc8aba21ab764233731f8d369bf394"
+checksum = "812e12b5285cc515a9c72a5c1d3b6d46a19dac5acfef5265968c166106e31dd3"
 
 [[package]]
 name = "blake2"
@@ -934,9 +910,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.2.41"
+version = "1.2.43"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ac9fe6cdbb24b6ade63616c0a0688e45bb56732262c158df3c0c4bea4ca47cb7"
+checksum = "739eb0f94557554b3ca9a86d2d37bebd49c5e6d0c1d2bda35ba5bdac830befc2"
 dependencies = [
  "find-msvc-tools",
  "jobserver",
@@ -946,9 +922,9 @@ dependencies = [
 
 [[package]]
 name = "cfg-if"
-version = "1.0.3"
+version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2fd1289c04a9ea8cb22300a459a72a385d7c73d3259e2ed7dcb2af674838cfa9"
+checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
 
 [[package]]
 name = "cfg_aliases"
@@ -1393,9 +1369,9 @@ dependencies = [
 
 [[package]]
 name = "deranged"
-version = "0.5.4"
+version = "0.5.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a41953f86f8a05768a6cda24def994fd2f424b04ec5c719cf89989779f199071"
+checksum = "ececcb659e7ba858fb4f10388c250a7252eb0a27373f1a72b8748afdd248e587"
 dependencies = [
  "powerfmt",
  "serde_core",
@@ -1488,9 +1464,9 @@ dependencies = [
 
 [[package]]
 name = "diesel"
-version = "2.3.2"
+version = "2.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e8496eeb328dce26ee9d9b73275d396d9bddb433fa30106cf6056dd8c3c2764c"
+checksum = "5e7624a3bb9fffd82fff016be9a7f163d20e5a89eb8d28f9daaa6b30fff37500"
 dependencies = [
  "bigdecimal",
  "bitflags",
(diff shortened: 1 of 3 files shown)


src/db/mod.rs
```diff
@@ -51,170 +39,177 @@ where
...
+    Sqlite(diesel::sqlite::SqliteConnection),
+    #[cfg(mysql)]
+    Mysql(diesel::mysql::MysqlConnection),
+    #[cfg(postgresql)]
+    Postgresql(diesel::pg::PgConnection),
+}
 
-        #[allow(non_camel_case_types)]
-        pub enum DbConnInner { $( #[cfg($name)] $name(PooledConnection<ConnectionManager< $ty >>), )+ }
+#[cfg(query_logger)]
```
weiznich: You can likely also replace the query logger dependency by setting the right instrumentation for either [the connection](https://docs.diesel.rs/2.2.x/diesel/connection/trait.Connection.html#tymethod.set_instrumentation) or at [global level](https://docs.diesel.rs/2.2.x/diesel/connection/fn.set_default_instrumentation.html). That can possibly help simplify the code a bit more.

src/db/mod.rs
```diff
@@ -51,170 +39,177 @@ where
...
-
-        pub struct DbConn {
-            conn: Arc<Mutex<Option<DbConnInner>>>,
-            permit: Option<OwnedSemaphorePermit>,
-        }
+#[cfg(not(query_logger))]
+#[derive(diesel::MultiConnection)]
+pub enum DbConnInner {
+    #[cfg(sqlite)]
+    Sqlite(diesel::sqlite::SqliteConnection),
```
weiznich: #If you use the provided establish method it is rather important to put the SQLite variant as last entry. The implementation of the establish is rather dumb, it just tries to connect to each variant in order and as SQLite accepts almost anything as database url that might cause issues connecting to the other backends. The other and likely better solution is to check the db url and construct the correct enum variant directly b calling establish on the inner connection. That should give you also better runtime error messages if something goes wrong.

src/db/mod.rs
```diff
@@ -232,144 +227,54 @@ impl DbConnType {
 
     pub fn default_init_stmts(&self) -> String {
         match self {
-            Self::sqlite => "PRAGMA busy_timeout = 5000; PRAGMA synchronous = NORMAL;".to_string(),
-            Self::mysql => String::new(),
-            Self::postgresql => String::new(),
+            #[cfg(sqlite)]
+            Self::Sqlite => "PRAGMA busy_timeout = 5000; PRAGMA synchronous = NORMAL;".to_string(),
```
weiznich: It might be meaningful to use the other settings from [here](https://docs.diesel.rs/2.2.x/diesel/prelude/struct.SqliteConnection.html#concurrency) as well?

src/db/mod.rs
```diff
@@ -232,144 +227,54 @@ impl DbConnType {
...
 
 // Reexport the models, needs to be after the macros are defined so it can access them
 pub mod models;
 
 /// Creates a back-up of the sqlite database
 /// MySQL/MariaDB and PostgreSQL are not supported.
-pub async fn backup_database(conn: &mut DbConn) -> Result<String, Error> {
-    db_run! {@raw conn:
+pub async fn backup_database(conn: &DbConn) -> Result<String, Error> {
+    db_run! {conn:
```
weiznich: For the backup implementation [this function](https://docs.diesel.rs/2.2.x/diesel/prelude/struct.SqliteConnection.html#method.serialize_database_to_buffer) might be more appropriated.

weiznich (commented):
I had a quick look at the core database code and wrote down a few things I noticed while reading that code. Feel free to ignore these notes if you don't feel like they are helpful.

BlackDex:
> I had a quick look at the core database code and wrote down a few things I noticed while reading that code. Feel free to ignore these notes if you don't feel like they are helpful.

Thanks! Ill take a look at them, and adjust where needed.

src/db/mod.rs
BlackDex: I did noticed yesterday during a test that i was missing the query parameters. But i haven't verified if this was because of the diesel_logger crate or something else.

I was looking into this though, i tried to also extract the file and line of where the call was placed, but that is a bit difficult hehe.

BlackDex: I implemented by own version now indeed. Only the binds are not shown currently. Have not yet looked into that specific issue.

src/db/mod.rs
BlackDex: I reordered all these throughout the whole code base, that is also a bit more consistent.

src/db/mod.rs
BlackDex: I have refactored this. I now also connect to the database directly without relying on a previous `conn` and i connect in read-only mode also. In the future we might be able to also let the users adjust the location of the backups, but for now ill leave it as the previous way-of-working.

This also fixed an issue which could cause the migrations to be run a second time, even though they should have been run already though.

src/db/query_logger.rs
```diff
@@ -0,0 +1,57 @@
+use dashmap::DashMap;
+use diesel::connection::{Instrumentation, InstrumentationEvent};
+use std::{
+    sync::{Arc, LazyLock},
+    thread,
+    time::Instant,
+};
+
+pub static QUERY_PERF_TRACKER: LazyLock<Arc<DashMap<(thread::ThreadId, String), Instant>>> =
```
dani-garcia: This might be easier and more performant with a thread_local, I think? https://doc.rust-lang.org/std/macro.thread_local.html

```rust
thread_local! {
    static QUERY_PERF_TRACKER: RefCell<HashMap<String, Instant>> = RefCell::new(HashMap::new());
}

QUERY_PERF_TRACKER.with_borrow_mut(|map| {
    map.insert(query_string, start);
});

QUERY_PERF_TRACKER.with_borrow_mut(|map| {
    if let Some(start) = map.remove(&query_string) {
        ...
    }
});

```

src/db/mod.rs
```diff
@@ -232,175 +237,102 @@ impl DbConnType {
...
-                        #[allow(unused)] use $crate::db::[<__ $db _schema>]::{self as schema, *};
-                        #[allow(unused)] use [<__ $db _model>]::*;
-                    }
-
-                    tokio::task::block_in_place(move || { $body }) // Run blocking can't be used due to the 'static limitation, use block_in_place instead
-                },
-            )+)+
-        }
+#[macro_export]
+macro_rules! db_run {
```
dani-garcia, line 322: I think we can remove some of the macro code now, if we wrap the code in a function like this:
```rust
impl DbConn {
    pub async fn run<T: Send>(&self, f: impl FnOnce(&mut DbConnInner) -> T + Send) -> T {
        let conn = Arc::clone(&self.conn);
        let mut conn = conn.lock_owned().await;
        let conn = conn.as_mut().expect("internal invariant broken: self.conn is Some");
        // Run blocking can't be used due to the 'static limitation, use block_in_place instead
        tokio::task::block_in_place(move || f(conn))
    }
}

#[macro_export]
macro_rules! db_run {
    ( $conn:ident: $body:block ) => {{
       $conn.run(move |$conn| $body).await
    }};

    ( $conn:ident: $( $($db:ident),+ $body:block )+ ) => {{
        $conn.run(move |$conn| {
            match $conn {
                $($(
                #[cfg($db)]
                pastey::paste!(&mut $crate::db::DbConnInner::[<$db:camel>](ref mut $conn)) => {
                    $body
                },
            )+)+}
        }).await
    }};
}
```

Then the basic invocation of the macro is basically just a single function, so we should be able to inline all the uses of that macro entirely. Though that should go as a separate PR: 
```rust
    pub async fn delete(state: &OIDCState, conn: &DbConn) -> EmptyResult {
        conn.run(|conn| {
            diesel::delete(sso_nonce::table.filter(sso_nonce::state.eq(state)))
                .execute(conn)
                .map_res("Error deleting SSO nonce")
        }).await
    }
```

src/db/query_logger.rs
BlackDex: Ah, i didn't knew this one, lets see if that works too!

BlackDex: Seems to work great.

src/db/mod.rs
BlackDex: As discussed via Matrix, we will leave this as-is for now. It bloats the binary because of monomorphization.
Well look into other ways to try and fully remove the need for the `db_run!` macro.

BlackDex: Ok, did some more research (And in this case also used an LLM to help), and there is a way to do this and not bloat the binary. It does create a lot of copies for the `conn.run()` function, but overall, less copies and less llvm-lines. It also looks like it lowered the compile time a bit too, as also lowered the binary size.

The only thing left would then still be updating all those functions, which can be done very easily by using a tool called [Comby](https://comby.dev/) and running it like this in the `src/` directory.

```bash
comby 'db_run! { conn: {:[body]}}' 'conn.run(move |conn| {:[body]}).await' .rs -i
```

I still think that it might be best to do that in an other PR and first merge a few other pending PR's which might affect it.

src/db/query_logger.rs
```diff
@@ -0,0 +1,60 @@
...
+
+thread_local! {
+    static QUERY_PERF_TRACKER: RefCell<HashMap<String, Instant>> = RefCell::new(HashMap::new());
+}
+
+pub fn simple_logger() -> Option<Box<dyn Instrumentation>> {
+    Some(Box::new(|event: InstrumentationEvent<'_>| match event {
+        // TODO: Figure out where the invalid connection errors are coming from
+        // There seem to be some invalid errors when connecting to a SQLite database
+        // Until the cause of this is found and resolved, disable the Connection logging
```
weiznich: That's likely caused by `#[derive(MultiConnection)]` and how it implements it's establish function. 

The establish function literally does something like:

```
  fn establish(database_url: &str) -> diesel::ConnectionResult<Self> {
                if let Ok(conn) = PgConnection::establish(database_url) {
                    return Ok(Self::Pg(conn));
                }
                if let Ok(conn) = diesel::SqliteConnection::establish(database_url) {
                    return Ok(Self::Sqlite(conn));
                }
                Err(
                    diesel::ConnectionError::BadConnection(
                        "Invalid connection url for multiconnection".into(),
                    ),
                )
            }
```

(https://docs.diesel.rs/2.3.x/diesel/derive.MultiConnection.html#expanded-code-1, search for "establish")

You can circumvent that by manually constructing the connection enum value from the connection url. So essentially something like that:

```
let conn = if url.starts_with("postgres") {
     DbConn::PgConnection(PgConnection::establish(url)?)
} else if url.starts_with("mysql") {
     DbConn::Mysql(MysqlConnection::establish(url)?)
} else {
     DbConn::SqliteConnection(SqliteConnection::establish(url)?)
}
```

This should also give much better error messages to the user if the connection url is incorrect.

For the pool that likely requires a custom impl of [`r2d2::ManageConnection`](https://docs.rs/r2d2/latest/r2d2/trait.ManageConnection.html) that wraps the custom establish logic and otherwise forwards to the diesel connection manager.

src/db/mod.rs
```diff
@@ -51,170 +44,182 @@ where
...
+}
 
-        impl Drop for DbConn {
-            fn drop(&mut self) {
-                let conn = Arc::clone(&self.conn);
-                let permit = self.permit.take();
+#[derive(Clone)]
+pub struct DbPool {
+    // This is an 'Option' so that we can drop the pool in a 'spawn_blocking'.
+    pool: Option<Pool<ConnectionManager<DbConnInner>>>,
```
weiznich: I would suggest using an async connection pool (like `deadpool-diesel`) in a next step. That should make the logic below much easier.

src/db/mod.rs
weiznich: I think that's a good way forward and that would also be what I would have suggested. I would expect that there are only a handful locations that really actually need to perform different actions for different database systems. I personally just matched the enum for those in our own code base. That works reasonably well.

src/db/query_logger.rs
BlackDex: I have implemented this now, and it indeed resolved the issue. Thanks for the pointers.

src/db/mod.rs
BlackDex: Yes, That is something to look into after this has been merged and probably after a new release with this PR.

src/db/query_logger.rs
weiznich: I still wonder if we can solve this in a better way in the generated `establish` configuration. I want to avoid looking at the type name there as this derive also needs to support third party connection impls, so it's just not possible to say: For `PgConnection` your Url needs to start with `postgres`. A possible solution might be to add a constant with the prefix to the internal [`MultiConnectionHelper`](https://docs.diesel.rs/2.3.x/diesel/connection/trait.MultiConnectionHelper.html) and utilize that in the generated code. Maybe that would work :thinking:

PR #6321, opened 2025-09-23
[Playwright] Improvements around node

Timshel:
In the context of the latest supply chain attack on npm (Shai-Hulud).

- Use fixed dependencies version (since it's not deployed using a slightly out of date version is way less of a risk than installing a compromised minor update).
- Replace `npm install` with `npm ci --ignore-scripts` to use the lock file (in the doc and docker image).

Additionally, updated the playwright docker image to `trixie` and backported the Keycloak update from the role PR.

### DIFF
--- a/playwright/compose/keycloak/Dockerfile
+++ b/playwright/compose/keycloak/Dockerfile
@@ -1,40 +1,19 @@
-FROM docker.io/library/debian:bookworm-slim as build
+FROM docker.io/library/debian:trixie-slim
 
-ENV DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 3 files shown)


BlackDex:
@dani-garcia, I can't merge this one, should be fine as no code is changed  here.

dani-garcia (approved):

dani-garcia:
> @dani-garcia, I can't merge this one, should be fine as no code is changed here.

Yeah I think we should remove the path limitations from the build workflow and let it run always, otherwise PRs get blocked on them and need admin bypass. It's pretty fast anyway

PR #6339, opened 2025-10-02
Optimizations and build speedup

BlackDex:
With this commit I have changed several components to be more efficient.
This can be less llvm-lines generated or less `clone()` calls.

 ### Config
- Re-ordered the `make_config` macro to be more efficient
- Created a custom Deserializer for `ConfigBuilder` less code and more efficient
- Use struct's for the `prepare_json` function instead of generating a custom JSON object. This generates less code and is more efficient.
- Updated the `get_support_string` function to handle the masking differently. This generates less code and also was able to remove some sub-macro-calls

 ### Error
- Added an extra new call to prevent duplicate Strings in generated macro code. This generated less llvm-lines and seems to be more efficient.
- Created a custom Serializer for `ApiError` and `CompactApiError` This makes that struct smaller in size, so better for memory, but also less llvm-lines.

 ### General
- Removed `once_lock` and replace it all with Rust's std LazyLock
- Added and fixed some Clippy lints which reduced `clone()` calls for example.
- Updated build profiles for more efficiency Also added a new profile specifically for CI, which should decrease the build check
- Updated several GitHub Workflows for better security and use the new `ci` build profile
- Updated to Rust v1.90.0 which uses a new linker `rust-lld` which should help in faster building
- Updated the Cargo.toml for all crates to better use the `workspace` variables
- Added a `typos` Workflow and Pre-Commit, which should help in detecting spell error's. Also fixed a few found by it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,9 +31,6 @@ on:
 jobs:
   build:
     name: Build and Test ${{ matrix.channel }}
-    permissions:
-      actions: write
-      contents: read
     runs-on: ubuntu-24.04
     timeout-minutes: 120
     # Make warnings errors, this is to prevent warnings slipping through.
@@ -69,9 +66,9 @@ jobs:
           CHANNEL: ${{ matrix.channel }}
         run: |
           if [[ "${CHANNEL}" == 'rust-toolchain' ]]; then
-            RUST_TOOLCHAIN="$(grep -oP 'channel.*"(\K.*?)(?=")' rust-toolchain.toml)"
+            RUST_TOOLCHAIN="$(grep -m1 -oP 'channel.*"(\K.*?)(?=")' rust-toolchain.toml)"
           elif [[ "${CHANNEL}" == 'msrv' ]]; then
-            RUST_TOOLCHAIN="$(grep -oP 'rust-version.*"(\K.*?)(?=")' Cargo.toml)"
+            RUST_TOOLCHAIN="$(grep -m1 -oP 'rust-version\s.*"(\K.*?)(?=")' Cargo.toml)"
           else
             RUST_TOOLCHAIN="${CHANNEL}"
           fi
@@ -81,7 +78,7 @@ jobs:
 
       # Only install the clippy and rustfmt components on the default rust-toolchain
       - name: "Install rust-toolchain version"
-        uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9 # master @ Aug 23, 2025, 3:20 AM GMT+2
+        uses: dtolnay/rust-toolchain@6d653acede28d24f02e3cd41383119e8b1b35921 # master @ Sep 16, 2025, 8:37 PM GMT+2
         if: ${{ matrix.channel == 'rust-toolchain' }}
         with:
           toolchain: "${{steps.toolchain.outputs.RUST_TOOLCHAIN}}"
@@ -91,7 +88,7 @@ jobs:
 
       # Install the any other channel to be used for which we do not execute clippy and rustfmt
       - name: "Install MSRV version"
-        uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9 # master @ Aug 23, 2025, 3:20 AM GMT+2
+        uses: dtolnay/rust-toolchain@6d653acede28d24f02e3cd41383119e8b1b35921 # master @ Sep 16, 2025, 8:37 PM GMT+2
         if: ${{ matrix.channel != 'rust-toolchain' }}
         with:
           toolchain: "${{steps.toolchain.outputs.RUST_TOOLCHAIN}}"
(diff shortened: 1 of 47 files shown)


BlackDex:
Also, together with #6279 this should help lower the compile time even more.

tessus:
Wow, this is a nice PR. I looked at it out of curiosity because seeing how something is done always helps to learn more about Rust.

Not being an expert in Rust made it a bit hard for me to put all the pieces together. This is not a critique but more a genuine question. How did you write this code? e.g. I usually create smaller commits, each of them as contained as possible. Almost all the items in your lists above could have been their own commit. A PR can always be squahed, but TBH it would not be necessary in this case. I squash commits that are related e.g. a back and forth in a review process and its resulting (additional) commits to resolve outstanding issues.

I am not asking you to create multiple commits or change the way you code, I just want to understand whether you wrote all the changes at the same time.

BlackDex:
@tessus, well it started out as some experimenting with different options. And i did several options together etc... and then it got messy. I tried to see if i could split it up again, but that would have taken much time again. So that is why i merged all in one. I should have tried to stick with one item and commit it per item.

tessus:
@BlackDex Thanks, that explains it. 

I also know people who create multiple commits during development and then squash before creating a PR (weird, but ok).
And yes, creating multiple commits from one big mess is a huge hassle and takes a lot of time. In such a case it's not worth it and I wouldn't bother either, unless the project requires it.

.github/workflows/build.yml
```diff
@@ -31,9 +31,6 @@ on:
 jobs:
   build:
     name: Build and Test ${{ matrix.channel }}
-    permissions:
```
williamdes, line 34: How does it work without contents read?

.github/workflows/check-templates.yml
```diff
@@ -6,8 +6,6 @@ on: [ push, pull_request ]
 jobs:
   docker-templates:
     name: Validate docker templates
-    permissions:
-      contents: read
```
williamdes, line 10: Add contents read to the top of the file permissions?

dfunkt:
There might be some problems when using this PR in combination with the release workflow. I've had a trial build that took over 1 hour till it was cancelled by GitHub.
image

BlackDex:
> There might be some problems when using this PR in combination with the release workflow. I've had a trial build that took over 1 hour till it was cancelled by GitHub. image

Strange, nothing changed that should cause this i think, ill see i i can test it my self too.

.github/workflows/build.yml
BlackDex: It works because the default is read anyway.

image

.github/workflows/check-templates.yml
BlackDex: Same here

dfunkt:
Maybe the changes require more RAM for the build? The CI runners could be running out of memory during linking or something like that.

Just finished a build with `thin` LTO instead of `fat`, looks like this did the trick.
Next up I'm going to test 2 vs 3 parallel builds. (been using 3 until now for my builds)

BlackDex:
> Maybe the changes require more RAM for the build? The CI runners could be running out of memory during linking or something like that.
> 
> Just finished a build with `thin` LTO instead of `fat`, looks like this did the trick. Next up I'm going to test 2 vs 3 parallel builds. (been using 3 until now for my builds)

Ah, that's it then, we use 2 parallel builds because of the memory usage.
It's the cost of building the most optimized binary, it will take a bit longer, but it should have the most optimizations and smaller binary.

dfunkt:
Possibly, it just caught me off guard since 3 parallel builds worked fine so far. Perhaps these new optimizations are more memory intensive during the build process.

EDIT: Yeah, the workflow finished successfully with 2 parallel builds. Looks like I'll have to make some changes for my own builds.

BlackDex:
@dfunkt, ow i think i have made an error in the build profiles. I actually enabled debugging in the release profile.
That shouldn't have happened. I'm verifying right now if that is indeed the case.

Good spot :)

BlackDex:
So, one, it probably did used more memory, and two, no clue why i added that there, because that actually increased the whole compile-time haha. It should be faster now by almost a minute maybe a bit less.

Cargo.toml
```diff
@@ -1,19 +1,26 @@
...
-license = "AGPL-3.0-only"
-publish = false
 build = "build.rs"
+resolver = "2"
+repository.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+publish.workspace = true
+
```
dfunkt, line 23: Unnecessary empty line

Cargo.toml
```diff
@@ -209,23 +213,14 @@ reqsign = { version = "0.16.5", optional = true }
 strip = "debuginfo"
 lto = "fat"
 codegen-units = 1
+debug = false
```
dfunkt, line 214: Unnecessary empty line

Cargo.toml
BlackDex: done

Cargo.toml
BlackDex: done

dani-garcia (approved):

PR #6349, opened 2025-10-08
Add output dir option to backup command

lukaspieper:
Closes #6347, #5973

I added the CLI option as discussed in #6347. As I mentioned, I'm not familiar with Rust. I tested my changes to confirm they work, but I don't know anything about Rust best practices.

While implementing this, I noticed that there seem to be three ways to trigger the backup process: via the CLI, the USR1 signal or an API call. Bearing this in mind, I'm wondering whether the current approach is sensible, given that only the CLI will benefit from the change. Might it be more future-proof to add a config entry for the backup path instead of the CLI option to handle all three variants in the same way? For this reason, I am opening this as a draft to discuss this question.

Thanks in advance for spending time on my PR/feature request.

### DIFF
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -799,7 +799,7 @@ async fn delete_config(_token: AdminToken) -> EmptyResult {
 #[post("/config/backup_db", format = "application/json")]
 fn backup_db(_token: AdminToken) -> ApiResult<String> {
     if *CAN_BACKUP {
-        match backup_sqlite() {
+        match backup_sqlite(None) {
             Ok(f) => Ok(format!("Backup to '{f}' was successful")),
             Err(e) => err!(format!("Backup was unsuccessful {e}")),
         }
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -385,7 +385,7 @@ pub mod models;
 /// Creates a back-up of the sqlite database
 /// MySQL/MariaDB and PostgreSQL are not supported.
 #[cfg(sqlite)]
-pub fn backup_sqlite() -> Result<String, Error> {
+pub fn backup_sqlite(output_dir: Option<String>) -> Result<String, Error> {
     use diesel::Connection;
     use std::{fs::File, io::Write};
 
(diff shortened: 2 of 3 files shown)


BlackDex:
I'm not against this feature, but there are multiple PR's which might have an impact on this.

Also, while the flexibility of a custom directory is nice, i think a per-determined fixed folder would be better. This for example could also help with the S3 feature (not enabled by default) to be useful for these kind of backups.

So, there are probably some changes which need to be made after some large PR's are merged, and probably also some changes to be done to let it work a bit more seamless with the S3 feature.

lukaspieper:
Hi,

don't worry, as mentioned in the PR description, I already figured out that there are alternative approaches to solve this issue. I don't mind having a fixed path, assuming we can then easily create a mount from the host to the fixed directory inside the container.

If you would like me to update the PR, please let me know once the other PRs are complete and what the solution should look like.

Caio-Nogueira:
What's the status on this? I'm trying to automate backups and think that this would be a really nice feature. 

> I already figured out that there are alternative approaches to solve this issue.

What are these alternatives? Do you mean using sqlite directly to create the backup?
IIUC the backup files are always created on the parent folder of the actual production db, which isn't very flexible. The consequence of this for me is having to run the systemd unit for backups with permissions to write on that parent directory, containing the actual db.

BlackDex:
@lukaspieper, all the PR's are merged. And I think an adjustment can be made also to make use of the OpenDal S3 option for the backup directory. That way you could keep the storage of the database on a local/external volume and have the Backup sent to an S3 bucket for offsite/remote-storage.

lukaspieper:
@BlackDex, just wanted to let you know that I'm planning to rebase this in January and look into your suggestion, as I'm leaving for 39C3 tomorrow.

BlackDex:
That is fine.  Viel Spaß!

lukaspieper:
So I rebased the PR in a first step. I tried looking into OpenDal S3 but couldn't find much more information than #5626. Could you give me some hints, please? Does "the OpenDal S3 option for the backup directory" already exist?

My understanding of the control flow is that the SQLite backup file will be created locally and then pushed to S3. In order to push something into a bucket, I need a uri, don't I?

Issue #6316, opened 2025-09-23
Invited Users Not Redirected to SSO Login with SSO_ONLY Mode


kaeferpsd:
### Prerequisites

- [x] I have searched the existing **Closed _AND_ Open** [Issues](https://github.com/dani-garcia/vaultwarden/issues?q=is%3Aissue%20) **_AND_** [Discussions](https://github.com/dani-garcia/vaultwarden/discussions?discussions_q=)
- [x] I have searched and read the [documentation](https://github.com/dani-garcia/vaultwarden/wiki/)

### Your environment (Generated via diagnostics page)

* Vaultwarden version: v1.34.3-a2ad1dc7
* Web-vault version: v2025.8.0
* OS/Arch: linux/x86_64
* Running within a container: true (Base: Debian)
* Database type: PostgreSQL
* Database version: PostgreSQL 15.13 (Debian 15.13-1.pgdg120+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 12.2.0-14) 12.2.0, 64-bit
* Uses config.json: false
* Uses a reverse proxy: true
* IP Header check: true (X-Real-IP)
* Internet access: true
* Internet access via a proxy: false
* DNS Check: true
* Browser/Server Time Check: true
* Server/NTP Time Check: true
* Domain Configuration Check: true
* HTTPS Check: true
* Websocket Check: true
* HTTP Response Checks: true

### Config & Details (Generated via diagnostics page)

<details><summary>Show Config & Details</summary>

**Config:**

```json
{
  "_duo_akey": null,
  "_enable_duo": true,
  "_enable_email_2fa": true,
  "_enable_smtp": true,
  "_enable_yubico": true,
  "_icon_service_csp": "",
  "_icon_service_url": "",
  "_ip_header_enabled": true,
  "_max_note_size": 10000,
  "_smtp_img_src": "***:",
  "admin_ratelimit_max_burst": 3,
  "admin_ratelimit_seconds": 300,
  "admin_session_lifetime": 20,
  "admin_token": "***",
  "allowed_connect_src": "",
  "allowed_iframe_ancestors": "",
  "attachments_folder": "data/attachments",
  "auth_request_purge_schedule": "30 * * * * *",
  "authenticator_disable_time_drift": false,
  "data_folder": "data",
  "database_conn_init": "",
  "database_idle_timeout": 600,
  "database_max_conns": 10,
  "database_min_conns": 2,
  "database_timeout": 30,
[...]

PR #6364, opened 2025-10-15
Fix Org identifier

Timshel:
In https://github.com/dani-garcia/vaultwarden/pull/6263 changed the identifier to reference `org_uuid` but forgot to replace it in some places.


Timshel:
https://github.com/dani-garcia/vaultwarden/issues/6316 made me think that the invitation should redirect to SSO Login as soon as `SSO_ENABLED=true` and not require `SSO_ONLY`.

PR #6380, opened 2025-10-20
Playwright against abitrary web-vault

Timshel:
The main goal of this PR is to modify the build script to allow pointing directly to a version of [vw_web_build]([https://github.com/vaultwarden/vw_web_builds) instead of a [bw_web_builds](https://github.com/dani-garcia/bw_web_builds).
This should allow to easily test unreleased web vault.

Added some other improvements and started to fix the tests for the `10.0` client but between the extension spam and https://github.com/dani-garcia/vaultwarden/pull/6364 they will remain broken.


BlackDex (approved):

dani-garcia:
LGTM, but this now has some conflicts with #6321

PR #6388, opened 2025-10-22
Add XoAuth2 support 

hnolde:
This pull request adds comprehensive support for SMTP OAuth2 (XOAUTH2) authentication, enabling the system to send emails via providers that require OAuth2 (like Gmail). It introduces new configuration options, admin panel endpoints, and user interface controls for managing OAuth2 tokens and authorization flows. The implementation includes secure handling of OAuth2 state for CSRF protection, token caching, and robust validation of configuration parameters.

**SMTP OAuth2 (XOAUTH2) Support**

- Added new configuration options for OAuth2, including `smtp_oauth2_client_id`, `smtp_oauth2_client_secret`, `smtp_oauth2_auth_url`, `smtp_oauth2_token_url`, `smtp_oauth2_refresh_token`, and `smtp_oauth2_scopes`. These allow the system to be configured for OAuth2-based SMTP authentication.
- Implemented validation logic to ensure all required OAuth2 parameters are set and correctly formatted when XOAUTH2 is enabled, including checks for valid URLs and required fields. [[1]](diffhunk://#diff-cba64c21ab992eaad29fce147a08f4560a4769bc14682b8a96081a5fd02dbecdR1051-R1056) [[2]](diffhunk://#diff-cba64c21ab992eaad29fce147a08f4560a4769bc14682b8a96081a5fd02dbecdR1148-R1188)

**Backend OAuth2 Flow and Token Management**

- Introduced OAuth2 authorization endpoints (`/admin/oauth2/authorize` and `/admin/oauth2/callback`) and a token refresh endpoint (`/admin/test/oauth2`) in the admin API, including secure state storage for CSRF protection during the OAuth2 flow. [[1]](diffhunk://#diff-48e73dac59e5044731073f602e5331265f839bd0ea9169e2a5368d16e2dbcebeR2-R10) [[2]](diffhunk://#diff-48e73dac59e5044731073f602e5331265f839bd0ea9169e2a5368d16e2dbcebeR65-R67) [[3]](diffhunk://#diff-48e73dac59e5044731073f602e5331265f839bd0ea9169e2a5368d16e2dbcebeR99-R101) [[4]](diffhunk://#diff-48e73dac59e5044731073f602e5331265f839bd0ea9169e2a5368d16e2dbcebeR343-R494)
- Added logic to exchange authorization codes for refresh tokens, persist them in configuration, and refresh access tokens as needed, with token caching to minimize unnecessary requests. [[1]](diffhunk://#diff-02d3b205a438920e7c90dc9a121d5f81f2f04940aea6f9c2b48c7adea12305dbR27-R112) [[2]](diffhunk://#diff-02d3b205a438920e7c90dc9a121d5f81f2f04940aea6f9c2b48c7adea12305dbL32-R121)

**SMTP Transport and Authentication Logic**

- Modified SMTP transport logic to support OAuth2: when configured, the system uses the OAuth2 access token as the SMTP password, falling back to traditional authentication if necessary, and preferring OAuth2 when both are present. [[1]](diffhunk://#diff-02d3b205a438920e7c90dc9a121d5f81f2f04940aea6f9c2b48c7adea12305dbL60-R185) [[2]](diffhunk://#diff-02d3b205a438920e7c90dc9a121d5f81f2f04940aea6f9c2b48c7adea12305dbL674-R798)

**Admin UI Enhancements**

- Added UI controls and JavaScript handlers to the admin settings page for initiating OAuth2 authorization and refreshing tokens, with checks to prevent actions if unsaved config changes are present. [[1]](diffhunk://#diff-afcf517ecd8a8db9f343dcbad52e2fb871838bd3f27a1e766881dc22ec426e00R29-R55) [[2]](diffhunk://#diff-afcf517ecd8a8db9f343dcbad52e2fb871838bd3f27a1e766881dc22ec426e00R255-R262)

**General Improvements**

- Made `update_config_partial` public to allow saving OAuth2 tokens from the OAuth2 callback handler.
- Added a convenience `From<&str>` implementation for the `Error` type to simplify error handling.

### DIFF
--- a/src/api/core/accounts.rs
+++ b/src/api/core/accounts.rs
@@ -1704,6 +1704,6 @@ pub async fn purge_auth_requests(pool: DbPool) {
     if let Ok(conn) = pool.get().await {
         AuthRequest::purge_expired_auth_requests(&conn).await;
     } else {
-        error!("Failed to get DB connection while purging trashed ciphers")
+        error!("Failed to get DB connection while purging auth requests")
     }
 }
--- a/src/api/identity.rs
+++ b/src/api/identity.rs
@@ -482,14 +482,18 @@ async fn authenticated_response(
         Value::Null
     };
 
-    let account_keys = json!({
-        "publicKeyEncryptionKeyPair": {
-            "wrappedPrivateKey": user.private_key,
-            "publicKey": user.public_key,
-            "Object": "publicKeyEncryptionKeyPair"
-        },
-        "Object": "privateKeys"
-    });
+    let account_keys = if user.private_key.is_some() {
+        json!({
+            "publicKeyEncryptionKeyPair": {
+                "wrappedPrivateKey": user.private_key,
+                "publicKey": user.public_key,
+                "Object": "publicKeyEncryptionKeyPair"
+            },
+            "Object": "privateKeys"
+        })
+    } else {
+        Value::Null
+    };
 
     let mut result = json!({
         "access_token": auth_tokens.access_token(),
--- a/src/static/templates/admin/diagnostics.hbs
+++ b/src/static/templates/admin/diagnostics.hbs
@@ -8,7 +8,7 @@
                 <dl class="row">
                     <dt class="col-sm-5">Server Installed
                         <span class="badge bg-success d-none abbr-badge" id="server-success" title="Latest version is installed.">Ok</span>
-                        <span class="badge bg-warning text-dark d-none abbr-badge" id="server-warning" title="There seems to be an update available.">Update</span>
+                        <span class="badge bg-warning text-dark d-none abbr-badge" id="server-warning" title="An update is available.">Update</span>
                         <span class="badge bg-info text-dark d-none abbr-badge" id="server-branch" title="This is a branched version.">Branched</span>
                     </dt>
                     <dd class="col-sm-7">
@@ -23,8 +23,8 @@
                     {{#if page_data.web_vault_enabled}}
                     <dt class="col-sm-5">Web Installed
                         <span class="badge bg-success d-none abbr-badge" id="web-success" title="Latest version is installed.">Ok</span>
-                        <span class="badge bg-warning text-dark d-none abbr-badge" id="web-warning" title="There seems to be an update available.">Update</span>
-                        <span class="badge bg-info text-dark d-none abbr-badge" id="web-prerelease" title="You seem to be using a pre-release version.">Pre-Release</span>
+                        <span class="badge bg-warning text-dark d-none abbr-badge" id="web-warning" title="An update is available.">Update</span>
+                        <span class="badge bg-info text-dark d-none abbr-badge" id="web-prerelease" title="You are using a pre-release version.">Pre-Release</span>
                     </dt>
                     <dd class="col-sm-7">
                         <span id="web-installed">{{page_data.active_web_release}}</span>


src/api/admin.rs
```diff
@@ -329,6 +340,151 @@ async fn test_smtp(data: Json<InviteData>, _token: AdminToken) -> EmptyResult {
...
+    <p class="success">The refresh token has been saved to your configuration.</p>
+    <p>You can now close this window and return to the admin panel.</p>
+    <p><a href="{}">Return to Admin Settings</a></p>
+</body>
+</html>"#,
+        admin_url()
+    );
+
+    Ok(Html(success_html))
+}
```
BlackDex: This should go into a handlebars template instead of embedding it like this.

src/api/admin.rs
```diff
@@ -329,6 +340,151 @@ async fn test_smtp(data: Json<InviteData>, _token: AdminToken) -> EmptyResult {
...
+
+    // Build authorization URL
+    let auth_url = format!(
+        "{}?client_id={}&redirect_uri={}&response_type=code&scope={}&state={}&access_type=offline&prompt=consent",
+        auth_url,
+        percent_encode(client_id.as_bytes(), NON_ALPHANUMERIC),
+        percent_encode(redirect_uri.as_bytes(), NON_ALPHANUMERIC),
+        percent_encode(scopes.as_bytes(), NON_ALPHANUMERIC),
+        percent_encode(state.as_bytes(), NON_ALPHANUMERIC)
+    );
```
BlackDex: It's probably safer to use the `url` crate for this and append/add each parameter. It will encode this and prevent invalid URL's

src/api/admin.rs
```diff
@@ -329,6 +340,151 @@ async fn test_smtp(data: Json<InviteData>, _token: AdminToken) -> EmptyResult {
...
+        ("client_secret", &client_secret),
+    ];
+
+    let client = reqwest::Client::new();
+    let response = client
+        .post(&token_url)
+        .form(&form_params)
+        .send()
+        .await
+        .map_err(|e| Error::new("OAuth2 Token Exchange Error", e.to_string()))?;
```
BlackDex: Please use the `http_client` mod which is a wrapper around reqwest with some basic settings and DNS safety.

src/mail.rs
```diff
@@ -21,6 +24,87 @@ use crate::{
...
+    let token_url = CONFIG.smtp_oauth2_token_url().ok_or("OAuth2 Token URL not configured")?;
+
+    let form_params = [
+        ("grant_type", "refresh_token"),
+        ("refresh_token", &refresh_token),
+        ("client_id", &client_id),
+        ("client_secret", &client_secret),
+    ];
+
+    let client = reqwest::Client::new();
```
BlackDex: Same here.
Please use the `http_client` mod which is a wrapper around reqwest with some basic settings and DNS safety.

BlackDex:
Sorry, didn't had much time to check and verify this PR yet in detail. I still have it on my list.

src/api/admin.rs
```diff
@@ -341,6 +352,139 @@ async fn test_smtp(data: Json<InviteData>, _token: AdminToken) -> EmptyResult {
...
+    // Extract refresh_token from response
+    let refresh_token =
+        token_response.get("refresh_token").and_then(|v| v.as_str()).ok_or("No refresh_token in response")?;
+
+    // Save refresh_token to configuration
+    let config_builder: ConfigBuilder = serde_json::from_value(json!({
+        "smtp_oauth2_refresh_token": refresh_token
+    }))
+    .map_err(|e| Error::new("ConfigBuilder serialization error", e.to_string()))?;
+    CONFIG.update_config_partial(config_builder).await?;
```
stefan0xC: I'm not really a fan of updating the config like this. Personally I think this should be done more dynamically and stored in the database and not as a config.

src/api/admin.rs
```diff
@@ -341,6 +352,139 @@ async fn test_smtp(data: Json<InviteData>, _token: AdminToken) -> EmptyResult {
     }
 }
 
+#[post("/test/oauth2")]
```
BlackDex: It might be better to rename this endpoint to `/oauth2/test` maybe? That matches more the other endpoints, and might make it possible for users to maybe add an exception for `/admin/oauth2` for some authentication tools.

src/config.rs
```diff
@@ -887,6 +887,18 @@ make_config! {
...
+        /// SMTP OAuth2 Client Secret |> OAuth2 Client Secret for XOAUTH2 authentication
+        smtp_oauth2_client_secret:     Pass,   true,   option;
+        /// SMTP OAuth2 Authorization URL |> OAuth2 Authorization Server URL
+        smtp_oauth2_auth_url:          String, true,   option;
+        /// SMTP OAuth2 Token URL |> OAuth2 Token Server URL for refreshing access tokens
+        smtp_oauth2_token_url:         String, true,   option;
+        /// SMTP OAuth2 Refresh Token |> OAuth2 Refresh Token for obtaining new access tokens
+        smtp_oauth2_refresh_token:     Pass,   true,   option;
+        /// SMTP OAuth2 Scopes |> Comma-separated list of OAuth2 scopes
+        smtp_oauth2_scopes:            String, true,   def,     "https://mail.google.com/".to_string();
```
BlackDex: I would suggest to leave this blank/empty instead of using a random predefined FQDN.
The main reason is that users might think it will automatically use Google, or if setting it to Microsoft something from Microsoft of course. It is fine to add an example in the comments though.

And then probably also check and validate the value in a later stage of course.

BlackDex: All these options including some more documentation should be added to the `.env.template` file too.

src/api/admin.rs
```diff
@@ -341,6 +352,139 @@ async fn test_smtp(data: Json<InviteData>, _token: AdminToken) -> EmptyResult {
...
+    let response = make_http_request(Method::POST, &token_url)?
+        .form(&form_params)
+        .send()
+        .await
+        .map_err(|e| Error::new("OAuth2 Token Exchange Error", e.to_string()))?;
+
+    if !response.status().is_success() {
+        let status = response.status();
+        let body = response.text().await.unwrap_or_else(|_| String::from("Unable to read response body"));
+        return Err(Error::new("OAuth2 Token Exchange Failed", format!("HTTP {}: {}", status, body)));
```
BlackDex: I think it's better to convert this to be using the `err!` macro as is done through the rest of the code.
So, this should work, and will also output an log message.

```rust
err!("OAuth2 Token Exchange Failed", format!("HTTP {}: {}", status, body))
```

BlackDex: The same goes btw for all other locations.

BlackDex:
Ok, this looks better indeed. I do wonder regarding the refresh token. That currently seems to be stored in de config and than saving it. That does implicate that everybody who is using this feature will generate a `config.json`.
That is however something i would like to avoid. My self for example, I want all my config to be done via environment variables, and I think other users would want that too.

If the refresh token is the only item which need to be stored/changed, and the other values are mostly static, then i would suggest to store this in the database using a new table like `xoauth2` or something like that. Which then can be used to store these details in the database, and have it updated via the database.

Also, I think a few of the new settings should be added to the `PRIVACY_CONFIG` located here: https://github.com/dani-garcia/vaultwarden/blob/0ab7784b06dc57b616c060ac8906871c65568900/src/config.rs#L414-L433

To prevent accidental exposure of secure/secret values when someone generates a support string.
`Pass` values will never be exposed, but some `String` items probably also contain sensitive values I think.

hnolde:
I have moved the storage of the SMTP OAuth2 refresh token from the configuration to the database. This avoids the automatic generation of config.json when using OAuth2 for SMTP, which was a concern for users who prefer environment-based configuration.

Key changes:

- Database Migration: Created a new xoauth2 table for SQLite, MySQL, and PostgreSQL.
- XOAuth2 Model: Implemented a new database model XOAuth2 to handle storage and retrieval of OAuth2 tokens.
- Global Database Pool: Introduced a global DB_POOL in src/db/mod.rs (initialized in src/main.rs) to allow the mail module to access the database without needing a connection passed through every function call.
- Updated Mail Logic: The mail module now prioritizes the refresh token from the database, falling back to the configuration (environment variables) if not found. It also automatically updates the database with new refresh tokens received during the refresh process.
- Admin API Update: The OAuth2 callback in src/api/admin.rs now saves the refresh token directly to the database instead of updating the configuration.
- Privacy Masking: Added sensitive OAuth2 configuration items (smtp_oauth2_client_id, smtp_oauth2_auth_url, smtp_oauth2_token_url) to PRIVACY_CONFIG in src/config.rs to ensure they are masked in support reports. Other sensitive values like secrets and tokens were already masked as Pass types.

src/mail.rs
```diff
@@ -16,11 +19,105 @@ use crate::{
...
+    static TOKEN_CACHE: LazyLock<RwLock<Option<OAuth2Token>>> = LazyLock::new(|| RwLock::new(None));
+
+    let cached_token = TOKEN_CACHE.read().unwrap().clone();
+
+    if let Some(token) = cached_token {
+        // Check if token is still valid (with 5 min buffer)
+        if let Some(expires_at) = token.expires_at {
+            let now = SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs();
+            if now + 300 < expires_at {
+                return Ok(token);
```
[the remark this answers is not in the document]
stefan0xC: Maybe we should use `chrono` to get the timestamp? Like we do for example in src/auth.rs:
https://github.com/dani-garcia/vaultwarden/blob/57bdab15504ff874f0ad6cb93f03292a70e4b365/src/auth.rs#L406-L409

src/config.rs
```diff
@@ -1244,6 +1263,44 @@ fn validate_config(cfg: &ConfigItems) -> Result<(), Error> {
...
+        if let Some(ref auth_url) = cfg.smtp_oauth2_auth_url {
+            if !auth_url.starts_with("http://") && !auth_url.starts_with("https://") {
+                err!("`SMTP_OAUTH2_AUTH_URL` must be a valid URL starting with http:// or https://");
+            }
+        }
+
+        // Validate that token URL is a valid URL
+        if let Some(ref token_url) = cfg.smtp_oauth2_token_url {
+            if !token_url.starts_with("http://") && !token_url.starts_with("https://") {
+                err!("`SMTP_OAUTH2_TOKEN_URL` must be a valid URL starting with http:// or https://");
```
stefan0xC: I think the recommended practice would be to use a secure connection. Any reason we might not want to require that?

hnolde: Changed it to require https.

src/mail.rs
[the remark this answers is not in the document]
hnolde: Changed to chrono.

src/api/admin.rs
```diff
@@ -341,6 +352,145 @@ async fn test_smtp(data: Json<InviteData>, _token: AdminToken) -> EmptyResult {
...
+    let auth_url = CONFIG.smtp_oauth2_auth_url().ok_or("OAuth2 Authorization URL not configured")?;
+    let scopes = CONFIG.smtp_oauth2_scopes();
+
+    // Generate a random state token for CSRF protection
+    let state = crate::crypto::encode_random_bytes::<32>(&BASE64URL_NOPAD);
+
+    // Store state with expiration (10 minutes from now)
+    // Multiple calls to Utc::now().timestamp() can return a negative value if the system time is set before the UNIX epoch.
+    // While extremely rare in practice, we handle this by using unwrap_or_default() to prevent huge values when casting to u64.
+    let now = u64::try_from(Utc::now().timestamp()).unwrap_or_default();
```
stefan0xC: Well, wouldn't it be better to expect an `i64` then? I don't think using a `u64` is required by any standard and this would make this a lot cleaner.

hnolde: makes sense, just did the changes.

septatrix:
I wonder if this might be better left to another service like https://github.com/simonrob/email-oauth2-proxy instead of pulling in this code here. XOAuth2 is generally used for personal email accounts and not something used by a service like vaultwarden

PR #6391, opened 2025-10-23
Switch to `fromJSON` digest extraction

dfunkt:
- removes the need for a separate digest extraction step

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -192,38 +192,35 @@ jobs:
             *.cache-from=${{ env.BAKE_CACHE_FROM }}
             *.cache-to=${{ env.BAKE_CACHE_TO }}
 
-      - name: Extract digest SHA
-        shell: bash
-        env:
-          BAKE_METADATA: ${{ steps.bake_vw.outputs.metadata }}
-          BASE_IMAGE: ${{ matrix.base_image }}
-        run: |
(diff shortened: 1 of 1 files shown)


BlackDex:
I noticed we probably do not have zizmor set to pedantic mode. Not sure if it will nag about not using an env in this case.

I'm not able to test this my self right now, probably able to do that tomorrow.

Other than that, looks ok i think.

dfunkt:
Doesn't seem to complain when using `matrix.base_image` directly, could change it to that if you want.
As far as pedantic mode goes, this pops up:
```
help[obfuscation]: obfuscated usage of GitHub Actions features
   --> vaultwarden/.github/workflows/release.yml:203:71
    |
203 | ...steps.bake_vw.outputs.metadata)[format('{0}-multi', env.BASE_IMAGE)]['containerimage.digest'] }}
    |                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ index expression is computed
    |
    = note: audit confidence → High
```

BlackDex:
Looking at this again, and this does kinda makes it harder to adjust the code when needed. Instead of one location extracting the SHA to be adjusted, it now needs to be adjusted at multiple locations. This makes it boilerplate code which i think shouldn't be needed.
One extra step to extract the SHA isn't that bad in my option.

dfunkt:
Fair enough, I might think about it a little longer to see if I can come up with a better solution.

PR #6433, opened 2025-11-03
Make JWT token refresh optional

Momi-V:
This implements a config option `disable_refresh_token_renewal`, that doesn't do refresh token renewal unless a full auth is performed.

**Goal:**
Only renew the JWT token that allows PIN unlock, bypassing 2FA, etc. if a "full authentication" is performed. This makes things like "require 2FA once every 30/90 days" possible

**Rationale:**
Currently it's only possible to require periodic reauthentication by setting clients to log-out instead of lock. This is inconvenient since it breaks PIN, offline functionality and requires full reauthentication for _every_ unlock. The current lock behavior on the other hand _only_ requires reauth if a client hasn't connected in 30/90 days, since the token is renewed even on a reduced login that used that same token. This means it effectively never expires, if the client just uses it's existing token more than once a month.

**Logic:**
If the new option is `false` the old logic is used (always renew the token). If a normal Password/SSO login is performed the old logic is used.
If the option is `true` the JWT isn't renewed in the `_refresh_login` code path, leaving it to expire and force a full auth after 30/90 days

**Code:**
Having `refresh_claims.sub.clone()` and `Some(&refresh_claims)` in the same call isn't amazing, but seems like the simplest way to implement this logic without a refactor to overload `AuthToken::new` into 2 separate, yet almost identical functions.

**Progress:**
Compiled and running, but not yet tested to actually behave as expected (currently ongoing).

Feedback welcome

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -1175,9 +1175,14 @@ impl AuthTokens {
     }
 
     // Create refresh_token and access_token with default validity
-    pub fn new(device: &Device, user: &User, sub: AuthMethod, client_id: Option<String>, existing_refresh_claims: Option<&RefreshJwtClaims>) -> Self {
+    pub fn new(
+        device: &Device,
+        user: &User,
+        sub: AuthMethod,
+        client_id: Option<String>,
+        existing_refresh_claims: Option<&RefreshJwtClaims>,
+    ) -> Self {
         let time_now = Utc::now();
-
         let access_claims = LoginJwtClaims::default(device, user, &sub, client_id);
 
         let validity = if device.is_mobile() {
@@ -1258,7 +1263,9 @@ pub async fn refresh_tokens(
         }
         AuthMethod::Sso => err!("SSO is now disabled, Login again using email and master password"),
         AuthMethod::Password if CONFIG.sso_enabled() && CONFIG.sso_only() => err!("SSO is now required, Login again"),
-        AuthMethod::Password => AuthTokens::new(&device, &user, refresh_claims.sub.clone(), client_id, Some(&refresh_claims)),
+        AuthMethod::Password => {
+            AuthTokens::new(&device, &user, refresh_claims.sub.clone(), client_id, Some(&refresh_claims))
+        }
         _ => err!("Invalid auth method, cannot refresh token"),
     };
 


src/config.rs
```diff
@@ -706,6 +706,10 @@ make_config! {
         /// Note that the checkbox would still be present, but ignored.
         disable_2fa_remember:   bool,   true,   def,    false;
 
+        /// Disable refresh token renewal |> If true, disables sliding window for refresh token expiry.
+        /// This only renews the token on a full login (Password (+2FA), SSO, etc.) forcing a full reauth every 30 days (90 for the native app)
```
BlackDex, line 710: Is there any documentation or link to upstream source where it mentions this 30 or 90 days?
Also, is fully disabling the only option here? Or can we for example let server-admin's configure a specific amount of days of this to be valid? (I have not deeply checked the workings, and I do not know this from the top of my head).

Momi-V: That's how I understood this: https://github.com/dani-garcia/vaultwarden/blob/ba5519167634ebe1e1f0fc10d610d10d1f405101/src/auth.rs#L30

(auth.rs line 30-33)

Based on my understanding of the logic flow it creates a new token (valid for 30 days in the Web UI and 90 days in the Apps (line 31)) every time a user authenticates (including authentication using the previous token).

I'm trying to find a way to (optionally) **stop** the existing token from being used to do a full authentication (including the refresh), because it's currently rendering 2FA useless as long as the user just opens the vault every couple weeks essentially **bypassing 2FA on every login but the initial one** (that's the current behavior and imo. not ideal).

The official Web UI has a Box with "don't ask again for 30 days" in the 2FA step, but with vaultwarden if that option is used it **never** asks for 2FA again (not after 30 days or 90) because the client just continuously reauthenticates using it's token.

The values of 30 and 90 days could probably be turned into config values too, but again based on my understanding they **only** apply if a client has not tried open the vault for 30/90 days **consecutively**. If it at any point unlocks (including less secure methods like just the password or even just the pin), for example after 20 days it will reauthenticate using it's existing token, triggering a refresh and with that refreshed one the next time 2FA could possibly be enforced is now 50 days after the initial login.

With the refresh disabled for the JWT based login flow that token would expire after it's 30/90 day validity no matter what and force a "proper" 2FA login at least that often.

BlackDex:
I'm looking at this, but I'm not really a fan of this way to be honest.
What currently happens is that we do not adhere to the 30 days remember checkbox, it will never ever re-ask to recheck the 2fa.

Adding yet another option to disable or bypass or whatever is not going to make security/auditing these options easier.

I think we should either adhere to the 30 days remember option and re-validate the 2fa as is what that checkbox is intended for, or if we do want to adjust this a bit, add an option to adjust the time-period of 30 days to something else, but never less then 1 day, and never more than 30.

I would need to check how this is done for devices/browser extensions, since i think that is where we made some kind of mistake maybe?

I'm also not sure how Bitwarden handles this, and I need to figure out how to test this without the need to wait 30 days haha.

BlackDex:
Ok, I have figured out how Bitwarden does this.
If the goal is to let the _Remember for 30 days_ actually work, as the text of Bitwarden mentions, then this PR is not the way to go.

We need to adjust our Remember Token code so that it is really only valid for 30 days, and not refresh indefinitely.
Since we already have a `DISABLE_2FA_REMEMBER`, there is no need for yet another config option, and i Think we should adhere to the same days as Bitwarden does to keep consistent.

I'm working on this now, so I'm not sure if you still need this PR then?

Momi-V:
The config option was just there to not change default behaviour. If you agree actually enforcing token expiration is the only correct option it's indeed not required.

And if you find a nicer way to implement this then even better. I'm not too familiar with rust or the wider codebase, so this was my attempt at "touching as littel as possible" while getting the option to work and not breaking existing logic. In this case by just making the creation routine return a cloned version of the original token (with the same expiration date) instead of a newly generated one.

But I'm sure someone more familiar with the codebase, without the constraint of preserving the current way it's behaving can do a much better job here with some cleaner and sompler logic.

BlackDex:
I do want to thank you for addressing this, as this is actually, in my eyes a bug, which a lot of people probably seem to find a nice feature. But to be fair, the 30 days might even be to long for some users also.

And we are talking about access to a system which holds all your passwords, and i think that a checkbox which mentions it will remember the token for 30 days, should actually remember the token for 30 days, and not indefinitely.

Momi-V:
This "behavior" came up while we were preparing for an audit and a colleague noticed they hadn't been asked for their 2FA in a while. Short term solution was DISABLE_2FA_REMEMBER, but having to login multiple times a day annoyed me enough to patch our instance. It appears to have worked, but I wasn't 100% confident (again, not familiar with the code base) and didn't have the time to rigorously test it, hence the draft status.

If this is just fixed for everyone by default that's great and we can just revert to the upstream (⁠^⁠^⁠)

BlackDex:
@Momi-V This is now fixed via #6929.

This does mean that once merged, all remember tokens will be invalidated in a newer version of Vaultwarden.
But that is probably more safe then just accepting the older format for a while.

PR #6459, opened 2025-11-11
Fix: admin theme emoji alignment

joepduin:
Emojis look different on each OS and device, so their size and spacing change.
A fixed inline-flex slot keeps every emoji the same width, so the text now always lines up.

Before:
before
After:
after

### DIFF
--- a/src/static/scripts/admin.css
+++ b/src/static/scripts/admin.css
@@ -65,3 +65,13 @@ img {
     flex: 0 0 1.75em;
     justify-content: center;
 }
+
+.theme-icon svg,
+.theme-icon-active svg {
+    width: 1.25em;
+    height: 1.25em;
+    min-width: 1.25em;
+    min-height: 1.25em;
+    display: block;
+    overflow: visible;
+}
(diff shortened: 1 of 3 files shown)


BlackDex:
Hmm, might need to fix that a bit more to not depend on the device/client on which icon's are shown and let them be more unified.

joepduin:
> Hmm, might need to fix that a bit more to not depend on the device/client on which icon's are shown and let them be more unified.

@BlackDex I'm going to look into it.

BlackDex:
Yea because those are to say it nicely, a bit ugly.

I Have this.
Screenshot_20251111-200132~2

joepduin:
@BlackDex What about this? I'm using SVG-sprite 
css-edit

BlackDex:
> @BlackDex What about this? I'm using SVG-sprite css-edit

That seems to look ok, not sure if sprites work that well on all systems, but last time i used that was ages ago, so it will probably be better now i think.

joepduin:
> 
> That seems to look ok, not sure if sprites work that well on all systems, but last time i used that was ages ago, so it will probably be better now i think.

Sprites do not work on every device, but the concept is widely supported across modern devices. Historically, dedicated hardware sprites were common in older gaming consoles and computers, but modern devices rely on the CPU to render sprites by drawing them into a frame buffer. Software-based sprites are used in nearly all modern applications, including web, mobile, and desktop apps, using technologies like CSS for web and specific rendering libraries for other platforms. 

**The primary compatibility issues were with very old browsers, such as older versions of Internet Explorer and Opera. These are rarely a concern in contemporary web development.**
 
Source: Google

BlackDex (approved):

PR #6472, opened 2025-11-16
Switch to multiple runners per arch

dfunkt:
- now uses arm64 native runners for faster compilation

Here's an example workflow run with all cache cleaned:
https://github.com/dfunkt/vaultwarden/actions/runs/19406766594
13m 10s for all Vaultwarden default target platforms (amd64, arm64, armv6 and armv7)

There's a lot of duplication in the second job that could be reduced via YAML anchors, but unfortunately zizmor has problems with parsing them atm (https://github.com/zizmorcore/zizmor/issues/1304)

And lastly, here's the reference documentation that I used for creating this: https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,6 +16,23 @@ concurrency:
   # Don't cancel other runs when creating a tag
   cancel-in-progress: ${{ github.ref_type == 'branch' }}
 
+defaults:
+  run:
+    shell: bash
+
+env:
+  # The *_REPO variables need to be configured as repository variables
+  # Append `/settings/variables/actions` to your repo url
+  # DOCKERHUB_REPO needs to be 'index.docker.io/<user>/<repo>'
+  # Check for Docker hub credentials in secrets
+  HAVE_DOCKERHUB_LOGIN: ${{ vars.DOCKERHUB_REPO != '' && secrets.DOCKERHUB_USERNAME != '' && secrets.DOCKERHUB_TOKEN != '' }}
+  # GHCR_REPO needs to be 'ghcr.io/<user>/<repo>'
+  # Check for Github credentials in secrets
+  HAVE_GHCR_LOGIN: ${{ vars.GHCR_REPO != '' && github.repository_owner != '' && secrets.GITHUB_TOKEN != '' }}
+  # QUAY_REPO needs to be 'quay.io/<user>/<repo>'
+  # Check for Quay.io credentials in secrets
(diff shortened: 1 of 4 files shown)


BlackDex:
It's this really that much faster? 
We only use QEMU for the final images, and it only is used for installing the packages needed for those images.

All the rest is cross-compiled from an x86_64 platform.

dfunkt:
Yeah, it's much faster than what's currently used. I've been using it for the past few weeks, it  certainly helps me a lot when I want to test something quickly.
Keep in mind that the result above is without cache, with cache the compile time goes even lower.

BlackDex:
Also, bake already runs multiple jobs, though we do limit those to 2 because of memory usage, and this would run 8 separate jobs and might speed it up of course.

I also think there are some other caching optimizations we could do.

dfunkt:
And here's another run, this time with cache:
https://github.com/dfunkt/vaultwarden/actions/runs/19407302056
10m 19s

dfunkt:
I guess I could also remove the max-parallelism part from here since it doesn't matter anymore with PR applied. It didn't have any downsides so I left it as it is for now, but if wanted I can remove it.
```
 # Also set max parallelism to 2, the default of 4 breaks GitHub Actions and causes OOMKills
        with:
          cache-binary: false
          buildkitd-config-inline: |
            [worker.oci]
              max-parallelism = 2   
```

BlackDex:
Thanks btw.

Ha, now i really have to keep building the Alpine build images as arm64, else this will break also, good to know.
I also am working on some changes on how i build my rust-musl containers, and i might be able to use some stuff for building the Vaultwarden images, but, still working onthe rust-musl workflow.

.github/workflows/release.yml
```diff
@@ -145,94 +161,99 @@ jobs:
...
-          push: true
           source: .
           files: docker/docker-bake.hcl
           targets: "${{ matrix.base_image }}-multi"
           set: |
             *.cache-from=${{ env.BAKE_CACHE_FROM }}
             *.cache-to=${{ env.BAKE_CACHE_TO }}
+            *.platform=linux/${{ matrix.arch }}
+            ${{ env.TAGS }}
+            *.output=type=image,push-by-digest=true,name-canonical=true,push=true,compression=zstd,force-compression=true
```
BlackDex: It's probably not a good idea to use `force-compression` here. The main reason is that this will re-compress known layers too, like Debian or Alpine, and will polute the Container Registry with duplicate caches, but different hashes while still being the same.
Setting the compression to zstd will still compress our own layers with zstd, but not the Alpine or Debian layers. This should also prevent users from downloading that same version/hash if they already have it via any other container.

dfunkt: Fixed

PR #6484, opened 2025-11-22
chore: prepare for edition 2024

reneleonhardt:
### Chores
- MSRV passed the 1.85 threshold, prepare for edition 2024

### DIFF
--- a/src/api/core/emergency_access.rs
+++ b/src/api/core/emergency_access.rs
@@ -746,22 +746,20 @@ pub async fn emergency_request_timeout_job(pool: DbPool) {
 
             if CONFIG.mail_enabled() {
                 // get grantor user to send Accepted email
-                let grantor_user =
-                    User::find_by_uuid(&emer.grantor_uuid, &conn).await.expect("Grantor user not found");
+                let grantor_user = User::find_by_uuid(&emer.grantor_uuid, &conn).await.expect("Grantor user not found");
 
                 // get grantee user to send Accepted email
-                let grantee_user =
-                    User::find_by_uuid(&emer.grantee_uuid.clone().expect("Grantee user invalid"), &conn)
-                        .await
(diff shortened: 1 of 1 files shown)


reneleonhardt:
### Notes
- All changes are fixing the stricter exhaustiveness rules
- TODO: `relative drop order changing in Rust 2024`
  https://doc.rust-lang.org/edition-guide/rust-2024/temporary-tail-expr-scope.html
  Run `cargo fix --edition --prepare-for=2024 --allow-dirty` and fix all warnings manually.

BlackDex:
I think we do not need any changes on our current code-base at all to switch to the 2024 edition.
No need to replace all `if let...` with bulky `match` statements as far as i know.

reneleonhardt:
Old code is still valid, but just try cargo fix yourself to see all warnings for changed rules.
https://doc.rust-lang.org/cargo/commands/cargo-fix.html#edition-migration

You can see how much I improved manually to reduce that bulkiness to a minimum, that's not the same code cargo fix generated.

BlackDex:
> Old code is still valid, but just try cargo fix yourself to see all warnings for changed rules. https://doc.rust-lang.org/cargo/commands/cargo-fix.html#edition-migration
> 
> You can see how much I improved manually to reduce that bulkiness to a minimum, that's not the same code cargo fix generated.

There is no need for `cargo fix` on our current code-base.
With Edition 2024 there are several other items which could be fixed like some `if` statements which can be merged, this can be seen when you run `clippy` on the code-base after setting the Edition to 2024.

All other fixes Cargo does automatically are only because it doesn't know if it will break the code or not, and just goes ahead and adjust it to a more safe default, which as far as I can tell, is not needed at all.

BlackDex:
Closing this as in my opinion items are changed which are not needed to be changed.

PR #6485, opened 2025-11-22
Update crates and Rust version

BlackDex:
- Update all crates (where possible) Adjusted code where needed
- Fixed some nightly clippy lints

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,6 +16,23 @@ concurrency:
   # Don't cancel other runs when creating a tag
   cancel-in-progress: ${{ github.ref_type == 'branch' }}
 
+defaults:
+  run:
+    shell: bash
+
+env:
+  # The *_REPO variables need to be configured as repository variables
+  # Append `/settings/variables/actions` to your repo url
+  # DOCKERHUB_REPO needs to be 'index.docker.io/<user>/<repo>'
+  # Check for Docker hub credentials in secrets
+  HAVE_DOCKERHUB_LOGIN: ${{ vars.DOCKERHUB_REPO != '' && secrets.DOCKERHUB_USERNAME != '' && secrets.DOCKERHUB_TOKEN != '' }}
+  # GHCR_REPO needs to be 'ghcr.io/<user>/<repo>'
+  # Check for Github credentials in secrets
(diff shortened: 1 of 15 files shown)


src/api/core/sends.rs
```diff
@@ -568,7 +568,7 @@ async fn post_access_file(
 async fn download_url(host: &Host, send_id: &SendId, file_id: &SendFileId) -> Result<String, crate::Error> {
     let operator = CONFIG.opendal_operator_for_path_type(&PathType::Sends)?;
 
-    if operator.info().scheme() == opendal::Scheme::Fs {
+    if operator.info().scheme() == String::from(opendal::Scheme::Fs) {
```
dani-garcia: I think we can save an allocation here by using:

```suggestion
    if operator.info().scheme() == <&'static str>::from(opendal::Scheme::Fs) {
```

https://docs.rs/opendal/0.55.0/src/opendal/types/scheme.rs.html#403-470

src/sso.rs
```diff
@@ -139,12 +139,7 @@ impl BasicTokenClaims {
 }
 
 fn decode_token_claims(token_name: &str, token: &str) -> ApiResult<BasicTokenClaims> {
-    let mut validation = jsonwebtoken::Validation::default();
-    validation.set_issuer(&[CONFIG.sso_authority()]);
-    validation.insecure_disable_signature_validation();
-    validation.validate_aud = false;
-
-    match jsonwebtoken::decode(token, &jsonwebtoken::DecodingKey::from_secret(&[]), &validation) {
+    match jsonwebtoken::dangerous::insecure_decode(token) {
```
dani-garcia: This was previously validating issuer, expiration and not before date, while now it doesn't. We should probably do that manually, I think?

src/api/core/sends.rs
BlackDex: Ah yes, that is better :).

src/sso.rs
BlackDex: I probably didn't checked that well, but indeed.

And that is a pain to do. It didn't checked the `nbf` since the default for that check is `false`.
Since the `BasicTokenClaims` has other attributes then needed to validate, I have created a `Validation` struct to validate first, and decode again afterwards. Creating a custom `impl From` didn't seem useful as the decode does all the heavy lifting already, yes it does need to base64 decode again, but i think that should be fine.

I did run the SSO test via Playwright and found no issues.

dani-garcia:
Oops, some conflicts to resolve :)

PR #6491, opened 2025-11-24
adds sso_identifier to /admin/users

### LINKED ISSUE #6413: "SSO Identifier" not returned from /admin/users API
### Prerequisites

* Vaultwarden version: v1.34.3-a2ad1dc7
* Web-vault version: v2025.8.0
* OS/Arch: linux/x86_64
* Database type: PostgreSQL
* Database version: PostgreSQL 15.14 on x86_64-redhat-linux-gnu, compiled by gcc (GCC) 11.5.0 20240719 (Red Hat 11.5.0-5), 64-bit
* Uses config.json: false
* Uses a reverse proxy: true
* IP Header check: false (X-Forwarded-For)
* Internet access: false
* DNS Check: true
* Server/NTP Time Check: n/a
* HTTPS Check: true
* Websocket Check: true
* HTTP Response Checks: true

**Config:**

```json
{
  "_duo_akey": null,
  "_enable_duo": true,
  "_enable_email_2fa": true,
  "_enable_smtp": true,
  "_enable_yubico": true,
  "_icon_service_csp": "",
  "_icon_service_url": "",
  "_ip_header_enabled": true,
  "_max_note_size": 10000,
  "_smtp_img_src": "***:",
  "admin_ratelimit_max_burst": 10,
  "admin_ratelimit_seconds": 60,
  "admin_session_lifetime": 20,
  "admin_token": "***",
  "allowed_connect_src": "",
  "attachments_folder": "/opt/vaultwarden/attachments",
  "data_folder": "/opt/vaultwarden",
  "database_conn_init": "",
  "database_idle_timeout": 600,
  "database_max_conns": 10,
  "database_min_conns": 2,
  "database_timeout": 30,
  "database_url": "**********://*************************************************************",
  "db_connection_retries": 15,
  "disable_admin_token": false,
  "domain_path": "",
  "domain_set": true,
  "duo_host": null,
  "duo_ikey": null,
  "duo_skey": null,
  "duo_use_iframe": false,
  "email_attempts_limit": 3,
  "email_change_allowed": true,
  "email_expiration_time": 600,
  "email_token_size": 6,
  "enable_db_wal": true,
  "enable_websocket": true,
  "events_days_retain": null,
  "extended_logging": false,
  "helo_name": null,
  "hibp_api_key": null,
  "icon_blacklist_regex": null,
  "icon_cache_folder": "/opt/vaultwarden/icon_cache",
  "icon_cache_negttl": 259200,
  "icon_cache_ttl": 2592000,
  "icon_download_timeout": 10,
  "icon_redirect_code": 302,
  "icon_service": "internal",
  "invitations_allowed": true,
  "ip_header": "X-Real-IP",
  "log_file": null,
  "log_level": "info",
  "org_attachment_limit": null,
  "org_creation_users": "****************",
  "org_events_enabled": false,
  "org_groups_enabled": false,
  "purge_incomplete_sso_nonce": "0 20 0 * * *",
  "push_enabled": false,
  "reload_templates": false,
  "rsa_key_filename": "/opt/vaultwarden/rsa_key",
  "sendmail_command": null,
  "sends_allowed": true,
  "sends_folder": "/opt/vaultwarden/sends",
  "show_password_hint": false,
  "signups_allowed": false,
  "signups_verify": false,
  "smtp_auth_mechanism": null,
  "smtp_debug": false,
  "smtp_embed_images": true,
  "smtp_explicit_tls": null,
  "smtp_host": "**********************",
  "smtp_password": null,
  "smtp_port": 25,
  "smtp_security": "off",
  "smtp_ssl": null,
  "smtp_timeout": 15,
  "smtp_username": null,
  "sso_allow_unknown_email_verification": true,
  "sso_audience_trusted": null,
  "sso_authority": "*****://******************",
  "sso_callback_path": "*****://***********************************************",
  "sso_client_id": "****************************************************************",
  "sso_client_secret": "***",
  "sso_debug_tokens": false,
  "sso_enabled": true,
  "sso_master_password_policy": null,
  "sso_only": false,
  "sso_pkce": true,
  "sso_scopes": "profile offline_access",
  "sso_signups_match_email": false,
  "templates_folder": "/opt/vaultwarden/templates",
  "tmp_folder": "/opt/vaultwarden/tmp",
  "use_sendmail": false,
  "use_syslog": true,
  "user_send_limit": null,
  "web_vault_enabled": true,
  "yubico_client_id": null,
  "yubico_secret_key": null,
  "yubico_server": null
}
```

</details>

### Vaultwarden Build Version

     1.34.3-a2ad1dc7 

### Deployment method

Official Container Image

### Custom deployment method

_No response_

### Reverse Proxy

n.a.

Linux

### Operating System Version

_No response_

### Clients

CLI

### Client Version

API

### Steps To Reproduce

1. make a request to /admin/users

### Expected Result

I expected the Admin API User's endpoint to return the full data available, including SSO parameters when SSO is used.

### Actual Result

The "SSO Identifier" column of the table at /admin/users/overview is missing from the JSON API

### Logs

```text

```

### Screenshots or Videos

_No response_

### Additional Context

_No response_

Katzenkralle:
Includes `sso_identifier` in JSON return of `/admin/users`

Fixes #6413

### DIFF
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -351,17 +351,17 @@ async fn get_users_property(users: Vec<(User, Option<SsoUser>)>, conn: &DbConn) 
     let mut users_json = Vec::with_capacity(users.len());
     for (u, sso_u) in users {
         let mut usr = u.to_json(conn).await;
-        usr["cipher_count"] = json!(Cipher::count_owned_by_user(&u.uuid, conn).await);
-        usr["attachment_count"] = json!(Attachment::count_by_user(&u.uuid, conn).await);
-        usr["attachment_size"] = json!(get_display_size(Attachment::size_by_user(&u.uuid, conn).await));
-        usr["user_enabled"] = json!(u.enabled);
-        usr["created_at"] = json!(format_naive_datetime_local(&u.created_at, DT_FMT));
-        usr["last_active"] = match u.last_active(conn).await {
+        usr["cipherCount"] = json!(Cipher::count_owned_by_user(&u.uuid, conn).await);
+        usr["attachmentCount"] = json!(Attachment::count_by_user(&u.uuid, conn).await);
+        usr["attachmentSize"] = json!(get_display_size(Attachment::size_by_user(&u.uuid, conn).await));
+        usr["userEnabled"] = json!(u.enabled);
+        usr["createdAt"] = json!(format_naive_datetime_local(&u.created_at, DT_FMT));
+        usr["lastActive"] = match u.last_active(conn).await {
             Some(dt) => json!(format_naive_datetime_local(&dt, DT_FMT)),
             None => json!("Never"),
         };
 
-        usr["sso_identifier"] = json!(sso_u.map(|u| u.identifier.to_string()).unwrap_or(String::new()));
+        usr["ssoIdentifier"] = json!(sso_u.map(|u| u.identifier.to_string()).unwrap_or(String::new()));
 
         users_json.push(usr);
     }
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -351,7 +351,7 @@ fn logout(cookies: &CookieJar<'_>) -> Redirect {
 async fn get_users_json(_token: AdminToken, conn: DbConn) -> Json<Value> {
     let users = User::get_all(&conn).await;
     let mut users_json = Vec::with_capacity(users.len());
-    for (u, _) in users {
+    for (u, sso_u) in users {
         let mut usr = u.to_json(&conn).await;
         usr["userEnabled"] = json!(u.enabled);
         usr["createdAt"] = json!(format_naive_datetime_local(&u.created_at, DT_FMT));
@@ -359,9 +359,10 @@ async fn get_users_json(_token: AdminToken, conn: DbConn) -> Json<Value> {
             Some(dt) => json!(format_naive_datetime_local(&dt, DT_FMT)),
             None => json!(None::<String>),
         };
+        usr["sso_identifier"] = json!(sso_u.map(|u| u.identifier.to_string()).unwrap_or(String::new()));
+
         users_json.push(usr);
     }
-
     Json(Value::Array(users_json))
 }
 
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -23,7 +23,7 @@ use crate::{
         backup_sqlite, get_sql_server_version,
         models::{
             Attachment, Cipher, Collection, Device, Event, EventType, Group, Invitation, Membership, MembershipId,
-            MembershipType, OrgPolicy, Organization, OrganizationId, SsoUser, TwoFactor, User, UserId
+            MembershipType, OrgPolicy, Organization, OrganizationId, SsoUser, TwoFactor, User, UserId,
         },
         DbConn, DbConnType, ACTIVE_DB_TYPE,
     },
@@ -347,7 +347,7 @@ fn logout(cookies: &CookieJar<'_>) -> Redirect {
     Redirect::to(admin_path())
 }
 
-async fn get_users_property(users: Vec<(User, Option<SsoUser>)>, conn: &DbConn) -> Vec<Value>  {
+async fn get_users_property(users: Vec<(User, Option<SsoUser>)>, conn: &DbConn) -> Vec<Value> {
     let mut users_json = Vec::with_capacity(users.len());
     for (u, sso_u) in users {
         let mut usr = u.to_json(conn).await;
@@ -365,7 +365,7 @@ async fn get_users_property(users: Vec<(User, Option<SsoUser>)>, conn: &DbConn) 
 
         users_json.push(usr);
     }
-    return users_json
+    users_json
 }
 
 #[get("/users")]
@@ -374,7 +374,6 @@ async fn get_users_json(_token: AdminToken, conn: DbConn) -> Json<Value> {
     let users_json = get_users_property(users, &conn).await;
     Json(Value::Array(users_json))
 }
-
 
 #[get("/users/overview")]
 async fn users_overview(_token: AdminToken, conn: DbConn) -> ApiResult<Html<String>> {
@@ -387,7 +386,7 @@ async fn users_overview(_token: AdminToken, conn: DbConn) -> ApiResult<Html<Stri
 #[get("/users/by-mail/<mail>")]
 async fn get_user_by_mail_json(mail: &str, _token: AdminToken, conn: DbConn) -> JsonResult {
     if let Some((u, sso)) = SsoUser::find_by_mail(mail, &conn).await {
-        let user_json = get_users_property(vec!((u, sso)), &conn).await[0].clone();
+        let user_json = get_users_property(vec![(u, sso)], &conn).await[0].clone();
         Ok(Json(user_json))
     } else {
         err_code!("User doesn't exist", Status::NotFound.code);
@@ -397,7 +396,7 @@ async fn get_user_by_mail_json(mail: &str, _token: AdminToken, conn: DbConn) -> 
 #[get("/users/<user_id>")]
 async fn get_user_json(user_id: UserId, _token: AdminToken, conn: DbConn) -> JsonResult {
     let u_sso = get_user_or_404(&user_id, &conn).await?;
-    let user_json = get_users_property(vec!(u_sso), &conn).await[0].clone();
+    let user_json = get_users_property(vec![u_sso], &conn).await[0].clone();
 
     Ok(Json(user_json))
 }
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -304,10 +304,6 @@ async fn get_user_or_404(user_id: &UserId, conn: &DbConn) -> ApiResult<User> {
     }
 }
 
-async fn get_sso_user(user_id: &UserId, conn: &DbConn) -> Option<SsoUser> {
-    SsoUser::find_by_uuid(user_id, conn).await.and_then(|user_and_sso| user_and_sso.1)
-}
-
 #[post("/invite", format = "application/json", data = "<data>")]
 async fn invite_user(data: Json<InviteData>, _token: AdminToken, conn: DbConn) -> JsonResult {
     let data: InviteData = data.into_inner();
@@ -400,7 +396,7 @@ async fn get_user_by_mail_json(mail: &str, _token: AdminToken, conn: DbConn) -> 
 #[get("/users/<user_id>")]
 async fn get_user_json(user_id: UserId, _token: AdminToken, conn: DbConn) -> JsonResult {
     let user = get_user_or_404(&user_id, &conn).await?;
-    let sso_user = get_sso_user(&user_id, &conn).await;
+    let sso_user = SsoUser::find_by_uuid(&user_id, &conn).await;
     let user_json = get_users_property(vec![(user, sso_user)], &conn).await[0].clone();
 
     Ok(Json(user_json))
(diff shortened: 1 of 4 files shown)


Timshel:
I had added it only to `/users/overview` since it's the only one neeeded for the admin table.

But it would probably make sense to return the same json object in all of `/users/overview`, `/users`, `/users/by-mail/<mail>` and `/users/<user_id>` ?

Katzenkralle:
Sure, all paths now return the same object

src/api/admin.rs
```diff
@@ -347,36 +347,16 @@ fn logout(cookies: &CookieJar<'_>) -> Redirect {
...
-        usr["attachment_count"] = json!(Attachment::count_by_user(&u.uuid, &conn).await);
-        usr["attachment_size"] = json!(get_display_size(Attachment::size_by_user(&u.uuid, &conn).await));
+        let mut usr = u.to_json(conn).await;
+        usr["cipher_count"] = json!(Cipher::count_owned_by_user(&u.uuid, conn).await);
+        usr["attachment_count"] = json!(Attachment::count_by_user(&u.uuid, conn).await);
+        usr["attachment_size"] = json!(get_display_size(Attachment::size_by_user(&u.uuid, conn).await));
         usr["user_enabled"] = json!(u.enabled);
         usr["created_at"] = json!(format_naive_datetime_local(&u.created_at, DT_FMT));
-        usr["last_active"] = match u.last_active(&conn).await {
+        usr["last_active"] = match u.last_active(conn).await {
```
stefan0xC: The only thing I noticed is that we currently return different identifiers
https://github.com/dani-garcia/vaultwarden/blob/bf37657c08aa3dd8b9c871d15d00c3a7bbcc756c/src/api/admin.rs#L356-L358
so this change would likely break for consumers of the `/admin/users` json endpoint. Accepting that I think this change is fine. (I think the alternative using `camelCase` while more consistent with the output of `u.to_json()` would require us to change the `admin/users.hbs` template and that would break any customized template.)

stefan0xC (approved):

src/api/admin.rs
Katzenkralle: > Accepting that I think this change is fine.

So is anything more needed from me to get this merged?

BlackDex:
Looking at this right now.
And i think we should not change the response key format, so that should be kept `camelCase`.
This does mean we need to adjust the template to match, else those will break, but since everything else is in `camelCase` we should stick with that and definitely not mix them.

Further, I also do not like that `get_user_or_404` now always returns a tuple with two items, while only one function call is really using this.

I think it's more idiomatic and less error prone to create a new function, something like `get_sso_user()` and call that separately at the only location where this is needed.

What do you think @stefan0xC?

stefan0xC:
I am not really the person to ask when it comes to idiomatic Rust as that is not my area of expertise but I agree that this can probably be improved and simplified. 

With regard to the change of the response key format I think it would also be fine to adapt the templates. Maybe we could mention it in the release notes that the templates have changed and/or that they will break for users that have customized them?

src/db/models/user.rs
```diff
@@ -527,6 +527,17 @@ impl SsoUser {
...
+        db_run! { conn: {
+            users::table
+                .left_join(sso_users::table)
+                .select(<(User, Option<Self>)>::as_select())
+                .filter(users::uuid.eq(uuid))
+                .first::<(User, Option<Self>)>(conn)
+                .ok()
+        }}
+    }
+
```
stefan0xC: I think this can now be simplified to just return the SsoUser entry?

src/api/admin.rs
```diff
@@ -304,6 +304,10 @@ async fn get_user_or_404(user_id: &UserId, conn: &DbConn) -> ApiResult<User> {
     }
 }
 
+async fn get_sso_user(user_id: &UserId, conn: &DbConn) -> Option<SsoUser> {
+    SsoUser::find_by_uuid(user_id, conn).await.and_then(|user_and_sso| user_and_sso.1)
+}
+
```
stefan0xC: This function would become obsolete if `SsoUser::find_by_uuid` were to return an `Option<SsoUser>`.

src/api/admin.rs
```diff
@@ -347,68 +351,59 @@ fn logout(cookies: &CookieJar<'_>) -> Redirect {
...
+        usr["cipherCount"] = json!(Cipher::count_owned_by_user(&u.uuid, conn).await);
+        usr["attachmentCount"] = json!(Attachment::count_by_user(&u.uuid, conn).await);
+        usr["attachmentSize"] = json!(get_display_size(Attachment::size_by_user(&u.uuid, conn).await));
         usr["userEnabled"] = json!(u.enabled);
         usr["createdAt"] = json!(format_naive_datetime_local(&u.created_at, DT_FMT));
-        usr["lastActive"] = match u.last_active(&conn).await {
+        usr["lastActive"] = match u.last_active(conn).await {
             Some(dt) => json!(format_naive_datetime_local(&dt, DT_FMT)),
-            None => json!(None::<String>),
+            None => json!("Never"),
```
stefan0xC: I think that it would probably be cleaner to return nothing here and have the template render `Never` if that is possible. That way the Admin template can also be translated.

src/api/admin.rs
```diff
@@ -347,68 +351,59 @@ fn logout(cookies: &CookieJar<'_>) -> Redirect {
     Redirect::to(admin_path())
 }
 
-#[get("/users")]
-async fn get_users_json(_token: AdminToken, conn: DbConn) -> Json<Value> {
-    let users = User::get_all(&conn).await;
+async fn get_users_property(users: Vec<(User, Option<SsoUser>)>, conn: &DbConn) -> Vec<Value> {
```
stefan0xC: Just as an idea but maybe we could rename the function so it's clearer what it does? E.g. `enrich_users_json` or something like that.

stefan0xC (changes requested):

src/db/models/user.rs
Katzenkralle: Sure, I originally had it keep returning User for all other find_by_XXX calls of SsoUser returned it, so I thought it might be beneficial for future use.

stefan0xC (approved):
Seems to work okay:
Image

Katzenkralle:
@BlackDex do I need to change anything else?

sistason:
this looks great. Roughly when can we expect a review, to plan ahead with integrating this in our workflow?

src/api/admin.rs
```diff
@@ -347,68 +347,59 @@ fn logout(cookies: &CookieJar<'_>) -> Redirect {
...
 
 #[get("/users/by-mail/<mail>")]
 async fn get_user_by_mail_json(mail: &str, _token: AdminToken, conn: DbConn) -> JsonResult {
-    if let Some(u) = User::find_by_mail(mail, &conn).await {
-        let mut usr = u.to_json(&conn).await;
-        usr["userEnabled"] = json!(u.enabled);
-        usr["createdAt"] = json!(format_naive_datetime_local(&u.created_at, DT_FMT));
-        Ok(Json(usr))
+    if let Some((u, sso)) = SsoUser::find_by_mail(mail, &conn).await {
+        let user_json = enrich_users_json(vec![(u, sso)], &conn).await[0].clone();
```
BlackDex: The `.await[0].clone()` could panic if the vec is empty, this is also not really idiomatic rust.
It's probably best to use something like `.into_iter().next()` which takes the first element.

src/api/admin.rs
```diff
@@ -347,68 +347,59 @@ fn logout(cookies: &CookieJar<'_>) -> Redirect {
...
 #[get("/users/<user_id>")]
 async fn get_user_json(user_id: UserId, _token: AdminToken, conn: DbConn) -> JsonResult {
-    let u = get_user_or_404(&user_id, &conn).await?;
-    let mut usr = u.to_json(&conn).await;
-    usr["userEnabled"] = json!(u.enabled);
-    usr["createdAt"] = json!(format_naive_datetime_local(&u.created_at, DT_FMT));
-    Ok(Json(usr))
+    let user = get_user_or_404(&user_id, &conn).await?;
+    let sso_user = SsoUser::find_by_uuid(&user_id, &conn).await;
+    let user_json = enrich_users_json(vec![(user, sso_user)], &conn).await[0].clone();
```
BlackDex: The same here regarding the `[0]`

src/static/templates/admin/users.hbs
```diff
@@ -43,22 +43,28 @@
...
+                            </span>
                         </td>
                         <td>
-                            <span class="d-block">{{cipher_count}}</span>
+                            <span class="d-block">{{cipherCount}}</span>
                         </td>
                         <td>
-                            <span class="d-block"><strong>Amount:</strong> {{attachment_count}}</span>
+                            <span class="d-block"><strong>Amount:</strong> {{attachmentCount}}</span>
                             {{#if attachment_count}}
```
BlackDex: This needs to be changed to `attachmentCount` too.

BlackDex (changes requested):
There are a few items to address.

src/api/admin.rs
Katzenkralle: https://github.com/dani-garcia/vaultwarden/pull/6491/changes/7debdbdf33d4a4ae31b6349764dc8af7beeca655#diff-48e73dac59e5044731073f602e5331265f839bd0ea9169e2a5368d16e2dbcebeR394

I hope this is fine, since there is no fitting error macro that dose not return

src/static/templates/admin/users.hbs
Katzenkralle: There is also `sso_enabled` and `page_data` still in snake case, but I did not touch these at all. So I will leave them unchanged

src/static/templates/admin/organizations.hbs
BlackDex, line 1: You edited the wrong file i guess, the org stuff isn't touched at all so probably revert this.

src/api/admin.rs
```diff
@@ -590,7 +591,7 @@ async fn organizations_overview(_token: AdminToken, conn: DbConn) -> ApiResult<H
         org["collection_count"] = json!(Collection::count_by_org(&o.uuid, &conn).await);
         org["group_count"] = json!(Group::count_by_org(&o.uuid, &conn).await);
         org["event_count"] = json!(Event::count_by_org(&o.uuid, &conn).await);
-        org["attachment_count"] = json!(Attachment::count_by_org(&o.uuid, &conn).await);
+        org["attachmentCount"] = json!(Attachment::count_by_org(&o.uuid, &conn).await);
```
BlackDex: This doesn't need to be changed in this PR. Lets leave it as was, same for the handlebars file linked to this function.

BlackDex (changes requested):

BlackDex:
This PR needs some rebasing and clippy fixes.

tessus:
@BlackDex this needs another workflow run...

src/api/admin.rs
```diff
@@ -351,68 +351,67 @@ fn logout(cookies: &CookieJar<'_>) -> Redirect {
...
+        usr["attachmentSize"] = json!(get_display_size(Attachment::size_by_user(&u.uuid, conn).await));
         usr["userEnabled"] = json!(u.enabled);
         usr["createdAt"] = json!(format_naive_datetime_local(&u.created_at, DT_FMT));
-        usr["lastActive"] = match u.last_active(&conn).await {
+        usr["lastActive"] = match u.last_active(conn).await {
             Some(dt) => json!(format_naive_datetime_local(&dt, DT_FMT)),
             None => json!(None::<String>),
         };
+
+        usr["sso_identifier"] = json!(sso_u.map_or(String::new(), |u| u.identifier.to_string()));
```
uedvt359, line 368: seems like this should be `ssoIdentifier`, c.f. https://github.com/dani-garcia/vaultwarden/pull/6491/changes#diff-40905218e594fec3b49fc332f17df9478fdbf21b69f307c21d4cafec0c69c1adR46

BlackDex:
This PR still needs some t.l.c. I'm still ok with adding this, but not as-is.

PR #6514, opened 2025-12-02
Org.put_policy type not in body anymore

Timshel:
With [web-v2025.11.3](https://github.com/bitwarden/clients/blob/web-v2025.11.3/libs/common/src/admin-console/models/request/policy.request.ts) the policy type is not sent in the request body anymore.

### DIFF
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -2057,8 +2057,6 @@ async fn get_policy(org_id: OrganizationId, pol_type: i32, headers: AdminHeaders
 #[derive(Deserialize)]
 struct PolicyData {
     enabled: bool,
-    #[serde(rename = "type")]
-    _type: i32,
     data: Option<Value>,
 }
 


stefan0xC:
If I try to check organization policies I get logged out due to a request to `/api/organizations/undefined/policies`. I don't think it's necessarily related to your issue but at least that's what keeps me from checking your PR.  (I have not found the time to check why this happens yet.)

```
[2025-12-02 17:54:42.757][request][INFO] GET /api/organizations/undefined/policies
[2025-12-02 17:54:42.761][auth][ERROR] Unauthorized Error: Error getting the organization id
[2025-12-02 17:54:42.761][vaultwarden::api::core::organizations::_][WARN] Request guard `AdminHeaders` failed: "Error getting the organization id".
```

Timshel:
@stefan0xC a yes had the same issue with latest client, had to add https://github.com/Timshel/oidc_web_vault/commit/36436c438437bf42b76552e98496f96693802fa4 :(.

Might try to open an issue directly on Bitwarden client repository but always hesitant :sweat_smile:

stefan0xC:
We could also ignore requests to that specific endpoint, so the `AdminHeaders` guard is not triggered. I just would not sure if it's better to return nothing or a 404 error message. (Looks like the next web-vault might have a refactored policy component https://github.com/bitwarden/clients/pull/17636, so it might not be an issue then. But I have not checked if this fixes the issue.)

Timshel:
Had a quick look, I would expect the refactor to fix the issue.

stefan0xC:
> Had a quick look, I would expect the refactor to fix the issue.

Indeed it does. Tested if that PR can be backported and while it will call the endpoint twice at least it doesn't request an `undefined` organization anymore.

dani-garcia (approved):

PR #6533, opened 2025-12-07
Further fixes for the release workflow

dfunkt:
Not all the fixes necessary were made in https://github.com/dani-garcia/vaultwarden/pull/6532, it still appends `debian` to the tags:
https://github.com/dani-garcia/vaultwarden/actions/runs/20004013122/job/57363788749#step:9:39
```
Creating manifest for index.docker.io/vaultwarden/server:testing-debian
Manifest created for index.docker.io/vaultwarden/server
#1 [internal] pushing docker.io/vaultwarden/server:testing-debian
```
Fixed variant:
```
Creating manifest for index.docker.io/vaultwarden/server:testing
Manifest created for index.docker.io/vaultwarden/server:testing
#1 [internal] pushing docker.io/vaultwarden/server:testing
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -362,8 +362,7 @@ jobs:
       - name: Create manifest list, push it and extract digest SHA
         working-directory: ${{ runner.temp }}/digests
         env:
-          BASE_IMAGE: "${{ matrix.base_image }}"
-          BASE_IMAGE_TAG: "${{ matrix.base_image == 'debian' && '' || format('-{0}', matrix.base_image) }}"
+          BASE_IMAGE_TAG: "${{ matrix.base_image != 'debian' && format('-{0}', matrix.base_image) || '' }}"
           BASE_TAGS: "${{ env.BASE_TAGS }}"
           CONTAINER_REGISTRIES: "${{ env.CONTAINER_REGISTRIES }}"
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -376,16 +375,16 @@ jobs:
 
               OUTPUT=$(docker buildx imagetools create \
                 -t "${img}:${tag}${BASE_IMAGE_TAG}" \
-                $(printf "${img}:${tag}-${BASE_IMAGE}@sha256:%s " *) 2>&1)
+                $(printf "${img}:${tag}${BASE_IMAGE_TAG}@sha256:%s " *) 2>&1)
```
BlackDex: Are you sure this is correct?
I thought the sha images are tagged with base-image and not without.

Also strange that the equal compare doesn't work, but ok.

BlackDex: Or, is that only used for the artifact reference?
Currently on my phone so a bit to difficult to check.

dfunkt: You made me realize something with this comment. Since the images pushed in the initial job aren't tagged with the base image/version (they're being referenced by digest), these shouldn't be tagged either since they're the ones being copied to the manifest. (which is tagged)

This can be seen in the reference implementation from docker:
https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners
```
env:
  REGISTRY_IMAGE: user/app
```
```
- name: Create manifest list and push
        working-directory: ${{ runner.temp }}/digests
        run: |
          docker buildx imagetools create $(jq -cr '.tags | map("-t " + .) | join(" ")' <<< "$DOCKER_METADATA_OUTPUT_JSON") \
            $(printf '${{ env.REGISTRY_IMAGE }}@sha256:%s ' *)
```

BlackDex: Ah, that is what i was thinking, but could check that well. Then i think is is ok.

BlackDex (approved):

PR #6551, opened 2025-12-12
Update crates and Rust

BlackDex:
- Updated all the crates
- Updated Rust to v1.92.0
- Updated to Alpine v3.23
- Adjusted some nightly clippy lints
- Updated Diesel to v2.3.5 which solves MariaDB issues, and the need for an older MariaDB library version.
- Update GHA Workflows

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -80,7 +80,7 @@ jobs:
 
       # Only install the clippy and rustfmt components on the default rust-toolchain
       - name: "Install rust-toolchain version"
-        uses: dtolnay/rust-toolchain@0b1efabc08b657293548b77fb76cc02d26091c7e # master @ Nov 20, 2025, 7:02 PM GMT+1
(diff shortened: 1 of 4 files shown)


src/api/icons.rs
```diff
@@ -797,8 +797,11 @@ impl Emitter for FaviconEmitter {
     fn emit_current_tag(&mut self) -> Option<html5gum::State> {
         self.flush_current_attribute(true);
         self.last_start_tag.clear();
-        if self.current_token.is_some() && !self.current_token.as_ref().unwrap().closing {
-            self.last_start_tag.extend(&*self.current_token.as_ref().unwrap().tag.name);
+        match &self.current_token {
+            Some(token) if token.closing => {
```
dani-garcia: Should this not be negated to match the previous logic?
```suggestion
            Some(token) if !token.closing => {
```

BlackDex: Indeed!

PR #6556, opened 2025-12-13
Commonize MariaDB Debian snapshot config

dfunkt:
- transform into macro

### DIFF
--- a/docker/Dockerfile.j2
+++ b/docker/Dockerfile.j2
@@ -27,6 +27,18 @@
 #     $ docker image inspect --format "{{ '{{' }}.RepoTags}}" docker.io/vaultwarden/web-vault@{{ vault_image_digest }}
 #     [docker.io/vaultwarden/web-vault:{{ vault_version }}]
 #
+{% macro debian_mariadb_snapshot_info() -%}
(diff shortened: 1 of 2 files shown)


dfunkt:
Added a missing slash from a path, apparently it was missing in the original template as well.

docker/Dockerfile.j2
```diff
@@ -91,6 +93,7 @@ RUN apt-get update && \
         "libc6-$(xx-info debian-arch)-cross" \
         "libc6-dev-$(xx-info debian-arch)-cross" \
         "linux-libc-dev-$(xx-info debian-arch)-cross" && \
+    {{ debian_mariadb_snapshot() }}
```
stefan0xC: Shouldn't the additional source be added before we run `apt-get update`?

stefan0xC (changes requested):

docker/Dockerfile.j2
dfunkt: Right, thanks for pointing that out.

BlackDex:
This is not needed anymore since Diesel v2.3.5

PR #6572, opened 2025-12-19
Add new accountKeys and masterPasswordUnlock fields

dani-garcia:
These fields are included in the login response now, and new versions of the mobile apps will require them.

I've also bumped the server version as I believe that the mobile apps will require server version to be >= 2025.8.0, though I haven't seen that in code anywhere yet.

### DIFF
--- a/src/api/identity.rs
+++ b/src/api/identity.rs
@@ -463,7 +463,6 @@ async fn authenticated_response(
 
     let master_password_policy = master_password_policy(user, conn).await;
 
-
     let has_master_password = !user.password_hash.is_empty();
     let master_password_unlock = if has_master_password {
         json!({


tessus:
Let me get this right, because I am not sure whether I understand this correctly.

The server now sends the master key and the public/private key pair to the client?

dani-garcia:
The (encrypted) master key and private key were always sent at the root of the response, the only difference is now they're also sent as part of the AccountKeys object. I think the original values at the root can now be considered deprecated and will be removed in future versions of the clients.

The only new addition is the public key, though I'm not sure why it's sent when it can be derived from the private key.

tessus:
> The (encrypted) master key and private key were always sent at the root of the response

Thanks, I didn't see that in the PR changes. I should have used the uncollapse button in the UI... ;-) (This is not the first time gh's collapsed section hid important info. I have to be more aware of that. I will work on it.)

> The only new addition is the public key, though I'm not sure why it's sent 

2 options: 1) the person who signed up on it, didn't know that you can get the public key from the private key 2) they want to save a few cycles not to calculate the pub key.
I seriously hope it's the latter. Haha.

PR #6578, opened 2025-12-20
Fix posting cipher with readonly collections

### LINKED ISSUE #6562: Broken ciphers created when collection is read-only
### Prerequisites

* Vaultwarden version: v1.34.3-f9751a0a
* Web-vault version: v2025.10.1
* OS/Arch: linux/x86_64
* Database type: PostgreSQL
* Database version: PostgreSQL 15.14 on x86_64-redhat-linux-gnu, compiled by gcc (GCC) 11.5.0 20240719 (Red Hat 11.5.0-5), 64-bit
* Uses config.json: true
* Uses a reverse proxy: true
* IP Header check: true (x-forwarded-for)
* Internet access: false
* Internet access via a proxy: true
* DNS Check: true
* Server/NTP Time Check: n/a
* HTTPS Check: true
* Websocket Check: true
* HTTP Response Checks: true

**Environment settings which are overridden:** DOMAIN, SIGNUPS_ALLOWED, ORG_CREATION_USERS, INVITATIONS_ALLOWED, ADMIN_TOKEN, IP_HEADER, SSO_ENABLED, SSO_SIGNUPS_MATCH_EMAIL, SSO_ALLOW_UNKNOWN_EMAIL_VERIFICATION, SSO_CLIENT_ID, SSO_CLIENT_SECRET, SSO_AUTHORITY, SSO_SCOPES, SSO_MASTER_PASSWORD_POLICY, SSO_AUTH_ONLY_NOT_SESSION, SMTP_HOST, SMTP_SECURITY, SMTP_PORT, SMTP_FROM

**Config:**

```json
{
  "_duo_akey": null,
  "_enable_duo": false,
  "_enable_email_2fa": false,
  "_enable_smtp": true,
  "_enable_yubico": false,
  "_icon_service_csp": "",
  "_icon_service_url": "",
  "_ip_header_enabled": true,
  "_max_note_size": 10000,
  "_smtp_img_src": "***:",
  "admin_ratelimit_max_burst": 10,
  "admin_ratelimit_seconds": 60,
  "admin_session_lifetime": 20,
  "admin_token": "***",
  "allowed_connect_src": "",
  "attachments_folder": "/opt/vaultwarden/attachments",
  "data_folder": "/opt/vaultwarden",
  "database_conn_init": "",
  "database_idle_timeout": 600,
  "database_max_conns": 10,
  "database_min_conns": 2,
  "database_timeout": 30,
  "database_url": "**********://*************************************************************",
  "db_connection_retries": 15,
  "disable_admin_token": false,
  "domain_path": "",
  "domain_set": true,
  "duo_host": null,
  "duo_ikey": null,
  "duo_skey": null,
  "duo_use_iframe": false,
  "email_attempts_limit": 3,
  "email_change_allowed": true,
  "email_expiration_time": 600,
  "email_token_size": 6,
  "emergency_access_allowed": false,
  "enable_db_wal": true,
  "enable_websocket": true,
  "events_days_retain": null,
  "extended_logging": true,
  "helo_name": null,
  "hibp_api_key": null,
  "icon_blacklist_regex": null,
  "icon_cache_folder": "/opt/vaultwarden/icon_cache",
  "icon_cache_negttl": 259200,
  "icon_cache_ttl": 2592000,
  "icon_download_timeout": 10,
  "icon_redirect_code": 302,
  "icon_service": "internal",
  "incomplete_2fa_time_limit": 0,
  "invitations_allowed": true,
  "ip_header": "x-forwarded-for",
  "log_file": null,
  "log_level": "info",
  "login_ratelimit_max_burst": 50,
  "login_ratelimit_seconds": 10,
  "org_attachment_limit": null,
  "org_creation_users": "******************************",
  "org_events_enabled": false,
  "org_groups_enabled": true,
  "purge_incomplete_sso_nonce": "0 20 0 * * *",
  "push_enabled": false,
  "reload_templates": false,
  "rsa_key_filename": "/opt/vaultwarden/rsa_key",
  "sendmail_command": null,
  "sends_allowed": true,
  "sends_folder": "/opt/vaultwarden/sends",
  "show_password_hint": false,
  "signups_allowed": false,
  "signups_verify": false,
  "smtp_auth_mechanism": null,
  "smtp_debug": false,
  "smtp_embed_images": true,
  "smtp_explicit_tls": null,
  "smtp_from_name": "***********",
  "smtp_host": "**********************",
  "smtp_password": null,
  "smtp_port": 25,
  "smtp_security": "off",
  "smtp_ssl": null,
  "smtp_timeout": 15,
  "smtp_username": null,
  "sso_allow_unknown_email_verification": true,
  "sso_audience_trusted": null,
  "sso_authority": "*****://******************",
  "sso_callback_path": "*****://***********************************************",
  "sso_client_id": "****************************************************************",
  "sso_client_secret": "***",
  "sso_debug_tokens": false,
  "sso_enabled": true,
  "sso_master_password_policy": "{\"minComplexity\":3,\"minLength\":12,\"requireLower\":true,\"requireNumbers\":true,\"requireSpecial\":false,\"requireUpper\":true}",
  "sso_only": true,
  "sso_pkce": true,
  "sso_scopes": "profile offline_access",
  "sso_signups_match_email": false,
  "templates_folder": "/templates",
  "tmp_folder": "/opt/vaultwarden/tmp",
  "use_sendmail": false,
  "use_syslog": true,
  "user_send_limit": null,
  "web_vault_enabled": true,
  "yubico_client_id": null,
  "yubico_secret_key": null,
  "yubico_server": null
}
```

</details>

### Vaultwarden Build Version

1.34.3-f9751a0a

### Deployment method

Official Container Image

### Custom deployment method

_No response_

### Reverse Proxy

n/a

Linux

### Operating System Version

rhel9

### Clients

CLI

### Client Version

api requests

### Steps To Reproduce

1. create a user for API access and join it to your ORG
2. add this user to a collection with permission view-only.
3. create a new cipher using the `/api/ciphers/create` endpoint, and set `collections` to the uuid of the read-only collection from step 2

### Expected Result

cipher creation should fail, and no half-finished entries should be written to the database

### Actual Result

cipher creation returns a 400 "No rights to modify the collection" error, but it creates an entry in the `ciphers` table, with the `data` column being the empty string.

later calls to `/api/sync` then cause an "Error parsing data field for <uuid>" log message for each cipher that was attemted to get created in a read-only collection.

### Logs

```text
(post_ciphers_create) POST /api/ciphers/create => 400 Bad Request
No rights to modify the collection



Error parsing data field for fe92b3a8-cbbf-47a8-b7d1-b90ae16b81fb
Error parsing data field for ff5764f2-1189-411d-869c-2d6d45b89f44
(sync) GET /api/sync?<data..> => 200 OK
```

### Screenshots or Videos

_No response_

### Additional Context

_No response_

BlackDex:
This fix will check if a collection is writeable for the user, and if not error out early instead of creating the cipher first and leaving it. It will also save some database transactions.

Fixes #6562

### DIFF
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -322,25 +322,6 @@ async fn post_ciphers_create(
 ) -> JsonResult {
     let mut data: ShareCipherData = data.into_inner();
 
-    // Check if there are one more more collections selected when this cipher is part of an organization.
-    // err if this is not the case before creating an empty cipher.
-    if let Some(org_id) = &data.cipher.organization_id {
-        if data.collection_ids.is_empty() {
-            err!("You must select at least one collection.");
-        } else {
-            for col_id in &data.collection_ids {
-                match Collection::find_by_uuid_and_org(col_id, org_id, &conn).await {
-                    None => err!("Invalid collection ID provided"),
-                    Some(collection) => {
-                        if !collection.is_writable_by_user(&headers.user.uuid, &conn).await {
-                            err!("No rights to modify the collection")
-                        }
-                    }
-                }
-            }
-        }
-    }
-
     // This check is usually only needed in update_cipher_from_data(), but we
     // need it here as well to avoid creating an empty cipher in the call to
     // cipher.save() below.
@@ -358,7 +339,11 @@ async fn post_ciphers_create(
     // or otherwise), we can just ignore this field entirely.
     data.cipher.last_known_revision_date = None;
 
-    share_cipher_by_uuid(&cipher.uuid, data, &headers, &conn, &nt, None).await
+    let res = share_cipher_by_uuid(&cipher.uuid, data, &headers, &conn, &nt, None).await;
+    if res.is_err() {
+        cipher.delete(&conn).await?;
+    }
+    res
 }
 
 /// Called when creating a new user-owned cipher.


src/api/core/ciphers.rs
```diff
@@ -324,8 +324,21 @@ async fn post_ciphers_create(
...
+    if let Some(org_id) = &data.cipher.organization_id {
+        if data.collection_ids.is_empty() {
+            err!("You must select at least one collection.");
+        } else {
+            for col_id in &data.collection_ids {
+                match Collection::find_by_uuid_and_org(col_id, org_id, &conn).await {
+                    None => err!("Invalid collection ID provided"),
+                    Some(collection) => {
+                        if !collection.is_writable_by_user(&headers.user.uuid, &conn).await {
+                            err!("No rights to modify the collection")
```
tessus:
```suggestion
                            err!("No permissions to modify the collection")
```

just an idea. not important. But I think I see the term `permissions` more often than `rights` in man pages and documentation.

BlackDex: shouldn't it then be `permission`?

BlackDex: Also, i copy/pasted it from the other location, so if, we would need to change it on the other functon(s) too.

tessus: > shouldn't it then be permission?

Yes and no. This one is a bit complicated. While it could be more than one permission that allows access, it is usually one permission that is the effective reason for the access. But lets take fs access as an example: `0644`. Any of the user, group, or other permission allows you read access. But when talking about the set you use `permissions`, when you talk about the specific read access for the user, you use `permission`.

In that sense, `permissions` is a term used for the concept and group of permissions, not a specific permission for a specific entity.
Otherwise you would have used the singular for `rights` as well.

But it is also valid to use the singular. It's up to you.

On the other side, if "rights" has been used in other parts of the code, ignore my squabble. No need to change it in multiple locations.

"Rights" is not wrong anyway. I am just more used to the other term. Who knows, maybe it's only subjective, idk. Sorry for the noise.

src/api/core/ciphers.rs
```diff
@@ -324,8 +324,21 @@ async fn post_ciphers_create(
 
     // Check if there are one more more collections selected when this cipher is part of an organization.
     // err if this is not the case before creating an empty cipher.
-    if data.cipher.organization_id.is_some() && data.collection_ids.is_empty() {
-        err!("You must select at least one collection.");
+    if let Some(org_id) = &data.cipher.organization_id {
+        if data.collection_ids.is_empty() {
+            err!("You must select at least one collection.");
```
dani-garcia: Checking for this seems mildly wasteful when `share_cipher_by_uuid` is basically checking the same thing. How about we just don't check here, and then check the result of `share_cipher_by_uuid` and delete the cipher if it's an error?

```rust
if let Err(error) = share_cipher_by_uuid(&cipher.uuid, data, &headers, &conn, &nt, None).await {
    cipher.delete().await?;
    return Err(e);
}
```

src/api/core/ciphers.rs
dani-garcia: Yeah we're a bit inconsitent on how we report these errors, we have a couple of `No rights to modify the collection`,  `no permission`, `user not allowed` around. Would be nice to go through the whole codebase and swap them for the same variant everywhere, probably on a separate PR though.

src/api/core/ciphers.rs
BlackDex: Thought about that too, but that makes even more database calls. First creating and saving, then a bunch of checking, then ow wait you aren't allowed to save, wait we will delete it. 

I actually think we need to rework a lot off these functions to lower tech debt and lower code duplication we currently have. 

But besides that, I'm fine  to delete afterwards instead of preventing to create.

BlackDex: Adjusted the code to delete on error. There are also a few more checks done there which might error out.
Since we need to bubble up the result, the `if let Err()` is not possible.

src/api/core/ciphers.rs
BlackDex: Agree.

src/api/core/ciphers.rs
dani-garcia: Yeah we're not being super efficient with database calls either way, but seems better to only make the extra calls on the error case, which leaves the success case a bit more optimized.

PR #6590, opened 2025-12-25
Add option to specify backup path

lukaskirner:
The current backup comannd will store the backup file in the same folder as the db itself. This MR gives the user the option to store backup files in a different folder (e.g. to an different SMB share).


BlackDex:
There already is an open PR for a feature like this, which needs some adjustments, and the committer is going to address those. So this for me is a duplicate PR.

Issue #6610, opened 2025-12-28
Logged out from every single client after last update


xJayMorex:
### Prerequisites

- [x] I have searched the existing **Closed _AND_ Open** [Issues](https://github.com/dani-garcia/vaultwarden/issues?q=is%3Aissue%20) **_AND_** [Discussions](https://github.com/dani-garcia/vaultwarden/discussions?discussions_q=)
- [x] I have searched and read the [documentation](https://github.com/dani-garcia/vaultwarden/wiki/)

### Your environment (Generated via diagnostics page)

* Vaultwarden version: v1.35.0
* Web-vault version: v2025.12.0
* OS/Arch: linux/x86_64
* Running within a container: true (Base: Debian)
* Database type: PostgreSQL
* Database version: PostgreSQL 18.1 (Debian 18.1-1.pgdg13+2) on x86_64-pc-linux-gnu, compiled by gcc (Debian 14.2.0-19) 14.2.0, 64-bit
* Uses config.json: false
* Uses a reverse proxy: true
* IP Header check: true (X-Real-IP)
* Internet access: true
* Internet access via a proxy: false
* DNS Check: true
* Browser/Server Time Check: false
* Server/NTP Time Check: false
* Domain Configuration Check: true
* HTTPS Check: true
* Websocket Check: true
* HTTP Response Checks: true

### Config & Details (Generated via diagnostics page)

<details><summary>Show Config & Details</summary>

**Config:**

```json
{
  "_duo_akey": null,
  "_enable_duo": true,
  "_enable_email_2fa": false,
  "_enable_smtp": true,
  "_enable_yubico": true,
  "_icon_service_csp": "",
  "_icon_service_url": "",
  "_ip_header_enabled": true,
  "_max_note_size": 10000,
  "_smtp_img_src": "***:",
  "admin_ratelimit_max_burst": 3,
  "admin_ratelimit_seconds": 300,
  "admin_session_lifetime": 20,
  "admin_token": "***",
  "allowed_connect_src": "",
  "allowed_iframe_ancestors": "",
  "attachments_folder": "data/attachments",
  "auth_request_purge_schedule": "30 * * * * *",
  "authenticator_disable_time_drift": false,
  "data_folder": "data",
  "database_conn_init": "",
  "database_idle_timeout": 600,
  "database_max_conns": 10,
  "database_min_conns": 2,
  "database_timeout": 30,
  "database_url": "**********://**********************************************************************************************************************************************************************",
[...]

PR #6629, opened 2025-12-29
Try old refresh token if we fail to decode jwt

dani-garcia:
The mobile apps don't seem to handle an error during the usage of `refresh_token` very well, so this code adds a backwards compatibility path to process the access token the old school way when it fails to decode as a JWT.

To test that this worked, what I've done is:
- Setup an instance of 1.34.3, log in to a web vault, lock it but don't close it.
- With dev tools open (and with `preserve log` enabled) try to unlock, which will make a successful /connect/token call with the refresh token
- Update to 1.35.0
- With dev tools open (and with `preserve log` enabled) try to unlock, this will return a 401, which will redirect you back to the login screen

With this patch, the connect call will succeed and return a new JWT refresh token.

Should fix https://github.com/dani-garcia/vaultwarden/issues/6610

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -1210,8 +1210,20 @@ pub async fn refresh_tokens(
 ) -> ApiResult<(Device, AuthTokens)> {
     let refresh_claims = match decode_refresh(refresh_token) {
         Err(err) => {
-            debug!("Failed to decode {} refresh_token: {refresh_token}", ip.ip);
-            err_silent!(format!("Impossible to read refresh_token: {}", err.message()))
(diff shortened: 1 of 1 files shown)


Gerardv514:
@dani-garcia is there a way to update the "old school" token to one that is generated as current?

dani-garcia:
If you updated the container to the latest `testing` that contains these changes, the next time the clients try to use the "old school" refresh token they'll get a newly generated token automatically.

If you mean if there's an easy manual way to do it without updating, then the answer is no. I'm planning a release of `1.35.1` sometime soon with the changes currently available in `testing`, so it's pretty safe to update.

PR #6626, opened 2025-12-29
Re-add `alpine` tag

dfunkt:
- fixes https://github.com/dani-garcia/vaultwarden/issues/6619
- also optimize the process while at it

PR #6632, opened 2025-12-30
Add dry run support to the release workflow

dfunkt:
- specific workflow name via `run_name` variable
- provides a "fake" tag to the manifest creation step (default 1.0.0, can be modified)
- added to the scheduled cleanup action

Suggested in https://github.com/dani-garcia/vaultwarden/pull/6626#issuecomment-3697577078
I left the artifacts & containers attestation enabled, can be removed if deemed unnecessary.
This could be modified to also support PRs, perhaps allowing for test images specific to that PR (could be done now or at a later date)

### DIFF
--- a/.github/workflows/cleanup.yml
+++ b/.github/workflows/cleanup.yml
@@ -5,7 +5,7 @@ on:
   workflow_dispatch:
     inputs:
       manual_trigger:
-        description: "Manual trigger buildcache cleanup"
+        description: "Manual trigger cleanup"
         required: false
         default: ""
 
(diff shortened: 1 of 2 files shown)


dfunkt:
This could also be done without pushing anything to the registry (stopping at the base tag generation step). Depends on whether the goal is to simulate the entire release process or not.

williamdes:
Nice, would this be the good time to provide an "edge" docker tag ?

dfunkt:
And how would this `edge` tag differ from the current `testing` tag that gets pushed after every commit? Or did you mean that this `edge` tag would only be added to the regular releases (like `1.34.3`,`1.35.0`)?

williamdes:
> And how would this `edge` tag differ from the current `testing` tag that gets pushed after every commit?

Sorry, I did not check :/
Perfect if we have the testing tag then 🎉

BlackDex:
First, sorry for my late check on this one.

A dry like this is probably not very useful i think. The main reason is we have to run it manually.
What might be better here is that we somehow are able to validate all tags will be outputed as wanted, before we actually build anything.

I have not yet tried this before, but i have like two ideas maybe?
1. Tag the `docker-build` with a yaml-link (or whatever it is called already)
2. Run that by default as dry-run by using a flag or env or something.
3. Validate the output in a second step, if that doesn't match what we want, abort.
4. Run that same `docker-build` again, since it worked and continue as normal.

Maybe using a workflow-import is better, but not sure though, there are some limitations to that if i'm correct.
But, i think that way, we still generate all the tags, can validate those, and continue if all is well?

That prevents manual checks, and will be a self-erroring-ci.

PR #6649, opened 2026-01-02
Add aria-label to token field on admin login page

martinstenrose:
This makes it possible for Bitwarden to autofill admin token by using a Custom Field [1] named "Admin token". Autofill for custom fields works by matching a form element's id, name, aria-label, or placeholder [2].

Currently, autofill can be configured using a custom field named "Token", which is somewhat generic and may be confused with other credentials such as an API token.

[1] https://bitwarden.com/help/custom-fields/
[2] https://bitwarden.com/help/auto-fill-custom-fields/

### DIFF
--- a/src/static/templates/admin/login.hbs
+++ b/src/static/templates/admin/login.hbs
@@ -13,7 +13,7 @@
             <small>Please provide it below:</small>
 
             <form class="form-inline" method="post" action="{{urlpath}}/admin">
-                <input type="password" autocomplete="password" class="form-control w-50 mr-2" name="token" placeholder="Enter admin token" autofocus="autofocus">
+                <input type="password" autocomplete="password" class="form-control w-50 mr-2" name="token" aria-label="Admin token" placeholder="Enter admin token" autofocus="autofocus">
                 {{#if redirect}}
                 <input type="hidden" id="redirect" name="redirect" value="/{{redirect}}">
                 {{/if}}


Gerardv514:
Hm I able to autofill the admin token without using a custom field, and by specifying the specific uri as a match. Do you also mean that autofill doesn't completely work for admin token, if a custom field isn't specified?

BlackDex:
As @Gerardv514 mentioned, this works just fine as it is just a password field, I use this my self regularly.
Since it isn't needed to add extra fields for this, I'm going to close this PR.

martinstenrose:
I wasn't able to do that, as it conflicts with the regular user login. Every other time Bitwarden fills the admin token (which matches the specific URI ending with /admin), and other times it fills the user login (which matches on the host).

I found that this can be worked around by adding a custom field named "token". Since that worked reliably, I created this PR to also allow matching on admin token, which felt more specific and less ambiguous.

As suggested by @Gerardv514, it also works if the user login entry is restricted to match only the specific URI `https://<vaultwarden-fqdn>/#/login`, and the admin token only the specific URI `https://<vaultwarden-fqdn>/admin`, which seems like a reasonable workaround as well.

Thanks for the review and the helpful comments.

Gerardv514:
Yes you'll need to set uri matches to exact 

My admin token vault item is set to match detect exact /admin

My vault - vault item also has a few uri's set to exact match, non of those have the /admin

So vault root uri set to exact match and then the same with /#/login

PR #6824, opened 2026-02-13
fix(send_invite): invite links

proofofcopilot:
Only include `orgSsoIdentifier` when the server is running in SSO-only mode, ensuring invites do not force SSO when password login is still allowed.

This PR is motivated by #6807

> I’m not deeply familiar with the Vaultwarden codebase yet, so I’d appreciate a review to confirm this change aligns with existing patterns and doesn’t introduce unintended side effects.

### DIFF
--- a/src/mail.rs
+++ b/src/mail.rs
@@ -302,10 +302,10 @@ pub async fn send_invite(
             .append_pair("organizationUserId", &member_id)
             .append_pair("token", &invite_token);
 
-        if CONFIG.sso_enabled() {
-            query_params.append_pair("orgUserHasExistingUser", "false");
+        if CONFIG.sso_enabled() && CONFIG.sso_only() {
             query_params.append_pair("orgSsoIdentifier", &org_id);
-        } else if user.private_key.is_some() {
+        }
+        if user.private_key.is_some() {
             query_params.append_pair("orgUserHasExistingUser", "true");
         }
     }


src/mail.rs
```diff
@@ -292,6 +296,7 @@ pub async fn send_invite(
         invited_by_email,
     );
     let invite_token = encode_jwt(&claims);
+    let user_exists = User::find_by_mail(&user.email, conn).await.is_some();
```
stefan0xC: In what way does the `User` returned by this call differ from the `user` that is passed to this function?

src/mail.rs
```diff
@@ -300,13 +305,11 @@ pub async fn send_invite(
...
             .append_pair("organizationUserId", &member_id)
-            .append_pair("token", &invite_token);
+            .append_pair("token", &invite_token)
+            .append_pair("orgUserHasExistingUser", if user_exists { "true" } else { "false" });
 
-        if CONFIG.sso_enabled() {
-            query_params.append_pair("orgUserHasExistingUser", "false");
+        if CONFIG.sso_enabled() && CONFIG.sso_only() {
             query_params.append_pair("orgSsoIdentifier", &org_id);
-        } else if user.private_key.is_some() {
```
stefan0xC, line 308: You have removed the actual check if a `user` has a private key, which they should have once they have registered.

stefan0xC: Might be a GitHub issue but I can't seem to mark those threads as resolved.
grafik

stefan0xC (approved):
Tested some more cases and the edge cases I could think of (i.e. without SSO enabled, with SSO enabled, and with SSO enforced for inviting both new accounts and already existing ones) and I think the change improves the situation even if there is still some ambiguity left due to the opposing requirements if you allow both password and SSO authentication that it's unclear whether the `orgSsoIdentifier` should be added or not.

With the parameter you effectively force users to signup using SSO so users without an SSO identity would have to manually remove the parameter to be able to join (the issue at hand). Without the parameter the signup link will not use SSO initially (unless you somehow manually added the parameter to the invite link) so we would create a normal account first and can only later associate the existing user to the SSO identity by matching the email address.

If we change it so that the parameter is only added with `SSO_ONLY` that also means that in case it's not enforced SSO login won't work with `SSO_SIGNUPS_MATCH_EMAIL=false` for users that created their account with an organization invite (or were invited via the `/admin` panel).
image

Personally I think this should be appropriate for most users that enable but don't enforce SSO as the option `SSO_SIGNUPS_MATCH_EMAIL` is already enabled by default and the current way would effectively enforce SSO for invited users.

stefan0xC:
@proofofcopilot I don't know if that is required by the frontend or how this would be handled in a selfhosted bitwarden/server instance or even what the Bitwarden Directory Connect does. To my understanding already existing for the purpose of the join link means that the user has already signed up (thus not requiring to set a master password just to login for confirmation).

btw: [That invited users are stored indefinitely was one of the first things](https://vaultwarden.discourse.group/t/how-long-are-invitations-kept/1166?u=stefan0xc) [I was confused about](https://vaultwarden.discourse.group/t/automatically-expire-invitations/1212?u=stefan0xc) [and frustrated with](https://github.com/dani-garcia/vaultwarden/pull/2795) when I started using Vaultwarden.

BlackDex:
> btw: [That invited users are stored indefinitely was one of the first things](https://vaultwarden.discourse.group/t/how-long-are-invitations-kept/1166?u=stefan0xc) [I was confused about](https://vaultwarden.discourse.group/t/automatically-expire-invitations/1212?u=stefan0xc) [and frustrated with](https://github.com/dani-garcia/vaultwarden/pull/2795) when I started using Vaultwarden.

And yet, no PR to fix that issue haha 😉

stefan0xC:
> And yet, no PR to fix that issue haha 😉

#2795 was my attempt if you don't recall 😛

BlackDex:
> > And yet, no PR to fix that issue haha 😉
> 
> #2795 was my attempt if you don't recall 😛

In have no active memory of that 🙈

stefan0xC:
> In have no active memory of that 🙈

Yeah, I mean I did not remember the entire conversation either only vaguely remembered that I wanted to change this behavior. Nowadays I would probably suggest that invitations should not be stored at all as user accounts until the invitations are complete but maybe only as temporary token that will be compared when you pass the invite token or something like that. I mean I have no idea if that would be feasible and I also don't have time to implement such a change either at the moment. So I still think that it would be better to follow the data minimization principle and to not store personal data in the first place if it's not necessary or at least not store it indefinitely. (Maybe I gave up to early and I should just revisit my PR? :smirk:)

PR #6828, opened 2026-02-14
Make available toggling webauthn allow_subdomains

Bert-Proesmans:
This patch makes it possible to have DNS records for subdomains of DOMAIN_URL to point to vaultwarden instances with working webauthn keys.

I verified this code works by running patched vaultwarden instances. This change is foolproof, allow_subdomains might as well have been set to true by default but opted for defensive approach.

Related discussion; https://github.com/dani-garcia/vaultwarden/discussions/6567

Before;

DOMAIN_URL = passwords.my.domain

1. RESET
1. Navigate to URL passwords.my.domain and login
2. Generate webauthn key and save
3. Attempt login and webauthn validation = OK
4. Navigate to URL \<anything>.passwords.my.domain and login
5. Attempt webauthn validation => FAILURE

Before;

DOMAIN_URL = passwords.my.domain

1. RESET
1. Navigate to URL \<anything>.passwords.my.domain and login
2. Generate webauthn key and save => FAILURE

=> webauthn_rs fails on the browser URL passed into auth_requests when doing stuff from hostname zone.passwords.my.domain

#
After;

DOMAIN_URL = passwords.my.domain
WEBAUTHN_ALLOW_SUBDOMAINS = true

* Generating webauthn keys from browser URL \<anything>.passwords.my.domain works
* Validating webauthn keys from browser URL \<anything>.passwords.my.domain works

### DIFF
--- a/src/api/core/two_factor/webauthn.rs
+++ b/src/api/core/two_factor/webauthn.rs
@@ -32,12 +32,17 @@ use webauthn_rs_proto::{
 static WEBAUTHN: LazyLock<Webauthn> = LazyLock::new(|| {
     let domain = CONFIG.domain();
     let domain_origin = CONFIG.domain_origin();
-    let rp_id = Url::parse(&domain).map(|u| u.domain().map(str::to_owned)).ok().flatten().unwrap_or_default();
-    let rp_origin = Url::parse(&domain_origin).unwrap();
+    let rp_id = Url::parse(&domain)
+        .map(|u| u.domain().map(str::to_owned))
+        .ok()
+        .flatten()
+        .expect("Invalid domain part for rp_id");
+    let rp_origin = Url::parse(&domain_origin).expect("Invalid domain_origin for rp_origin");
 
     let webauthn = WebauthnBuilder::new(&rp_id, &rp_origin)
         .expect("Creating WebauthnBuilder failed")
         .rp_name(&domain)
+        .allow_subdomains(CONFIG.webauthn_allow_subdomains())
         .timeout(Duration::from_millis(60000));
 
     webauthn.build().expect("Building Webauthn failed")
--- a/src/config.rs
+++ b/src/config.rs
@@ -793,6 +793,9 @@ make_config! {
         /// Prefer IPv6 (AAAA) resolving |> This settings configures the DNS resolver to resolve IPv6 first, and if not available try IPv4
         /// This could be useful in IPv6 only environments.
         dns_prefer_ipv6: bool, true, def, false;
+
+        /// Accept passkeys bound to subdomains |> This setting controls if passkeys bound to sub-domains of <Domain URL> are accepted for authentication.
+        webauthn_allow_subdomains: bool, false, def, false;
     },
 
     /// OpenID Connect SSO settings


BlackDex:
How is this related to #6900?
And would #6900 fix these issues without making these changes and having yet another config option?

Bert-Proesmans:
> How is this related to #6900? And would #6900 fix these issues without making these changes and having yet another config option?

can't say. that toggle looks like a client feature flag, and there is no obvious link to documentation. i assume #6900 is about passkeys inside the vault while this pr is about passkeys for vault login.

this is my first time reading about the "related origin requests" functionality. i gather from [the spec](https://w3c.github.io/webauthn/#sctn-related-origins) that the functionality is implemented on the user's password manager client and the website's authentication system (backend).
i'd say this doesn't provide compatible passkeys for vaultwarden for multiple domains out-of-the box. webauthn_rs still needs to validate the domain discrepancy somehow.
the suggested flow is creating redirects to the login page on the actual origin of the presented passkey and use unspecified cross domain authorization data to continue the user login (aka amazon.co.uk -> amazon.com [login] -> amazon.co.uk [logged id]). too much room to mess up, i don't think this is a good approach for selfhosted complexity-level and rather change the key domain (RP ID).

Maybe @Firstyear wants to take a quick look at open PRs related to webauthn and provide some insights?

Firstyear:
Related origin requests exist to resolve an issue with stateful (resident) keys. That is, the key which has client side state may only encode a single relyingParty Identifier. 

This means that when you change domain name you can NOT update the rp-id on the client without the client either re-enroling the key, or somehow negotiating the change with the client.

And of course, if you update the domain name, you can't just magically signal to all your clients to update their rp-id. They aren't all online, or on your site or whatever. 

related origin requests solve this by allowing a domain (the rp-id) to signal to an alternate domain that the credential *can* be used on that site. This firstly allows the key for domain A to be "found" (aka discovered) when you go to domain B, but also allows a call to navigator.credentials.get() to have a divergent rp-id from the origin in use. 


This is *different* to allowing subdomains. Allowing subdomains means that you define the rp-id as `example.com` and then you can have subdomains such as `a.example.com` or `b.example.com` both use that. This is valid as rp-id is an effective domain of both the subdomains. This is validated both in the browser and on the webauthn server side. 


So each of these solve a different problem.

Subdomains allow multiple urls as @Bert-Proesmans is attempting to do here.

Related origin requests allow you to *migrate* between different domains that do not have a common effective domain.

Bert-Proesmans:
Thanks for chippin' in!

I have some specific follow-up questions related to this PR;

- Would you appreciate it if WebAuthnBuilder is constructed with allow_subdomains set to true by default without dedicated configuration option?
- Am I dooming us all making rp_id configurable through a dedicated config parameter? (Besides the fact that valid values are restricted to domain suffixes of the DOMAIN config value)

Firstyear:
> Thanks for chippin' in!
> 
> I have some specific follow-up questions related to this PR;
> 
>     * Would you appreciate it if WebAuthnBuilder is constructed with allow_subdomains set to true by default without dedicated configuration option?

It should NOT be allowed by default as there are potential security issues that people need to be aware of (specifically that you need to be careful about subdomains and applications then). 

> 
>     * Am I dooming us all making rp_id configurable through a dedicated config parameter? (Besides the fact that valid values are restricted to domain suffixes of the DOMAIN config value)

The better question is "what are you trying to achieve".

Bert-Proesmans:
> The better question is "what are you trying to achieve".

Multiple database-synced instances of vaultwarden DNS bound to individual FQDNs for example, instance1.vault.example.com and instance2.vault.example.com. Without needing yet another HTTP system responding at URL vault.example.com.

\- If the individual FQDN is set as DOMAIN config value, then the RP_ID is also set[1] to the individual FQDN, then security keys bound to one instance are not usable on the other instance.

\- If the common domain-suffix vault.example.com is set as DOMAIN config value, then the RP_ID is also set[1] to vault.example.com, then security keys bound to one instance _are_ usable on the other instance.
BUT now URL generation and CSP policies use "vault.example.com". There is no physical instance bound to URL "vault.example.com" and you're forced to introduce a third system that handles this (HTTP forwarder or backend loadbalancer).

The question then explores applying the individual FQDN as DOMAIN config value, with RP_ID manually set to the domain-suffix "vault.example.com".
This comes from not accepting that I'm supposed to "embrace loadbalancers" in this multiple instance scenario, maybe the quick and right answer is just that.


[1] This is how vaultwarden currently implements WebauthnBuilder.
https://github.com/dani-garcia/vaultwarden/blob/9c7df6412c088832e1aeb673000716f5801dc8ed/src/api/core/two_factor/webauthn.rs#L33-L38

Firstyear:
So I think then in that case allow-subdomains is the correct answer here, but I think it should be opt-in (aka off by default). 

Allowing rp-id to be set manually is fine, but just ensure that you have an "origin" vs "rp-id" parameter where:

* Origin is the absolute URL of *this* server instance
* rp-id is equivalent to the domain of origin, or it is a parent domain of origin. 

This way vaultwarden can use "origin" to generate csp/urls and direct the user to "this instance" where rp-id allows the credentials to be used between instances. We have a similar concept in Kanidm to address this (origin vs domain) to enable exactly this scenario.

Bert-Proesmans:
The proposed "origin" and "domain" have to come from the configuration struct. I foresee that a semantical change of the keywords will be necessary (part of the discussion in #3870) and doing this in a backwards compatible way will be very hairy.

A temporary way forward could be to define 2 new config options with a default value `frontend_origin = domain` and `rp_domain = domain` (names not final). But the use case this change enables is not officially supported, which also applies to this PR, and it feels like a bad direction.

To be transparent, I have no intention to further participate in discussion and code changes working towards the goal described in the first paragraph. I'll close this PR to get it off everyone's watch list.

PR #6833, opened 2026-02-15
SSO config to skip 2FA on login

kowalski7cc:
Allow users to skip 2FA on SSO, userful when an external IdP has already multiple 2FA options.

Tested on Web Vault, Chrome Extension and Android app.

### DIFF
--- a/src/api/identity.rs
+++ b/src/api/identity.rs
@@ -275,7 +275,11 @@ async fn _sso_login(
         Some((mut user, sso_user)) => {
             let mut device = get_device(&data, conn, &user).await?;
 
-            let twofactor_token = twofactor_auth(&mut user, &data, &mut device, ip, client_version, conn).await?;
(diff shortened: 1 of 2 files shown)


kowalski7cc:
@BlackDex @dani-garcia is this an acceptable change? Hope you can review this pull 😃

stefan0xC:
If we want to skip 2FA for SSO shouldn't this be done properly i.e. by requesting a [specific `acr` level](https://openid.net/specs/openid-connect-core-1_0.html#acrSemantics) when authenticating with SSO so we make sure that the IdP is actually using 2FA?

kowalski7cc:
Well it could be a tri-state where you can skip always (as workaround or testing), automatically (acr level) or never (current behavior). To add the automatically I need to read a bit of documentation about the acr 😅

PR #6843, opened 2026-02-17
Update Rust and Crates and GHA

BlackDex:
- Update Rust to v1.93.1
- Updated all the crates
  Adjust changes needed for the newer `rand` crate
- Updated GitHub Actions

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,7 +78,7 @@ rmpv = "1.3.1" # MessagePack library
 dashmap = "6.1.0"
 
 # Async futures
-futures = "0.3.31"
+futures = "0.3.32"
 tokio = { version = "1.49.0", features = ["rt-multi-thread", "fs", "io-util", "parking_lot", "time", "signal", "net"] }
 tokio-util = { version = "0.7.18", features = ["compat"]}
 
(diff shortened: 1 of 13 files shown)


Cargo.toml
```diff
@@ -172,7 +172,7 @@ pastey = "0.2.1"
 governor = "0.10.4"
 
 # OIDC for SSO
-openidconnect = { version = "4.0.1", features = ["reqwest", "native-tls"] }
```
Timshel, line 175: Hey,
Any reason to switch to `rustls` ? 
Because this will break setups where people use self-signed certificate added to the OS store.

Edit: just seen that `rustls` appear to use `rustls-platform-verifier` in `reqwest` now.
Edit2: but unless I'm wrong that's not the case for the `0.12.28` version ?

BlackDex: It shouldn't break, since that still works because of the openssl probe which is done.

Also, everything currently is using rustls, so that is the main reason. And looking at reqwest, it's going to switch to aws-lc in the near future.

Timshel: Looking at https://doc.rust-lang.org/cargo/reference/features.html#feature-unification I'm guessing `rustls-tls-native-roots` should be applied since `openidconnect` depends on `reqwest` `0.12` too ? (sorry for the spam).

BlackDex: That feature is already applied to reqwest v0.12, which then also enables it for openidconnect.
Also, since newer versions of `rustls` support using certs in `SSL_CERT_DIR` without rehashing by-default now. And that feature is enabled via the `rustls-tls-native-roots` which enables `rustls-tls-native-roots-no-provider`, which in turn enables `dep:rustls-native-certs` and which will apply on every crate used within the Vaultwarden Project.

Timshel: Sorry about that, meant `should` as in should work :sweat:.
I was unsure and only later on confirmed with `cargo tree -f "{p} {f}"` that the feature was active for transitive dependency.

BlackDex: no prob, better an extra check and be sure right 😄

PR #6844, opened 2026-02-17
feat: add official Helm chart

### LINKED ISSUE #316: Helm Chart for Kubernetes deployment.
Hi,

Just a bit of shameless self-promotion, and a plea for cooperation in case someone-else considered using Helm deploying bitwarden_rs.

Based upon #234 and [kubernetes-bitwarden_rs](https://github.com/icicimov/kubernetes-bitwarden_rs), originally discussed in https://github.com/mprasil/bitwarden_rs/pull/3, I've developed a Helm Chart for deployment with [Helm](https://docs.helm.sh/).

The result can be found [here](https://github.com/Skeen/helm-bitwarden_rs)

Rohmilchkaese:
## Summary

Adds a production-ready Helm chart under `helm/vaultwarden/`. Closes #316.

Vaultwarden has no official Helm chart — [issue #316](https://github.com/dani-garcia/vaultwarden/issues/316) from 2019 only linked community charts, all of which are outdated or unmaintained. This chart provides a first-party, maintained deployment option for Kubernetes users.

### Highlights

- **Database flexibility**: SQLite default (matching vaultwarden defaults), PostgreSQL and MySQL selectable. Supports both full `DATABASE_URL` from a secret and composing the URL from parts via separate `secretKeyRef` (ideal for Zalano Postgres Operator, CloudNativePG, etc.)
- **Security hardened by default**: non-root (UID 1000), `readOnlyRootFilesystem: true`, `capabilities.drop: ["ALL"]`, `seccompProfile: RuntimeDefault`, `runAsNonRoot: true`, `automountServiceAccountToken: false`, tmpfs `/tmp` mount
- **All credentials via `secretKeyRef`**: Admin token, SMTP, SSO/OIDC, database, push notifications, Yubico — all support `existingSecret` for external secret management (ExternalSecrets, Sealed Secrets, etc.)
- **Ingress**: `ingressClassName` (not deprecated annotation), annotations, labels, multi-host TLS
- **Persistence**: `helm.sh/resource-policy: keep` (prevents data loss on uninstall), `storageClassName` 3-way handling (nil=cluster default, `"-"`=disable dynamic provisioning, `"name"`=explicit class), custom annotations/labels
- **Template validation**: `fail` on misconfiguration (e.g. admin enabled without token, postgresql without credentials)
- **Extensible**: `extraEnv`, `extraVolumes`, `extraVolumeMounts`, `initContainers`, `topologySpreadConstraints`, `priorityClassName`, `startupProbe`

### Chart structure

```
helm/vaultwarden/
├── .helmignore
├── Chart.yaml
├── README.md
├── values.yaml
└── templates/
    ├── _helpers.tpl
    ├── configmap.yaml
    ├── deployment.yaml
    ├── ingress.yaml
    ├── NOTES.txt
    ├── pvc.yaml
    ├── secret.yaml
    ├── service.yaml
    ├── serviceaccount.yaml
    └── tests/
        └── test-connection.yaml
```

### All env var names verified against source

Every environment variable in the chart was verified against `src/config.rs` in the vaultwarden source.

## Test plan

- [x] `helm lint --strict` — passes
- [x] `helm template` — tested with 6 configurations (SQLite default, PostgreSQL URL, PostgreSQL compose-from-parts, Ingress+TLS, SSO+SMTP, all existingSecret refs)
- [x] Template validation — confirmed `fail` triggers for: admin without token, postgresql without credentials, postgresql without any DB config
- [x] Deployed on real Kubernetes cluster (k3s v1.35.0):
  - Pod starts, `1/1 Running`, 0 restarts
  - Liveness and readiness probes pass
  - `readOnlyRootFilesystem: true` works (with `/tmp` tmpfs and `/data` PVC)
  - Vaultwarden logs show clean startup
  - `helm test` passes (busybox wget to `/alive`)
  - PVC retained on `helm uninstall` (resource-policy: keep)

### DIFF
--- a/helm/vaultwarden/templates/deployment.yaml
+++ b/helm/vaultwarden/templates/deployment.yaml
@@ -154,7 +154,20 @@ spec:
                   name: {{ include "vaultwarden.yubicoSecretName" . }}
                   key: {{ .Values.vaultwarden.yubico.existingSecretSecretKeyKey | default "yubico-secret-key" }}
             {{- end }}
-            {{- /* Extra env vars */}}
+            {{- /* Plain env vars from env map */}}
+            {{- range $name, $value := .Values.env }}
+            - name: {{ $name }}
+              value: {{ $value | quote }}
+            {{- end }}
+            {{- /* Secret env vars from secretEnv map */}}
+            {{- range $name, $ref := .Values.secretEnv }}
+            - name: {{ $name }}
+              valueFrom:
+                secretKeyRef:
+                  name: {{ $ref.secretName }}
+                  key: {{ $ref.secretKey }}
+            {{- end }}
+            {{- /* Raw extra env vars */}}
             {{- with .Values.extraEnv }}
             {{- toYaml . | nindent 12 }}
             {{- end }}
--- a/helm/vaultwarden/templates/deployment.yaml
+++ b/helm/vaultwarden/templates/deployment.yaml
@@ -1,0 +1,215 @@
+{{- /* Validation */}}
+{{- if and .Values.vaultwarden.admin.enabled (not .Values.vaultwarden.admin.token) (not .Values.vaultwarden.admin.existingSecret) }}
+  {{- fail "vaultwarden.admin.enabled is true but neither admin.token nor admin.existingSecret is set" }}
+{{- end }}
+{{- if and (ne .Values.database.type "sqlite") .Values.database.host (not .Values.database.credentialsSecret) }}
+  {{- fail "database.host is set but database.credentialsSecret is not — provide the secret containing database credentials" }}
+{{- end }}
+{{- if and (ne .Values.database.type "sqlite") (not .Values.database.host) (not .Values.database.url) (not .Values.database.existingSecret) }}
+  {{- fail "database.type is not sqlite but no database connection is configured — set database.host (with credentialsSecret), database.url, or database.existingSecret" }}
+{{- end }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "vaultwarden.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "vaultwarden.labels" . | nindent 4 }}
+spec:
+  replicas: {{ .Values.replicaCount }}
+  {{- if .Values.revisionHistoryLimit }}
+  revisionHistoryLimit: {{ .Values.revisionHistoryLimit }}
+  {{- end }}
+  strategy:
+    type: Recreate
+  selector:
+    matchLabels:
+      {{- include "vaultwarden.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      annotations:
+        checksum/configmap: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+        checksum/secret: {{ include (print $.Template.BasePath "/secret.yaml") . | sha256sum }}
+        {{- with .Values.podAnnotations }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+      labels:
+        {{- include "vaultwarden.labels" . | nindent 8 }}
+        {{- with .Values.podLabels }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+    spec:
+      serviceAccountName: {{ include "vaultwarden.serviceAccountName" . }}
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.podSecurityContext }}
+      securityContext:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- if .Values.priorityClassName }}
+      priorityClassName: {{ .Values.priorityClassName }}
+      {{- end }}
+      terminationGracePeriodSeconds: {{ .Values.terminationGracePeriodSeconds }}
+      {{- with .Values.initContainers }}
+      initContainers:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      containers:
+        - name: vaultwarden
+          image: "{{ .Values.image.repository }}:{{ include "vaultwarden.imageTag" . }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          ports:
+            - containerPort: {{ .Values.vaultwarden.rocketPort }}
+              name: http
+              protocol: TCP
+          {{- with .Values.securityContext }}
+          securityContext:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+          envFrom:
+            - configMapRef:
+                name: {{ include "vaultwarden.fullname" . }}
+          env:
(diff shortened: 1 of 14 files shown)


BlackDex:
I'm not very keen on keeping the helm chart in this repo to be honest, the same for any other Distro build scripts etc... besides the current Docker build scripts which we use to build and publish new versions.

Also, there is a helm chart already which is up-to-date https://github.com/guerzon/vaultwarden
Last updated 5 days ago, and also mentioned in the wiki.

Rohmilchkaese:
Thanks for the honest take @BlackDex.

I hear you on not wanting to turn this repo into a packaging monorepo — that makes total sense, and I agree distro packages / Ansible roles etc. don't belong here.

But I do think vaultwarden needs an official Helm chart somewhere, and here's why:

Every env var vaultwarden exposes is defined in `src/config.rs`. When config options get added, renamed, or deprecated, a community chart has to notice, reverse-engineer, and catch up — on its own timeline, with no guarantee it happens at all. An official chart maintained by people close to the project can stay in sync.

The track record of community charts for vaultwarden illustrates this point. The k8s-at-home chart was archived in 2022. The TrueCharts one is primarily designed for TrueNAS SCALE and explicitly doesn't guarantee standalone Helm usage. The guerzon chart is solid work, but it's maintained independently by a single contributor with no formal relationship to this project — and the wiki currently sends Kubernetes users to a dependency the maintainers don't control.

This isn't just my take — Discussion #6315 raised the same concern. A university IT department noted that community charts don't carry the same trust as something under the vaultwarden org, and another chart maintainer (gissilabs) offered to contribute theirs as well.

I'm not married to having it in this repo. Projects like Traefik, Gitea, and Grafana all maintain their official Helm charts in a dedicated repo under their org — not in the main application repo. Something like `vaultwarden/helm-chart` would keep this repo lean while giving Kubernetes users an officially maintained deployment path.

The maintenance overhead is minimal — Helm charts are mostly static YAML templates that only change when config options do.

Would a separate repo under the vaultwarden org work?

qianlongzt:
I don't think Helm is a good choice. It's hard to maintain, and if you have a lot of options, it becomes very difficult to manage. I think the best option is just to use plain Kubernetes YAML.

https://levelup.gitconnected.com/the-good-the-bad-and-the-ugly-of-templating-yaml-in-kubernetes-82fc5ce43fec

ilbarone87:
@qianlongzt that article specifically says that helm is great for wider adoption and less effective for static deployments. 
The goal of this PR is exactly having a wider adoption and consistency with an helm chart that can stay in sync with the project.

helm/vaultwarden/templates/deployment.yaml
```diff
@@ -0,0 +1,215 @@
...
+              name: http
+              protocol: TCP
+          {{- with .Values.securityContext }}
+          securityContext:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+          envFrom:
+            - configMapRef:
+                name: {{ include "vaultwarden.fullname" . }}
+          env:
```
qianlongzt, line 74: It needs to copy a lot of content when adding a new option in `configmap.yaml`, `deployment.yaml`, and `values.yaml`.

And it makes users have to translate between `values.yaml` and `.env`. 

We can simply allow users to define additional environment variables via `.Values.extraEnv` or rename to `.Values.env`

qianlongzt:
> The goal of this PR is exactly having a wider adoption and consistency with an helm chart that can stay in sync with the project.

This won’t happen automatically. Who will maintain and synchronize the Helm chart with the project?

Rohmilchkaese:
Fair point. I've added two new layers in the latest push that address this directly:

**`env`** — plain key-value map, zero boilerplate:

```yaml
env:
  SIGNUPS_ALLOWED: "true"
  INVITATION_ORG_NAME: "My Org"
  SENDS_ALLOWED: "true"
```

**`secretEnv`** — shorthand for secretKeyRef, no nested YAML:

```yaml
secretEnv:
  ADMIN_TOKEN:
    secretName: my-admin-secret
    secretKey: admin-token
  DATABASE_URL:
    secretName: my-db-secret
    secretKey: database-url
```

With these, users can set *any* vaultwarden env var without the chart needing changes. New options in `config.rs` never require a chart update — users just add them to `env` or `secretEnv`.

The structured values (`vaultwarden.smtp.*`, `database.*`, etc.) remain available for the cases where they add real value: template-time validation (`fail` on misconfiguration), grouped `existingSecret` references (one field covers both SMTP username and password), and the database compose-from-parts feature (Zalano/CNPG operator support). But they're optional — you can skip them entirely and use `env` + `secretEnv` for everything.

Rohmilchkaese:
Two fair concerns, let me address both:

**"Who will maintain it?"** — I will. But the maintenance burden is genuinely minimal here. With the `env`/`secretEnv` maps added in the latest push, the chart doesn't need updating when vaultwarden adds new config options — users just add the new env var directly. The structured values (`smtp.*`, `database.*`, etc.) are a convenience layer on top — if they ever fall behind, nothing breaks.

The only regular maintenance is bumping `appVersion` + `version` in `Chart.yaml` when a new vaultwarden release comes out. That's a two-line change — and with a GitHub Actions workflow using [chart-releaser](https://github.com/helm/chart-releaser-action), even that can be automated: bump the tag, CI packages the chart and publishes it to GitHub Pages, ArtifactHub picks it up automatically. At that point 99% of releases are just an image tag bump.

**"Just use plain YAML"** — Plain YAML works great for a single deployment you manage yourself. But it doesn't solve distribution. Every Kubernetes user deploying vaultwarden today is either copy-pasting YAML from the wiki and hand-editing it, or depending on a community chart with no connection to this project. There's no versioning, no upgrade path, no way to share configuration patterns.

The article you linked actually makes this distinction — Helm is less effective for static, single-environment deployments, but strong for wider adoption and distributing software to users with different environments. That's exactly the use case here: vaultwarden users running different ingress controllers, different database backends, different secret management, with or without SSO. A Helm chart encodes those choices as values rather than asking each user to understand and modify raw YAML.

That said, the two approaches aren't mutually exclusive. `helm template` renders plain YAML — users who prefer that workflow can use the chart as a YAML generator and apply the output directly with `kubectl`.

qianlongzt:
> "Just use plain YAML" 

I just want to avoid maintenance overhead. Using YAML with Go templates is quite cumbersome.

> With the env/secretEnv maps added in the latest push, the chart doesn't need updating when vaultwarden adds new config options 

This is ok for me now.

Timshel:
If the maintainers are not using it I don't believe it makes sense to integrate it, especially since presenting it as ` production-ready` would imply it is vouched by them.

And with:
> With the env/secretEnv maps added in the latest push, the chart doesn't need updating when vaultwarden adds new config options

Then the need to integrate it in the project is diminished.

BlackDex:
I'm still not in favor of adding this as a part of this repo it self. It would indicate it's an official supported helm chart, which, even though I work with K8s and helm charts, still do not find it should be part of this repo.

Therefor, I'm going to close this PR. I'm still thinking on maybe creating a community or contrib repo where stuff like this might be added.

For now, not going to add it to this repo.

PR #6853, opened 2026-02-20
Add cxp-import-mobile and cxp-export-mobile: feature flags on mobile

phoeagon:
add feature flags to turn on credential exchange protocol on mobile clients, notably iOS credential import/exporting with native Passwords

Bitwarden server side:
https://github.com/bitwarden/server/blob/a9616269573895640b2dad5baaa7b3a03498b2a5/src/Core/Constants.cs#L226

This should be a client feature, that allows the Bitwarden app to import/export with the iOS Passwords app and Chrome via iOS system level APIs.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -372,6 +372,7 @@
 ## Note that clients cache the /api/config endpoint for about 1 hour and it could take some time before they are enabled or disabled!
 ##
 ## The following flags are available:
+## - "pm-5594-safari-account-switching": Enable account switching in Safari. (Needs Safari >=2026.2.0)
 ## - "inline-menu-positioning-improvements": Enable the use of inline menu password generator and identity suggestions in the browser extension.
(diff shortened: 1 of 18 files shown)


billyprice1:
I was running stock Bitwarden client and connecting it to my self hosted vaultwarden. I didn’t set the feature flag as I couldn’t see a way to do it on IOS in the end I exported my vaultwarden json imported it into Bitwarden and that allowed me to move my passkeys  hope this helps

phoeagon:
@billyprice1 I can confirm that this will work on iOS. You can either review my changes and compile them to run, or use your reverse proxy to rewrite api/config to include the flags.

BlackDex (dismissed):
LGTM

waterdrops:
@phoeagon Many thanks, it really helped.

PR #6878, opened 2026-02-26
fix: grant manage permission to Manager on collection creation

### LINKED ISSUE #6871: Permission Editing not possible
### Prerequisites

While editing permissions as a Manager, you get logged out after trying to Edit Permissions on a Collection.
Might be caused by the recent Vulnerability-Fix

### Vaultwarden Build Version

1.35.4

### Deployment method

Official Container Image

### Custom deployment method

_No response_

### Reverse Proxy

nginx

Linux

### Operating System Version

_No response_

### Clients

Web Vault

### Client Version

_No response_

### Steps To Reproduce

1. Go to Web Vault with a Manager Account
2. Create New Collection
3. Select Collection to Edit
4. Edit Something
5. Click on Save
6. You get logged out

With existing Collections it works to edit the Rights

### Expected Result

The Setting will be saved.

### Actual Result

You get logged out with Error: No Permission

### Logs

```text
vaultwarden  | [2026-02-25 06:36:45.154][auth][ERROR] Unauthorized Error: The current user isn't a manager for this collection
vaultwarden  | [2026-02-25 06:36:45.154][vaultwarden::api::core::organizations::_][WARN] Request guard `ManagerHeaders` failed: "The current user isn't a manager for this collection".
vaultwarden  | [2026-02-25 06:36:45.154][response][INFO] (put_organization_collection_update) PUT /api/organizations/<org_id>/collections/<col_id> => 401 Unauthorized
```

### Screenshots or Videos

_No response_

### Additional Context

_No response_

rafaelfariasbsb:
## Summary

- Fixes a bug where a Manager without `access_all` could create a new collection but was unable to edit it afterward
- The `CollectionUser` record was being saved with `manage=false`, causing the `ManagerHeaders` guard to reject edit requests with "The current user isn't a manager for this collection"
- Changed the `manage` parameter from `false` to `true` when saving the `CollectionUser` for the creating Manager

## Details

In `src/api/core/organizations.rs` (line 524), the collection creation endpoint saves a `CollectionUser` entry for the Manager:

```rust
CollectionUser::save(&headers.membership.user_uuid, &collection.uuid, false, false, false, &conn).await?;
```

The last parameter (`manage`) was `false`, so subsequent calls to `Collection::is_coll_manageable_by_user()` via the `ManagerHeaders` guard would fail since:
1. `users_collections.manage` is `false`
2. `access_all` is `false` (condition to enter this code block)
3. The user is a Manager, not Admin/Owner

This was not reproduced when testing with a Manager that had `Manage all collections` enabled, because `access_all=true` bypasses this code path entirely.

## Test plan

- [ ] Create a Manager user **without** `Manage all collections`
- [ ] Grant the Manager permission to create collections
- [ ] Log in as the Manager and create a new collection
- [ ] Verify the Manager can edit the newly created collection
- [ ] Verify the Manager can update permissions on the collection
- [ ] Verify existing collection behavior is unchanged for Admins/Owners

Fixes #6871

### DIFF
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -521,7 +521,7 @@ async fn post_organization_collections(
     }
 
     if headers.membership.atype == MembershipType::Manager && !headers.membership.access_all {
-        CollectionUser::save(&headers.membership.user_uuid, &collection.uuid, false, false, false, &conn).await?;
+        CollectionUser::save(&headers.membership.user_uuid, &collection.uuid, false, false, true, &conn).await?;
     }
 
     Ok(Json(collection.to_json_details(&headers.membership.user_uuid, None, &conn).await))


stefan0xC:
Should a Manager that does not have access to all collections be allowed to create collections?

BlackDex:
> Should a Manager that does not have access to all collections be allowed to create collections?

No it should not. AI made a mistake.

uedvt359:
> Should a Manager that does not have access to all collections be allowed to create collections?

yes, they should. we migrated from bitwarden to vaultwarden last year, and with bitwarden, any manager is allowed to create (and delete) collections.

BlackDex:
> > Should a Manager that does not have access to all collections be allowed to create collections?
> 
> yes, they should. we migrated from bitwarden to vaultwarden last year, and with bitwarden, any manager is allowed to create (and delete) collections.

Vaultwarden is currently not using the exact same role-based-access as Bitwarden.
We abuse the `Custom` role to be an old standard manager, and when the `Manage all collections` is selected we internally use the old `access_all` functionality.

PR #6890 will also prevent any further possibilities for a non-access-all manager to be able to create a collection.
They are still allowed to edit/update a collection on which they are assigned to be a manager though.

Therefor I'm going to close this PR.

If you would like these specific features like multiple roles etc.. then please do not hesitate to create a well written PR.
But, maybe it would be better to start a topic in the `Ideas` Discussion section and discuss how this should be handled, since it will effect a lot of functions and code and queries used through the whole code.

PR #6880, opened 2026-02-27
Fix favicon fetching to check all icon links instead of just the first one

Shocker:
Fix logic in `download_icon()`, got broken (most likely accidentally) in https://github.com/dani-garcia/vaultwarden/commit/27dc67fadd3d45b9f7d8d37407cef9453b8f5802#diff-c0f730ed83eb02623c5c9bde35b0366115b8961fb596c27f98613140cba6d2a7L714 which caused the downloader to stop after first error

~~Also don't limit max checked icons to 5, allow all icons in the list~~

### DIFF
--- a/src/api/icons.rs
+++ b/src/api/icons.rs
@@ -534,7 +534,9 @@ async fn download_icon(domain: &str) -> Result<(Bytes, Option<&str>), Error> {
 
     use data_url::DataUrl;
 
-    for icon in icon_result.iconlist.iter().take(5) {
+    for (i, icon) in icon_result.iconlist.iter().enumerate() {
+        let is_last = i == icon_result.iconlist.iter().count() - 1;
+
         if icon.href.starts_with("data:image") {
             let Ok(datauri) = DataUrl::process(&icon.href) else {
                 continue;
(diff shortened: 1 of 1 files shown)


BlackDex:
The reason we only select the first 5 items is because we do not want to try and download every icon a site has listed.
It could be 1 or 20, and calling a site 20 times might also trigger a block/ban even.

The 5 items from the list are already sorted in a most preferred item is at the top of the list.

Are there specific sites on which you experience issues?
Since, i have run a local script which compares icons between Vaultwarden, Bitwarden, Google and DuckDuckGo.
And there are just a few which do not work currently, but also a lot which Vaultwarden is able to fetch an icon where Bitwarden, Google or DuckDuckGo (Either one or more) is not.

Shocker:
Regarding the 5 limit: I don't have any specific site with issues. To be honest it was just a weird seemingly undocumented line in the code which didn't make sense to have an arbitrary limit of 5.

What about the main issue being fixed with the PR?

BlackDex:
Which main issue do you refer to?
Is it that if one item fails it stop and aborts?

Shocker:
> Which main issue do you refer to? Is it that if one item fails it stop and aborts?

Yes.

src/api/icons.rs
```diff
@@ -534,7 +534,9 @@ async fn download_icon(domain: &str) -> Result<(Bytes, Option<&str>), Error> {
 
     use data_url::DataUrl;
 
-    for icon in icon_result.iconlist.iter().take(5) {
+    for (i, icon) in icon_result.iconlist.iter().take(5).enumerate() {
+        let is_last = i == icon_result.iconlist.iter().take(5).count() - 1;
```
BlackDex: Hmm, this `is_last` is now done during every loop, which seems to be wasted resources.
Maybe using `peekable()` here instead of `enumerate()` could help.
Then the `is_last` can be removed here, as this also is never needed for `data:images` either.

And you could use `Err(e) if icon.peek().is_some() => return`

BlackDex (changes requested):

src/api/icons.rs
Shocker: added a new commit

BlackDex (approved):
Seems to work ok. Thanks!

dani-garcia (approved):

PR #6881, opened 2026-02-27
Add Google S2 icon cache fallback for favicon fetching

Shocker:
**LATER EDIT: ignore the request but see the comments below**

In case the favicon download fails due to anti-bot/anti-crawler techniques (most likely HTTP code 403 or even 444) fallback to Google icon cache which is almost sure to have the favicon cached.

While it may seem a bad idea to hardcode links like this, an increasingly amount of sites started using such techniques, especially aggressive Cloudflare-protected sites, causing Vaultwarden to fail in fetching favicons

P.S.: For this pull request to work, https://github.com/dani-garcia/vaultwarden/pull/6880 also needs to be accepted/merged

### DIFF
--- a/src/api/icons.rs
+++ b/src/api/icons.rs
@@ -425,6 +425,9 @@ async fn get_icon_url(domain: &str) -> Result<IconUrlResult, Error> {
         iconlist.push(Icon::new(40, format!("{httpdomain}/apple-touch-icon.png")));
     }
 
+    // Add Google S2 icon cache fallback
+    iconlist.push(Icon::new(255, format!("https://www.google.com/s2/favicons?sz=32&domain={domain}")));
+
     // Sort the iconlist by priority
     iconlist.sort_by_key(|x| x.priority);
 


BlackDex:
I'm not in favor of this.
The whole reason to have a self-working favicon downloader is to not use external sites like Google or any other favicon provider.

If you want to use Google, DuckDuckGo or even Bitwarden's version, you can already configure this.
Most people do not want companies like Google or which ever to track these kind of items.

So, from my point of view this is not going to be implemented.

Shocker:
> I'm not in favor of this. The whole reason to have a self-working favicon downloader is to not use external sites like Google or any other favicon provider.
> 
> If you want to use Google, DuckDuckGo or even Bitwarden's version, you can already configure this. Most people do not want companies like Google or which ever to track these kind of items.
> 
> So, from my point of view this is not going to be implemented.

I apologize for not knowing about the ICON_SERVICE. I mostly agree with what you said. Then what about adding an extra config to have ICON_SERVICE act as a fallback to the internal downloader instead of replacing it completely, and have it work only when the internal downloader fails?

larena1:
I already tried to add some missing favicons by hand but Vaultwarden manages the directory and would just delete them after a while. 

Might make sense to have a user managed fallback favicon directory for that?

stefan0xC:
> I already tried to add some missing favicons by hand but Vaultwarden manages the directory and would just delete them after a while.

Have you read the [wiki article that explains how to use custom icons](https://github.com/dani-garcia/vaultwarden/wiki/Using-custom-website-icons#how-the-icon-cache-expiration-works)? It has a section about cache expiration and explains how to prevent your custom placed items from getting deleted.

Shocker:
I am proposing https://github.com/dani-garcia/vaultwarden/pull/6882 instead

PR #6882, opened 2026-02-27
Implement Fallback Icon service

Shocker:
Following the discussion from here https://github.com/dani-garcia/vaultwarden/pull/6881 add a separate config to allow fallback to external icon services icon URL (bitwarden, duckduckgo, google) when using the internal icon service in case all the default icon links fail to be downloaded (possibly due to anti-bot/anti-crawler techniques)

~~P.S.: For this pull request to work, https://github.com/dani-garcia/vaultwarden/pull/6880 MUST be accepted/merged. After that this PR will require a small code change to accomodate for those new changes~~ resolved


Shocker:
ready for review

Roki100 (approved):

Shocker:
ready for review again. merge it please, thank you

Boer2333:
Hi maintainers, just adding a real-world use case to (hopefully) help get this reviewed and merged.

**Scenario:** I self-host Vaultwarden on my NAS behind a reverse proxy, using `ICON_SERVICE=duckduckgo`. Public sites (binance, google, etc.) get crisp icons fine. But I also store logins for my **own self-hosted services** (e.g. `mima.mydomain.xyz`, `home.mydomain.xyz`, `tv.mydomain.xyz`). Those domains don't exist in duckduckgo's / google's icon indices at all, so they always fall back to the generic default icon — even though the sites themselves serve a perfectly good `/favicon.ico` that the internal fetcher *can* retrieve (I verified this: with `internal` the icon was fetched and cached; with an external service it 404s and shows the placeholder).

So the exact combination this PR proposes — **internal first, fall back to external (bitwarden/duckduckgo/google) when internal download fails** — would give us both:
- crisp icons for public sites via external services, and
- working icons for self-hosted/internal domains via the internal fetcher.

This is a pretty common pain point for the self-hosting crowd (multiple threads on the forum and reddit about "no icons for self-hosted services"). The feature seems complete and has been "ready for review" for a while — if the fallback order could also be made configurable (e.g. `external` first with `internal` fallback, or the reverse), that would cover both directions of preference.

Really appreciate the work on this. Thanks!

BlackDex:
> Hi maintainers, just adding a real-world use case to (hopefully) help get this reviewed and merged.
> 
> **Scenario:** I self-host Vaultwarden on my NAS behind a reverse proxy, using `ICON_SERVICE=duckduckgo`. Public sites (binance, google, etc.) get crisp icons fine. But I also store logins for my **own self-hosted services** (e.g. `mima.mydomain.xyz`, `home.mydomain.xyz`, `tv.mydomain.xyz`). Those domains don't exist in duckduckgo's / google's icon indices at all, so they always fall back to the generic default icon — even though the sites themselves serve a perfectly good `/favicon.ico` that the internal fetcher _can_ retrieve (I verified this: with `internal` the icon was fetched and cached; with an external service it 404s and shows the placeholder).
> 
> So the exact combination this PR proposes — **internal first, fall back to external (bitwarden/duckduckgo/google) when internal download fails** — would give us both:
> 
> * crisp icons for public sites via external services, and
> * working icons for self-hosted/internal domains via the internal fetcher.
> 
> This is a pretty common pain point for the self-hosting crowd (multiple threads on the forum and reddit about "no icons for self-hosted services"). The feature seems complete and has been "ready for review" for a while — if the fallback order could also be made configurable (e.g. `external` first with `internal` fallback, or the reverse), that would cover both directions of preference.
> 
> Really appreciate the work on this. Thanks!

In your case, you should disable the block non global ips.

https://github.com/dani-garcia/vaultwarden/blob/46d71107f5094460dd5ecbe1dbac6e6c71e5189a/.env.template#L382-L384

BlackDex:
I actually think this would make everything more complex and unpredictable.
The main reason for the external icon fetching services is to not let those sites know you are using Vaultwarden, this would end up requesting an icon from those external services internally and exposes your server.

Adding this will only add more calls to an already large list of calls done by Vaultwarden to try and fetch an icon.
I know it will be an option and not something that will be done by default, but still.

BlackDex:
I have thought about this a bit longer, and really do not want to add any more complexity to this process.
We already do a lot of http calls trying to fetch a favicon, and adding one extra, even though it's users choice still makes it more complex.

Maybe ill think about it differently somewhere in the future, but for now, not really.
There is a self-hosted tool which kinda does this already, https://github.com/vemetric/favicon-api, and i know, you now have to run a second tool again just for this. But i would like to keep Vaultwarden from this kind of complexity.

Therefor, going to close this one.

PR #6891, opened 2026-03-03
Feat(config): add feature flag for Safari account switching

DerPlayer2001:
This enables the use of the Feature from this PR https://github.com/bitwarden/clients/pull/18339

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1027,12 +1027,14 @@ fn validate_config(cfg: &ConfigItems) -> Result<(), Error> {
     }
 
     // Server (v2025.6.2): https://github.com/bitwarden/server/blob/d094be3267f2030bd0dc62106bc6871cf82682f5/src/Core/Constants.cs#L103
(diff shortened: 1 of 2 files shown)


src/config.rs
```diff
@@ -1046,6 +1046,8 @@ fn validate_config(cfg: &ConfigItems) -> Result<(), Error> {
         "anon-addy-self-host-alias",
         "simple-login-self-host-alias",
         "mutual-tls",
+        // Web Team
```
DerPlayer2001: I'm not 100% sure about the Team Comment please advise if this is correct or if I should adjust this.

stefan0xC: What would be the point of the comment?

DerPlayer2001: The feature Flags above are grouped by those comments so I assumed i should add a group for web since none of the existing groups fit that feature flag

stefan0xC: But why "Web Team" and not "[Auth](https://github.com/bitwarden/clients/blob/a2fefe804d8c9b4a56c42f9904512c5c5821e2f6/libs/common/src/enums/feature-flag.enum.ts#L19)"? It should also probably be added to the top?

PR #6900, opened 2026-03-05
Add Webauthn related origins flag to known flags.

pasarenicu:
support pm-30529-webauthn-related-origins flag

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1046,6 +1046,8 @@ fn validate_config(cfg: &ConfigItems) -> Result<(), Error> {
         "anon-addy-self-host-alias",
         "simple-login-self-host-alias",
         "mutual-tls",
+        // Webauthn Related Origins
+        "pm-30529-webauthn-related-origins",
     ];
     let configured_flags = parse_experimental_client_feature_flags(&cfg.experimental_client_feature_flags);
     let invalid_flags: Vec<_> = configured_flags.keys().filter(|flag| !KNOWN_FLAGS.contains(&flag.as_str())).collect();


BlackDex:
How is this related to #6828?
And is there nothing else needed to be changed or added on the server side?
How does the client known/generate/add the allowed related origins?

pasarenicu:
> How is this related to #6828? And is there nothing else needed to be changed or added on the server side? How does the client known/generate/add the allowed related origins?

yes, nothing else. Just add the flag.  Browser extension needs this flag to be sent from the server. Tested just the way it is and it works with bitwarden client extension for Chrome and Edge 2026.2.0,  (NOT 2026.1.1)
Firefox does not support Related Origins yet in browser.

i saved some info for this

7 Ian 2026 | PR #18242 created in  bitwarden/clients
27 Ian 2026 | Feature flag added in bitwarden/server (commit 0f43235d)
11 Feb 2026 | PR #18242 merged in main (merge commit 29e2be0d) 
18 Feb 2026 | browser-v2026.1.1 release (456e7deb) — diverged from main (PR #18242 NOT IN  browser-v2026.1.1 only in  browser-v2026.2.0)

 more info here.
 https://passkeys.dev/docs/advanced/related-origins/

BlackDex:
> Yeah this is entirely a client side feature. The clients do a request to `$DOMAIN/.well-known/webauthn` to get the list of related origins so it doesn't require changes on our side. Seems like a resasonable addition here.

@dani-garcia 
Well, we currently do not have this endpoint, should we add this?

dani-garcia:
@BlackDex This is used when the clients use the passkeys stored in the user's vault to authenticate to other sites, so I don't think it's needed. It would be up to those sites to provide the endpoint, or the clients will fallback to the older style rp == origin check.

It might be useful for things like #6828 that are related to authenticating with Vaultwarden itself, but I haven't had the time to test that.

pasarenicu:
> Yeah this is entirely a client side feature. The clients do a request to `$DOMAIN/.well-known/webauthn` to get the list of related origins so it doesn't require changes on our side. Seems like a resasonable addition here.

Yes. This support the new related origins feature from clients.

PR #6916, opened 2026-03-09
Add support for archiving items

matt-aaron:
Adds support for archiving items: https://github.com/dani-garcia/vaultwarden/discussions/6372

I'm new to the codebase, so let me know if I have missed anything. Happy to apply any feedback you may have. Thank you!


src/api/core/ciphers.rs
```diff
@@ -1703,6 +1707,36 @@ async fn delete_all(
...
+#[put("/ciphers/unarchive", data = "<data>")]
+async fn unarchive_cipher_selected(
+    data: Json<CipherIdsData>,
+    headers: Headers,
+    conn: DbConn,
+    nt: Notify<'_>,
+) -> JsonResult {
+    _set_archived_multiple_ciphers(data, &headers, false, &conn, &nt).await
+}
+
```
BlackDex, line 1759: Looks ok, but i would like to not use an `_` as function prefix. I know we currently have this on several locations, but it's not really idomatic Rust since functions or variables starting with `_` are normally used for functions or variables not really used, or need to be ignored during clippy lints.

BlackDex (changes requested):

src/api/core/ciphers.rs
matt-aaron: Updated

src/config.rs
```diff
@@ -1046,6 +1046,7 @@ fn validate_config(cfg: &ConfigItems) -> Result<(), Error> {
         "anon-addy-self-host-alias",
         "simple-login-self-host-alias",
         "mutual-tls",
+        "pm-19148-innovation-archive",
```
stefan0xC: @BlackDex I have not tested it (and I am not sure how stable this feature is yet) but could we not just enable this feature directly instead of adding a new feature flag that you have to opt in?

Also the feature flags we set have already been removed for a few releases, so they can be safely replaced in my opinion.
https://github.com/dani-garcia/vaultwarden/blob/ba5519167634ebe1e1f0fc10d610d10d1f405101/src/api/core/mod.rs#L208-L212

stefan0xC (changes requested):

src/config.rs
matt-aaron: With it being included in the 2026.2.1 release announcement, I think it'd be stable enough to enable directly

qianlongzt:
I’m just wondering why we don’t simply add a field to the `ciphers` table.

https://github.com/bitwarden/server/blob/main/src/Sql/dbo/Vault/Tables/Cipher.sql#L16
ref: https://github.com/bitwarden/server/pull/6578

matt-aaron:
> I’m just wondering why we don’t simply add a field to the `ciphers` table.
> 
> https://github.com/bitwarden/server/blob/main/src/Sql/dbo/Vault/Tables/Cipher.sql#L16 ref: [bitwarden/server#6578](https://github.com/bitwarden/server/pull/6578)

They originally did that, but later added the `archives` column to that table. This enables users the archives to be per-user instead of impacting other users within an org, for example.

matt-aaron:
Refactored to properly support importing items that were archived

src/config.rs
matt-aaron: I have enabled the flag directly and removed some of the older flags (keeping `email-verification` and `mobile-error-reporting`).

@BlackDex Let me know if this needs further adjustments

dcelasun:
@BlackDex any further thoughts on this? Seems fairly straightforward and previous comments were addressed.

BlackDex:
I need time to check and verify this, or @dani-garcia must have some time of course.
But I am busy with a lot of other stuff which needs attention, and for me, new feature are nice, especially this one, but they do come in a lower rank then some other items.

src/config.rs
```diff
@@ -1485,6 +1485,8 @@ pub const SUPPORTED_FEATURE_FLAGS: &[&str] = &[
     "cxp-export-mobile",
     // Platform Team
     "pm-30529-webauthn-related-origins",
+    // Innovation Team
+    "pm-19148-innovation-archive",
```
BlackDex: This is not needed when adding it as a default flag, since those can't be adjusted by users anyway.

src/db/schema.rs
```diff
@@ -341,6 +341,14 @@ table! {
...
 }
 
+table! {
+    archives (user_uuid, cipher_uuid) {
+        user_uuid -> Text,
+        cipher_uuid -> Text,
+        archived_at -> Timestamp,
+    }
+}
+
```
BlackDex, line 351: Looking at the migration sql's i think there are some `joinable!()` lines missing between the `archives` and `users`/`ciphers` tables. Without those Diesel doesn't treat those as joinable records during queries or might be unable to optimize the queries.

BlackDex (changes requested):

src/config.rs
matt-aaron: Ah, true. Removed

src/db/schema.rs
matt-aaron: Added the `joinable!()` lines and also included `archives` in the `allow_tables_to_appear_in_same_query!()` macro

src/config.rs
BlackDex: This is resolved i think, as this flag is no longer supported upstream.

migrations/mysql/2026-03-09-005927_add_archives/down.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE archives;
```
BlackDex: Probably best to add `IF EXISTS` here too.
Looks like we didn't do that in the past, but it's probably safer in case something went wrong trying to create it.

migrations/postgresql/2026-03-09-005927-add_archives/down.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE archives;
```
BlackDex: Same here.

migrations/sqlite/2026-03-09-005927-add_archives/down.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE archives;
```
BlackDex: And here

BlackDex (changes requested):

migrations/mysql/2026-03-09-005927_add_archives/down.sql
matt-aaron: Added to all 3

src/db/models/archive.rs
```diff
@@ -0,0 +1,97 @@
...
+                            .filter(archives::cipher_uuid.eq(cipher_uuid))
+                    )
+                    .execute(conn)
+                    .map_res("Error unarchiving")
+                }}
+            }
+            // Otherwise, the archived status is already what it should be
+            _ => Ok(()),
+        }
+    }
```
BlackDex: This whole function looks really complex.
It also causes extra database queries just to update a cipher.

There are better options to do this like how we update to which folder a cipher belongs.
https://github.com/dani-garcia/vaultwarden/blob/b04ed75f9f278c57ede12f078fca645584f8bec5/src/db/models/folder.rs#L145-L165

It's probably better to adapt something like that which is more idiomatic Rust and SQL.

Also, we prefer a separation between save and delete, this function seems to merged them.

Please split the two functions into a `fn save` and `fn delete_by_cipher` to better describe what is going to be done.

src/api/core/ciphers.rs
```diff
@@ -1966,12 +2079,16 @@ impl CipherSyncData {
 
                 // Generate a HashSet of all the Cipher UUID's which are marked as favorite
                 cipher_favorites = Favorite::get_all_cipher_uuid_by_user(user_id, conn).await.into_iter().collect();
+
+                // Generate a HashMap with the Cipher UUID as key and the archived date time as value
+                cipher_archives = Archive::find_by_user(user_id, conn).await.into_iter().collect();
             }
             // Organization Sync does not support Folders and Favorites.
```
BlackDex: Please adjust the comments to reflect the change below

src/api/core/ciphers.rs
```diff
@@ -1933,6 +1979,71 @@ async fn _delete_cipher_attachment_by_id(
...
+            &cipher.update_users_revision(conn).await,
+            &headers.device,
+            None,
+            conn,
+        )
+        .await;
+    }
+
+    Ok(Json(cipher.to_json(&headers.host, &headers.user.uuid, None, CipherSyncType::User, conn).await?))
+}
```
BlackDex: This function also has a dual function it looks like, and it uses a boolean to tell what it should do.
Please split this into two functions one for archiving and one for un-archiving, this makes it more clear what the intention is.

That probably also means you can remove the `Option` around the `archived_at` since that currently is only used to determine if it should archive or un-archive the item.

migrations/postgresql/2026-03-09-005927_add_archives/down.sql
BlackDex, line 1: Could you rename this file to use an `_` after `005927` and before `add`. This should be the same as like all other migration folders, and as the `mysql` branch  you already have named correctly

migrations/sqlite/2026-03-09-005927_add_archives/down.sql
BlackDex, line 1: Could you rename this file to use an `_` after `005927` and before `add`. This should be the same as like all other migration folders, and as the `mysql` branch  you already have named correctly

src/api/core/ciphers.rs
```diff
@@ -533,6 +538,17 @@ pub async fn update_cipher_from_data(
...
+        Some(dt_str) => match NaiveDateTime::parse_from_str(&dt_str, "%+") {
+            Ok(dt) => Some(dt),
+            Err(err) => {
+                warn!("Error parsing ArchivedDate '{dt_str}': {err}");
+                None
+            }
+        },
+        None => None,
+    };
+    cipher.set_archived_at(archived_at, &headers.user.uuid, conn).await?;
```
BlackDex: I think you can ignore this data. As far as i can tell the clients do not use this endpoint to archive or un-archive an item. This also adds extra database overhead which is probably not needed in this case.

BlackDex (changes requested):
I have taken a better look and this is what i found so far.
I might see other items later one, or after you have made some changed though.

src/api/core/ciphers.rs
matt-aaron: This is to re-archive the items when importing

src/db/models/archive.rs
matt-aaron: Split into save/delete fns

src/api/core/ciphers.rs
matt-aaron: Updated comments

src/api/core/ciphers.rs
matt-aaron: Split the functions

migrations/postgresql/2026-03-09-005927_add_archives/down.sql
matt-aaron: Renamed both

BlackDex (approved):
I have tested this using the v2026.3.1 web-vault and all seems to be working as far as i can test it.

dani-garcia (approved):

Snuupy:
has anyone been able to get this working on mobile clients? I see it on the web, browser extension and desktop but not mobile.

I also don't see an option to archive all selected items like as shown here: https://bitwarden.com/help/managing-items/#archive

matt-aaron:
> has anyone been able to get this working on mobile clients? I see it on the web, browser extension and desktop but not mobile.
> 
> 
> 
> I also don't see an option to archive all selected items like as shown here: https://bitwarden.com/help/managing-items/#archive

My phone took a while to pick up the new config. Either wait a few hours, or log out and back in to force it to fetch the updated server config.

BJReplay:
> or log out and back in to force it to fetch the updated server config.

That fixed it for me instantly.  Thanks for the hint!

Snuupy:
> My phone took a while to pick up the new config. Either wait a few hours, or log out and back in to force it to fetch the updated server config.

I actually did manually log out and log back in prior to posting my comment and my phone still does not show the archive option. On bitwarden on android (pixel 9) v2026.4.0 (21434). I just tried logging out and back in again (for a 2nd time), still no archive option.

BJReplay:
> I just tried logging out and back in again


My Vaultwarden server picked up the new release pretty quickly as it turns out, as it pulled the new version only an hour after it was released.  That's just coincidence in the timing of my watchtower schedule and the release, but it did mean that it had been running for a while when I first logged in this morning.

For me, my experience was:

- Desktop (windows) app seemed to have picked it up automagically (about 8 hours after the update).
- Browser extension required logout / log in to show the feature.
- Phone required log out / log in to show the feature, and on log in, popped up a pop-up with a splash with info about the new archive feature.

Phone Samsung S26, Android 16, same Bitwarden app version as you.

Snuupy:
> Phone required log out / log in to show the feature, and on log in, popped up a pop-up with a splash with info about the new archive feature.

@BJReplay Thanks! I played with it a bit more and got it working, here's what I had to do for my phone:

1. log out in app
2. go to multitask -> app info -> force stop
3. log in, select bitwarden.com server, then type in my self-hosted server again (not sure if this is required)
4. it works now! 🎉 

simply logging out and logging back in did not activate this feature for me, so thanks for the hint!

PR #6981, opened 2026-03-21
Update Feature Flags

BlackDex:
Added new feature flags which could be supported without issues. Removed all deprecated feature flags and only match supported flags. Do not error on invalid flags during load, but do on config save via admin interface. During load it will print a `WARNING`, this is to prevent breaking setups when flags are removed, but are still configured.

There are no feature flags anymore currently needed to be set by default, so those are removed now.

Also added the invalid flags into the Admin Diagnostics and Support string if they are filled, else the check is not shown.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -372,19 +372,22 @@
 ## Note that clients cache the /api/config endpoint for about 1 hour and it could take some time before they are enabled or disabled!
 ##
 ## The following flags are available:
-## - "pm-5594-safari-account-switching": Enable account switching in Safari. (Needs Safari >=2026.2.0)
-## - "inline-menu-positioning-improvements": Enable the use of inline menu password generator and identity suggestions in the browser extension.
-## - "inline-menu-totp": Enable the use of inline menu TOTP codes in the browser extension.
-## - "ssh-agent": Enable SSH agent support on Desktop. (Needs desktop >=2024.12.0)
-## - "ssh-key-vault-item": Enable the creation and use of SSH key vault items. (Needs clients >=2024.12.0)
-## - "pm-25373-windows-biometrics-v2": Enable the new implementation of biometrics on Windows. (Needs desktop >= 2025.11.0)
-## - "export-attachments": Enable support for exporting attachments (Clients >=2025.4.0)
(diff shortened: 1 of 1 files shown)


larena1:
Is this still WIP? There still seem to be some now invalid feature flags like ssh-agent and others in.

BlackDex:
> Is this still WIP? There still seem to be some now invalid feature flags like ssh-agent and others in.

If you check the links located near the feature flags you will see that those flags are still available for 2026.2.1. they only have been removed since 2026.3.0, which is only just released. And between 2026.2.0 and 2026.2.1 there were no significant flags added or removed which are relevant for Vaultwarden and thus the baseline is 2026.2.1 which we do not yet even have the web-vault for.

That is why those flags are still in the code.

larena1:
Are you in the right year? The SSH flags have been removed in 2025.3.

https://github.com/bitwarden/clients/pull/13506

And the base for the flags should be bundled web vault version?

BlackDex:
> Are you in the right year? The SSH flags have been removed in 2025.3.
> 
> [bitwarden/clients#13506](https://github.com/bitwarden/clients/pull/13506)
> 
> And the base for the flags should be bundled web vault version?

Yes I'm at the right tag, if you would have checked the links, but I'll post them here to the exact lines without a commit hash so you can see it.

https://github.com/bitwarden/server/blob/v2026.2.1/src/Core/Constants.cs#L174

https://github.com/bitwarden/server/blob/v2026.2.1/src/Core/Constants.cs#L200

Note that I'm looking mostly at the server list, as that is what Vaultwarden is mainly copying of course.

larena1:
Yeah, server still has them. Maybe they remove them there 12 months after they have been removed from clients, which seems quite long but understandable if you want to mimic this.

You might want to add ssh-agent-v2 though. I'm considering to give that a try to see if it fixes issues I'm seeing with the current agent and Dropbear.

BlackDex:
> You might want to add ssh-agent-v2 though. I'm considering to give that a try to see if it fixes issues I'm seeing with the current agent and Dropbear.

I added it, including checks at the Diagnostics page.

pquantin:
Itncould be great to update the .env.template file accordingly do that it helps discoverablity.

BlackDex:
> Itncould be great to update the .env.template file accordingly do that it helps discoverablity.

Good catch! I forgot that one.

dani-garcia (approved):

PR #6986, opened 2026-03-22
improve CI

TriplEight:
1. remove dead BASE_TAGS reference in release bake step.
  steps.determine-version doesn't exist in docker-build; the expression
  resolves to empty string. The HCL default (testing) would have
  applied, but it's moot - the bake uses push-by-digest=true so tags are
  only set in merge-manifests. Dead code.

2. replace unsecured curl hadolint download with an official action.
  hadolint/hadolint-action uses a Docker-based runner with hadolint
  pre-bundled in ghcr.io/hadolint/hadolint:v2.14.0-debian,so no binary
  downloaded at runtime. Pinning the action to a commit SHA covers the
  Dockerfile that specifies the image version, closing the supply-chain
  gap from the previous unverified curl | sudo install.

  Split {debian,alpine}: the action takes a single dockerfile argument,
  so debian and alpine are linted separately.

3. pin ubuntu-latest to ubuntu-24.04 in merge-manifests and zizmor.
  ubuntu-latest is a moving target that can silently change the runner OS
  on the next GitHub-side update. All other jobs in this repo already pin
  to ubuntu-24.04; this makes merge-manifests and zizmor consistent.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -188,6 +188,15 @@ jobs:
       # End Run cargo fmt
 
 
+      # Run cargo deny (supply-chain: advisories, licenses, bans, sources)
+      # Only run on rust-toolchain channel — deny.toml pins the feature set.
+      - name: "cargo deny"
+        id: deny
+        if: ${{ !cancelled() && matrix.channel == 'rust-toolchain' }}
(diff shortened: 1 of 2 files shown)


TriplEight:
closing - made #6991 cleaner

PR #6990, opened 2026-03-22
ci: add Dependabot config for github-actions and cargo.

TriplEight:
Weekly schedule (Mondays), 7-day cooldown, grouped updates to keep PR
noise low. Actions updates prefixed `ci`, Cargo prefixed `deps`. Major
version bumps for Cargo crates are ignored - those are for manualreview
given they can introduce breaking API or behaviour changes.

After merging this, go to settings -> Advanced security and turn everything on.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,39 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+      time: "06:00"
+      timezone: "Europe/Amsterdam"
+    open-pull-requests-limit: 10
+    cooldown:
+      days: 7
+    groups:
+      github-actions:
+        patterns:
+          - "*"
+    commit-message:
+      prefix: "ci"
+
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+      time: "06:00"
+      timezone: "Europe/Amsterdam"
+    open-pull-requests-limit: 10
+    cooldown:
+      days: 7
+    groups:
+      cargo-dependencies:
+        patterns:
+          - "*"
+    commit-message:
+      prefix: "deps"
+    ignore:
+      # Ignore major version bumps — review these manually
+      - dependency-name: "*"
+        update-types: ["version-update:semver-major"]


BlackDex:
We would like to manage these dependencies our self. There were other PR's which added this and we didn't merged them.

I personally want to check and verify the versions.

TriplEight:
Thanks for the clarification. I appreciate the concern about scope and maintenance cost, and I understand why that matters. I first searched any reasons of why did you discontinue using dependabot here in ~2023 and didn't find any.

PR #6993 is a concrete example from my side: after installing Dependabot, I caught a high severity vulnerability and opened that PR to address it (https://github.com/TriplEight/vaultwarden/security/dependabot/2)

That’s why I wanted to bring this up - not as extra busywork, but as something that can help surface important issues earlier. If this is still considered out of scope, I respect that - Feel free to close both.

TriplEight:
As a follow-up question - CodeQL scan found a bunch of issues as well:
image

BlackDex:
I appreciate the effort here, but we have closed these kind of PR's in the past.
From my point of view, this only adds a lot of noise and we have too much to do already to check and verify.
I also have found that sometimes breaking changes were part of a Dependabot update, and that only brings in more complexity during checking.

As mentioned, i would rather have 1 PR with multiple updates then having dependabot keep telling there are updates or security issues.

Therefor I'm going to close this one for now. If someone has a good argument for still adding this, please let us know and we might re-open.

TriplEight:
@BlackDex 
Supply-chain attacks are very popular these days, it's important to stay up-to-date with at least High and Critical tiers. I as an active user of Vaultwarden would sleep better knowing that everything is done for the security of my secrets. Hence the PR. Otherwise I'll have to keep tracking my fork that has them on.

Dependabot & CodeQL don't do your job reviewing, they just facilitate it and help with alerting - to prioritise if there's something urgent. With the grouping setup (and PR limit if you want to) - even less noisy. Otherwise - how do you know about new threats? - I don't see your local alerting setup.

BlackDex:
If you check our history, we have updated and kept track of this relatively fast from my point of view.
A warning that a package might need an update would probably be fine, but not every day or every hour or whatever.
Having a lot of dependabot PR's is noise in my opinion, especially if I'm going to check/vet those updates my self, and maybe also need to update other stuff. Then i need to close those PR's in favor of another PR etc.. etc...

It's just the workflow we currently use.

A CI Job which runs every week or so, or every 3 or 4 days might be fine too and errors out if there is something with a high or medium vulnerability.

But again, having noise of multiple PR's is just not what we want, we didn't disabled Dependabot just because, but we did that for a reason which still counts.

PR #6991, opened 2026-03-22
improve CI

TriplEight:
1. remove dead `BASE_TAGS` reference in release bake step.
  steps.determine-version doesn't exist in docker-build; the expression
  resolves to empty string. The HCL default (testing) would have
  applied, but it's moot - the bake uses `push-by-digest=true` so tags are
  only set in merge-manifests. Dead code.

2. replace unsecured curl hadolint download with an official action.
  `hadolint/hadolint-action` uses a Docker-based runner with hadolint
  pre-bundled in `ghcr.io/hadolint/hadolint:v2.14.0-debian`,so no binary
  downloaded at runtime. Pinning the action to a commit SHA covers the
  Dockerfile that specifies the image version, closing the supply-chain
  gap from the previous unverified `curl | sudo install`.

  Split `{debian,alpine}`: the action takes a single dockerfile argument,
  so debian and alpine are linted separately.

3. pin `ubuntu-latest` to `ubuntu-24.04` in `merge-manifests` and `zizmor`.
  `ubuntu-latest` is a moving target that can silently change the runner OS
  on the next GitHub-side update. All other jobs in this repo already pin
  to `ubuntu-24.04`; this makes `merge-manifests` and `zizmor` consistent.


.github/workflows/release.yml
```diff
@@ -183,7 +183,6 @@ jobs:
         id: bake_vw
         uses: docker/bake-action@82490499d2e5613fcead7e128237ef0b0ea210f7 # v7.0.0
         env:
-          BASE_TAGS: "${{ steps.determine-version.outputs.BASE_TAGS }}"
```
BlackDex, line 190: This should not be removed as it is used by bake.

.github/workflows/hadolint.yml
```diff
@@ -46,8 +38,17 @@ jobs:
       # End Checkout the repo
 
       # Test Dockerfiles with hadolint
-      - name: Run hadolint
-        run: hadolint docker/Dockerfile.{debian,alpine}
+      # Uses the Docker-based action (hadolint pre-bundled in ghcr.io/hadolint/hadolint:v2.14.0-debian)
+      # so no binary is downloaded at runtime. Pinned by commit SHA for supply-chain safety.
+      - name: Run hadolint on Dockerfile.debian
+        uses: hadolint/hadolint-action@2332a7b74a6de0dda2e2221d575162eba76ba5e5 # v3.3.0
```
BlackDex, line 44: Please use a lowercase `v`, which is what pinact would also use.

.github/workflows/hadolint.yml
```diff
@@ -46,8 +38,17 @@ jobs:
...
+        with:
+          dockerfile: docker/Dockerfile.debian
+
+      - name: Run hadolint on Dockerfile.alpine
+        uses: hadolint/hadolint-action@2332a7b74a6de0dda2e2221d575162eba76ba5e5 # v3.3.0
```
BlackDex, line 49: Same as above

BlackDex (changes requested):

TriplEight:
done @BlackDex

BlackDex (approved):

dani-garcia (approved):

PR #6992, opened 2026-03-22
cargo deny

TriplEight:
## Add cargo-deny config for supply-chain checks.

Covers advisories (RustSec), license compliance, duplicate detection,
  and source allowlist (crates.io only).

  License allowlist reflects the full transitive dep tree: MIT, Apache-2.0,
  ISC, BSD-*, 0BSD, Unlicense, Zlib, BSL-1.0, MPL-2.0, Unicode-3.0,
  LGPL-2.1-or-later (r-efi, Windows-only), CDLA-Permissive-2.0 (webpki-roots).

  Three known advisories are ignored:
  - RUSTSEC-2023-0071: rsa Marvin Attack, no upstream fix available
  - RUSTSEC-2025-0134: rustls-pemfile unmaintained, blocked on rustls upgrade
  - RUSTSEC-2026-0049: rustls-webpki CRL bug, fix blocked by rustls 0.21.x chain

  Duplicate versions are warned rather than denied - all are transitive.

# Run cargo deny in build workflow

Adds a cargo-deny step (EmbarkStudios/cargo-deny-action@v2)
  after clippy and fmt, gated to the rust-toolchain matrix channel only.
  Wires the outcome into the existing failure summary table.


0x484558:
> RUSTSEC-2023-0071: rsa Marvin Attack, no upstream fix available

What is the purpose of compliance check if they end up ignored? Just because it is unfixed upstream does not mean it should be allowed to pass. I think it is intentional and by design that compliance check on Vaultwarden must fail until this particular issue is resolved upstream, or until Vaultwarden ensures that its supply chain has no known-but-unpatched vulnerability.

Instead of silencing this otherwise legitimate issue, it should be possible to switch to AWS-LC library by swapping `"rust_crypto"` feature of `jsonwebtoken` to `"aws_lc_rs"` with no code changes. At least this will reduce the surface of impact on users who do not use OpenID Connect / SSO features.

BlackDex:
Switching to aws-lc for just one library does bring in a lot of dependencies, and probably also a lot of duplicates.

Then, also depending on which library or which depending library version it is it might not be that easy, of that version doesn't support that feature, or is depending on an older version which still is vulnerable for example.

Best would be, if possible create a PR for the library which is depending on an older version at that crates repo. And hope they still accept PR's.

Also, currently we mostly use rustls for almost everything.
The only crates using openssl libraries are MySQL and PostgreSQL i think. This is the same for both Debian and Alpine builds.

I would like to prevent multiple different TLS/SSL libraries, that might even make it more difficult to keep track of security risks.

TriplEight:
@BlackDex  shall I remove advisories.ignore for this PR?

BlackDex:
I took some more time to check this. I was looking at cargo-deny my self last month or earlier maybe, and didn't come around adding it for some reason.

deny.toml
```diff
@@ -0,0 +1,75 @@
...
+    "BSD-3-Clause",
+    "BSL-1.0",           # Boost Software License (ryu)
+    "CDLA-Permissive-2.0", # Community Data License Agreement (webpki-roots)
+    "ISC",
+    "LGPL-2.1-or-later", # r-efi (Windows-only UEFI crate; dynamically linked)
+    "MIT",
+    "MPL-2.0",           # webauthn-rs crates (file-level copyleft, compatible)
+    "Unicode-3.0",
+    "Unlicense",
+    "Zlib",
```
BlackDex, line 46: This list can be shorter if I'm correct. I see no need in adding not used license forms.
The comments behind the forms are not really needed i think.

The list i came-up with would be this, and that should be enough.
```
    "0BSD",
    "Apache-2.0",
    "BSD-3-Clause",
    "CC0-1.0",
    "CDLA-Permissive-2.0",
    "ISC",
    "MIT",
    "MPL-2.0",
    "Unicode-3.0",
    "Zlib",
```

deny.toml
```diff
@@ -0,0 +1,75 @@
...
+# Skip license checks for unpublished workspace crates (vaultwarden, macros).
+ignore = true
+
+# =============================================================================
+# Bans — duplicate versions and disallowed crates
+# =============================================================================
+[bans]
+# Warn on duplicate versions; many arise from transitive deps and require
+# upstream changes to resolve. Promote to "deny" once the dep tree stabilises.
+multiple-versions = "warn"
```
BlackDex: I would like to set this to `allow`. There are unfortunately a lot of crate duplicates, and that will be impossible to fix, and therefor probably not as useful to report this every time.

deny.toml
```diff
@@ -0,0 +1,75 @@
...
+# Warn on wildcard version requirements. "deny" would also flag internal
+# workspace path dependencies (macros = { path = "./macros" }), which have
+# no version specifier by design, so "warn" is used instead.
+wildcards = "warn"
```
BlackDex: I would like to set this to `deny`. You can add `allow-wildcard-paths = true` which should prevent erroring out on the `macro` path crate.

deny.toml
```diff
@@ -0,0 +1,75 @@
...
+
+    # `rustls-pemfile` v1.x is unmaintained (archived Aug 2025). It is a thin
+    # wrapper around rustls-pki-types and pulled in transitively via rustls 0.21.x.
+    # Removing it requires upstream crates to drop their rustls 0.21 dependency.
+    { id = "RUSTSEC-2025-0134", reason = "Transitive via rustls 0.21.x compat chain; blocked on upstream upgrade." },
+
+    # CRL validation bug in rustls-webpki: fix requires >=0.103.10, but rustls-webpki
+    # 0.101.x is pinned by the rustls 0.21.x compatibility stack.
+    # Track upstream rustls upgrade to resolve both occurrences.
+    { id = "RUSTSEC-2026-0049", reason = "rustls-webpki 0.101.x locked by rustls 0.21.x compat chain; track upstream rustls upgrade." },
```
BlackDex: This exclusion isn't needed it seems, as it does not exists or not being triggered for our repo.

.github/workflows/build.yml
```diff
@@ -188,6 +188,15 @@ jobs:
       # End Run cargo fmt
 
 
+      # Run cargo deny (supply-chain: advisories, licenses, bans, sources)
+      # Only run on rust-toolchain channel — deny.toml pins the feature set.
+      - name: "cargo deny"
+        id: deny
+        if: ${{ !cancelled() && matrix.channel == 'rust-toolchain' }}
+        uses: EmbarkStudios/cargo-deny-action@3fd3802e88374d3fe9159b834c7714ec57d6c979 # v2
```
BlackDex: Could you update to `175dc7fd4fb85ec8f46948fb98f44db001149081` and make sure the comment is referring to the exact version, that is also what pinact would use, so with this new hash that would be `v2.0.16` and also use a lowercase `v`.

BlackDex (changes requested):

TriplEight:
@0x484558 @BlackDex The `RUSTSEC-2023-0071` ignore stays for now. Switching to `aws-lc` was considered but BlackDex noted it pulls in a large dependency chain for a single crate. The exposure is narrow (JWT RS256 signing, not network-facing decryption). When an upstream fix lands in the `rsa` crate or `jsonwebtoken` migrates away from it, this entry comes out.

TriplEight:
@BlackDex addressed all of your review points

BlackDex (changes requested):
Looks like we need to add a new one `RUSTSEC-2026-0099`

TriplEight:
@BlackDex here you go, added a bit more.

PR #6994, opened 2026-03-22
ci: create GitHub Release with binaries on tag push.

TriplEight:
Adds a create-release job that runs after merge-manifests completes,
  gated to tag pushes only. Downloads all 8 binary artifacts (4 arch ×
  2 base_image), renames them to release-friendly names, and publishes
  a GitHub Release via gh CLI with auto-generated notes.

  Asset naming convention:
    vaultwarden-linux-amd64        (debian, glibc)
    vaultwarden-linux-amd64-musl   (alpine, musl)
    vaultwarden-linux-armv6
    vaultwarden-linux-armv6-musl

  Replaces 1-day-retention Actions artifacts as the permanent home for
  release binaries. Uses --generate-notes for auto changelog from PRs
  and commits since the previous tag.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -379,3 +379,55 @@ jobs:
           subject-name: ${{ vars.QUAY_REPO }}
           subject-digest: ${{ env.DIGEST_SHA }}
           push-to-registry: true
+
+  create-release:
+    name: Create GitHub Release
+    runs-on: ubuntu-24.04
+    needs: merge-manifests
+    # Only create a release when a version tag is pushed, not on branch pushes.
+    if: ${{ github.ref_type == 'tag' }}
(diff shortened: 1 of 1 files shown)


BlackDex:
I'm not sure if we want to release these via the release pages. The main thing is that the container images have everything included to ensure a propper working environment, that includes the web-vault for example, which we have tested and should be working fine.

Providing these binaries from the release page might cause more trouble for users who think it's a download and run binary, and i want to prevent that.

Those binaries are only extracted/uploaded during the release process as artifacts for possible troubleshooting.

If there are good use cases for this and other users might also want this, it's something to think about.
But, for now, I'm not really a fan of adding this.

TriplEight:
Fair, I agree, let's close it.

PR #6995, opened 2026-03-22
ci: add daily scheduled RustSec advisory scan

TriplEight:
cargo deny check advisories runs daily at 11:17 UTC (offset from Trivy
at 11:08 to spread DB load). This catches new CVEs published against
Cargo dependencies between code pushes, matching the coverage Trivy
provides for container vulnerabilities.

Runs advisories-only so duplicate/license warnings don't obscure newly
published CVEs - the full cargo deny check already runs in build.yml on
every push/PR. Guarded to the upstream repo to avoid fork hammering the
RustSec DB. Supports workflow_dispatch for manual triggering.

### DIFF
--- a/.github/workflows/cargo-deny-scheduled.yml
+++ b/.github/workflows/cargo-deny-scheduled.yml
@@ -1,0 +1,41 @@
+name: Scheduled Advisory Scan
+permissions: {}
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: false
+
+on:
+  schedule:
+    # Daily at 11:17 UTC — offset from Trivy (11:08) to spread advisory DB load.
+    - cron: '17 11 * * *'
+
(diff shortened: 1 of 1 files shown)


hqnicolas:
[Please take care guis](https://snyk.io/pt-BR/articles/poisoned-security-scanner-backdooring-litellm/)

TriplEight:
@dani-garcia @BlackDex anything you'd like me to improve here?

BlackDex:
This will only produce noise from my point of view. I understand that it could warn on potential issues.
Also, there are 2 advisories which will be triggered every day which we can not fix.

That would mean we either need to add them to an exclude list, which is also not very nice of course, but that is then the only option. I'm not sure we want to enable this in the current form.

BlackDex:
As mentioned before, having this as a daily check is too much i think, therefor going to close this for now.

PR #6998, opened 2026-03-22
ci: add integration test workflow with live PostgreSQL and MySQL

TriplEight:
## Two jobs - one per db - each spinning up a service container

1. diesel migration run --migration-dir migrations/<db> Verifies every migration applies cleanly to a fresh instance. This is the primary gap: a broken migration would otherwise only surface on production deployment.

2. cargo test --features <db> with DATABASE_URL set Builds and runs the test suite against the live engine. Existing tests are unit-level (no DB access), but DATABASE_URL is wired in so any future integration tests work without further infrastructure changes.

Service containers: postgres:16, mysql:8.4 (utf8mb4).
diesel CLI binary is cached keyed on Cargo.lock hash to avoid recompiling it on every run.
Triggers on the same path set as build.yml (src/**, migrations/**, Cargo.*, rust-toolchain.toml).

## Two bugs exposed by the integration test workflow

1. MySQL migration: groups_users and collections_groups used UNIQUE instead of PRIMARY KEY. Diesel requires primary keys on all tables for schema introspection (print-schema). PostgreSQL and SQLite migrations already used PRIMARY KEY correctly.

2. diesel.toml has [print_schema] configured, so `diesel migration run` rewrites src/db/schema.rs with backend-specific types after running. This corrupted the checked-in schema before cargo test could compile, causing E0277 CompatibleType errors for every query.

## add MySQL migration to fix missing primary keys on group tables

groups_users and collections_groups were created with UNIQUE constraints instead of PRIMARY KEY in the 2022-07-27 migration. Add a new migration that promotes the unique indexes to primary keys.
  
MySQL auto-names a UNIQUE constraint after its first column, so the index names to drop are 'groups_uuid' and 'collections_uuid' respectively.


BlackDex:
I'm not sure this really is needed here.
I have 0 issues with MySQL v8 if i disable foreign-key checks, which is needed during migrations, and the diesel-migrations isn't doing this as far as i know.

For some reason I'm not able to let that same command work nicely on a MariaDB container either v11 or v12.
But starting Vaultwarden using those same containers works just fine. I think MariaDB did something strange with the foreign-key checks and ignores them in some cases.

TriplEight:
@BlackDex thanks for looking into this.
Are you saying that if the original migration chain only succeeds in MySQL or MariaDB when foreign key checks are disabled, then this CI job could fail for reasons unrelated to the new primary-key migration? So you mean fixing the migration issues was unnecessary? It just stopped me adding a CI pipeline.

The CI job is meant to verify that the raw Diesel migration chain can be replayed on a fresh DB, which is stricter than “Vaultwarden starts successfully” because startup can hide backend-specific migration behaviour. So if MySQL/MariaDB needs FK checks disabled for part of the chain, that is a valid concern about the workflow path, but it is separate from the schema bug this PR also fixes.

The `groups_users` / `collections_groups` change itself looks sensible: those are join tables, and using a composite `PRIMARY KEY` is the right shape. Diesel CLI does not properly handle tables without primary keys when generating schema, so the existing MySQL-only `UNIQUE` definition is a real inconsistency that should be fixed regardless of whether the integration workflow needs an extra FK-check workaround.

- new MySQL migration is valid on its own.
- integration workflow is useful
- and if the migration runner also needs f`oreign_key_checks=0` for MySQL/MariaDB replay-from-scratch, that should probably be handled explicitly as a follow-up

BlackDex:
I'm working on some changes regarding playwright, those tests will be better then just a single database test.
Therefor going to close this.

PR #7008, opened 2026-03-24
Handle `SIGTERM` and `SIGQUIT` shutdown signals.

0x484558:
By default, graceful shutdown is wired to `ctrl_c()` which only handles `SIGINT`. Service can also be terminated with `SIGQUIT` and `SIGTERM` and without handling them, the app will not terminate gracefully.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -558,6 +558,11 @@ async fn launch_rocket(pool: db::DbPool, extra_debug: bool) -> Result<(), Error>
     let basepath = &CONFIG.domain_path();
 
     let mut config = rocket::Config::from(rocket::Config::figment());
+
+    // We install our own signal handlers below; disable Rocket's built-in handlers
+    config.shutdown.ctrlc = false;
+    config.shutdown.signals.clear();
+
     config.temp_dir = canonicalize(CONFIG.tmp_folder()).unwrap().into();
     config.cli_colors = false; // Make sure Rocket does not color any values for logging.
     config.limits = Limits::new()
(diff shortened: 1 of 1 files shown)


0x484558:
Currently this just leads to graceful Rocket termination, but it is still desirable and important for any more sophisticated graceful shutdown infrastructure going forward, which is especially relevant on serverless platforms that may scale instances of Vaultwarden to zero and shut them down with these signals instead of `SIGINT`.

BlackDex:
Rocket already handled `SIGTERM` it looks like, at least, if i test using that signal, it even mentions it is stopping like that. But, besides that, the `SIGQUIT` isn't handled now, and mostly only used in cases where the application doesn't want to stop. It wouldn't hurt that much to add this i think.

src/main.rs
```diff
@@ -621,6 +617,35 @@ async fn launch_rocket(pool: db::DbPool, extra_debug: bool) -> Result<(), Error>
...
+        let mut sigterm = signal(SignalKind::terminate()).expect("Error setting SIGTERM handler");
+        let mut sigquit = signal(SignalKind::quit()).expect("Error setting SIGQUIT handler");
+
+        let signal_name = tokio::select! {
+            _ = sigint.recv() => "SIGINT",
+            _ = sigterm.recv() => "SIGTERM",
+            _ = sigquit.recv() => "SIGQUIT",
+        };
+
+        info!("Received {signal_name}, starting graceful shutdown");
```
BlackDex: I find using `starting` in a shutdown sentence a bit weird. And i think it better to use something like `initiating` probably.

So `initiating graceful shutdown` it probably better.

0x484558: I've changed the message as requested and rebased latest changes onto my branch.

BlackDex (approved):

PR #7013, opened 2026-03-24
Do not display unavailable 2FA options

0x484558:
- Prevent YubiKey option availability if Yubico credentials are not configured; 
- Prevent Duo option availability if there is no valid per-user data or fully configured global Duo keys; 
- Prevent WebAuthn if domain does not match one that generates challenge payload; 
- Prevent E-mail option if email 2FA is disabled by admin.

### DIFF
--- a/src/api/core/two_factor/mod.rs
+++ b/src/api/core/two_factor/mod.rs
@@ -56,7 +56,7 @@ pub fn is_twofactor_provider_usable(provider_type: i32, provider_data: Option<&s
         x if x == TwoFactorType::YubiKey as i32 => {
             CONFIG._enable_yubico() && CONFIG.yubico_client_id().is_some() && CONFIG.yubico_secret_key().is_some()
         }
-        x if x == TwoFactorType::Webauthn as i32 => CONFIG.domain_set(),
+        x if x == TwoFactorType::Webauthn as i32 => CONFIG.is_webauthn_2fa_supported(),
         x if x == TwoFactorType::Remember as i32 => !CONFIG.disable_2fa_remember(),
         x if x == TwoFactorType::RecoveryCode as i32 => true,
         _ => false,
--- a/src/api/core/two_factor/mod.rs
+++ b/src/api/core/two_factor/mod.rs
@@ -2,6 +2,7 @@ use chrono::{TimeDelta, Utc};
 use data_encoding::BASE32;
 use rocket::serde::json::Json;
 use rocket::Route;
+use serde::Deserialize;
 use serde_json::Value;
 
 use crate::{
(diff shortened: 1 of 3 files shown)


src/api/core/two_factor/mod.rs
```diff
@@ -31,6 +32,37 @@ pub mod protected_actions;
...
+
+pub fn is_twofactor_provider_usable(provider_type: i32, provider_data: Option<&str>) -> bool {
+    #[derive(Deserialize)]
+    struct DuoProviderData {
+        host: String,
+        ik: String,
+        sk: String,
+    }
+
+    match provider_type {
```
BlackDex, line 48: I think we can use `match TwoFactorType::from_i32(provider_type)` here, it will cause the arms to have Some(x) but i think it will be more clear. And we already use this format in to other places, so it would make it more consistent.

Also, i think it will catch new types if those are not defined here. For that to work you need to remove the `_`/`Some(_)` and add a `None` i think from the top of my head.

Maybe it is even possible to adjust the function call to only accept the `TwoFactorType` and convert it from an `i32` at the call site? That would make it even more nice.

src/api/core/two_factor/mod.rs
```diff
@@ -53,7 +85,11 @@ pub fn routes() -> Vec<Route> {
 #[get("/two-factor")]
 async fn get_twofactor(headers: Headers, conn: DbConn) -> Json<Value> {
     let twofactors = TwoFactor::find_by_user(&headers.user.uuid, &conn).await;
-    let twofactors_json: Vec<Value> = twofactors.iter().map(TwoFactor::to_json_provider).collect();
+    let twofactors_json: Vec<Value> = twofactors
+        .iter()
+        .filter(|tf| is_twofactor_provider_usable(tf.atype, Some(&tf.data)))
+        .map(TwoFactor::to_json_provider)
```
BlackDex: I think you can merge the filter and map into one call, that would be better.

src/api/identity.rs
```diff
@@ -739,8 +741,22 @@ async fn twofactor_auth(
 
     TwoFactorIncomplete::mark_incomplete(&user.uuid, &device.uuid, &device.name, device.atype, ip, conn).await?;
 
-    let twofactor_ids: Vec<_> = twofactors.iter().map(|tf| tf.atype).collect();
+    let twofactor_ids: Vec<_> = twofactors
+        .iter()
+        .filter(|tf| tf.enabled && is_twofactor_provider_usable(tf.atype, Some(&tf.data)))
+        .map(|tf| tf.atype)
```
BlackDex: And the same here, merge filter and map.

BlackDex (changes requested):

src/api/core/two_factor/mod.rs
0x484558: Oh right, I should have done that from the start.

src/api/core/two_factor/mod.rs
0x484558: I've changed to filter_map but kept inner part only half-way towards fully functional form to keep it human-comprehensible.

src/api/identity.rs
0x484558: I've changed to filter_map but kept inner part only half-way towards fully functional form to keep it human-comprehensible.

BlackDex (approved):

dani-garcia (approved):

PR #7033, opened 2026-03-29
Disable deployments for release env

dfunkt:
As according to the docs:
https://docs.github.com/en/actions/how-tos/deploy/configure-and-manage-deployments/control-deployments#using-environments-without-deployments

> This is useful when you want to use environments for:
> 
> Organizing secrets—group related secrets under an environment name without creating deployment records.
> Access control—restrict which branches can use certain secrets via environment branch policies, without deployment tracking.
> CI and testing jobs—reference an environment for its configuration without adding noise to the deployment history.


BlackDex:
I think the envs are moved already by @dani-garcia

PR #7034, opened 2026-03-29
Add trusted device verification

rwjack:
This kind of sucks. I was hoping to make it work so the whole thing is idiot proof.

I guess friends and family will still have to type their master password at least once, and will still have to remember some sort of pin at the very least, for unlocking the extension, but at least this eliminates the need for typing the master password multiple times during new device registration for example.

### DIFF
--- a/src/api/user_decryption.rs
+++ b/src/api/user_decryption.rs
@@ -1,0 +1,218 @@
+//! `UserDecryptionOptions` (login) and `userDecryption` (sync) payloads for Bitwarden-compatible clients.
+//!
+//! References: Bitwarden `UserDecryptionOptionsBuilder`, `TrustedDeviceUserDecryptionOption`, and
+//! `libs/common/.../user-decryption-options.response.ts` in bitwarden/clients.
+
+use serde_json::{json, Value};
+
+use crate::db::models::{Device, Membership, SsoUser, User, UserId};
+use crate::db::DbConn;
+use crate::CONFIG;
+
+/// Device types that may approve “login with device” / trusted-device flows (see Bitwarden `LoginApprovingClientTypes`).
+pub fn device_type_can_approve_trusted_login(atype: i32) -> bool {
+    !matches!(atype, 21..=25) // SDK, Server, CLIs
+}
+
+async fn has_login_approving_device(user_uuid: &UserId, current: &Device, conn: &DbConn) -> bool {
+    let devices = Device::find_by_user(user_uuid, conn).await;
+    devices.iter().any(|d| {
+        d.uuid != current.uuid
+            && device_type_can_approve_trusted_login(d.atype)
+    })
+}
+
+fn has_valid_reset_password_key(m: &Membership) -> bool {
+    m.reset_password_key.as_ref().is_some_and(|s| !s.trim().is_empty())
+}
+
+/// Owner or Admin (Vaultwarden does not persist custom-role JSON for `manageResetPassword` on members).
+fn membership_has_manage_reset_password(m: &Membership) -> bool {
+    matches!(m.atype, 0 | 1)
+}
+
+async fn aggregate_trusted_device_flags(user: &User, device: &Device, conn: &DbConn) -> (bool, bool, bool) {
+    let members = Membership::find_confirmed_by_user(&user.uuid, conn).await;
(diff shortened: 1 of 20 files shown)


rwjack:
I'd like to work on an open source version of the key connector (making it from scratch basically), but I have a crucial question before I start. Do you guys think this should be an external service/container, or does it make more sense to build it into VW?

My goal is to have SSO only logins, while only prompting the user for a master password during account creation.

BlackDex:
> I'd like to work on an open source version of the key connector (making it from scratch basically), but I have a crucial question before I start. Do you guys think this should be an external service/container, or does it make more sense to build it into VW?
> 
> My goal is to have SSO only logins, while only prompting the user for a master password during account creation.

I would suggest to read the comments posted here: https://github.com/dani-garcia/vaultwarden/pull/6949

In my opinion, Key Connector shouldn't be something supported by Vaultwarden. The chances of having security issues in either Vaultwarden or the custom Key Connector are always possible and makes the whole logic keeping everything safe and encrypted almost nill.

src/util.rs
```diff
@@ -100,7 +100,7 @@ impl Fairing for AppHeaders {
                     form-action 'self'; \
                     media-src 'self'; \
                     object-src 'self' blob:; \
-                    script-src 'self' 'wasm-unsafe-eval'; \
+                    script-src 'self' 'wasm-unsafe-eval' 'sha256-ZswfTY7H35rbv8WC7NXBoiC7WNu86vSzCDChNWwZZDM='; \
```
stefan0xC: What script do you allow here? Is this an intentional change? If so this should be properly documented.

rwjack:
> I would suggest to read the comments posted here: #6949
> 
> In my opinion, Key Connector shouldn't be something supported by Vaultwarden. The chances of having security issues in either Vaultwarden or the custom Key Connector are always possible and makes the whole logic keeping everything safe and encrypted almost nill.

Hmmm, I see. I totally agree with the security implications. My main use case is to have "something of value", so I can have better arguments while convincing friends and family to switch from cloud based password managers. As you already know, those "just work", and nobody cares to enter a password these days, let alone be forced to do so on every log in. 

This is why I consider my local only VW instance, even with a "potentially vulnerable" key-connector, to still be a better option, while literally being a better option user-flow wise (Paired with PocketID in my case, which is passkeys only, so it can't get more idiot proof than that).

Now, regarding my security (and the security of other like-minded folks out there), the idea was to have a per user toggle, which would allow/deny key connector usage for that account - making the whole thing optional, while being a cool feature nevertheless. I'm sure I wouldn't use the key connector for my main VW account, but for the phone account, friends, family - it just makes sense.

And of course a global env var would be required to turn on the feature at all. Either way, architecture wise, I'm leaning towards keeping the key connector in a separate container / as a separate service.

My main dilemma is that the per-user config toggle would likely require modifying the settings page, and since I want this change to be upstreamable, I know UI patches create potential maintenance problems down the road, so I still haven't figured out if that toggle (or an entire custom user settings page, if there's more than one setting we'd like to control) should be separate from the original BW web vault, or if I could just patch it in there.

src/util.rs
rwjack: Forgot to remove that, had some errors in the console, so I was just testing if that would stop them.

Timshel:
Hey,
Played with the PR, broke some stuff and discovered it's missing the `/devices/lost-trust` endpoint.

Tested changing the activation condition to require only that the account already has a masterPassword which make using the feature possible without having to setup an Organization.

~~But wonder if it might make sense to gate it behind a separate setting and keep the official logic as the default in case the instance admin does not really trust the users to store their master password (since it's almost never required it increase the risk of losing it).~~

src/api/user_decryption.rs
```diff
@@ -0,0 +1,218 @@
...
+                .filter(|s| !s.is_empty())
+                .map(|s| json!(s))
+                .unwrap_or(Value::Null),
+        )
+    } else {
+        (Value::Null, Value::Null)
+    };
+
+    json!({
+        "HasAdminApproval": has_admin_approval,
```
Timshel, line 77: Should probably always be  `false` since it's missing the logic to handle admin approval.

anthonyalayo:
Hopefully either this or the derivative PR make it in. This would be a great feature for everyone.

PR #7042, opened 2026-03-31
Create build-macos-m4.yml

yaogh99123:


BlackDex:
We do not really have a large audience for macOS and also not specifically for the M4 CPU.

Adding this will only increase our footprint into running build containers without too much usage.

This also would mean we need to maintain this, which would add more maintenance burden.

BlackDex:
Closing this PR since we will not start building for specific macOS systems.

PR #7047, opened 2026-04-01
Fix logout push identifiers and send logout before clearing devices

qaz741wsd856:
This fixes two issues in the logout push flow.

1. `push_logout()` currently sends the acting device UUID in both `deviceId` and `identifier`. That does not match the other push entry points, where `deviceId` is the push UUID and `identifier` is the device UUID.

2. `post_sstamp()` and `disable_user()` delete all devices **before** calling `send_logout()`. Since `check_user_has_push_device()` filter was introduced in #3578, those flows will always skip the push because the device lookup returns `false` after the delete.

This patch:

- passes the acting `Device` through `send_logout()` into `push_logout()`
- sends logout pushes with `deviceId = device.push_uuid` and `identifier = device.uuid`
- moves `Device::delete_all_by_user()` in `post_sstamp()` and `disable_user()` to after logout sending
- leaves `deauth_user()` unchanged, because it already sends the logout notification before deleting devices

### DIFF
--- a/src/api/notifications.rs
+++ b/src/api/notifications.rs
@@ -358,7 +358,7 @@ impl WebSocketUsers {
         }
     }
 
-    pub async fn send_logout(&self, user: &User, device: Option<&Device>, conn: &DbConn) {
+    pub async fn send_logout(&self, user: &User, acting_device: Option<&Device>, conn: &DbConn) {
         // Skip any processing if both WebSockets and Push are not active
         if *NOTIFICATIONS_DISABLED {
             return;
--- a/src/api/notifications.rs
+++ b/src/api/notifications.rs
@@ -363,7 +363,7 @@ impl WebSocketUsers {
         if *NOTIFICATIONS_DISABLED {
             return;
         }
-        let acting_device_id = device.map(|d| d.uuid.clone());
+        let acting_device_id = acting_device.map(|d| d.uuid.clone());
         let data = create_update(
             vec![("UserId".into(), user.uuid.to_string().into()), ("Date".into(), serialize_date(user.updated_at))],
             UpdateType::LogOut,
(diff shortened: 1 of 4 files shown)


src/api/notifications.rs
```diff
@@ -358,23 +358,24 @@ impl WebSocketUsers {
         }
     }
 
-    pub async fn send_logout(&self, user: &User, acting_device_id: Option<DeviceId>, conn: &DbConn) {
+    pub async fn send_logout(&self, user: &User, device: Option<&Device>, conn: &DbConn) {
```
BlackDex: I would prefer to keep the variable name having `acting` to make it clear it should be the current device making the request.

src/api/push.rs
```diff
@@ -188,15 +188,13 @@ pub async fn push_cipher_update(ut: UpdateType, cipher: &Cipher, device: &Device
     }
 }
 
-pub async fn push_logout(user: &User, acting_device_id: Option<DeviceId>, conn: &DbConn) {
-    let acting_device_id: Value = acting_device_id.map(|v| v.to_string().into()).unwrap_or_else(|| Value::Null);
-
+pub async fn push_logout(user: &User, device: Option<&Device>, conn: &DbConn) {
```
BlackDex: And the same here, so `acting_device` is fine 😄

BlackDex (changes requested):
Thanks for you contribution.

I'm almost certain that Bitwarden did it different before too, but i could be wrong, and it would also be strange if they did, and would probably be a bug then.

Just one small remark if you could adjust that.

qaz741wsd856:
Sure. I've changed it.

BlackDex:
> Sure. I've changed it.

But you forgot to change the variable inside the functions it self I think.
I did started the workflow testing, but those will probably fail.

qaz741wsd856:
SRY, might something wrong with the rename function in my editor.

Fixed.

BlackDex (approved):

dani-garcia (approved):

PR #7057, opened 2026-04-06
Allow SQLite to be linked against dynamically

ISSOtm:
Keeping the default behaviour of SQLite being built statically, so as not to break anyone's workflow, but allowing for downstream packagers to link dynamically against SQLite (where it's fine because that's the point of package managers).

Note that SQLite is still *not* enabled by default, thanks to the `?` operator.


BlackDex (changes requested):
This could still break someones workflow if they use `--no-default-features`
Also, i would like to have no features enabled by default at all, since that might also cause weird issues for other users or package maintainers.

While i have to say, i have no idea why the `bundled` feature is used by default since the start of Vaultwarden, maybe @dani-garcia can still remember why?

But, that doesn't matter for the current outcome.

I would rather have a new feature `sqlite_system` which is a current copy of the `sqlite` feature, and than adjust the `sqlite` feature to add `"libsqlite3-sys/bundled"`, and remove that feature from the `libsqlite3-sys` crate config.

That would make this a truly non-breaking change.

Edit:
That does mean you also need to adjust the `build.rs` and i think you can still set the `println!("cargo:rustc-cfg=sqlite");` if the `sqlite_system` feature is enabled, which should keep everything working in the code-base.

Cargo.toml
```diff
@@ -94,8 +96,10 @@ diesel_migrations = "2.3.1"
 derive_more = { version = "2.1.1", features = ["from", "into", "as_ref", "deref", "display"] }
 diesel-derive-newtype = "2.1.2"
 
-# Bundled/Static SQLite
-libsqlite3-sys = { version = "0.36.0", features = ["bundled"], optional = true }
+# SQLite, statically compiled if the `sqlite_bundled` feature is enabled
+# We do not depend on a specific version ourselves, instead that is done by downstream crates;
+# this is only to toggle a feature for this crate
+libsqlite3-sys = { version = "*", optional = true }
```
BlackDex: That is not going to happen, we force specific versions, also Diesel doesn't allow a wide range of versions anyway. So, please revert this.

BlackDex (changes requested):

dani-garcia:
> While i have to say, i have no idea why the bundled feature is used by default since the start of Vaultwarden, maybe @dani-garcia can still remember why?

My guess is we had issues with cross-compilation using the system libs and that was the easiest way to fix them. Just followed the recommendation on diesel's page: https://docs.rs/diesel/2.3.7/diesel/#crate-feature-flags

Cargo.toml
ISSOtm: Well, as the comment says this only removes the requirement placed by `vaultwarden` itself, but the version is still constrained by Diesel, and thus resolves just like before. Put another way, this is no override, and cannot cause breakage with Diesel. This blanket requirement is fine for `vaultwarden` itself, because it does not depend directly on `libsqlite3-sys`.

Now, if the goal is to enforce a minimum version in `vaultwarden` because e.g. some earlier versions of `libsqlite3-sys` are known to have vulns or whatnot, then that would make sense, but it would be a good idea to write down that rationale in a comment ^^

ISSOtm:
Sure, I can add `sqlite_system`, but it would cause dependency duplication with the current `sqlite` feature:

```toml
sqlite = ["diesel/sqlite", "diesel_migrations/sqlite", "libsqlite3-sys/bundled"]
sqlite_system = ["diesel/sqlite", "diesel_migrations/sqlite", "dep:libsqlite3-sys"]
```

This could be solved with an intermediate feature:

```toml
__sqlite_common = ["diesel/sqlite", "diesel_migrations/sqlite"]
sqlite = ["__sqlite_common", "libsqlite3-sys/bundled"]
sqlite_system = ["__sqlite_common", "dep:libsqlite3-sys"]
```

...but I'm not sure this is better. I'll let you choose, though—personally, I'd feel fine with `--no-default-features` since that set is currently empty and thus I guess the likelihood of practical breakage to be low, but that's my own opinion and I'll conform to your final say.

Cargo.toml
BlackDex: We want to show the version used in our Cargo.toml. We never use `*` for anything.

Even though Vaultwarden it self doesn't depend on that crate, it does define the required and wanted version. Having it set to `*` leaves it open to anything.

So, to keep everything clear which versions are used or at least required as a minimum please revert that commit.

BlackDex:
> Sure, I can add `sqlite_system`, but it would cause dependency duplication with the current `sqlite` feature:
How would this cause dependency duplication?

The other option you provided would work too, but it can be adjusted a bit i think.
Though, not sure if it would all work great with the current build checks and our `#[cfg(sqlite)]` flags in the code.

ISSOtm:
Sorry, I meant that it would cause duplication *in the dependency specifications*, not of the deps (their code, ...) themselves. That was poor wording ^^'

I can go with the `__sqlite_common` route and use that in the `build.rs`, then.

Cargo.toml
ISSOtm: Even with `"*"`, `vaultwarden`'s `Cargo.toml` does define the required and wanted version, through `diesel`. Another example is how Vaultwarden depends on the `hex` crate, but without defining its version explicitly:

```console
.../vaultwarden% cargo tree --invert hex
hex v0.4.3
└── webauthn-rs-core v0.5.4
    ├── vaultwarden v1.0.0 (.../vaultwarden)
    └── webauthn-rs v0.5.4
        └── vaultwarden v1.0.0 (.../vaultwarden)
```

Here, the version is constrained [by `webauthn-rs-core`](https://github.com/kanidm/webauthn-rs/blob/197cddf8487a2dbdd5d374e50d80aa1e4682f3ab/Cargo.toml#L94), so `vaultwarden` itself implicitly inherits a `hex = "0.4.3"` constraint, but doesn't (and **shouldn't**!) specify it itself.

Further, I'd argue that specifying a version number here is actually a liability, because it can cause breakage from a routine crate bump:

1. Assume that `diesel v2.3.8` (compatible with our current `2.3.7` requirement) bumps its `libsqlite3-sys` dependency to `v0.37.0` (which is incompatible with the current `v0.36.0`); this is not semver-major breakage due to `diesel` not publicly re-exporting any of that crate's types, so it wouldn't be a fault of diesel's.
2. Cargo will resolve that to one instance of 0.36 *and* one instance of 0.37; for any regular crate this would “just” create two copies, but since this one has a `links =` field, instead the build would fail.

Thus, since `vaultwarden` doesn't **directly** call into this crate, it should not **directly** require a specific version of it. In fact, `version = "*"` is *more* robust when all we're trying to do is configure indirect dependencies.

Note also that `Cargo.toml` is only a specification of which crates the project requires at a minimum; the *exact* versions used are stored in `Cargo.lock`. And, yes, it's possible to specify the minimum `libsqlite3-sys` version required by our dependencies ourselves, but that's creating a second source of truth and creates brittleness such as in the way I've outlined above.

ISSOtm:
I actually found a solution that doesn't require that weird extra feature, let me know what you think.

Cargo.toml
BlackDex: Your point 1 is not valid, since if we are updating Diesel, we will also see that they bumped those libraries, so that is a non-argument.

Your second argument is also not valid. If i set `0.34.0` Cargo will keep that version, and will not update to anything else. Diesel sets a range of allowed versions, not a specific version `">=0.17.2, <0.38.0"`

You can check it your self by setting `libsqlite3-sys` to a lower version number and run `cargo update`, it will switch to that lower version number.

Using `*` will only let Cargo try to auto-select the best match if there are any other crates also using the `libsqlite3-sys` crate which does have a pinned version or a different range. Since that is not the case, we want a version specified.

And, Rust also doesn't recommend using `*`, and we want a traceability for these versions in our `Cargo.toml` if we define these crates in that file. Therefor please add the version back instead of `*`.

ISSOtm: > Using `*` will only let Cargo try to auto-select the best match if there are any other crates also using the `libsqlite3-sys` crate which does have a pinned version or a different range. Since that is not the case, we want a version specified.

Well, this is the case of `diesel`: with this change, 0.36.0 gets picked, not the available 0.37.0.

But, this is not a hill I want to die on, so I've reverted the change.

BlackDex: I'm not sure what you mean. `0.36` is the current latest supported version of the diesel version we currently have. I'm not sure why you think `0.37` should be used? There is no diesel version released yet that supports it.

BlackDex: And if we want to keep `0.35` for what ever reason, it stays there, and it's clear that that version is used.

Cargo.toml
```diff
@@ -94,8 +96,8 @@ diesel_migrations = "2.3.1"
 derive_more = { version = "2.1.1", features = ["from", "into", "as_ref", "deref", "display"] }
 diesel-derive-newtype = "2.1.2"
 
-# Bundled/Static SQLite
-libsqlite3-sys = { version = "0.36.0", features = ["bundled"], optional = true }
+# SQLite, statically compiled if the `sqlite_bundled` feature is enabled
```
BlackDex: Not to be a pain or something. But i think this sentence is a bit wrong, and is a leftover from the previous commits. Can you adjust it so that it mentions the right feature flags and what it does?

BlackDex (changes requested):
Ok, just one small change in the comments. The rest looks good and works without issues.

Cargo.toml
ISSOtm: Is this better?

BlackDex (approved):

dani-garcia (approved):

ISSOtm:
Thanks!

Issue #1910, opened 2021-08-16
`[vaultwarden::db::sqlite_migrations][ERROR] Error creating database directory` when using Postgres


cschmatzler:
### Subject of the issue
Using a `DATABASE_URL` env variable starting with `postgres://`, Vaultwarden still tries to run SQLite migrations and fails at https://github.com/dani-garcia/vaultwarden/blob/main/src/db/mod.rs#L291. Data directory is writeable since the keys are created successfully.

### Deployment environment

Debian Bullseye with Docker `version 20.10.8, build 3967b7d`

* vaultwarden version: Docker image `vaultwarden/server:1.22.2`

* Install method:

```
docker run --rm --name vaultwarden \
                        --log-driver=none \
                        --user 1500:1500 \
                        --cap-drop=ALL \
                        --cap-add=DAC_OVERRIDE \
                        --tmpfs=/tmp:rw,noexec,nosuid,size=100m \
                        --network={redacted} \
                        --env-file=/{redacted}/vaultwarden/env-vaultwarden \
                        --mount type=bind,src=/{redacted}/vaultwarden/data,dst=/data \
                        --mount type=bind,src=/etc/passwd,dst=/etc/passwd,ro \
                        --mount type=bind,src=/etc/timezone,dst=/etc/timezone,ro \
                        --mount type=bind,src=/etc/localtime,dst=/etc/localtime,ro \
                        vaultwarden/server:1.22.2
```

* MySQL/MariaDB or PostgreSQL version: `postgres:13.3`

* Other relevant details:
```
> ls -l data
total 12
drwxr-xr-x 2 1500 1500 4096 Aug 16 20:44 icon_cache
-rw-r--r-- 1 1500 1500 1675 Aug 16 20:44 rsa_key.pem
-rw-r--r-- 1 1500 1500 451 Aug 16 20:44 rsa_key.pub.pem
```

env-vaultwarden:
```
ROCKET_PORT=8080

DATABASE_URL="postgresql://{redacted}:{redacted}@postgres/vaultwarden"
```

Running it under root uid works.

### Steps to reproduce
Create the env file, a postgres database, data dir with correct owner and run the above command to launch the docker image.

### Expected behaviour
SQLite migrations should be skipped and container should launch.

### Actual behaviour
Container stops at SQLite migrations failing

### Troubleshooting data
```
[2021-08-16 20:43:49.406][vaultwarden][INFO] Private key created correctly.
[...]

Issue #860, opened 2020-02-13
`DATABASE_URL` is meant for MySQL or Postgres, while this server is meant for SQLite


mzch:
### Subject of the issue

Container can't access to PostgreSQL server.

### Your environment

* Bitwarden_rs version: latest
* Install method: docker
* Version of mysql/postgresql: PostgreSQL 11.6

### Steps to reproduce

- docker-compose up
docker-compose.yml is as below:
```yaml
version: '3.3'
services:
  application:
    image: bitwardenrs/server:latest
    container_name: bitwarden
    ports:
      - 8080:80
      - 3012:3012
    volumes:
      - /srv/bitwarden/data:/data
    environment:
      - DATABASE_URL=postgresql://bitwarden:password@192.168.207.11:5432/bitwarden?sslmode=disable
      - ENABLE_DB_WAL=false
      - WEBSOCKET_ENABLED=true
      - WEBSOCKET_ADDRESS=0.0.0.0
      - WEBSOCKET_PORT=3012
      - SIGNUPS_ALLOWED=true
      - SIGNUPS_VERIFY=true
      - SIGNUPS_VERIFY_RESEND_TIME=3600
      - SIGNUPS_VERIFY_RESEND_LIMIT=6
      - DOMAIN=https://bitsteward.com
      - SMTP_HOST=mx1.mail-services.net
      - SMTP_FROM=no-repky@bitsteward.com
      - SMTP_FROM_NAME=BitSteward
      - SMTP_PORT=465
      - SMTP_SSL=true
      - SMTP_USERNAME=admin@example.com
      - SMTP_PASSWORD=password
      - SMTP_AUTH_MECHANISM="Plain"
      - SMTP_TIMEOUT=15
```
### Expected behaviour

Bitwarden_rs starts normally.

### Actual behaviour

Abort.

### Relevant logs

```log
Feb 13 02:27:58 n21 docker-compose[15501]: #033[36mbitwarden      |#033[0m Error loading config:
Feb 13 02:27:58 n21 docker-compose[15501]: #033[36mbitwarden      |#033[0m #011`DATABASE_URL` is meant for MySQL or Postgres, while this server is meant for SQLite
```

mzch:
Sorry, I didn't notice it would need to rebuild a docker image.

mprasil:
[...]

PR #7061, opened 2026-04-07
Reject unrecognised DATABASE_URL instead of silent SQLite fallback

### LINKED ISSUE #7062: Misconfigured DATABASE_URL silently falls back to SQLite, risking data loss
## Description

When `DATABASE_URL` does not start with `mysql:` or `postgresql:`, vaultwarden silently treats it as a SQLite file path due to the catch-all `else` branch in `DbConnType::from_url`. This means any misconfiguration that does not contain a `/` (or whose path happens to resolve to an existing directory) results in an empty SQLite database being quietly created rather than an error.

In containerised or ephemeral environments, this causes data loss on restart. The user believes they are connected to their intended database whilst secrets are actually written to a throwaway SQLite file that is destroyed when the container restarts.

## Steps to reproduce

1. Set `DATABASE_URL` to any value that is not a valid database URI, e.g. `DATABASE_URL=foobar`
2. Start vaultwarden in a container
3. Observe no error. Vaultwarden starts and creates a local SQLite database named `foobar`
4. Store some secrets
5. Restart the container
6. Secrets are gone

## Prior art

This has been reported before in #2835 and #1910. The partial fix in #2873 (checking the parent directory exists) helps when the misconfigured URL would create a new directory, but does not cover cases where the resolved path sits within an existing directory (such as the working directory itself).

The suggestions from contributors in #2873 (checking for colons, checking for quote characters) were never implemented.

## Proposed fix

See #7061. Require an explicit `sqlite://` prefix for new SQLite deployments. Bare paths without a recognised scheme are still accepted for backwards compatibility, but only when the database file already exists. Otherwise the process panics with a clear error message.

mfw78:
## Summary

When `DATABASE_URL` does not match `mysql:` or `postgresql:`, vaultwarden silently treats it as a SQLite file path. Any misconfiguration (typo, quoting issue, garbage value) quietly creates an empty SQLite database instead of erroring. In containerised environments this means data loss on restart, which has bitten me personally and others (#2835, #1910, #860).

The partial fix in #2873 catches some cases by checking the parent directory, but does not help when the path resolves within an existing directory.

The core issue is that SQLite selection is implicit. Database selection for a secrets manager should be explicit, not a fallback. An operator who misconfigures their database URL should get a clear error, not a silent downgrade to a different database engine entirely.

## Changes

- `DbConnType::from_url` now recognises an explicit `sqlite:` scheme. URLs without any recognised scheme are still accepted as bare-path SQLite for backwards compatibility, but only if the file already exists. Otherwise the process panics with a clear message.
- Default `DATABASE_URL` changed to `sqlite://{data_folder}/db.sqlite3` so new deployments use the explicit scheme.
- `backup_sqlite` and `validate_config` updated to handle the `sqlite://` prefix.
- `.env.template` updated accordingly.

## Test plan

- [ ] Fresh deployment with default config creates database normally
- [ ] Existing bare-path `DATABASE_URL=data/db.sqlite3` (file exists) still works
- [ ] Misconfigured URL (e.g. `foobar`) panics with a helpful message
- [ ] SQLite backup works with `sqlite://` prefix
- [ ] MySQL and PostgreSQL URLs unaffected

Fixes #7062

### DIFF
--- a/src/db/mod.rs
+++ b/src/db/mod.rs
@@ -272,13 +272,32 @@ impl DbConnType {
             #[cfg(not(postgresql))]
             err!("`DATABASE_URL` is a PostgreSQL URL, but the 'postgresql' feature is not enabled")
 
-        //Sqlite
-        } else {
+        // Sqlite (explicit)
+        } else if url.len() > 7 && &url[..7] == "sqlite:" {
             #[cfg(sqlite)]
             return Ok(DbConnType::Sqlite);
 
             #[cfg(not(sqlite))]
-            err!("`DATABASE_URL` looks like a SQLite URL, but 'sqlite' feature is not enabled")
+            err!("`DATABASE_URL` is a SQLite URL, but the 'sqlite' feature is not enabled")
+
+        // No recognized scheme — assume legacy bare-path SQLite, but the database file must already exist.
+        // This prevents misconfigured URLs (typos, quoted strings) from silently creating a new empty SQLite database.
+        } else {
+            #[cfg(sqlite)]
+            {
+                if std::path::Path::new(url).exists() {
+                    return Ok(DbConnType::Sqlite);
+                }
+                panic!(
+                    "`DATABASE_URL` does not match any known database scheme (mysql://, postgresql://, sqlite://) \
+                     and no existing SQLite database was found at '{url}'. \
+                     If you intend to use SQLite, use an explicit `sqlite://` prefix in your `DATABASE_URL`. \
+                     Otherwise, check your DATABASE_URL for typos or quoting issues."
+                );
+            }
+
+            #[cfg(not(sqlite))]
+            err!("`DATABASE_URL` does not match any known database scheme (mysql://, postgresql://, sqlite://)")
         }
     }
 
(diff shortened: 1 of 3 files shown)


tessus:
May I ask why a panic and not a proper error message and clean exit with a specific error code?

mfw78:
Good catch, and fair point. The panic was a bit short-sighted on my part; this issue had actually bitten me personally so I was perhaps a touch too eager to make it loud. I've since had a proper look through the codebase and can see the established convention is to use `err!()` which propagates up through `from_config()` and gets caught by `create_db_pool()` with a clean `exit(1)`. That's clearly the right approach.

I've pushed a follow-up commit that swaps the `panic!` for `err!()` to bring it in line with how the rest of the config validation works.

src/db/mod.rs
```diff
@@ -272,13 +272,32 @@ impl DbConnType {
...
+        } else {
+            #[cfg(sqlite)]
+            {
+                if std::path::Path::new(url).exists() {
+                    return Ok(DbConnType::Sqlite);
+                }
+                err!(format!(
+                    "`DATABASE_URL` does not match any known database scheme (mysql://, postgresql://, sqlite://) \
+                     and no existing SQLite database was found at '{url}'. \
+                     If you intend to use SQLite, use an explicit `sqlite://` prefix in your `DATABASE_URL`. \
```
BlackDex: I think instead of the word `prefix` we should use `scheme` because that matches the exact meaning.

BlackDex (changes requested):
Just a small text change. But i think that is all, the rest LGTM.

src/db/mod.rs
mfw78: Good point, updated to use 'scheme' instead of 'prefix' in both the error message and the .env.template. Thanks!

BlackDex (approved):

BlackDex (approved):

PR #7068, opened 2026-04-09
fix: return Err instead of panic on unknown cipher atype in to_json()

### LINKED ISSUE #7067: Cipher::to_json() panics on unknown atype instead of returning an error
## Bug Description

`Cipher::to_json()` in `src/db/models/cipher.rs` returns `Result<Value, Error>`, but the match arm for unknown `atype` values calls `panic!("Wrong type")` rather than returning a proper `Err`.

**Affected code** (around line 395–402):

```rust
let key = match self.atype {
    1 => "login",
    2 => "secureNote",
    3 => "card",
    4 => "identity",
    5 => "sshKey",
    _ => panic!("Wrong type"),   // ← panics instead of returning Err
};
```

## Why This Matters

Because the whole function signature is `-> Result<Value, crate::Error>`, the intent is clearly to propagate failures to callers — but a `panic!` bypasses that entirely and terminates the Rocket worker thread (or the whole process, depending on the panic hook).

A cipher with an unexpected `atype` can end up in the database via:
- Direct database edits / migrations
- Corruption or a future Bitwarden protocol extension that adds a new type before vaultwarden is updated

When a user with such a cipher calls `/sync`, vaultwarden crashes instead of returning a 500/error response, which also affects all other concurrent users on the same instance.

## Expected Behavior

Return an `Err` (using the existing `err!` macro already used elsewhere in the same file) so the error is logged and the request fails gracefully without crashing the server.

## Suggested Fix

```rust
_ => err!(format!("Cipher {} has an invalid type {}", self.uuid, self.atype)),
```

I have a PR ready with this one-line fix.

mango766:
Fixes #7067

## What

`Cipher::to_json()` returns `Result<Value, Error>` but the wildcard arm of the `atype` match called `panic!("Wrong type")`. This means if a cipher with an unrecognised type exists in the database, the whole request (and potentially the server) panics instead of failing gracefully.

## Change

One-line fix — replace `panic!` with `err!`, which is already used elsewhere in the same file:

```diff
-            _ => panic!("Wrong type"),
+            _ => err!(format!("Cipher {} has an invalid type {}", self.uuid, self.atype)),
```

This lets callers receive a proper `Err`, the error gets logged, and the sync request returns a failure response without crashing the server.

## How it can happen

- A cipher row is edited directly in the database
- A future Bitwarden protocol version introduces a new type that vaultwarden doesn't yet recognise
- Data migration from a third-party tool writes an unexpected value

In all these cases the current code would terminate the Rocket worker thread for every sync request that includes that cipher, affecting all concurrent users.


BlackDex (approved):
While these panic's will not crash the server, since Rocket handles them in a good way, having the error message a bit more nice and easier to troubleshoot via the log messages, this is a good change

dani-garcia (approved):

PR #7070, opened 2026-04-09
fix: return error instead of panicking on invalid date input in parse_date()

### LINKED ISSUE #7069: parse_date() panics on invalid user-supplied date strings, crashing the server
## Bug Description

`parse_date()` in `src/util.rs` calls `.unwrap()` directly on the result of parsing a user-controlled date string from HTTP query parameters. If a request is sent with a malformed RFC3339 date, the unwrap panics and takes down the server process.

```rust
// src/util.rs
pub fn parse_date(date: &str) -> NaiveDateTime {
    DateTime::parse_from_rfc3339(date).unwrap().naive_utc()  // panics on invalid input!
}
```

## Affected Endpoints

All three event query endpoints and the event collection endpoint call `parse_date()` with values taken directly from HTTP query parameters or request body:

- `GET /api/organizations/{id}/events?start=<user_input>&end=<user_input>`
- `GET /api/ciphers/{id}/events?start=<user_input>&end=<user_input>`
- `GET /api/organizations/{id}/users/{id}/events?start=<user_input>&end=<user_input>`
- `POST /events/collect` (date field in JSON body)

Any authenticated user with access to the event log (admin/manager of an org) can trigger this.

## Steps to Reproduce

```bash
# Authenticated as an org admin/manager
curl -H "Authorization: Bearer <token>" \
  "https://your-vaultwarden/api/organizations/<org_id>/events?start=not-a-date&end=also-not-a-date"
```

This causes the Rocket thread handling the request to panic with something like:
```
thread 'rocket-worker-1' panicked at 'called `Result::unwrap()` on an `Err` value: ParseError(...)'
```

## Expected Behavior

The server should return a `400 Bad Request` with a descriptive error message instead of panicking.

## Fix

Change `parse_date()` to return `Result<NaiveDateTime, Error>` and propagate the error to callers using `?`. The single call site using a hardcoded literal in `sends.rs` can safely use `.expect()`.

mango766:
Fixes #7069

## What's happening

`parse_date()` in `src/util.rs` calls `.unwrap()` on a `DateTime::parse_from_rfc3339()` result, where the input comes directly from HTTP query parameters or JSON body fields. Sending a malformed RFC3339 date string to any of the event endpoints panics the request handler thread.

Affected endpoints:
- `GET /api/organizations/{id}/events?start=...&end=...`
- `GET /api/ciphers/{id}/events?start=...&end=...`  
- `GET /api/organizations/{id}/users/{id}/events?start=...&end=...`
- `POST /events/collect`

## Fix

Changed `parse_date()` to return `Result<NaiveDateTime, Error>` and updated all callers in `events.rs` to propagate the error with `?`, which results in a `400 Bad Request` response instead of a panic.

The one call site with a hardcoded literal (`sends.rs`, epoch timestamp used for an anonymous push device) now uses `.expect()` with an explanatory message since that string is always valid.

## Changes

- `src/util.rs`: `parse_date()` now returns `Result<NaiveDateTime, crate::Error>` and uses pattern matching instead of `.unwrap()`
- `src/api/core/events.rs`: all `parse_date()` calls use `?` to propagate errors
- `src/api/core/sends.rs`: hardcoded constant call-site uses `.expect()` with explanation

### DIFF
--- a/src/api/core/events.rs
+++ b/src/api/core/events.rs
@@ -41,11 +41,11 @@ async fn get_org_events(org_id: OrganizationId, data: EventRange, headers: Admin
     let events_json: Vec<Value> = if !CONFIG.org_events_enabled() {
         Vec::with_capacity(0)
     } else {
-        let start_date = parse_date(&data.start);
+        let start_date = parse_date(&data.start)?;
         let end_date = if let Some(before_date) = &data.continuation_token {
-            parse_date(before_date)
+            parse_date(before_date)?
         } else {
-            parse_date(&data.end)
+            parse_date(&data.end)?
         };
 
         Event::find_by_organization_uuid(&org_id, &start_date, &end_date, &conn)
@@ -71,11 +71,11 @@ async fn get_cipher_events(cipher_id: CipherId, data: EventRange, headers: Heade
     } else {
         let mut events_json = Vec::with_capacity(0);
         if Membership::user_has_ge_admin_access_to_cipher(&headers.user.uuid, &cipher_id, &conn).await {
-            let start_date = parse_date(&data.start);
+            let start_date = parse_date(&data.start)?;
             let end_date = if let Some(before_date) = &data.continuation_token {
-                parse_date(before_date)
+                parse_date(before_date)?
             } else {
-                parse_date(&data.end)
+                parse_date(&data.end)?
             };
 
             events_json = Event::find_by_cipher_uuid(&cipher_id, &start_date, &end_date, &conn)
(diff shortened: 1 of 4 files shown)


BlackDex:
I'm not sure this really really is needed, though it might be a bit nicer.
But your Claude Code is mixing PR's or isn't using the right branch. You now also have the _fix_ of your other PR included here. That is probably not what you want.

BlackDex (approved):

BlackDex:
Going to close this, it's not a real issue, and no response regarding the mixing of PR's

PR #7085, opened 2026-04-12
Fix MFA Remember

BlackDex:
Fix MFA Remember
And update some crates to the latest version.


dfunkt:
You also included changes from `Cargo.lock` in your commit, was that intended?

BlackDex:
> You also included changes from `Cargo.lock` in your commit, was that intended?

Yes, i thought, while at it, I can probably just update them too.

dani-garcia (approved):

PR #7108, opened 2026-04-17
Update Rust, Crates, GHA and fix a DNS issue

BlackDex:
- Updated Rust to v1.95.0
- Updated all the crates
- Update GitHub Actions

With the crate updates, hickory-resolver was updated which needed some changes. During testing I found a bug with the fallback resolving from Tokio. The resolver doesn't work if it receives only a `&str`, it needs a `port` too. This fixed the resolving if Hickory failed to load.

Also, Hickory switched the resolving to prefer IPv6. While this is nice, it could break or slowdown resolving for IPv4 only environments. Since we already have a flag to prefer IPv6, we check if this is set, else resolve IPv4 first and IPv6 afterwards.

Also, we returned just 1 IpAddr record, and ignored the rest. This could mean, a failed attempt to connect if the first IP endpoint has issues. Same if the first records is IPv6 but the server doesn't support this, it never tried a possible returned IPv4 address.

We now return a full list of the resolved records unless one of the records matched a filtered address, than the whole resolving is ignored as was previously the case.


src/http_client.rs
```diff
@@ -184,12 +184,15 @@ impl CustomDnsResolver {
...
-                    builder.options_mut().ip_strategy = hickory_resolver::config::LookupIpStrategy::Ipv6thenIpv4;
+                // Hickory's default since v0.26 is `Ipv6AndIpv4`, which sorts IPv6 first
+                // This might cause issues on IPv4 only systems or containers
+                // Unless someone enabled DNS_PREFER_IPV6, use Ipv4AndIpv6, which returns IPv4 first which was our previous default
+                if !CONFIG.dns_prefer_ipv6() {
+                    builder.options_mut().ip_strategy = hickory_resolver::config::LookupIpStrategy::Ipv4AndIpv6;
                 }
-                let resolver = builder.build();
+                let resolver = builder.build().expect("Building Hickory resolver failed");
                 Arc::new(Self::Hickory(Arc::new(resolver)))
```
dani-garcia: Can this fail? Should we instead do something like:
```rust

match builder.build() {
    Ok(resolver) => Arc::new(Self::Hickory(Arc::new(resolver))),
    Err(e) => {
        warn!("Error building Hickory resolver, falling back to default: {e:?}");
        Arc::new(Self::Default())
    }
}
```

src/http_client.rs
```diff
@@ -200,19 +203,19 @@ impl CustomDnsResolver {
...
-            Self::Default() => tokio::net::lookup_host(name).await?.next(),
-            Self::Hickory(r) => r.lookup_ip(name).await?.iter().next().map(|a| SocketAddr::new(a, 0)),
+        let results: Vec<SocketAddr> = match self {
+            Self::Default() => tokio::net::lookup_host((name, 0)).await?.collect(),
+            Self::Hickory(r) => r.lookup_ip(name).await?.iter().map(|i| SocketAddr::new(i, 0)).collect(),
         };
 
-        if let Some(addr) = &result {
+        for addr in &results {
             post_resolve(name, addr.ip())?;
```
dani-garcia, line 212: This will error if any of the addresses are non-global, should we instead just filter them out?

something like `results.iter().filter_map(|a| post_resolve(name, addr.ip()).ok()).collect()`

src/http_client.rs
BlackDex: Hmm, that might be a better option indeed.
It might, as they changed it to Result for a reason i guess.

src/http_client.rs
BlackDex: I did that at first. But if a DNS has one record which points to something not wanted, it probably isn't a nice record anyway. Though maybe someone has two addresses one public and another remote, but that is generally bad practice.

That is why i opted for ignoring the whole resolving.

dani-garcia: Yeah that's fair, we can leave this as is for now then

src/http_client.rs
BlackDex: I have changed it to use chaining logic. This way only one `Self::Default()` part is needed instead of defining it twice.

dani-garcia (approved):

mrmelon54:
I have implied from the discussions about IPv4/IPv6 that the implementation doesn't use [Happy Eyeballs](https://en.wikipedia.org/wiki/Happy_Eyeballs). Would it not make more sense to try all addresses in a parallel but partially-staggered order to speed up all connections.

BlackDex:
@mrmelon54 Happy Eyeballs is used, but even then, the first IP address in the list is still the first item tested as far as i have seen. `Ipv4AndIpv6` is also the default used by Reqwest, and `Ipv6AndIpv4` is the default of Hickory.
Since we override the resolver of Reqwest, we needed to use the same defaults, Reqwest will handle the rest.

PR #7129, opened 2026-04-24
Add send_template_email and deduplicate common email handling

tjps:


stefan0xC:
Why? What problem does this solve?

tjps:
I like to read the source for things I used and noticed a repeated pattern. For things I maintain I'm always happy to see a patch that maintains functionality and reduces LoC, and assumed that would be the case here.

BlackDex:
This doesn't seem like an improvement too me.
It makes it less clear which data is sent from where, and also adds another macro.

If anything, this also could have been solved much much easier by just adding those common fields into the `get_text` function and merge it with provided one. But leaving those items at the specific mail functions provides a clear notice which template values are used/needed i think.

PR #7137, opened 2026-04-25
Update secure context requirements in README.md

realfunnyguy:
It is my understanding that most browsers require https to use web crypto apis and these apis wont function over localhost at all. With this in mind, i don't see the need to keep this port 8000 blurb here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ A nearly complete implementation of the Bitwarden Client API is provided, includ
 
 > [!IMPORTANT]
 > The web-vault requires the use a secure context for the [Web Crypto API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API).
-> That means it will only work via `http://localhost:8000` (using the port from the example below) or if you [enable HTTPS](https://github.com/dani-garcia/vaultwarden/wiki/Enabling-HTTPS).
+> That means it will only work via [HTTPS](https://github.com/dani-garcia/vaultwarden/wiki/Enabling-HTTPS).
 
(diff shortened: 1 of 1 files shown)


DorianCoding:
Localhost is a trusted url and can therefore use Web Crypto: https://w3c.github.io/webappsec-secure-contexts/#potentially-trustworthy-url

stefan0xC:
> Localhost is a trusted url and can therefore use Web Crypto: https://w3c.github.io/webappsec-secure-contexts/#potentially-trustworthy-url

Due to https://github.com/bitwarden/clients/pull/17095 HTTPS is required in all clients ([since `v2025.11.0`](https://bitwarden.com/help/releasenotes/#2025-11-0:~:text=Web%20clients%20require%20https%20configuration)). So a secure context is not enough anymore.

stefan0xC (changes requested):
> It is my understanding that most browsers require https to use web crypto apis and these apis wont function over localhost at all.

A secure context is still necessary but not sufficient anymore because the Bitwarden clients require the self-hosted environments to use [an URL that start with `https://`](https://github.com/bitwarden/clients/blob/b9658a795ff0aa10f927667f8b286c4ec1552d9a/libs/common/src/services/api.service.ts#L1318-L1321). Without a secure context the subtle crypto API fails to load and our [patched web-vault will display a warning](https://github.com/vaultwarden/vw_web_builds/pull/19) that you need to enable HTTPS; if you set up a secure context without HTTPS like localhost our warning will not be displayed (because most browsers consider it a secure context) but instead the login will fail because of the forementioned requirement by the client. I hope this clears this up a bit and maybe you could rephrase the whole paragraph to be more accurate?

BlackDex:
Going to close this as i have address this in #7171

PR #7144, opened 2026-04-27
Apply `duration_suboptimal_units` lint findings

dfunkt:
Quote from lint description:
"Using a smaller unit for a duration that is evenly divisible by a larger unit reduces readability. Readers have to mentally convert values, which can be error-prone and makes the code less clear."


BlackDex:
Nice, strange way to notate them at all like this though.

BlackDex (approved):

PR #7154, opened 2026-04-28
Dummy identifier need to pass for a guid

Timshel:
Since client version `2026-3.1` the `organizationId` need to pass the `isId` check [cf](https://github.com/bitwarden/clients/blob/web-v2026.3.1/apps/web/src/app/auth/organization-invite/accept-organization.component.ts#L137)
Which mean passing `guidRegex = /^[0-9a-f]{8}-(?:[0-9a-f]{4}-){3}[0-9a-f]{12}$/i`

### DIFF
--- a/src/sso.rs
+++ b/src/sso.rs
@@ -17,7 +17,7 @@ use crate::{
     CONFIG,
 };
 
-pub static FAKE_SSO_IDENTIFIER: &str = "vaultwarden-dummy-oidc-identifier";
+pub static FAKE_SSO_IDENTIFIER: &str = "00000000-01DC-01DC-01DC-000000000000";
 
 static SSO_JWT_ISSUER: LazyLock<String> = LazyLock::new(|| format!("{}|sso", CONFIG.domain_origin()));
 


BlackDex:
Hmm, i have not encountered it during my tests.
Via which way would this happen?

I'm fine with merging this though, just curious.

BlackDex (approved):

Timshel:
A yes sorry,
 it's triggered for an admin invite when SSO is enabled ([cf](https://github.com/dani-garcia/vaultwarden/blob/7cf0c5d67eb81c8b4f2e86b5c8d030bb330faa28/src/api/admin.rs#L319)).
 
 Edit: you can test it even without a provider since it will fail when opening the link.

BlackDex:
@Timshel I might have been to quick with merging. You also need to update that special endpoint so it will not trigger the OrgMember check.

If you could create a followup on this, Ill merge that quickly.

Timshel:
@BlackDex o yeah :(, I'll make it now

PR #7157, opened 2026-04-28
Fix hardcoded sso identifier

Timshel:
Correct the modified value from https://github.com/dani-garcia/vaultwarden/pull/7154


BlackDex (approved):

BlackDex:
Thanks! I tested it locally, and that works just fine

PR #7163, opened 2026-04-29
Several SSO Fixes

BlackDex:
### Ensure SSO token is only usable on the same client
This commit adds an extra check via cookies to ensure the same browser/client is used to request and provide the SSO token.
Previously it would be able to provide a custom link which attackers could use to steal data.
While an attacker would still need the Master Password to be able to decrypt or execute specific actions, they were able to fetch encrypted data.

### Check email-verified on SSO login/create
This commit prevents possible account takeover via SSO which doesn't check/validate or provide validated status of the email.
It was checked at other locations, but was skipped here.

### Prevent data disclosure via SSO endpoints
This commit prevents some data disclosure and user enumeration by only returning the fake SSO identifier.
Since we do not check the identifier anywhere useful, returning the fake one is just fine.

During an invite to an org, that link contains the correct UUID and will be used for the master password requirements.
For anything else, server admins should set the `SSO_MASTER_PASSWORD_POLICY` env variable.

### Misc
Adjust admin layout to fix issues when SSO is enabled

### DIFF
--- a/migrations/mysql/2026-04-25-120000_sso_auth_binding/down.sql
+++ b/migrations/mysql/2026-04-25-120000_sso_auth_binding/down.sql
@@ -1,0 +1,1 @@
+ALTER TABLE sso_auth DROP COLUMN binding_hash;
--- a/migrations/mysql/2026-04-25-120000_sso_auth_binding/up.sql
+++ b/migrations/mysql/2026-04-25-120000_sso_auth_binding/up.sql
@@ -1,0 +1,1 @@
+ALTER TABLE sso_auth ADD COLUMN binding_hash TEXT;
--- a/migrations/postgresql/2026-04-25-120000_sso_auth_binding/down.sql
+++ b/migrations/postgresql/2026-04-25-120000_sso_auth_binding/down.sql
@@ -1,0 +1,1 @@
+ALTER TABLE sso_auth DROP COLUMN binding_hash;
--- a/migrations/postgresql/2026-04-25-120000_sso_auth_binding/up.sql
+++ b/migrations/postgresql/2026-04-25-120000_sso_auth_binding/up.sql
@@ -1,0 +1,1 @@
+ALTER TABLE sso_auth ADD COLUMN binding_hash TEXT;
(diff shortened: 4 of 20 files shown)


dani-garcia (approved):

Timshel:
Hey,
I'm discovering this issue.

> Ensure SSO token is only usable on the same client

I'm looking at it, but it looks redundant with the PKCE check ? Was it judged insufficient ?
If a different client is used to redeem the sso token not sure how it would be able to provide the `client_challenge` ?

BlackDex:
@Timshel, there was a security issue reported regarding this, no matter if PKCE was used or not.
The only feasible option was using a cookie, this is btw the same as Bitwarden does as far as i could see.

Timshel:
@BlackDex is there anything published ?  and how it bypassed PKCE ? 

Because the whole point of pkce is to ensure that the client redeeming the token is the same as the one which initiated the flow.

Sry if I'm misunderstanding something :(

Timshel:
> Prevent data disclosure via SSO endpoints

The value returned in `get_org_domain_sso_verified` is used by `get_auto_enroll_status` and `get_master_password_policy`. This means it becomes impossible to change the policy at the org level and use different policy per organization. With `enforceOnLogin` it does not apply only when creating the account (It was indicated as not working in the env template, but I believe it should work now).

BlackDex:
@Timshel I have sent you the relevant reports via email as we have not yet disclosed them while waiting for the CVE's.

BlackDex:
@Timshel. If users receive an invite from their org, that will contain the correct Org UUID, and that flow works just fine. If users are not invited, they the server admin's should probably configure the Global SSO Master Password Policy via the config.

For all other items i would suggest to read the reports and if you have another solution which is safe and secure, I'm pleased to see a PR regarding this.

BlackDex:
@Timshel, also, If you would like to discuss this in a chat, you can find me on Matrix in our Vaultwarden channel.

Timshel:
~~Had a second look for the disclosure issue with the skip email issue.
Realized that in the SSO only `web-vault` client version I maintain I added a small patch in case the identifier returned by `get_auto_enroll_status` is different [cf](https://github.com/Timshel/oidc_web_vault/blob/787c173b6d97536ac2cb8a1a239748466598a6c9/patches/oidc_override.patch#L159)~~
Checked again, it was for a use case not applicable to Vaultwarden, and it appears unnecessary now anyway.


Edit: and I was mixing stuff the `enforceOnLogin` use the policy returned in `authenticated_response` so no issue here.

PR #7174, opened 2026-05-01
Update `reqwest` to 0.13.3

dfunkt:
- since version 0.13 `reqwest` defaults to `aws-lc-rs` as the CryptoProvider, so in order to use `ring` switch to rustls-no-provider as mentioned in the link below https://github.com/seanmonstar/reqwest/blob/5b5ac174bc1c8f62de5b6bd200c943719e18fb88/src/tls.rs#L46-L66
- form had to be manually specified now, no longer enabled by default


BlackDex:
The reason if not updated it, is because it will make reqwest have two versions now. Both 0.12 and 0.13. that is something we would like to prevent.

PR #7180, opened 2026-05-03
option to use encrypted rsa key

pieska:
I'm not a rust pro, but tried to implement creating/loading of encrypted rsa key:

- no passphrase set, no keyfile found -> unencrypted pkcs1 created, no change

- no passphrase set, key unencrypted -> key will be loaded, no change

- no passphrase set, key encrypted -> error and exit

- passhrase set, no keyfile found  -> encrypted pkcs1 created

- passphrase set, key unencrypted -> issue warnung but key will be loaded

- passphrase set, key encrypted (pkcs1 or pkcs8) -> key will be loaded and decrypted

passphrase can be set in RSA_KEY_PASSPHRASE or in RSA_KEY_PASSPHRASE_FILE 

encrypt existing keys:

- pkcs1
openssl rsa -in rsa_key.pem -aes256 -out rsa_key_encrypted.pem -traditional

- pkcs8
openssl rsa -in rsa_key.pem -aes256 -out rsa_key_encrypted.pem

- check
openssl rsa -in rsa_key_encrypted.pem -check

- check with passphrase file
openssl rsa -in rsa_key_encrypted.pem -check -passin file:rsa_key.pass

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -69,14 +69,59 @@ pub async fn initialize_keys() -> Result<(), Error> {
         Err(e) => return Err(e.into()),
     };
 
-    let (priv_key, priv_key_buffer) = if let Some(priv_key_buffer) = priv_key_buffer {
-        (Rsa::private_key_from_pem(priv_key_buffer.to_vec().as_slice())?, priv_key_buffer.to_vec())
+    let passphrase = CONFIG.rsa_key_passphrase();
+
+    let (priv_key, priv_key_buffer) = if let Some(stored_buffer) = priv_key_buffer {
+        let mut passphrase_error: Option<Error> = None;
+        let mut key_is_unencrypted = true;
+
+        // The callback is only invoked if the key file is encrypted. For unencrypted keys it is never called.
+        let rsa = Rsa::private_key_from_pem_callback(stored_buffer.to_vec().as_slice(), |buf| {
+            key_is_unencrypted = false;
+            if passphrase.is_empty() {
+                // Only reached when key is encrypted. Return any error to abort the callback;
+                // the actual error message is stored in passphrase_error above.
+                passphrase_error = Some(Error::other(
+                    "Private RSA key is encrypted but RSA_KEY_PASSPHRASE is not configured",
+                ));
+                return Err(openssl::error::ErrorStack::get());
+            }
+            let bytes = passphrase.as_bytes();
+            buf[..bytes.len()].copy_from_slice(bytes);
+            Ok(bytes.len())
+        });
+
+        let rsa = match rsa {
+            Ok(r) => r,
+            Err(_) if passphrase_error.is_some() => return Err(passphrase_error.unwrap().into()),
+            Err(e) => return Err(e.into()),
+        };
+
+        if key_is_unencrypted && !passphrase.is_empty() {
+            warn!(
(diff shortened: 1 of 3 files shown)


BlackDex:
What are you trying to solve here?
Having an encrypted RSA key for which you need to provide a plain/text password isn't going to solve anything.

From my point of view this only adds more complexity to something which isn't going to add more security.

pieska:
I would like to backup the system and don't worry about the key in the backup unencrypted (compliance regulations, etc.). I backup everything excluding such passkey files which reside in a specific filesystem location monitored by auditd. If you have the key encrypted you just have to solve the problem how to feed the passphrase into vaultwarden.

Some ideas:

- pre/post grants and removes access to the pass key file outside the data dir, so after starting up the user running vaultwarden has no access to the key anymore, works with the keyfile itself too, of course

- you set up a fifo as pass key file and pre writes the passkey into the fifo and post reads from the fifo to make sure, if vaultwarden did not start, the passphrase is removed from the fifo, the passphrase is only available at startup, this is what I usually do

- using the fifo method you could even start vaultwarden with en empty fifo, fifo read blocks vaultwarden until an admin entered the passphrase by cat > fifo or whatever tool you like

you don't have to use encrypted keys but having the option provides options to solve some requirements about secure key  handling

BlackDex:
The key isn't used for encryption of anything in the database. It is used for JWT tokens. If the key is lost, people need to login again in the worst case scenario.

Also, if you do a backup unencrypted, including the database, then there are still items inside that database that might be of interest, and therefor should be encrypted in a remote backup i think.

pieska:
I know what the key is for. Won’t there be consequences if the key falls into the wrong hands? If someone gets their hands on the key, couldn’t they use it to do something malicious?

But if you don’t see any need to protect the key, that’s fine with me.

BlackDex:
> I know what the key is for. Won’t there be consequences if the key falls into the wrong hands? If someone gets their hands on the key, couldn’t they use it to do something malicious?
> 
> But if you don’t see any need to protect the key, that’s fine with me.

How would someone get that key? By having access to the server? Or a full backup?
In that case you probably also have other issues to think about, and when they have access to the server they probably also have the password used for the key anyway.

pieska:
The question isn't how someone gets the key, it's how to limit the impact if they do, it's called defense-in-depth.

For reference, Bitwarden does to my knowledge exactly this for the same key, identity.pfx is a password-protected PKCS12 by default. I'd assumed parity with that threat model would be a reasonable starting point.

If the project doesn't see value in the option, fair enough. I'll leave the PR open in case anyone else finds it useful and won't push further. Have a good one.

BlackDex:
In my opinion, it's better to store the key secretly and maybe use Docker Secrets or In K8s using SecretStore stuff.

If you do it like that, no need to backup it up including in the other content.

You are only shifting the security to a different value.
Also, if the key gets leaked just replace it. 

For me it's security theater instead of actually improving security. 
And, to be honest, i have never seen any project which promotes to use it like this, and most say to store the key it self at a secure location and not use a password to protect it.

BlackDex:
I'm going to close this one, as this is not something we want to encourage. There are other ways to securely distribute the RSA file without the need to encrypt it, and have a plain text password somewhere else.

PR #7197, opened 2026-05-08
sso_auth improvements

Timshel:
Improve SSO auth flow.

Was a bit lazy with the `code_response_error` column and stored both `error` and `error_description` in it.
Could be written as two columns but preferred to have the `Option` of the tuple instead of two separate `Option`.

For `mysql` deleted the `sso_auth` table since the `code_response` contained and whole json serialization and could overflow the new `VARCHAR(768)` limit (due to the index).

### DIFF
--- a/migrations/mysql/2026-05-05-120000_sso_auth_error/up.sql
+++ b/migrations/mysql/2026-05-05-120000_sso_auth_error/up.sql
@@ -1,0 +1,1 @@
+ALTER TABLE sso_auth ADD COLUMN code_response_error TEXT;
--- a/migrations/mysql/2026-05-05-120000_sso_auth_error/down.sql
+++ b/migrations/mysql/2026-05-05-120000_sso_auth_error/down.sql
@@ -1,0 +1,1 @@
+ALTER TABLE sso_auth DROP COLUMN code_response_error;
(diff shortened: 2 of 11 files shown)


migrations/mysql/2026-05-05-120000_sso_auth_error/up.sql
```diff
@@ -0,0 +1,4 @@
+DELETE FROM sso_auth;
+ALTER TABLE sso_auth ADD COLUMN code_response_error TEXT;
+ALTER TABLE sso_auth MODIFY COLUMN code_response VARCHAR(768);
```
BlackDex: I think it is better to keep this the same as the rest and keep it `TEXT` as it was before.
Was there a specific reason to make it a varchar?

BlackDex (changes requested):
I think overall it looks good, and it works fine.
Just one question.

migrations/mysql/2026-05-05-120000_sso_auth_error/up.sql
Timshel: MySQL still can't index TEXT I believe ? (Don't remember which version I tested against but was not latest).

BlackDex: @Timshel, the question then is, is an index really needed?
If this table is cleaned every time, an index on it isn't that useful probably. The amount of records will probably be low. And, we do not have many indexes at all currently..

Timshel: Yes unless someone disable the job cleaning it and spam unfinished login flow it should have almost no row.

BlackDex (approved):

PR #7209, opened 2026-05-12
Update docker cli example to match docker compose example

Mathsoum:
Little PR

Adding a '.' to the docker cli example in README.md make it match with the docker-compose example right below it.  
It may also help people that simply copy-paste the command for tests purposes to not try to write to a directory at the root of their FS


BlackDex:
This is incorrect, that will not work with docker cli.

PR #7243, opened 2026-05-20
http_client: don't panic on invalid block regex; use RwLock

nghiphaam:
`should_block_address_regex` called `Regex::new(&block_regex).unwrap()` on every miss. Config validation (`config.rs`) already rejects invalid patterns, but a slip-through (e.g. test path or a race) would crash the server mid-request. Swap the `unwrap()` for a logged warning + skip-block fallback.

Also switched the cache from `Mutex` to `RwLock` so concurrent outbound HTTP checks don't serialise on the same lock when the regex is unchanged.

### DIFF
--- a/src/http_client.rs
+++ b/src/http_client.rs
@@ -1,7 +1,7 @@
 use std::{
     fmt,
     net::{IpAddr, SocketAddr},
-    sync::{Arc, LazyLock, Mutex},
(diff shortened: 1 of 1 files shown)


BlackDex:
Is this AI generated? Since the claim is invalid. The regex is checked during startup and config update, so this never fails.

BlackDex:
Invalid

PR #7265, opened 2026-05-27
Fix migration for MariaDB 12.2.2

Timshel:
I had missed the Mariadb fix so encountered the error with latest `mariadb` version.
It appears they changed the naming convention for foreign key :(.

But with the current fix it ends up with a double definition:

```sql
> show create table sso_users
CREATE TABLE `sso_users` (
  `user_uuid` char(36) NOT NULL,
  `identifier` varchar(768) NOT NULL,
  `created_at` timestamp NOT NULL DEFAULT current_timestamp(),
  PRIMARY KEY (`user_uuid`),
  UNIQUE KEY `identifier` (`identifier`),
  CONSTRAINT `1` FOREIGN KEY (`user_uuid`) REFERENCES `users` (`uuid`),
  CONSTRAINT `2` FOREIGN KEY (`user_uuid`) REFERENCES `users` (`uuid`) ON DELETE CASCADE ON UPDATE CASCADE
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_uca1400_ai_ci
```

Edit: Wonder if it would make sense to add a new migration to cleanup and add it as a named constraint ?

### DIFF
--- a/migrations/mysql/2024-03-13-170000_sso_users_cascade/up.sql
+++ b/migrations/mysql/2024-03-13-170000_sso_users_cascade/up.sql
@@ -1,15 +1,31 @@
--- Dynamically create DROP FOREIGN KEY
--- Some versions of MySQL or MariaDB might fail if the key doesn't exists
--- This checks if the key exists, and if so, will drop it.
-SET @drop_sso_fk = IF((SELECT true FROM information_schema.TABLE_CONSTRAINTS WHERE
-    CONSTRAINT_SCHEMA = DATABASE() AND
-    TABLE_NAME = 'sso_users' AND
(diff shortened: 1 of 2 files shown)


BlackDex:
Uch, MariaDB is a star in adjusting stuff not really compatible or nicely migrated in some way.
Ill see if i have some time this week to check and verify.

flipreverse:
@Timshel How do I apply your fix?

Timshel:
@flipreverse Just rebased the branch on top of `main` so you can build it.

Alternatively you can test-it with the fork I maintain.

BlackDex (approved):

PR #7270, opened 2026-05-28
Admin password recovery endpoint change

Timshel:
Since `web-v2026.4.2` appears to have switched to `recover-account`: [cf](https://github.com/bitwarden/clients/blob/web-v2026.4.2/libs/admin-console/src/common/organization-user/services/default-organization-user-api.service.ts#L296).

More change should come cf https://github.com/bitwarden/clients/blob/web-v2026.4.2/apps/web/claude-plan.md :laughing:

### DIFF
--- a/playwright/tests/organization.smtp.spec.ts
+++ b/playwright/tests/organization.smtp.spec.ts
@@ -40,6 +40,16 @@ test('Invite users', async ({ page }) => {
     await createAccount(test, page, users.user1, mail1Buffer);
 
     await orgs.create(test, page, 'Test');
+
+    await test.step(`Set account recovery`, async () => {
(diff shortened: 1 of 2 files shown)


BlackDex (dismissed):

BlackDex:
We have to way to merge until we also have a PR for the web-vault to bump to 2026.6.0.

Timshel:
A yes, probably should have set the fields with a default `false` value this way it would stay compatible with current clients ?
Edit: especially since without it it makes no sense keeping the old endpoint since it would be broken.

BlackDex:
Ah, yes, that is also true, ill recheck.

BlackDex (approved):

Issue #7050, opened 2026-04-02
Google OIDC, users can create new accounts with the  "signups_allowed": false,


capsel22:
### Prerequisites

- [x] I have searched the existing **Closed _AND_ Open** [Issues](https://github.com/dani-garcia/vaultwarden/issues?q=is%3Aissue%20) **_AND_** [Discussions](https://github.com/dani-garcia/vaultwarden/discussions?discussions_q=)
- [x] I have searched and read the [documentation](https://github.com/dani-garcia/vaultwarden/wiki/)

### Your environment (Generated via diagnostics page)

* Vaultwarden version: v1.35.4
* Web-vault version: v2026.1.1
* OS/Arch: linux/x86_64
* Running within a container: true (Base: Debian)
* Database type: SQLite
* Database version: 3.50.2
* Uses config.json: true
* Uses a reverse proxy: true
* IP Header check: true (X-Real-IP)
* Internet access: true
* Internet access via a proxy: false
* DNS Check: true
* TZ environment: Europe/London
* Browser/Server Time Check: false
* Server/NTP Time Check: true
* Domain Configuration Check: true
* HTTPS Check: true
* Websocket Check: true
* HTTP Response Checks: true

### Config & Details (Generated via diagnostics page)

<details><summary>Show Config & Details</summary>

**Environment settings which are overridden:** DOMAIN, ADMIN_TOKEN

**Config:**

```json
{
  "_duo_akey": null,
  "_enable_duo": false,
  "_enable_email_2fa": false,
  "_enable_smtp": true,
  "_enable_yubico": false,
  "_icon_service_csp": "",
  "_icon_service_url": "",
  "_ip_header_enabled": true,
  "_max_note_size": 10000,
  "_smtp_img_src": "***:",
  "admin_ratelimit_max_burst": 3,
  "admin_ratelimit_seconds": 300,
  "admin_session_lifetime": 20,
  "admin_token": "***",
  "allowed_connect_src": "",
  "allowed_iframe_ancestors": "",
  "attachments_folder": "data/attachments",
  "auth_request_purge_schedule": "30 * * * * *",
  "authenticator_disable_time_drift": false,
  "data_folder": "data",
  "database_conn_init": "",
  "database_idle_timeout": 600,
  "database_max_conns": 10,
  "database_min_conns": 2,
  "database_timeout": 30,
  "database_url": "***************",
  "db_connection_retries": 15,
  "disable_2fa_remember": false,
[...]

Issue #2028, opened 2021-10-08
SIGNUPS_ALLOWED not stopping new user creation


JBFUK:
### Subject of the issue

SIGNUPS_ALLOWED not stopping new user creation

### Deployment environment

Latest Docker Image

* Install method: Docker Compose

* Clients used: Web-Vault

* Reverse proxy and version: Nginx

* MySQL/MariaDB or PostgreSQL version: MariaDB

* Other relevant details:

### Steps to reproduce
Set environment variable SIGNUPS_ALLOWED=false via docker-compose file

### Expected behaviour
Expect new user creation from the main login page to fail with an error (as described in WIKI)

### Actual behaviour
Allows new user creation and subsequent login

jjlin:
You probably have `SIGNUPS_DOMAINS_WHITELIST` set as well.

See https://github.com/dani-garcia/vaultwarden/wiki/Disable-registration-of-new-users#restricting-registrations-to-certain-email-domains.

jjlin:
These settings were implemented by different people at different times, so the design is not always the most cohesive.

`SIGNUPS_DOMAINS_WHITELIST` is documented as

https://github.com/dani-garcia/vaultwarden/blob/9930a0d75224202dbc355e945724aa545bedebe2/.env.template#L179-L181

The original implementation required `SIGNUPS_ALLOWED=false` to be set. Since the default is `SIGNUPS_ALLOWED=true`, a quite common issue was to set only `SIGNUPS_DOMAINS_WHITELIST` and expect that to limit the signups, which wasn't how it worked. So later `SIGNUPS_DOMAINS_WHITELIST` was changed to simply override `SIGNUPS_ALLOWED`, because why would you set `SIGNUPS_DOMAINS_WHITELIST` if you didn't expect that to control signups?

If you're interested in proposing and implementing another design that is still backward compatible, feel free.

pepa65:
Backward compatible meaning: we keep the same variables and the current results need to stay as they are? And then introduce new variables to attain the same range of outcomes more sensibly??

I think the outcomes I listed are fairly useful, and with the 3 boolean variables, there are only 2 outcomes that are overlapping, ie. if SIGNUPS_ALLOWED [S] and INVITATIONS_ALLOWED are both false, the value of SIGNUPS_DOMAINS_WHITELIST is irrelevant, and that makes sense.

@jjlin My proposal would be along the lines of the list of outcomes in my previous post. Do you think that is workable?

[...]

PR #7272, opened 2026-05-28
Add SSO_SIGNUPS_ALLOWED

Timshel:
Add configuration to disable SSO signups, useful when using a public provider.

Discussed in https://github.com/dani-garcia/vaultwarden/issues/7050

There is a bit of an issue with Organization invitation, if the invitation is deleted then the stub account will still be present and there is no way to know it when processing the login.

Could reuse the `Invitation` logic used when email are disabled but didn't want to make more change without discussion.

### DIFF
--- a/src/api/identity.rs
+++ b/src/api/identity.rs
@@ -285,7 +285,7 @@ async fn sso_login(
     // Will trigger 2FA flow if needed
     let (user, mut device, twofactor_token, sso_user) = match user_with_sso {
         None => {
-            if !CONFIG.sso_signups_allowed() {
+            if !CONFIG.is_sso_signup_allowed(&user_infos.email) {
                 if CONFIG.signups_domains_whitelist().is_empty() {
                     err!(
                         "Signups are disabled. You will need an invitation",
@@ -293,14 +293,13 @@ async fn sso_login(
                             event: EventType::UserFailedLogIn
                         }
                     );
-                } else if !CONFIG.is_email_domain_allowed(&user_infos.email) {
-                    err!(
-                        "Email domain not allowed",
-                        ErrorEvent {
-                            event: EventType::UserFailedLogIn
-                        }
-                    );
                 }
+                err!(
+                    "Email domain not allowed",
+                    ErrorEvent {
+                        event: EventType::UserFailedLogIn
+                    }
+                );
             }
 
             match user_infos.email_verified {
--- a/src/api/identity.rs
+++ b/src/api/identity.rs
@@ -220,6 +220,24 @@ async fn sso_login(
                     }
                 )
             }
+            Some((user, None))
+                if user.private_key.is_none()
+                    && !CONFIG.sso_signups_allowed()
+                    && !CONFIG.is_email_domain_allowed(&user.email)
+                    && !CONFIG.mail_enabled()
+                    && Invitation::find_by_mail(&user.email, conn).await.is_none() =>
+            {
+                error!(
+                    "Login failure ({}), no invitation with email ({}) was found",
+                    user_infos.identifier, user.email
+                );
+                err_silent!(
+                    "Missing invitation",
+                    ErrorEvent {
+                        event: EventType::UserFailedLogIn
+                    }
+                )
+            }
             Some((user, None)) if user.private_key.is_some() && !CONFIG.sso_signups_match_email() => {
                 error!(
                     "Login failure ({}), existing non SSO user ({}) with same email ({}) and association is disabled",
@@ -267,13 +285,22 @@ async fn sso_login(
     // Will trigger 2FA flow if needed
     let (user, mut device, twofactor_token, sso_user) = match user_with_sso {
         None => {
-            if !CONFIG.is_email_domain_allowed(&user_infos.email) {
-                err!(
-                    "Email domain not allowed",
-                    ErrorEvent {
-                        event: EventType::UserFailedLogIn
-                    }
-                );
+            if !CONFIG.sso_signups_allowed() {
+                if CONFIG.signups_domains_whitelist().is_empty() {
+                    err!(
+                        "Signups are disabled. You will need an invitation",
+                        ErrorEvent {
+                            event: EventType::UserFailedLogIn
+                        }
+                    );
+                } else if !CONFIG.is_email_domain_allowed(&user_infos.email) {
+                    err!(
+                        "Email domain not allowed",
+                        ErrorEvent {
+                            event: EventType::UserFailedLogIn
+                        }
+                    );
+                }
             }
 
             match user_infos.email_verified {
--- a/src/api/identity.rs
+++ b/src/api/identity.rs
@@ -299,7 +299,7 @@ async fn sso_login(
     // Will trigger 2FA flow if needed
     let (user, mut device, twofactor_token, sso_user) = match user_with_sso {
         None => {
-            if !CONFIG.sso_signups_allowed() {
+            if !CONFIG.is_sso_signup_allowed(&user_infos.email) {
                 if CONFIG.signups_domains_whitelist().is_empty() {
                     err!(
                         "Signups are disabled. You will need an invitation",
@@ -307,14 +307,13 @@ async fn sso_login(
                             event: EventType::UserFailedLogIn
                         }
                     );
-                } else if !CONFIG.is_email_domain_allowed(&user_infos.email) {
-                    err!(
-                        "Email domain not allowed",
-                        ErrorEvent {
-                            event: EventType::UserFailedLogIn
-                        }
-                    );
                 }
+                err!(
+                    "Email domain not allowed",
+                    ErrorEvent {
+                        event: EventType::UserFailedLogIn
+                    }
+                );
             }
 
             match user_infos.email_verified {
--- a/src/api/identity.rs
+++ b/src/api/identity.rs
@@ -234,6 +234,24 @@ async fn sso_login(
                     }
                 )
             }
+            Some((user, None))
+                if user.private_key.is_none()
+                    && !CONFIG.sso_signups_allowed()
+                    && !CONFIG.is_email_domain_allowed(&user.email)
+                    && !CONFIG.mail_enabled()
+                    && Invitation::find_by_mail(&user.email, conn).await.is_none() =>
+            {
+                error!(
+                    "Login failure ({}), no invitation with email ({}) was found",
+                    user_infos.identifier, user.email
+                );
+                err_silent!(
+                    "Missing invitation",
+                    ErrorEvent {
+                        event: EventType::UserFailedLogIn
+                    }
+                )
+            }
             Some((user, None)) if user.private_key.is_some() && !CONFIG.sso_signups_match_email() => {
                 error!(
                     "Login failure ({}), existing non SSO user ({}) with same email ({}) and association is disabled",
@@ -281,13 +299,22 @@ async fn sso_login(
     // Will trigger 2FA flow if needed
     let (user, mut device, twofactor_token, sso_user) = match user_with_sso {
         None => {
-            if !CONFIG.is_email_domain_allowed(&user_infos.email) {
-                err!(
-                    "Email domain not allowed",
-                    ErrorEvent {
-                        event: EventType::UserFailedLogIn
-                    }
-                );
+            if !CONFIG.sso_signups_allowed() {
+                if CONFIG.signups_domains_whitelist().is_empty() {
+                    err!(
+                        "Signups are disabled. You will need an invitation",
+                        ErrorEvent {
+                            event: EventType::UserFailedLogIn
+                        }
+                    );
+                } else if !CONFIG.is_email_domain_allowed(&user_infos.email) {
+                    err!(
+                        "Email domain not allowed",
+                        ErrorEvent {
+                            event: EventType::UserFailedLogIn
+                        }
+                    );
+                }
             }
 
             match user_infos.email_verified {
--- a/.env.template
+++ b/.env.template
@@ -494,6 +494,9 @@
 ## Prevent users from logging in directly without going through SSO
 # SSO_ONLY=false
 
+## Allow SSO flow to create account. You probably want to disable it when using a public provider.
+# SSO_SIGNUPS_ALLOWED=true
+
 ## On SSO Signup if a user with a matching email already exists make the association
 # SSO_SIGNUPS_MATCH_EMAIL=true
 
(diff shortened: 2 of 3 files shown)


src/api/identity.rs
```diff
@@ -267,13 +285,22 @@ async fn sso_login(
...
+                        }
+                    );
+                } else if !CONFIG.is_email_domain_allowed(&user_infos.email) {
+                    err!(
+                        "Email domain not allowed",
+                        ErrorEvent {
+                            event: EventType::UserFailedLogIn
+                        }
+                    );
+                }
```
stefan0xC: Is this correct? Does this not skip the check for whether an email domain is allowed if `SSO_SIGNUPS_ALLOWED=true`?

berglh: Hi @stefan0xC I don't write rust, but I'm assuming the first condition is negated, we are hoping to have a way to disable signups completely in SSO_ONLY mode and only allow existing users to attempt to login, it doesn't impact when `SSO_SIGNUPS_ALLOWED=true` as far as I can see, which is the intended default setting.

berglh: Oh, I see the, original `!CONFIG.is_email_domain_allowed` is removed, I think you are indeed correct @stefan0xC **Edit**: I guess you mean it should be something like this.

```rust
    let (user, mut device, twofactor_token, sso_user) = match user_with_sso {
        None => {
            if !CONFIG.sso_signups_allowed() {
                if CONFIG.signups_domains_whitelist().is_empty() {
                    err!(
                        "Signups are disabled. You will need an invitation",
                        ErrorEvent {
                            event: EventType::UserFailedLogIn
                        }
                    );
                } else if !CONFIG.is_email_domain_allowed(&user_infos.email) {
                    err!(
                        "Email domain not allowed",
                        ErrorEvent {
                            event: EventType::UserFailedLogIn
                        }
                    );
                } 
            } else if CONFIG.sso_signups_allowed() && !CONFIG.is_email_domain_allowed(&user_infos.email) {
                err!(
                    "Email domain not allowed",
                    ErrorEvent {
                        event: EventType::UserFailedLogIn
                    }
                );
            }
```

One comment I would have is that it seems to me that if `!CONFIG.signups_domains_whitelist().is_empty()`, then it will still allow the signup to occur, which might not be what users would expect if `!CONFIG.sso_signups_allowed()`.

**Edit** Maybe something like this is better, I'm not sure the edge case would be expected behaviour.

```rust
    let (user, mut device, twofactor_token, sso_user) = match user_with_sso {
        None => {
            if !CONFIG.sso_signups_allowed() {
                err!(
                    "SSO signups are disabled. You will need an invitation",
                    ErrorEvent {
                        event: EventType::UserFailedLogIn
                    }
                );
            }
            if !CONFIG.is_email_domain_allowed(&user_infos.email) {
                err!(
                    "Email domain not allowed",
                    ErrorEvent {
                        event: EventType::UserFailedLogIn
                    }
                );
            }
```

stefan0xC: Yes but I think the true question is whether `signups_domains_whitelist` should also automatically supercede `sso_signups_allowed` or not? If not we could probably make this easier by always just checking if an email is allowed (which already checks the domain whitelist) and also whether signup via sso is allowed.

berglh: Agreed the simplification is likely the expected behaviour.

Timshel: @stefan0xC O crap, ~~think what make it harder than necessary is the `whitelist.is_empty()` in `is_email_domain_allowed`, had added `is_email_domain_in_allowlist` then rollback.~~

Timshel: I think it makes sense to use the same allowlist for both MP/SSO settings.
So I just duplicated the `is_signup_allowed` in a `is_sso_signup_allowed`.

Timshel: As for supersede I believe maybe it should apply only when signup is disabled and not when enabled ?
But I think it should have it's own issue,

src/api/identity.rs
```diff
@@ -285,22 +285,21 @@ async fn sso_login(
     // Will trigger 2FA flow if needed
     let (user, mut device, twofactor_token, sso_user) = match user_with_sso {
         None => {
-            if !CONFIG.sso_signups_allowed() {
+            if !CONFIG.is_sso_signup_allowed(&user_infos.email) {
                 if CONFIG.signups_domains_whitelist().is_empty() {
```
berglh, line 289: Is this extra condition check `CONFIG.signups_domains_whitelist().is_empty()` superfluous if you're already testing it here [src/config.rs](https://github.com/dani-garcia/vaultwarden/pull/7272/changes/17eb54e0c533ce121a51c8060a9f6d75a6585ac6#diff-cba64c21ab992eaad29fce147a08f4560a4769bc14682b8a96081a5fd02dbecdR1523)?

Timshel: ? it's to return a different error message.

src/api/identity.rs
stefan0xC: > As for supersede I believe maybe it should apply only when signup is disabled and not when enabled ? But I think it should have it's own issue,

For whatever reason `SIGNUPS_ALLOWED` is superseded by `SIGNUPS_DOMAINS_WHITELIST`. This is already confusing for many users so I think that this should not be the case for normal registration in the first place. (I'll have to look up why we can't change this behavior but that's a separate point.)

Signup via SSO happens when you login so it should not care about `SIGNUPS_ALLOWED` at all since the mechanism is completely different from normal registration flow hence the need for a separate configuration option. Given the separate mechanism I don't see a reason to do the same for `SSO_SIGNUPS_ALLOWED` since you would only set `SSO_SIGNUPS_ALLOWED=false` if you want to completely disable signups for SSO (so a user could only finish registration when they have been invited).

edit: maybe if you would call this option something else like `SSO_ALLOW_ACCOUNT_CREATION_ON_LOGIN` the meaning would become clearer?

stefan0xC:
> There is a bit of an issue with Organization invitation, if the invitation is deleted then the stub account will still be present and there is no way to know it when processing the login.
> 
> Could reuse the `Invitation` logic used when email are disabled but didn't want to make more change without discussion.

I think that issue is also present in the normal invitation flow? At least with mails disabled and invitations allowed, if you invite someone and later rescind the invitation both an user account and also an entry in the invitations table will stay present, so even if you later disable invitations and signups the invited mail address can still be used to register an account. (I am not sure if this is possible with mail enabled but I have not yet tried it.)

src/api/identity.rs
Timshel: Since `SIGNUPS_DOMAINS_WHITELIST` is stricter and more involved to configure I can see the logic of always applying it as soon as it's defined. 
And it has the advantage that if the user misunderstands the logic it ends up being stricter than expected not making the instance public.

Changing the logic would probably require blocking application startup to prevent making it public (would be a good time to rename it to `SIGNUPS_DOMAINS_ALLOWLIST`).

> you would only set SSO_SIGNUPS_ALLOWED=false if you want to completely disable signups for SSO

Or in combination with `SIGNUPS_DOMAINS_WHITELIST` to allow only those domains.
I would not expect the allowlist to work when signup is allowed and even if I knew I would probably not rely on it.

> mechanism is completely different

I agree that they are different flow and require a different setting to control, but in the end both have onboarding to allow the same feature of creating an account with a Master password.
And in the current version they have the same interaction with the allowlist, as such it made sense to me to reuse the same naming convention.
Otherwise, I would wonder at the difference between `SIGNUPS` and `ACCOUNT_CREATION`.
But I don't really care ^^.

Timshel:
> and also an entry in the invitations table will stay present

Just checked, and it appears you are right, it's missing from the delete [endpoint](https://github.com/dani-garcia/vaultwarden/blob/d626ea81abf8504b7e54be14b0cfaf45134171b6/src/api/core/organizations.rs#L1677).
I'll make a PR to remove it.

src/api/identity.rs
stefan0xC: Well, with SSO it's often expected that you can just login and an account is created automatically (hence my suggestion `SSO_ALLOW_ACCOUNT_CREATION_ON_LOGIN`).

With an allowlist you restrict which domains are allowed for a user to register with if registrations are allowed (and also who can be invited if invitations are allowed). Whereas with signups_allowed you control whether registration via create new account is allowed or not.

Currently the allow list will override this setting, so you can always register an account if a domain has been allowed. https://github.com/dani-garcia/vaultwarden/blob/d626ea81abf8504b7e54be14b0cfaf45134171b6/.env.template#L272-L273

If we change the logic users won't be able to register with the create account link unless signups are allowed. Since the `is_email_domain_allowed` check will still be performed and thus the domain restriction would be in effect as well (unless invited via the `/admin` panel which skips this check). So to me this behavior would be more strict and also less confusing. Vaultwarden admins that want users with a specific domain to be able to register (the current behavior) would also have to allow registration again if we change that of course so maybe that's why we have not made this change, but I think that this would be fine because it would not allow anyone to register that was not able to register before.

Like I said I don't see a reason to reimplement this confusing behavior in SSO as well because if you want to restrict the allowed domains you would still be able to. Or in other words: If you link those configuration options like that you can't just disable account creation on login temporarily without also unconfiguring the allowed domain list.

Timshel: > SSO it's often expected that you can just login and an account is created automatically

Here there is some onboarding to do and after it unless it's an SSO only instance you can use the account without SSO.

> With an allowlist you restrict which domains are allowed for a user to register with if registrations are allowed 

Well we see it differently :sweat_smile:, I see it more first you block everyone then you add back who is allowed. Like for network traffic you first deny all traffic then add back what is allowed.
I guess we kind of differ on if the blocking itself is done by another setting (to me `SIGNUPS_ALLOWED`) or the presence of the setting itself.
 
 > I don't see a reason to reimplement this confusing behavior in SSO as well
 
 Biggest reason I see is not to have two similar logic which are configured differently. I believe it would be even more confusing.
 
 > you can't just disable account creation on login temporarily without also unconfiguring the allowed domain list
 
 It is a downside, but I find it more uncomfortable and error-prone to have to set `SIGNUPS_ALLOWED=true` for the allowlist to work.
 
 But I'll stop arguing more on it ^^.
 
 Action I can see:
 
 1. Keep it as is (same logic and naming convention), logic could be changed later on
 2. Wait for a change to be made to the existing logic to keep a similar logic between the two
 3. Switch to `SSO_ALLOW_ACCOUNT_CREATION_ON_LOGIN` with the allowlist being applied only if signups are allowed.
 
If you believe 3 make more sense then I'll stop arguing and make the change :) (and sorry for wasting your time).

stefan0xC: Why would the allow list not work / be applied only if you disable signups?

Timshel: If we are talking about 3.
I meant signups as `SSO_ALLOW_ACCOUNT_CREATION_ON_LOGIN=true`.

Since you mentioned that:

> you can't just disable account creation on login temporarily without also unconfiguring the allowed domain list.

Which would be `SSO_ALLOW_ACCOUNT_CREATION_ON_LOGIN=false`

stefan0xC: > It is a downside, but I find it more uncomfortable and error-prone to have to set SIGNUPS_ALLOWED=true for the allowlist to work.

I was responding to this statement, which is confusing to me.

Timshel: I mean if you are passing the allowlist setting through an env variable for example and make a typo in the variable name.
If it's only applied when `SIGNUPS_ALLOWED=true` then you'll test with a valid domain and the signup will work, and you might not realize that there is an issue.
As opposed to applying it when `SIGNUPS_ALLOWED=false`  where if there is a mistake you won't be able to log.

stefan0xC: Yes? If you make a typo, you can't register unless your email address happens to pass the domain check. However registration currently will be allowed **even if** `SIGNUPS_ALLOWED=false` because a non-empty allow list completely ignores the setting. https://github.com/dani-garcia/vaultwarden/blob/d626ea81abf8504b7e54be14b0cfaf45134171b6/src/config.rs#L1507-L1516 So as an admin you are wondering why people can still register even though you thought you have disabled it.

```rust
// My suggestion would be simply
pub fn is_signup_allowed(&self, email: &str) -> bool { 
    self.signups_allowed() && self.is_email_domain_allowed(email)
}
```
On the other hand if we were to implement my suggestion, you control whether people can signup with `SIGNUPS_ALLOWED`. If you further restrict the allowed emails to certain domains that would still be a requirement but you can be sure that even if you made a typo in the allow list no one can register (without an invitation) if you have disabled signups.

edit: I've created a PR #7287 for my suggestion

berglh: @stefan0xC @Timshel  This is the behaviour I expected when configuring `SIGNUPS_ALLOWED=false`.

> So as an admin you are wondering why people can still register even though you thought you have disabled it.
> ```rust
> // My suggestion would be simply
> pub fn is_signup_allowed(&self, email: &str) -> bool { 
>     self.signups_allowed() && self.is_email_domain_allowed(email)
> }
> ```
> 
> On the other hand if we were to implement my suggestion, you control whether people can signup with `SIGNUPS_ALLOWED`. If you further restrict the allowed emails to certain domains that would still be a requirement but you can be sure that even if you made a typo in the allow list no one can register (without an invitation) if you have disabled signups.

Timshel: @stefan0xC I was making the change and realized that `SSO_SIGNUPS_MATCH_EMAIL` already exists. 
Do you prefer to switch to `SSO_ALLOW_ACCOUNT_CREATION_ON_LOGIN` or keep `SSO_SIGNUPS_ALLOWED` ?

jjlin: FYI there is more background in #2028

manelrodero:
In which version is this variable?

We are using version 1.36.0 and have configured OIDC with Keycloak.

We have the same problem explained here: https://github.com/dani-garcia/vaultwarden/issues/7050.

Even though we have:

"signups_allowed": false,
"signups_verify": false,

any user who authenticates with SSO is redirected to the screen to create the Master Password, and the user is created in Vaultwarden.

This is not ideal. Our Keycloak provides access to the entire organization (a university), and this Vaultwarden installation is for a single department. We cannot use the allowed domains because we all have the same one.

Our idea was to prevent the automatic creation of users simply by authenticating with SSO. We would prefer to create/invite users who are authorized to log in and then use SSO for them to authenticate.

If this isn't possible, the team managing Keycloak suggests they could return some user attributes (unitCode and unit). However, Vaultwarden would then need to be able to access these attributes and only allow logins for those we specify (e.g., unitCode=10 or unit=XXX).

Any suggestions would be welcome.

Thanks.

Timshel:
@manelrodero It's a bit stalled but the role mapping from https://github.com/dani-garcia/vaultwarden/pull/6158 is kind of made for this, where you would grant the user role only to specific department.

BlackDex (approved):

PR #7283, opened 2026-06-01
fix(sends): emit hideEmail as non-null boolean in sync response

kvdb:
### Problem

`GET /api/sync` returns `"hideEmail": null` for Sends whose `hide_email` column is NULL. The `sends.hide_email` column is `Nullable<Bool>` with no default (added in migration `2021-05-11-205202_add_hide_email`), so it is NULL for older Sends and any code path that leaves it unset.

The Bitwarden Android client deserializes `SyncResponseJson.Send.hideEmail` as a non-null Kotlin `Boolean` and throws a `JsonDecodingException`, aborting the **entire** sync. Web, desktop and CLI clients coerce null to `false`, so only accounts with at least one Send are affected, and only on Android.

The official Bitwarden server treats `hideEmail` as non-nullable.

### Fix

Default `None` to `false` at the serialization boundary in `Send::to_json()`:

```rust
"hideEmail": self.hide_email.unwrap_or(false),
```

The field stays `Option<bool>` internally. No DB migration is required, and this fixes both existing NULL rows and any future NULLs.

### DIFF
--- a/src/db/models/send.rs
+++ b/src/db/models/send.rs
@@ -161,7 +161,7 @@ impl Send {
             "password": self.password_hash.as_deref().map(|h| BASE64URL_NOPAD.encode(h)),
             "authType": if self.password_hash.is_some() { SendAuthType::Password as i32 } else { SendAuthType::None as i32 },
             "disabled": self.disabled,
-            "hideEmail": self.hide_email,
+            "hideEmail": self.hide_email.unwrap_or(false),
 
             "revisionDate": format_date(&self.revision_date),
             "expirationDate": self.expiration_date.as_ref().map(format_date),


src/db/models/send.rs
```diff
@@ -379,3 +379,36 @@ pub mod id {
...
+        assert_eq!(json["hideEmail"], Value::Bool(false));
+        assert!(!json["hideEmail"].is_null());
+    }
+
+    #[test]
+    fn to_json_hide_email_some_round_trips() {
+        assert_eq!(minimal_send(Some(true)).to_json()["hideEmail"], Value::Bool(true));
+        assert_eq!(minimal_send(Some(false)).to_json()["hideEmail"], Value::Bool(false));
+    }
+}
```
BlackDex: For me this test is too much. Especially if this is only testing a single value.
So from my point of view, this can be removed.

Or, it should check and validate the whole response, not just the `hideEmail` that would be way to narrow.

BlackDex (changes requested):

src/db/models/send.rs
kvdb: I'm not at home in rust, so I asked AI to help me there. I think we should at least have one test to make sure this functionality stays working in the future. So I collapsed it into one test, with as you mentioned the full response. Hope it's better this way.
I already tested the patch locally for at least a week now at my vaultwarden instance, no sync problems so far.

BlackDex: The current test is useless, so, if you used AI, then just remove the whole test, it doesn't add anything useful.

BlackDex (approved):

PR #7284, opened 2026-06-01
Org membership delete remove Invitation

Timshel:
While discussing https://github.com/dani-garcia/vaultwarden/pull/7272 realized that the `Invitiation` is not deleted when the membership is deleted through the Orgnaization membership.

And while testing I realized that if the user stub is already present then the `Invitation` is not created again.

Additionally, I wonder if the invitations check should not be done again when the user stub is present ? cf:

```rust 
                if !CONFIG.invitations_allowed() {
                    err!(format!("User does not exist: {email}"))
                }
                if !CONFIG.is_email_domain_allowed(email) {
                    err!("Email domain not eligible for invitations")
                }
``` 

And realized that the revoked flow does not handle the `Invitations`  creation/deletion nor check again the invitation/domain is still allowed on restoration.

Wanted some feedback before making more modifcations.

### DIFF
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -1089,9 +1089,13 @@ async fn send_invite(
                     err!(format!("User already in organization: {email}"))
                 }
 
-                // automatically accept existing users if mail is disabled
-                if !CONFIG.mail_enabled() && !user.password_hash.is_empty() {
-                    member_status = MembershipStatus::Accepted as i32;
+                if !CONFIG.mail_enabled() {
+                    if user.password_hash.is_empty() {
+                        Invitation::new(email).save(&conn).await?;
+                    } else {
+                        // automatically accept existing users if mail is disabled
+                        member_status = MembershipStatus::Accepted as i32;
+                    }
                 }
                 user
             }
@@ -1713,6 +1717,15 @@ async fn delete_member_impl(
 
     if let Some(user) = User::find_by_uuid(&member_to_delete.user_uuid, conn).await {
         nt.send_user_update(UpdateType::SyncOrgKeys, &user, headers.device.push_uuid.as_ref(), conn).await;
+
+        if !CONFIG.mail_enabled()
+            && !Membership::find_invited_by_user(&user.uuid, conn)
+                .await
+                .into_iter()
+                .any(|m| m.uuid != member_to_delete.uuid)
+        {
+            Invitation::take(&user.email, conn).await;
+        }
     }
 
     member_to_delete.delete(conn).await


stefan0xC:
> Additionally, I wonder if the invitations check should not be done again when the user stub is present ?

Well, if we add those checks we could not invite someone immediately to an existing organizations anymore if they were just invited via the `/admin` panel (to circumvent those restrictions) so they would first have to finish registration before we could invite them.

Personally I would (still) like to see something that automatically deletes stale invites (and also stub user accounts) after a couple of days. I mean @BlackDex was against it if I remember correctly but maybe we can make this configurable (since the invitation token life span is configurable it would have to be at least `invitation_expiration_hours` long). So maybe I'll give #2795 another attempt.

BlackDex:
I agree with @stefan0xC, adding those checks is probably not useful.

@stefan0xC, thinking about this again I'm not really against it. But then we would at least somehow log this somewhere. Maybe we could use the `OrganizationUser_Removed` event https://github.com/bitwarden/server/blob/v2026.6.0/src/Core/Dirt/Enums/EventType.cs to at least log it somewhere. And also just output it as an info message into the default logging? That way there at least is an audit trail that the user was invited and later removed if it is for an org invite of course.

For other invites adding it to the logs is probably just fine then. It does cleanup the database and does not leave dangling invites there, also maybe for someone else to takeover in someway.

BlackDex (approved):

maximilize:
The `send_invite` hunk looks right and fixes a real latent issue: re-inviting an email that already has a stub user with mail disabled previously created no `Invitation` row, leaving the invitee unable to register when signups are off. `Invitation::save` is idempotent, so no double-insert risk.

One thing worth weighing on the new delete hunk before the invite lifecycle is finalized (separate from the re-invite check @stefan0xC raised): the `invitations` table is keyed by email and **shared** with emergency access (and admin invites), but `Invitation::take` in `delete_member_impl` decides purely from org-membership status, so it can remove a row that a pending emergency-access invite still needs.

Path: mail disabled + invitations allowed → an emergency-access invite for a new email creates the same row (`emergency_access.rs:236`, re-created at `:318`) plus a stub user. If that email is also org-invited and the org membership is later removed, this block fires — `find_invited_by_user` only returns the membership being deleted, so `any(uuid != ...)` is false and `take` runs — deleting the shared row. `register_verification_email` (`identity.rs:1050-1051`) then rejects the grantee, because unlike `accounts.rs:272-274` it has no emergency-access fallback. This deletion is new behavior introduced by this hunk.

Might be worth gating the `take` on "no pending emergency-access invite for this email" too, or centralizing the shared-invitation lifecycle.

PR #7296, opened 2026-06-03
[v2026.5.0] PutPolicy now using vnext format

Timshel:
It appears that the put policy logic now use the old endpoint with the new format.

### DIFF
--- a/src/api/core/organizations.rs
+++ b/src/api/core/organizations.rs
@@ -2020,18 +2020,27 @@ struct PolicyData {
     data: Option<Value>,
 }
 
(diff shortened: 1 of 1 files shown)


BlackDex (approved):

BlackDex:
We have to way to merge until we also have a PR for the web-vault to bump to 2026.6.0.

Timshel:
I believe current client are using the `/organizations/<org_id>/policies/<pol_type>/vnext` which should be unchanged.
So it should be safe to merge now.

PR #7248, opened 2026-05-24
Fix playwright: get the test suite back to green against bundled web-vault v2026.4.1

zaid-marji:
## Summary

The bundled web vault (`v2026.4.1` per `docker/DockerSettings.yaml`) refuses
the playwright suite's interactions for several compounding reasons: stale
label selectors, stale org-nav selectors, a renamed invitation-acceptance
toast, an SSO existing-account flow that no longer auto-redirects to the
IdP, and the web vault refusing plain HTTP. `login.spec.ts` and every spec
calling `createAccount` / `logUser` / `activateTOTP` / `orgs.*` /
`*.smtp.spec.ts` org-invite flows fails before doing anything useful. This
PR restores the suite to green against the current bundled vault, and along
the way centralises the selectors and 2FA-challenge flow so the next
web-vault bump has one place to update.

Eight commits, each with a single concern (commit bodies have the full
motivation):

- **`fix stale Master password selectors`** — the input lost its
  `(required)` suffix. Switch to `getByLabel('Master password')` and anchor
  the ambiguous new/confirm fields by `formcontrolname`.
- **`fix TOTP setup flow`** — Continue race with Angular `updateOn: 'blur'`
  (submit via `Enter`), Key selector ambiguous with Yubico SVG aria-label
  (anchor `{ exact: true }`), Close button transiently disabled during the
  success-dialog transition (drop the click; dialog auto-closes).
- **`serve the test Vaultwarden over HTTPS`** — the web vault refuses
  registration over HTTP. Self-signed cert generated in the test image,
  `ROCKET_TLS` pointed at it, `DOMAIN` switched to `https://`,
  `ignoreHTTPSErrors: true` set on every playwright context.
- **`gitignore: catch stray playwright artifacts at the repo root`** —
  adds `test-results/` and `playwright-report/` to root `.gitignore`.
- **`fix stale org-nav selectors`** — product-switch nav links and the
  in-org "Members" entry are icon-only; switch to
  `getByRole('link', { name: '…' })` and force-click past the hover
  tooltip that intercepts the click.
- **`fix invitation-accepted toast text + SSO existing-account redirect`**
  — toast is now `"Successfully accepted your invitation"`; the SSO
  invited-with-existing-account flow no longer auto-redirects to Keycloak
  (explicit "Use single sign-on" click required).
- **`centralize web-vault selectors into shared setup helpers`** — extracts
  the avatar menu, registration MP-fill, user-verification gate, and the
[...]

PR #6820, opened 2026-02-12
Add webauthn login

samos667:
Inspired from https://github.com/dani-garcia/vaultwarden/pull/5929 when trying to address raised issues, work for firefox and chrome. I have not yet tested the migration.

PR #7297, opened 2026-06-03
Add webauthn login (consolidated)

zaid-marji:
## Summary

Adds the long-requested **"Log in with passkey"** feature: passwordless
account login via a discoverable WebAuthn credential, with optional
PRF-based vault unlock. This consolidates the prior efforts in #6820
(@samos667) and #5929 (@zUnixorn) into a single review-ready PR, as the
maintainers suggested on those threads. The foundational commit retains
@samos667 as author and credits @zUnixorn via `Co-authored-by:`; the
remaining commits are my own.

## What it does

- **Passwordless login** (`grant_type=webauthn`): a discoverable, user-verified
  WebAuthn assertion authenticates the user. Per Bitwarden's model the
  authenticator's user-verification is the second factor, so TOTP is not
  re-prompted.
- **PRF vault unlock**: a passkey can carry a wrapped key set so the vault can
  be unlocked with the passkey. Exposed via `webAuthnPrfOptions` on `/sync`
  and `UserDecryptionOptions` at login, matching the Bitwarden web client.
- **Passkey management**: enroll, list, and delete account passkeys
  (`/api/webauthn*`), each gated by password/OTP.
- **Key rotation**: `post_rotatekey` re-wraps the account-key-wrapped PRF
  material so rotating the account key no longer silently breaks passkey unlock.
- **`prfStatus`** implemented (Enabled / Supported / Unsupported), replacing
  the previous placeholder.
- Availability advertised via the `pm-2035-passkey-unlock` config flag and
  suppressed under `SSO_ONLY` or a WebAuthn-incompatible `DOMAIN`.

## Threat model

The login and key-handling paths defend against:

- **Replay** of a captured assertion (single-use transactional login
  challenges; signature-counter advance with CAS rollback protection).
- **Account/credential enumeration** via the unauthenticated grant
  (uniform `Passkey authentication failed` response across every failure
  branch — unknown user, missing credential, verification failure, account
  disabled or unverified — gated by an IP-shared login rate limit).
- **Cross-user credential misuse** (every passkey CRUD and read is
  per-user-scoped at the DB layer; credential ID hashes are unique per user
  rather than global).
- **Passkey-unlock breakage on key rotation** (`post_rotatekey` re-wraps each
  PRF credential's account-key-wrapped material under the new account key as
  part of the rotation; a concurrent passkey delete during rotation is
  tolerated as a best-effort degraded no-op).
- **Audit-log poisoning by anonymous probes** (pre-verification failures
  deliberately omit the `UserFailedLogIn` event so an attacker submitting
  forged user-handles cannot create audit entries on the victim's account).

## Hardening

A few changes also tighten **shared** paths beyond passkey login — flagged here
since they affect more than the new feature:

- **2FA-policy enforcement is now fail-closed** (`enforce_2fa_policy`): a
  transient DB error while loading a user's Require-2FA org memberships
  previously collapsed to an empty set and silently skipped enforcement; it now
  propagates. This closes the same gap for password login and recovery-code use.
- **SSO_ONLY is enforced deny-by-default** at the token endpoint across all
  grant types, replacing the previous password-only per-grant check.
- **The 2FA-WebAuthn endpoints and `is_webauthn_2fa_supported` are hardened
  against a malformed `DOMAIN`** that could previously panic a worker thread
  (`.expect()` → graceful refusal); every WebAuthn entry point is now gated.
- **Several shared DB lookups now return `Result`** instead of swallowing
  errors into `None`, so a transient failure surfaces rather than being misread
  as "not found".

## Database

Two new **additive** tables (`web_authn_credentials`,
`web_authn_login_challenges`) with migrations for all three backends. No
existing data is migrated — `webauthn-rs` 0.5 is already in `main`, so the
2FA-object migration discussed on #5929 does not apply here.

## Testing

- Builds on #7248, which brought the Playwright suite back to green against the
  current bundled web-vault; the new E2E specs run on that harness, so whoever
  reviews/merges this may want to take #7248 alongside it. The first 8 commits
  of this PR's diff are #7248's; the webauthn-specific work starts at
  `1cdf0b7b`.
- `cargo fmt` / `clippy` / `test` and the Playwright E2E suite pass. New Rust
  unit tests cover the PRF/UserDecryption response shapes, challenge
  freshness + single-use, and passkey-rotation validation.
- Playwright coverage: passkey login, PRF enrolment + unlock, the full
  enrol → login → lock → unlock → rotate → remove lifecycle, the rejection
  paths (disabled account, deleted credential, runtime-unusable 2FA), SSO_ONLY
  denial, and malformed-input rejection.

## Compatibility

Wire shapes (`UserDecryptionOptions`, `webAuthnPrfOptions`, `GetPrfStatus`)
follow the upstream Bitwarden server so the official web client works
unchanged.

---

I'm glad to answer questions or make any changes needed to help move this
along — just let me know.


BJReplay:
Well set out PR and commit chain.

You have clearly set out the work you have done, so hopefully any gaps are easy to spot - but it looks like you have done a lot of work to try to make sure that there aren't any. 

I hope that the previous contributors that you have recognised as having done so much of the work -  and that you have credited as authors or co-authors get behind this PR.

BlackDex:
Ok, this PR for me is a mess, AI merged two stuff and added some extra's?
Checking and vetting this is going to take a long time from my point of view, I'm not sure if this really helps getting this feature into Vaultwarden t.b.h.

I don't mind the usage of AI in general, as a tool to help figure out stuff or as a sparring partner for an idea. But having this, that for me is too much, I'm not sure what others think about this.

And, I'm not sure, and do not take this the wrong way, but it looks like your rust experience in coding is none or very little.

Also this PR is way out-of-style compared to everything else inside the whole code base.
And, i do not mind tests, but some tests here are just too much.

And this was only from a quick glance over this generated code.

zaid-marji:
Thank you for the feedback @BlackDex. Yes, I am not an experienced Rust developer. However, I am a developer, and I wrote this code with the help of Claude Opus 4.8 in agentic mode.

I have been carefully reviewing the decisions made by the agent. I have also been working since yesterday on reverting some hardening choices that were too aggressive with the intention of submitting them as separate PRs to reduce load on reviewers. [Which have just been pushed. No further edits are planned, unless requested.]

I can upon request reduce this PR to a minimal working PR, and leave hardening to separate PRs. As I said, I have been extensively discussing design choices with Claude Opus 4.8, so this code was written with close human oversight.

As for the tests, yes, I tend to come from a background of using Test-Driven Development (was not followed strictly in this PR), so I made sure tests were providing extensive coverage.

The first 8 commits are intended to be a separate PR to bring the Playwright test suite up to speed (it is not passing on main). However, since that PR has not been merged, I decided it will be at the root of this PR, since I have written many Playwright tests, and those fixes are required to ensure it passes.

zaid-marji:
Just to reduce the apprehension of any potential reviewer. This PR's added lines are composed of:
- Rust code (production)	1,459
- Rust tests	1,103
- TypeScript (tests)	1,379
- Comments	1,698
- Blank lines	600
- Other	126

So, the actual code diff is less than 25% of the diff lines. The rest are tests and comments.

an1ndra:
All changes looks like generated by Copilot 😕

gh10ltz:
As much as I endorse these features and for as long as I've been following the related PRs, I do not trust this author or the code quality to be comfortable with seeing this merged. This PR is a liability, risk, and hazard, in my opinion.

Mikilio:
Given the lack of options I believe this would be at least worth a review. A 6.5k diff is very huge though and from what I read there isn't really a reason to have it be that large. Would you be open to split it up? That would make it easier to reason about it. AI or not it's the closest we have to a working feature.

Yeradon:
> Given the lack of options I believe this would be at least worth a review. A 6.5k diff is very huge though and from what I read there isn't really a reason to have it be that large. Would you be open to split it up? That would make it easier to reason about it. AI or not it's the closest we have to a working feature.

There has been a quick first impression review by a maintainer already. The author himself claimed that he does not have a lot experience with rust and thus not with this codebase either. This is not a topic that should be tackled by someone who lacks the experience to critically review the output of an ai on that matter. 

And the question for review is basically asking a maintainer to do the heavy lifting here. As we all know going from working to merge ready, can be the majority of the work.

PR #7308, opened 2026-06-06
Fix invalid SSH key sync for Bitwarden 2026.5

ix64:
## Description

  Fixes sync responses for invalid SSH key cipher data on Bitwarden 2026.5 clients.

  SDK-backed clients expect SSH key cipher data to include string values for `keyFingerprint`, `privateKey`, and `publicKey`. Returning `null` for invalid SSH key data can make SDK/WASM deserialization fail before the user can edit or delete the bad item.

  This changes invalid SSH key handling to return encrypted placeholder values for the mandatory SSH key fields instead of `null`, allowing clients to continue processing the cipher.

  ## Evidence

  In `bitwarden/clients`, SSH key cipher type is `5`:
  https://github.com/bitwarden/clients/blob/8652881a5019a800381274566b9b67b77f8c6043/libs/common/src/vault/enums/cipher-type.ts#L1-L7

  The client API model defines `privateKey`, `publicKey`, and `keyFingerprint` as string fields:
  https://github.com/bitwarden/clients/blob/8652881a5019a800381274566b9b67b77f8c6043/libs/common/src/vault/models/api/ssh-key.api.ts#L5-L17

  The Bitwarden SDK SSH key view also models the decrypted fields as non-optional `String` values, and parsing `CipherSshKeyModel` requires all three encrypted fields:
  https://sdk-api-docs.bitwarden.com/src/bitwarden_vault/cipher/ssh_key.rs.html#35
  https://sdk-api-docs.bitwarden.com/src/bitwarden_vault/cipher/ssh_key.rs.html#89

### DIFF
--- a/src/db/models/cipher.rs
+++ b/src/db/models/cipher.rs
@@ -94,6 +94,20 @@ impl Cipher {
         }
     }
 
+    fn ssh_key_data_has_mandatory_fields(type_data_json: &Value) -> bool {
+        type_data_json["keyFingerprint"].as_str().is_some_and(|s| !s.is_empty())
+            && type_data_json["privateKey"].as_str().is_some_and(|s| !s.is_empty())
+            && type_data_json["publicKey"].as_str().is_some_and(|s| !s.is_empty())
+    }
+
+    fn invalid_ssh_key_data_placeholder(encrypted_fallback: &str) -> Value {
+        json!({
+            "keyFingerprint": encrypted_fallback,
+            "privateKey": encrypted_fallback,
+            "publicKey": encrypted_fallback,
+        })
+    }
+
     pub fn validate_cipher_data(cipher_data: &[CipherData]) -> EmptyResult {
         let mut validation_errors = serde_json::Map::new();
         let max_note_size = CONFIG._max_note_size();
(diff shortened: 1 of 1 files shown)


strau0106:
It would be lovely if this could be merged! I cant access my Vault from the auto-updating browser extensions currently. This PR would appear to fix that issue... let me know, dear maintainers, if i could help with anything to get this merged.

Blacks-Army:
Is this maybe related to this issue https://github.com/issues/created?issue=bitwarden%7Cclients%7C21199 ?

davidhidvegi:
Same here, I had to downgrade the extension to `Version 2026.4.0` for it to work.

strau0106:
I was able to finally fix my issue, which in the end does not seem to be related to this pr. Empty ssh keys appeared to be "stuck" in the bin. After deleting them through CLI and web interface all browser extensions appear to function properly again.

src/db/models/cipher.rs
```diff
@@ -94,6 +94,20 @@ impl Cipher {
...
+            && type_data_json["publicKey"].as_str().is_some_and(|s| !s.is_empty())
+    }
+
+    fn invalid_ssh_key_data_placeholder(encrypted_fallback: &str) -> Value {
+        json!({
+            "keyFingerprint": encrypted_fallback,
+            "privateKey": encrypted_fallback,
+            "publicKey": encrypted_fallback,
+        })
+    }
```
BlackDex, line 109: I'm not seeing the usefulness of a separate function here. Adding this in-line is probably better in my opinion.

src/db/models/cipher.rs
```diff
@@ -1185,3 +1194,46 @@ impl Cipher {
...
+
+    #[test]
+    fn invalid_ssh_key_data_placeholder_uses_encrypted_fallback_for_required_fields() {
+        let placeholder = Cipher::invalid_ssh_key_data_placeholder("2.encrypted-name");
+
+        assert_eq!(placeholder["keyFingerprint"], "2.encrypted-name");
+        assert_eq!(placeholder["privateKey"], "2.encrypted-name");
+        assert_eq!(placeholder["publicKey"], "2.encrypted-name");
+    }
+}
```
BlackDex, line 1239: I'm not against tests, but this seems too much. This would mean we need to create a test for all other items too, like URI, fido etc.. etc... etc...
Which could all be moving targets. Since we basically do not try to validate everything which is sent we keep it flexible, but that also means we could allow invalid stuff, or new stuff which we do not check, and that will provide a false sense of correctness.

Now, i do not think a lot will change for the ssh keys, since this is the basic part needed. But i think these tests can be removed.

BlackDex (changes requested):
I have not tested if this will cause any other issues, but i at least left some remarks

PR #7346, opened 2026-06-17
2026.6.0 send support

Timshel:
With `2026.6.0` the client now start by requiring an access_token.

Some remarks:

- Added `auth/send.rs` since I though that `auth.rs` was already long enough.
- Added the `silent` property to the `Error` to be use `ErrorKind::Json` without logging the error.
  Probably could be injected in some place like the `Error::empty` but wanted to keep modifications to a minimum.
  
  While adding another set of optional values to `ConnectData` I wondered if it could be converted to an `Enum` with a value for each `grant_type` ? I think it would be cleaner, but maybe I'm missing something ?

### DIFF
--- a/src/api/core/sends.rs
+++ b/src/api/core/sends.rs
@@ -12,7 +12,7 @@ use serde_json::Value;
 use crate::{
     CONFIG,
     api::{ApiResult, EmptyResult, JsonResult, Notify, UpdateType},
-    auth::{ClientIp, Headers, Host},
+    auth::{ClientIp, Headers, Host, SendHeaders},
     config::PathType,
     db::{
         DbConn, DbPool,
@@ -48,7 +48,9 @@ pub fn routes() -> Vec<rocket::Route> {
         post_send,
         post_send_file,
         post_access,
+        post_access_legacy,
         post_access_file,
+        post_access_file_legacy,
         put_send,
         delete_send,
         put_remove_password,
@@ -371,7 +373,7 @@ pub struct SendFileData {
 }
 
 // https://github.com/bitwarden/server/blob/9ebe16587175b1c0e9208f84397bb75d0d595510/src/Api/Tools/Controllers/SendsController.cs#L195
-#[post("/sends/<send_id>/file/<file_id>", format = "multipart/form-data", data = "<data>")]
+#[post("/sends/<send_id>/file/<file_id>", format = "multipart/form-data", data = "<data>", rank = 2)]
 async fn post_send_file_v2_data(
     send_id: SendId,
     file_id: SendFileId,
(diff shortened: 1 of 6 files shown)


BlackDex:
@Timshel, which clients btw? I seem to be able to use the old implementation using the web-vault just fine.
I have not tested other clients yet though.

Timshel:
@BlackDex Hum :thinking:
It was changed in the client I built based on `2026.6.0` (EX: https://github.com/Timshel/oidc_web_vault/releases/tag/v2026.6.0-3)

BlackDex:
Could be a PR/Commit has been missed on our side then? @stefan0xC?
I used the artifact from here: https://github.com/dani-garcia/bw_web_builds/actions/runs/27478766767

Timshel:
Hum both client appears to be based on https://github.com/Timshel/oidc_web_vault/commit/924ba4b03c28028e34e75ca92965d4c2ba28cdbb / https://github.com/vaultwarden/vw_web_builds/commit/924ba4b03c28028e34e75ca92965d4c2ba28cdbb

stefan0xC:
I think creation works just fine the problem is viewing the send? I get an error 
```[2026-06-21 16:46:53.209][vaultwarden::api::identity][ERROR] Invalid type. send_access```

image

**Comment by Timshel** [comment]
A yes sorry was not clear in the issue the problem is with viewing :sweat_smile:

**Comment by BlackDex** [comment]
Strange, didn't had issue with that either.
Maybe i still had the old vault loaded then?

File: src/api/core/sends.rs [hunk 1]
```diff
@@ -479,6 +495,13 @@ async fn post_access(
...
     }
 
+    // Files are incremented during the download
+    if send.atype == SendType::Text as i32 {
+        send.access_count += 1;
+    }
+
+    send.save(&conn).await?;
+
     if send.password_hash.is_some() {
```
BlackDex: @Timshel What is the reason for moving this before the password check?
This now could cause someone who typed the wrong password, and access count is set to 1 to be unable to access it anymore even with a valid password.

src/auth/send.rs
```diff
@@ -0,0 +1,172 @@
...
+        })
+    }
+}
+
+pub struct SendHeaders {
+    #[expect(dead_code)]
+    pub host: String,
+
+    #[expect(dead_code)]
+    pub ip: ClientIp,
```
BlackDex: If these are not used, then, we probably do not need them.
Is there a specific reason you added these?

It seems to work fine without them.

BlackDex (changes requested):
LGTM, apart from a few questions.

src/api/core/sends.rs
Timshel: No reason, probably moved the code when adding the `process_access` and should have checked the diff again.

src/auth/send.rs
Timshel: I think I was expecting to use them to log some errors and forgot to clean it later on.

BlackDex (approved):

PR #7347, opened 2026-06-18
Add Vaultwarden OIDC Post

puria:
## Problem

Vaultwarden SSO uses the `openidconnect` crate which defaults to HTTP Basic Auth (`client_id:client_secret` in the `Authorization` header) for token endpoint requests.

[Stalwart Mail Server](https://stalw.art)'s OIDC provider only accepts client credentials as `client_id` and `client_secret` POST body parameters — it does not support the Basic Auth header method. This causes the token exchange to fail when using Stalwart as the OIDC identity provider for Vaultwarden SSO.

## Fix

Set `.set_auth_type(openidconnect::AuthType::RequestBody)` on the client, so credentials are sent as `application/x-www-form-urlencoded` POST body parameters instead of in the `Authorization` header.

This is consistent with RFC 6749 Section 2.3.1 which says clients using the "client password" authentication method may include credentials in the request body. The `openidconnect` crate supports both methods — the fix simply selects the one that Stalwart (and other providers that follow the body-only convention) expects.

## Testing

Tested with Stalwart OIDC as the identity provider. Token exchange succeeds after this change where it previously failed with an authentication error.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/sso_client.rs
+++ b/src/sso_client.rs
@@ -114,7 +114,8 @@ impl Client {
             Ok(metadata) => metadata,
         };
 
-        let base_client = CoreClient::from_provider_metadata(provider_metadata, client_id, Some(client_secret));
+        let base_client = CoreClient::from_provider_metadata(provider_metadata, client_id, Some(client_secret))
+            .set_auth_type(openidconnect::AuthType::RequestBody);
 
         let token_uri = if let Some(uri) = base_client.token_uri() {
             uri.clone()


BlackDex:
This of course will break providers that do not support this. 
So this is probably bad for all those clients.

puria:
> This of course will break providers that do not support this. So this is probably bad for all those clients.

Thanks @BlackDex for your quick reply! I was trying to use stalwart as the OIDC provider as per https://stalw.art/docs/auth/openid/provider/ and didn't went smotth with vaultwarden ;p

Some vibe coding helped to deploy the service and with the fix seems, acceptable. Surely can be improved!

stefan0xC:
Since [OpenID Connect Core 1.0](https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication) specifies `client_secret_basic` as the default method if no other method is optionally defined I would consider it a requirement. Especially if you look up the referenced [RFC6749 Section 2.3.1](https://www.rfc-editor.org/info/rfc6749/#section-2.3.1) which requires that

> The authorization server MUST support the HTTP Basic authentication scheme for authenticating clients that were issued a client password.

Whereas support for the `client_secret_post` method of client authentication is only optional:

> Alternatively, the authorization server MAY support including the client credentials in the request-body [...]

BlackDex:
Well, as @stefan0xC mentioned, it looks like your OIDC Provider doesn't uphold the RFC.
Their might be room for improvement maybe, but clearly vibe-coding didn't do a great job in keeping it compatible with other OIDC Providers. Which is probably fine for your use-case, but we have to also support other endpoints.

Maybe @Timshel has some ideas about this. But since this PR is vibe-coded, and i suspect you have no Rust experience at all, coding this in a way Vaultwarden is coded is probably going to be too much.

Therefor I'm going to close this for now.

puria:
Yep ok! Thanks for your time <3

Timshel:
Had a similar request recently but was closed before I had a look at it ^^. https://github.com/Timshel/OIDCWarden/issues/75

Unless you are against adding another configuration, setting it depending on a simple boolean like `SSO_AUTHORIZE_BODY` could do the trick.

BlackDex:
Are there any other options besides these two? I think I saw someone mentioning something like a Bearer token also? But that might not be something similar regarding this?

Regarding adding this feature, if people need this for the SSO to work, and their OIDC Provider doesn't support the RFC needed protocols, then it might be an option of course.

Also, reading a bit more might it also be an option to read the `.well-known/openid-configuration` our self and extract the supported `token_endpoint_auth_methods_supported` and see if either `client_secret_basic` or `client_secret_post` is listed, and then use one of the two?

That way no flag is needed i think?

Timshel:
@BlackDex hum good catch was a bit lazy :sweat_smile: 
I'll check it :)

BlackDex:
Cool, not sure if there are other auth's we could support, but if none of the those two are listed, then we can also error with a nice message stating no auth method in the list is supported or something.

PR #7354, opened 2026-06-20
enable rocket/mtls feature

nagisa:
I want mTLS for the communication between my vaultwarden instance and the reverse proxy that's fronting vaultwarden. For various reasons my reverse proxy is hosted on a different machine within the same LAN than vaultwarden and I don't want to put another reverse proxy on the machine with vaultwarden just for this.

Once vaultwarden is built with rocket/mtls enabled, mTLS seems to be working. I used the following steps/commands to verify this works:

```bash
$ openssl req -x509 -nodes -days 365 -newkey rsa:2048 \
  -subj "/CN=Test-Local-CA" \
  -keyout ca.key -out ca.crt
$ openssl req -nodes -newkey rsa:2048 \
  -subj "/CN=localhost" \
  -keyout server.key -out server.csr
$ openssl x509 -req -days 365 -in server.csr \
  -CA ca.crt -CAkey ca.key -CAcreateserial \
  -extfile <(printf "subjectAltName=DNS:localhost,IP:127.0.0.1\nextendedKeyUsage=serverAuth") \
  -out server.crt
$ openssl req -nodes -newkey rsa:2048 \
  -subj "/CN=test-client" \
  -keyout client.key -out client.csr
$ openssl x509 -req -days 365 -in client.csr \
  -CA ca.crt -CAkey ca.key \
  -extfile <(printf "extendedKeyUsage=clientAuth") \
  -out client.crt
$ env WEB_VAULT_ENABLED=false \
  ROCKET_TLS='{certs="server.crt", key="server.key", mutual={ca_certs="ca.crt",mandatory=true}}' \
  ./target/debug/vaultwarden
$ curl --cert client.crt --key client.key --cacert ca.crt https://localhost:8000/admin
The admin panel is disabled, please configure the 'ADMIN_TOKEN' variable to enable it
$ curl --cacert ca.crt https://localhost:8000/admin
curl: (56) OpenSSL SSL_read: OpenSSL/3.6.2: error:0A00045C:SSL routines::tlsv13 alert certificate required, errno 0
$ curl -k http://localhost:8000/admin
curl: (1) Received HTTP/0.9 when not allowed
```

The downside here is that `rocket` seems to be pulling a relatively old version of `x509-parser` crate, and so there are significant new duplicate crates in the lockfile, including return of the syn 1.0 series. However I'm not sure how fixable this is, given the maintenance status of the rocket framework, so I figured I'll post the PR first and see if this is at all palatable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -99,12 +99,28 @@ dependencies = [
 
 [[package]]
 name = "asn1-rs"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "30ff05a702273012438132f449575dbc804e27b2f3cbe3069aa237d26c98fa33"
+dependencies = [
+ "asn1-rs-derive 0.1.0",
+ "asn1-rs-impl 0.1.0",
+ "displaydoc",
+ "nom 7.1.3",
+ "num-traits",
+ "rusticata-macros",
+ "thiserror 1.0.69",
+ "time",
+]
+
+[[package]]
+name = "asn1-rs"
 version = "0.6.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5493c3bedbacf7fd7382c6346bbd66687d12bbaad3a89a2d2c303ee6cf20b048"
 dependencies = [
- "asn1-rs-derive",
- "asn1-rs-impl",
+ "asn1-rs-derive 0.5.1",
+ "asn1-rs-impl 0.2.0",
  "displaydoc",
  "nom 7.1.3",
  "num-traits",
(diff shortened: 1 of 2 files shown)


BlackDex:
It's better to use a reverse proxy instead. I'm not a fan of enabling this, also because of these old versions.

Therefore I'm leaning to closing this and suggest a reverse proxy instead.

BlackDex:
I have thought about this a bit longer, and for mTLS I think we have the same advice as for running Vaultwarden anyway. Use a reverse proxy, since those are better supporting TLS, newer features and security wise better then the TLS option of Rocket.

Therefor I'm going to close this PR.

Besides that. Only enabling the `mtls` feature will not help, there needs to be code in place which provides the correct location of the certificates etc... So this will also never ever work as-is.

nagisa:
> Only enabling the mtls feature will not help, there needs to be code in place which provides the correct location of the certificates etc... So this will also never ever work as-is.

The certificate necessary for this can already be provided via the existing `ROCKET_TLS` environment variable, as I demonstrated in the PR description.

But I do understand you not wanting to support or maintain this. Fair enough and its entirely up to you.

PR #7363, opened 2026-06-22
Add Sends email verification support

Timshel:
Built on top of https://github.com/dani-garcia/vaultwarden/pull/7346

Remarks:

- Did not reuse the existing `protected_action` template but it could make sense

Include some refactoring on the `send` table (split in a separate commit)

- Remove the `organization_uuid` column since it's not used and I could not find mention of Sends for org.
- Set `user_uuid` as `NOT NULL`
- Set `hide_email` as  `NOT NULL` , will default to `false`.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -187,6 +187,10 @@
 ## Cron schedule of the job that cleans sso auth from incomplete flow
 ## Defaults to daily (20 minutes after midnight). Set blank to disable this job.
 # PURGE_INCOMPLETE_SSO_AUTH="0 20 0 * * *"
+#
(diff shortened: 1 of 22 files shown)


Timshel:
Tested with the client from https://github.com/dani-garcia/bw_web_builds/actions/runs/27478766767

GlisseManTV:
Very interesting feature. Which is available through the desktop app but didn't work.
First, I thought it was a bug but finally I found this PR.

When can we expect to have it merged ?

marvineimer-kw:
Is there a chance of this getting merged? I've been waiting for over year that Bitwarden adds this feature...

PR #7367, opened 2026-06-23
Add Podman Quadlet instructions to Readme

4liceD:
Added example for using VW as Podman Quadlet to Readme. 
Added 'docker compose up -d' as context to docker compose example.


BlackDex:
This belongs to the wiki, not the default readme which just provides a basic example.

PR #7370, opened 2026-06-23
Add passwordless webauthn login

RaphaelRoumezin:
This PR is my attempt at providing passwordless login, taking heavy inspiration from the work of @samos667 (#6820), who I believe has its work inspired from @zUnixorn (#5929).

I'm looking forward to your feedback on this work !

## Working features
*(same as #5929 and #6820)*
- Login with passkey without using it for encryption
- Login with passkey while using it for encryption
- Adding a new passkey
- Listing all registered passkeys
- Deleting a passkey

*New*
- The `passkey_login_allowed` backend option was added, which can disable passkey-related endpoints and remove passkey-related settings and passkey login button from the web vault
- The vault can be unlocked with all the encryption passkeys of the account, and not only the one used for logging in

## Main differences from #6820
- Refactored `web_authn` to `webauthn`, and `WebAuthn` to `Webauthn` for all related symbols
- Switched to a stateless architecture for auth flow. The authentication state is sent to the client through JWT, despite the warning given by webauthn_rs. This is because the state does not hold more information than what is already given to the client, and i don't think the replay attack weakness is a problem in the bitwarden context, as this allows to bypass authentication but keeps vault content encrypted (and it isn't more replayable than the password auth flow). As for the passkey registration flow, the state is stored in the `TwoFactor` table (with type `TwoFactorType::WebauthnPasskeyRegisterChallenge`), as a replay could allow a deleted passkey to be recreated.
- The webauthn authentication flow does not check for a verified email like the password flow, as creating a passkey already requires the user to be able to use the web vault, and thus having a verified email in the related contexts.
- Relocated webauthn-related endpoints from `src/api/core/mod.rs` to `src/api/core/webauthn.rs`
- The `get_prf_status()` function was added, following the C# bitwarden implementation.

## Open questions
- The authentication state is stored in JWT, and the registration state is stored in the table `TwoFactor`. The safest would be to store everything in the DB, but I think stateless auth would be preferrable if its secure. Using a dedicated table for webauthn_rs states is an option too.


kengzzzz:
I built and ran this branch (at `224ad8a`) against a Helium browser + YubiKey. Login works, but I hit one issue getting passkey **login** working from the browser extension — sharing the root cause + the patch I'm carrying in case it helps.

### Browser-extension login fails with `InvalidRPOrigin`

Passkey login works from the web vault, but from the Bitwarden *browser extension* the assertion is signed correctly yet rejected server-side. The verifier in `src/api/core/two_factor/webauthn.rs` only allows the vault domain as an origin, while the extension asserts from a `chrome-extension://<id>` origin — so `clientDataJSON.origin` never matches. Official Bitwarden works because it allowlists those origins.

`webauthn-rs` `origins_match()` short-circuits on exact `Url` equality before the opaque-origin branch, so appending the origin verbatim is enough (no special scheme handling needed):

```rust
let mut webauthn = WebauthnBuilder::new(&rp_id, &rp_origin)
    .expect("Creating WebauthnBuilder failed")
    .rp_name(&domain)
    .timeout(tokio::time::Duration::from_mins(1));

for origin in [
    "chrome-extension://nngceckbapebfimnlniiiahkandclblb", // Chrome Web Store
    "chrome-extension://jbkfoedolllekgbhcbcoahefnbanhhlh", // Edge Add-ons
] {
    let origin = url::Url::parse(origin).expect("origin should parse");
    webauthn = webauthn.append_allowed_origin(&origin);
}
webauthn.build().expect("Building Webauthn failed")
```

Hardcoding matches what Bitwarden's server does.

---

Environment: Helium browser + YubiKey, self-hosted.

RaphaelRoumezin:
Thanks for your message @kengzzzz.
I implemented your fix, and also had to split the webauthn context between 2FA and passwordless, since 2FA does not require this fix as Webauthn 2FA is performed in a browser's page and not directly in the extension.

kengzzzz:
I found one more Bitwarden client compatibility issue while testing PRF/encrypted passkey login.

The current login response adds:

- `UserDecryptionOptions.WebAuthnPrfOption.EncryptedPrivateKey`
- `UserDecryptionOptions.WebAuthnPrfOption.EncryptedUserKey`

but the Bitwarden clients also expect the PRF option to include:

- `CredentialId`
- `Transports`

Without those fields, login with an encrypted passkey can succeed, but locked-vault passkey unlock fails because the client cannot map the PRF unlock option back to the WebAuthn credential.

I also observed that official Bitwarden returns PRF unlock metadata from `/sync` as:

- `userDecryption.webAuthnPrfOptions[]`

For enabled PRF credentials, the server can derive `CredentialId` from the stored serialized `Passkey` credential id, and `Transports: []` appears to match official server behavior.

So the server-side shape should include these fields in both places:

```json
{
  "CredentialId": "...",
  "Transports": [],
  "EncryptedPrivateKey": "...",
  "EncryptedUserKey": "..."
}
```

This only needs to include credentials where PRF is enabled and both encrypted keys are present.

RaphaelRoumezin:
Thank you for your report @kengzzzz.
I implemented your fix at https://github.com/RaphaelRoumezin/vaultwarden/commit/8b05d3b3444adbecd379d65696d7c9a4ac1ee17d, which fixes the problem for browser extension clients, but not the web vault for some reason, and I can't figure out why.
Complying with the Bitwarden `/sync` endpoint implementation didn't seem to change anything, and it seems like the browser receives the credential ID correctly, as it can be seen in the browser's session storage. Perhaps the web vault 2026.4.1 has a different behavior than Bitwarden production version ?

Any idea you have on this, feel free to let me know, or even make your own commit on top of https://github.com/RaphaelRoumezin/vaultwarden/commit/8b05d3b3444adbecd379d65696d7c9a4ac1ee17d.

kengzzzz:
I put together a small patch against `3b669264bb247351129916bcbbf2bf125b6c939d` with these changes:

- reuse one helper to build `UserDecryptionOptions.WebAuthnPrfOption` for WebAuthn login
- include `CredentialId`, `Transports`, `EncryptedPrivateKey`, and `EncryptedUserKey` in that option
- add `userDecryption.webAuthnPrfOptions[]` to `/sync` when the user has PRF-enabled credentials
- expose `pm-2035-passkey-unlock` from `/config` when passkey login is allowed
- only emit PRF unlock options for credentials whose PRF status is enabled and whose encrypted unlock keys are present

I tested this on my vaultwarden server and passkey unlock works with the web vault.

The patch applies cleanly on top of the PR head commit:

```bash
git checkout 3b669264bb247351129916bcbbf2bf125b6c939d
git apply vaultwarden-prf-unlock-fix.patch
```

[vaultwarden-prf-unlock-fix.patch](https://github.com/user-attachments/files/29472658/vaultwarden-prf-unlock-fix.patch)

RaphaelRoumezin:
I made a new commit (https://github.com/RaphaelRoumezin/vaultwarden/commit/8d005c02a347c26b6fb3885d9c1227a9912ed7e2) mainly based on your patch, but also fixes the divergent JSON fields case, which is PascalCase for `/sync` and camelCase for `/token` in Bitwarden's implementation.
It appears the core of the issue was that the missing feature flag, but the client still use the contents of `/sync` to fill the userDecryptionOptions Angular state with the other passkeys of the account, as opposed to doing nothing of it (which I thought it did), so I left that in.

I'm not super satisfied with the way I provide 2 different naming cases in `webauthn_prf_option`, but it does the job i guess.
If nothing more comes up I'll cherry-pick the commit into this PR in the evening.
Thanks again for your work @kengzzzz !

kengzzzz:
I tested https://github.com/RaphaelRoumezin/vaultwarden/commit/8d005c02a347c26b6fb3885d9c1227a9912ed7e2 on my vaultwarden server and can confirm it fixes the issue for my case.

RaphaelRoumezin:
I've merged the main branch into this PR to keep it up to date.
Honestly I feel like this branch is a pretty clean implementation of the "login with passkey" feature, and even - if i dare say - review worthy.
But the rest is up to the maintainers, and I'd love to get feedback on what we've made so far. I feel like we are (or at least pretty close from being) mergeable.
What do you think @BlackDex @dani-garcia ?

BJReplay:
Watching with interest @RaphaelRoumezin.

Just one comment - and I'm not sure how you can (or if you can) fix it at this point, but your first commit https://github.com/dani-garcia/vaultwarden/pull/7370/changes/7711b02153d01386a9543b19d5141e2f8e99b04a is showing as unverified - possibly because it is tagged as co-authored.  Not sure if that will be an issue or not if this moves forward (and I hope it will).

image

RaphaelRoumezin:
> Watching with interest @RaphaelRoumezin.
> 
> Just one comment - and I'm not sure how you can (or if you can) fix it at this point, but your first commit

@BJReplay it's not really an issue. It's just because I used the code from another PR, so I added the person as co-author, but he has vigilant mode enabled on GitHub. That's not really fixable either, since commits can only hold one PGP signature. It's just GitHub being GitHub

jcadduono:
I've rebased this over the 1.37.0 tag and built a container with `docker/bake.sh` and ran it. It shows Allow passkey login as true under General settings in my `/admin` page, but I can't for the life of me figure out how to set up a login passkey. I expected the fields to be available on `/settings/security/password` (the master password page), but it's not there. Where should I look?

Update: Later on another PC, it showed up where expected, so must have been the browser cache. This patch works beautifully with my YubiKey 5C NFC! Thank you.

Amsal1:
Eagerly waiting for this to get merged 🫡🚀

RaphaelRoumezin:
PR author here. Quick design question:
```rust
pub static WEBAUTHN_PASSWORDLESS: LazyLock<Webauthn> = LazyLock::new(|| {
    let domain = CONFIG.domain();
    let domain_origin = CONFIG.domain_origin();
    let rp_id = url::Url::parse(&domain).map(|u| u.domain().map(str::to_owned)).ok().flatten().unwrap_or_default();
    let rp_origin = url::Url::parse(&domain_origin).unwrap();

    let mut webauthn = WebauthnBuilder::new(&rp_id, &rp_origin)
        .expect("Creating WebauthnBuilder failed")
        .rp_name(&domain)
        .timeout(tokio::time::Duration::from_mins(1));

    for origin in [
        "chrome-extension://nngceckbapebfimnlniiiahkandclblb", // Chrome Web Store
        "chrome-extension://jbkfoedolllekgbhcbcoahefnbanhhlh", // Edge Add-ons
    ] {
        let origin = url::Url::parse(origin).unwrap();
        webauthn = webauthn.append_allowed_origin(&origin);
    }

    webauthn.build().expect("Building Webauthn failed")
});
```

The `WEBAUTHN_PASSWORDLESS` config structure is wrapped in a LazyLock, which means it is initialized once and never again for the vaultwarden instance at runtime. The thing is, it utilizes runtime dynamic variables, mainly the `DOMAIN` value from config which can be edited from the admin panel, and if those are edited, the webauthn config structure keeps using the old ones until the vaultwarden instance is restarted.
Having those values set correctly is important, because the `webauthn_rs` requires (and rightfully so) that the response from the browser indicates the correct origin to match the correct credential, and to enforce that check, the `DOMAIN` config value is set as the expected origin.

Now, I'm not sure how this should be fixed. The `WEBAUTHN_PASSWORDLESS` should be reset every time the config is updated, but that's not really a feature of `LazyLock`. Let me know your ideas.

It's also good to note that this issue also exists in `main`, within the `WEBAUTHN` structure used for 2FA, known as `WEBAUTHN_2FA` within this PR.

src/api/core/webauthn.rs
```diff
@@ -0,0 +1,286 @@
...
+        user.display_name(),
+        Some(existing_cred_ids),
+    )?;
+
+    // For passkey login, we need discoverable credentials (resident keys).
+    // start_passkey_registration() defaults to require_resident_key=false, but passkey login
+    // requires the credential to be discoverable (resident) so the authenticator can find it
+    // without the server providing allowCredentials.
+    if let Some(asc) = challenge.public_key.authenticator_selection.as_mut() {
+        asc.user_verification = UserVerificationPolicy::Discouraged_DO_NOT_USE;
```
dfunkt: Just tested this PR and I encountered the following: 
I needed to use ```UserVerificationPolicy::Preferred``` here for Google Password Manager to work.
Otherwise this happened:
```
[error][ERROR] Webauthn.
[CAUSE] UserNotVerified
[response][INFO] (post_webauthn) POST /api/webauthn => 400 Bad Request
```
If I understand correctly, this is how it's supposed to be in Bitwarden:
https://github.com/bitwarden/clients/blob/51c7e4c650cec8d7f0cf53e297c484625ef5f210/libs/key-management-ui/src/lock/services/default-webauthn-prf-unlock.service.ts#L149

BlackDex: Since we see that `preferred` also doesn't seem to cause issues with WebAuthn for 2fa, it's probably best if we fully remove `Discouraged`.

RaphaelRoumezin: Fix pushed. I didnt touch the 2FA UVP configuration though, as I believe it is being worked on in #7500.

michaelbeaumont:
> It's also good to note that this issue also exists in main, within the WEBAUTHN structure used for 2FA, known as WEBAUTHN_2FA within this PR.

Well I'd say it's not really a problem for this PR then!

PR #7382, opened 2026-06-29
add Zenith Hosting badge

l1mey112:
Hey, i'm part of a startup called Zenith Hosting; where we frequently use Vaultwarden to help safeguard our creds. We wanted to make it easier for other people who don't want to use heavyweight/non-free password managers, so we added Vaultwarden as a one-click deploy on our platform. 

this PR simply adds a small deploy badge to the README; which links to https://zenith.hosting/host/vaultwarden where people can just spin up a fresh vw instance with persistent storage and auto backups without the hassle of setting up Docker or a reverse proxy/ufw themselves.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,8 @@ services:
 
 <br>
 
+[![Deploy with Zenith](https://cdn.zenith.hosting/buttons/deploy-with-zenith.svg)](https://zenith.hosting/host/vaultwarden?ref=gh)
+
 ## Get in touch
 
 Have a question, suggestion or need help? Join our community on [Matrix](https://matrix.to/#/#vaultwarden:matrix.org), [GitHub Discussions](https://github.com/dani-garcia/vaultwarden/discussions) or [Discourse Forums](https://vaultwarden.discourse.group/).


xuyeqi:
@codex review

stefan0xC:
I'm against this because we currently don't advocate any hosting platform like this and while there are some deployments providers advocated on the [Deployment examples page ](https://github.com/dani-garcia/vaultwarden/wiki/Deployment-examples#paas-hosting)I don't think any of them should be added to the README. (In fact I think that the wiki might be missing a disclaimer at the top of the page that warns users that all links are provided by users and are not checked or endorsed by the project.)

l1mey112:
> I'm against this because we currently don't advocate any hosting platform like this and while there are some deployments providers advocated on the [Deployment examples page ](https://github.com/dani-garcia/vaultwarden/wiki/Deployment-examples#paas-hosting)I don't think any of them should be added to the README. (In fact I think that the wiki might be missing a disclaimer at the top of the page that warns users that all links are provided by users and are not checked or endorsed by the project.)

Sounds good, sorry for this.

PR #7397, opened 2026-07-01
Add Custom role with granular permissions (users, groups, policies, collections, event logs, reports, import/export)

tom27052006:
## Summary

Adds a real, persisted **Custom** organization role with nine independently grantable
permissions, enforced server-side.

Until now Vaultwarden had no Custom role: members who were meant to be Custom were stored
as the legacy `Manager` type, and the permission checkboxes were hidden because the server
could not enforce them. This PR replaces that compatibility layer with a persisted role and
real authorization, and folds the legacy `Manager` role into `Custom` (wire value `3` is
still accepted and normalized).

Role rank is `User < Custom < Admin < Owner`.

## Bitwarden compatibility

The authorization semantics are aligned directly with the current Bitwarden server
implementation rather than with a reduced Vaultwarden-specific interpretation. Earlier
revisions of this PR narrowed several permissions beyond what Bitwarden does; those
restrictions were removed in favour of the actual upstream behaviour.

Concretely:

- The permission guards mirror upstream's `BasePermissionRequirement`: a confirmed Owner or
  Admin, or a Custom member holding the matching flag.
- `manageUsers` covers the member lifecycle — invite, confirm, revoke, restore, remove, role
  change — **including** a member's collection assignments and group memberships, matching
  upstream's `OrganizationUsersController`.
- `manageGroups` covers group creation, update and deletion including the group's collection
  access and its member list, matching upstream's `GroupsController`. The group *details*
  list is guarded by `manageUsers` **or** `manageGroups`, as upstream does.
- `accessImportExport` grants the organization-wide cipher scope Bitwarden gives it: the
  organization export, and organization import into existing or new collections.
- `accessReports` grants the organization-wide cipher list clients compute reports from.
  Upstream's `CanAccessAllCiphersAsync` admits `accessImportExport`, `editAnyCollection` or
  `accessReports`; this PR applies the same three-way condition. Vaultwarden has no
  server-side report endpoint, so the permission is enforced where that cipher list is served.
- `editAnyCollection` grants the administrative cipher/collection scope but deliberately does
  not widen the regular vault scope, matching upstream's split between the admin context and
  the user context.

## Permissions

`manageUsers`, `manageGroups`, `managePolicies`, `createNewCollections`,
`editAnyCollection`, `deleteAnyCollection`, `accessEventLogs`, `accessImportExport`,
`accessReports`.

Stored as nine boolean columns on `users_organizations`, all defaulting to `FALSE`.
`manageAllCollections` stays what it is upstream: a client-side aggregate of the three
collection permissions, not a persisted flag.

## Security / authorization

- `Custom` is a real persisted role, not a re-labelled `Manager`.
- Permission flags are only read for Custom memberships, so a stale flag left behind by a role
  change can never grant anything.
- Every permission check also requires `Confirmed` status.
- An unknown stored role value grants no authority and fails closed everywhere; only an Owner
  may revoke or delete such a membership.
- Organization binding is re-checked server-side on every guarded route.
- A Custom member cannot delegate a permission they do not hold themselves, and may only
  manage `User` and `Custom` memberships. Admins cannot act on Owners.
- Holding `managePolicies` does not exempt a member from policy enforcement; only Owners and
  Admins are excluded, plus the acting member, so enabling a policy cannot revoke the caller
  mid-request.
- A per-collection `manage` grant must be explicit. Neither membership nor group `access_all`
  manufactures one.

## Migration

One migration per backend replaces the membership-level `access_all` column with the role and
its nine permissions:

- `Manager` becomes `Custom`.
- A Manager that carried membership `access_all` receives `createNewCollections`,
  `editAnyCollection` and `deleteAnyCollection` — that flag was the "Manage all collections"
  checkbox and covered exactly those three.
- All management and access permissions start at `FALSE`. Nothing they unlock was ever a
  Manager capability, so granting one here would create a new privilege.
- Owners and Admins are untouched; they hold every permission by role.

The group-level `groups.access_all` flag is a separate, still-supported feature and is
**not** read by the migration. It is deliberately not materialized into permanent Custom
membership permissions such as `editAnyCollection` or `deleteAnyCollection`, because a
dynamic group grant — one that ends when the group relationship or the flag ends — should not
be converted into a permanent individual right.

The trade-off is explicit: a legacy Manager whose management reach came *only* from an
`access_all` group loses that reach on upgrade and has to be granted the Custom permissions
explicitly. That is accepted, because the guiding principle for this migration is that an
upgrade must not create new permanent privileges. This is a deliberate Vaultwarden-specific
migration decision, not a claim of 1:1 parity with Bitwarden.

One state cannot be converted and is refused before the first mutation: a plain `User`
carrying membership `access_all`, a row only older Vaultwarden versions could write. It means
read/write reach over every collection with no management authority, which the new model
cannot express. A startup preflight evaluates the same condition and prints a recovery
procedure; `LEGACY_USER_ACCESS_ALL_MIGRATION` (`refuse` | `drop` | `materialize`) lets an
owner decide once for the instance. Default is `refuse`.

## Testing

Unit tests cover the role authority matrix, permission parsing and delegation (no privilege
escalation), the collection-operation and cipher-access scope matrices, the administrative vs.
regular vault scope split, the migration's Manager conversion and its preflight decision
table. Manual verification was done per backend (SQLite, MySQL, PostgreSQL), including
multi-organization cases and upgrade paths from the legacy `access_all` states.


RaphaelRoumezin:
A few questions:
- Do you plan to also implement `accessEventLogs`, `accessImportExport` and `accessReports` permissions ?
- Why not get rid of the `access_all` flag, which was until then a patch for not having permissions implemented ?
- Since Manager is a legacy role, why not just refactor Manager to Custom instead of adding Custom along ? Since Custom was mapped to Manager internally anyway

tom27052006:
> A few questions:
> 
> - Do you plan to also implement `accessEventLogs`, `accessImportExport` and `accessReports` permissions ?
> 
> - Why not get rid of the `access_all` flag, which was until then a patch for not having permissions implemented ?
> 
> - Since Manager is a legacy role, why not just refactor Manager to Custom instead of adding Custom along ? Since Custom was mapped to Manager internally anyway

1: I still want to add them, I just wanted to Focus on the most important permissions first
2: Yes I agree with you, removing access_all would just be a bit more work, but if the maintainers want it to be removed completely, maybe @BlackDex can confirm, I would try to remove it completly
3: I kept Manager for old data, old clients and downgrades. Existing Manager users are already moved to Custom, but like the point above if the maintainers want it done differently, I would change this too

I think I would try to start working on These changes, especially point 2 and maybe point 3. But I would still like to hear what the maintainers think about it

tom27052006:
Pushed the remaining Custom-role work, which completes this PR.

**Three new Custom permissions**

- accessEventLogs — gates the organization and per-member event log endpoints. Per-cipher event reads are scoped to the caller's own organization.
- accessImportExport — gates the organization export and import. The export is scoped to the collections the caller can actually reach unless they already have full access, so the permission decides whether a member may export, never what they may read.
- accessReports — opens GET /ciphers/organization-details. Since Vaultwarden has no server-side reports and the clients compute them locally from that list, this is by design a full organization read permission, matching Bitwarden's CanAccessAllCiphersAsync. This is documented at the call site so the trade-off is explicit.

**Membership access_all removed**
The flag is now fully expressed by the role model: Owners/Admins reach every collection implicitly, a Custom member does so via editAnyCollection. groups.access_all is a separate feature and is untouched.

**Legacy Manager role removed**
Manager is no longer a variant — it is never persisted or emitted, and an incoming wire value of 3 is folded onto Custom for backward compatibility. Stored Manager rows are migrated to Custom. MembershipType now orders by authorization rank (User < Custom < Admin < Owner) and fails closed on unknown values.

**Migrations (SQLite, PostgreSQL, MySQL/MariaDB)**
A Manager that held access_all keeps the equivalent create/edit/delete-any-collection grants. A Manager whose full access came from an access_all group keeps deriving it from that group rather than getting a static copy, so removing the group still revokes the access. A startup preflight refuses to migrate ambiguous legacy state (a plain User carrying access_all) instead of guessing. The down-migration chain deliberately blocks an automatic downgrade, because nine independent permissions cannot be represented by the old schema without silent data loss.

tom27052006:
I think the PR is in good shape now, but there are a few open questions where
opinions may differ:

1. **Deleting collections.** We tell clients that deleting collections is
   restricted, but in practice anyone with "manage" rights on a single
   collection can delete it. That's how main already behaves, so I left it
   alone. Bitwarden does not allow this. Now that "Delete any collection" is its
   own checkbox, the gap is easier to notice. Leave it as is, or change it?

2. **What "Manage users" and "Manage groups" may do.** In Bitwarden these two
   permissions also let someone change which collections a user or group can
   see, which indirectly hands out access to passwords. I did not allow that
   here: they can manage people, not data. Should I follow Bitwarden instead?

3. **"Access import/export".** In Bitwarden this permission also lets you export
   the entire organization vault. Here it only lets you export what you can
   already see. So keep it limited or match Bitwarden?

   `accessReports` right now goes the other way. It follows Bitwarden and does
   open the full organization cipher list, so in practice it is a read-everything
   permission. 

4. **Old Managers who got their rights from a group.** After the migration they
   still reach the data through that group, but they can no longer manage
   collections directly. The upside is that removing them from the group now
   really takes the rights away. Is that acceptable?

5. **Going back to an older version.** The upgrade removes a database column, so
   an older Vaultwarden build cannot run against the new database. Is "restore
   your backup if you need to roll back" good enough, or should I keep the old
   column around for one release first?
   
Happy to hear any opinion on these, from maintainers or anyone else following
the PR :)

jmueller42:
First of all, thank you for your work.

Here my opinion on your questions:

1: I would suggest to leave it as it is.
2: I would stick to your solution. I think this makes it more clear which access which roles provides.
3: I would suggest to only be able to export the entries, which the user has access to.
4: I think this is good, such that the current behavior is still the case after the update.
5: Maybe an easy rollback would be good.

tom27052006:
> First of all, thank you for your work.
> 
> Here my opinion on your questions:
> 
> 1: I would suggest to leave it as it is. 2: I would stick to your solution. I think this makes it more clear which access which roles provides. 3: I would suggest to only be able to export the entries, which the user has access to. 4: I think this is good, such that the current behavior is still the case after the update. 5: Maybe an easy rollback would be good.

Thanks @jmueller42 for your opinion :) Where things stand:

1. Left as is — a `manage` grant on a collection still allows deleting it.

2. Left as is — they manage people, not data. Collection assignments still need
   Admin/Owner or `editAnyCollection`.

3. Export stays limited to what the user can already see. I made `accessReports`
   work the same way, since it still opened the whole organization.

4. They now keep collection edit/delete as well. It is derived live from the
   `access_all` group instead of being written into the permission columns, so
   removing them from the group still takes the rights away.

5. `tools/custom_role_rollback/` now has a script per database that restores the
   old schema, so an older build starts against the database again. The README
   lists what the old schema cannot represent. The old column is not kept for a
   release, since that would undo the `access_all` removal

jmueller42:
> > First of all, thank you for your work.
> > Here my opinion on your questions:
> > 1: I would suggest to leave it as it is. 2: I would stick to your solution. I think this makes it more clear which access which roles provides. 3: I would suggest to only be able to export the entries, which the user has access to. 4: I think this is good, such that the current behavior is still the case after the update. 5: Maybe an easy rollback would be good.
> 
> Thanks @jmueller42 for your opinion :) Where things stand:
> 
>     1. Left as is — a `manage` grant on a collection still allows deleting it.
> 
>     2. Left as is — they manage people, not data. Collection assignments still need
>        Admin/Owner or `editAnyCollection`.
> 
>     3. Export stays limited to what the user can already see. I made `accessReports`
>        work the same way, since it still opened the whole organization.
> 
>     4. They now keep collection edit/delete as well. It is derived live from the
>        `access_all` group instead of being written into the permission columns, so
>        removing them from the group still takes the rights away.
> 
>     5. `tools/custom_role_rollback/` now has a script per database that restores the
>        old schema, so an older build starts against the database again. The README
>        lists what the old schema cannot represent. The old column is not kept for a
>        release, since that would undo the `access_all` removal

Thank you for all the effort you put in this.

So this seems complete for me. Or is something missing?

tom27052006:
> > > First of all, thank you for your work.
> > > Here my opinion on your questions:
> > > 1: I would suggest to leave it as it is. 2: I would stick to your solution. I think this makes it more clear which access which roles provides. 3: I would suggest to only be able to export the entries, which the user has access to. 4: I think this is good, such that the current behavior is still the case after the update. 5: Maybe an easy rollback would be good.
> > 
> > 
> > Thanks @jmueller42 for your opinion :) Where things stand:
> > ```
> > 1. Left as is — a `manage` grant on a collection still allows deleting it.
> > 
> > 2. Left as is — they manage people, not data. Collection assignments still need
> >    Admin/Owner or `editAnyCollection`.
> > 
> > 3. Export stays limited to what the user can already see. I made `accessReports`
> >    work the same way, since it still opened the whole organization.
> > 
> > 4. They now keep collection edit/delete as well. It is derived live from the
> >    `access_all` group instead of being written into the permission columns, so
> >    removing them from the group still takes the rights away.
> > 
> > 5. `tools/custom_role_rollback/` now has a script per database that restores the
> >    old schema, so an older build starts against the database again. The README
> >    lists what the old schema cannot represent. The old column is not kept for a
> >    release, since that would undo the `access_all` removal
> > ```
> 
> Thank you for all the effort you put in this.
> 
> So this seems complete for me. Or is something missing?

Yes, ive implented a few more fixes, but now it should be complete.

A quick note on the diff size: GitHub currently shows 10,718 changed lines (+10,085/−633). Of those, about 2,500 are comment-only lines and about 800 are blank lines. Around 6,400 lines cover migrations, recovery/rollback logic, and related tests—much of which handles the possibility that someone has already deployed an earlier revision of this still-unmerged PR. If this compatibility is not needed, I can simply remove it and simplify the migrations; I just do not know whether anyone is already running an earlier revision of this PR.

bigbgh:
@dani-garcia please look at this. The community is very interested to see this merged.

NeffIsBack:
> A quick note on the diff size: GitHub currently shows 10,718 changed lines (+10,085/−633). Of those, about 2,500 are comment-only lines and about 800 are blank lines. Around 6,400 lines cover migrations, recovery/rollback logic, and related tests—much of which handles the possibility that someone has already deployed an earlier revision of this still-unmerged PR. If this compatibility is not needed, I can simply remove it and simplify the migrations; I just do not know whether anyone is already running an earlier revision of this PR.

Imho there should only be support for backwards compatibility for officially released builds. If lines of code (and therefore review overhead) can be removed by eliminating support for every possible state of this PR that should be done. Imho this is the responsibility of people themselves if they decide to clone and migrate their production database to some state of some PR, and not the responsibility of the maintainer(s).

tom27052006:
> > A quick note on the diff size: GitHub currently shows 10,718 changed lines (+10,085/−633). Of those, about 2,500 are comment-only lines and about 800 are blank lines. Around 6,400 lines cover migrations, recovery/rollback logic, and related tests—much of which handles the possibility that someone has already deployed an earlier revision of this still-unmerged PR. If this compatibility is not needed, I can simply remove it and simplify the migrations; I just do not know whether anyone is already running an earlier revision of this PR.
> 
> Imho there should only be support for backwards compatibility for officially released builds. If lines of code (and therefore review overhead) can be removed by eliminating support for every possible state of this PR that should be done. Imho this is the responsibility of people themselves if they decide to clone and migrate their production database to some state of some PR, and not the responsibility of the maintainer(s).

Yeah, that´s what I thought. I’ve now removed all the unnecessary migrations that were only there for older states of the PR and also shortened some of the very long comments. This should reduce the amount of code a bit and make the PR easier to review

tom27052006:
Small update: I've reworked this quite a bit. The Custom-role authorization now follows the
current Bitwarden server implementation much more closely — earlier revisions locked down
member collection assignments, group collection access and the org-wide cipher scope behind
import/export and reports tighter than upstream actually does. Those extra restrictions are
gone; the guards now line up with upstream's `BasePermissionRequirement` and the matching controllers.

One spot where I deliberately went my own way: legacy Managers whose reach came from a group
with `groups.access_all` don't get that turned into permanent Custom permissions like
`editAnyCollection`. That flag is a dynamic grant, and baking it into a permanent individual
right on upgrade felt wrong — better that it has to be granted explicitly than that a migration
quietly hands out new permanent privileges. Everything the old membership-level `access_all`
covered migrates as before.

Also a lot less to read through now: ~5.7k added lines instead of ~8.2k, and roughly 1.1k of
that is pure unit tests.

PR #7412, opened 2026-07-11
chore: enable the `s3` feature within the Dockerfiles

ProbstDJakob:
Since the PR #6127 has been merged, enabling the `s3` feature for released container images.

### DIFF
--- a/docker/Dockerfile.alpine
+++ b/docker/Dockerfile.alpine
@@ -79,15 +79,16 @@ COPY ./macros ./macros
 
 ARG CARGO_PROFILE=release
 
-# Configure the DB ARG as late as possible to not invalidate the cached layers above
+# Configure the DB and FEATURES ARGs as late as possible to not invalidate the cached layers above
+ARG DB=sqlite,mysql,postgresql
 # Enable MiMalloc to improve performance on Alpine builds
-ARG DB=sqlite,mysql,postgresql,enable_mimalloc
+ARG FEATURES=s3,enable_mimalloc
 
 # Builds your dependencies and removes the
 # dummy project, except the target folder
 # This folder contains the compiled dependencies
 RUN source /env-cargo && \
-    cargo build --features ${DB} --profile "${CARGO_PROFILE}" --target="${CARGO_TARGET}" && \
+    cargo build --features "${DB}" --features "${FEATURES}" --profile "${CARGO_PROFILE}" --target="${CARGO_TARGET}" && \
     find . -not -path "./target*" -delete
 
 # Copies the complete project
@@ -102,7 +103,7 @@ RUN source /env-cargo && \
     # Also do this for build.rs to ensure the version is rechecked
     touch build.rs src/main.rs && \
     # Create a symlink to the binary target folder to easy copy the binary in the final stage
-    cargo build --features ${DB} --profile "${CARGO_PROFILE}" --target="${CARGO_TARGET}" && \
+    cargo build --features "${DB}" --features "${FEATURES}" --profile "${CARGO_PROFILE}" --target="${CARGO_TARGET}" && \
     if [[ "${CARGO_PROFILE}" == "dev" ]] ; then \
         ln -vfsr "/app/target/${CARGO_TARGET}/debug" /app/target/final ; \
     else \
--- a/docker/Dockerfile.alpine
+++ b/docker/Dockerfile.alpine
@@ -44,8 +44,6 @@ ARG TARGETARCH
 ARG TARGETVARIANT
 ARG TARGETPLATFORM
 
-SHELL ["/bin/bash", "-o", "pipefail", "-c"]
-
 # Build time options to avoid dpkg warnings and help with reproducible builds.
 ENV DEBIAN_FRONTEND=noninteractive \
     LANG=C.UTF-8 \
@@ -70,7 +68,7 @@ RUN echo "export CARGO_TARGET=${RUST_MUSL_CROSS_TARGET}" >> /env-cargo && \
     # Output the current contents of the file
     cat /env-cargo
 
-RUN source /env-cargo && \
+RUN . /env-cargo && \
     rustup target add "${CARGO_TARGET}"
 
 # Copies over *only* your manifests and build files
@@ -87,7 +85,7 @@ ARG FEATURES=s3,enable_mimalloc
 # Builds your dependencies and removes the
 # dummy project, except the target folder
 # This folder contains the compiled dependencies
-RUN source /env-cargo && \
+RUN . /env-cargo && \
     cargo build --features "${DB}" --features "${FEATURES}" --profile "${CARGO_PROFILE}" --target="${CARGO_TARGET}" && \
     find . -not -path "./target*" -delete
 
@@ -98,13 +96,13 @@ COPY . .
 ARG VW_VERSION
 
 # Builds again, this time it will be the actual source files being build
-RUN source /env-cargo && \
+RUN . /env-cargo && \
     # Make sure that we actually build the project by updating the src/main.rs timestamp
     # Also do this for build.rs to ensure the version is rechecked
     touch build.rs src/main.rs && \
     # Create a symlink to the binary target folder to easy copy the binary in the final stage
     cargo build --features "${DB}" --features "${FEATURES}" --profile "${CARGO_PROFILE}" --target="${CARGO_TARGET}" && \
-    if [[ "${CARGO_PROFILE}" == "dev" ]] ; then \
+    if test "${CARGO_PROFILE}" = "dev"; then \
         ln -vfsr "/app/target/${CARGO_TARGET}/debug" /app/target/final ; \
     else \
         ln -vfsr "/app/target/${CARGO_TARGET}/${CARGO_PROFILE}" /app/target/final ; \
--- a/docker/Dockerfile.debian
+++ b/docker/Dockerfile.debian
@@ -42,8 +42,6 @@ ARG TARGETARCH
 ARG TARGETVARIANT
 ARG TARGETPLATFORM
 
-SHELL ["/bin/bash", "-o", "pipefail", "-c"]
-
 # Build time options to avoid dpkg warnings and help with reproducible builds.
 ENV DEBIAN_FRONTEND=noninteractive \
     LANG=C.UTF-8 \
@@ -80,7 +78,7 @@ RUN mkdir -pv "${CARGO_HOME}" && \
 RUN USER=root cargo new --bin /app
 WORKDIR /app
 
-RUN source /env-cargo && \
+RUN . /env-cargo && \
     rustup target add "${CARGO_TARGET}"
 
 # Copies over *only* your manifests and build files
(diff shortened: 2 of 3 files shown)


BlackDex:
This will bloat the binary and container a lot by default, that is the main reason to not have it enabled by default.

ProbstDJakob:
Actually it is only ~5MiB for a ~200MiB image. I personally would say that this is not much of a deal. For reference, the base `docker.io/library/alpine:3.24` image is ~8.3MiB big.

```console
$ podman build --format docker -f Dockerfile.alpine ..
[...]
6bbe0a62fec4702a9928a4dd8663d45ee463c6bfc7e626ec0d6852a9a129cbb2
$ podman image save --format docker-archive -o with-s3.tar 6bbe0a62fec4702a9928a4dd8663d45ee463c6bfc7e626ec0d6852a9a129cbb2
[...]
$ podman build --format docker -f Dockerfile.alpine --no-cache --build-arg FEATURES=enable_mimalloc ..
[...]
f311f128551f28e9eceb4f4c6ea496fc63cf700b48d491e53fd72a5fca02bf1f
$ podman image save --format docker-archive -o without-s3.tar f311f128551f28e9eceb4f4c6ea496fc63cf700b48d491e53fd72a5fca02bf1f
[...]
$ du -b with-s3.tar without-s3.tar
213288960	with-s3.tar
207968256	without-s3.tar
$ du -h with-s3.tar without-s3.tar
204M	with-s3.tar
199M	without-s3.tar
```

BlackDex:
Its more the binary bloat, the binary will be about 1/3rd larger, or in percentage 33%. Which from my point of view is a lot.

ProbstDJakob:
Either you mean another binary or its "only" ~11%. Nonetheless it is only enabled for the container images not as a default feature and for that it is only a increase of ~2.5% for alpine (and 1.8% for debian).

```console
$ podman run --rm 6bbe0a62fec4702a9928a4dd8663d45ee463c6bfc7e626ec0d6852a9a129cbb2 du -h /vaultwarden
52.0M	/vaultwarden
$ podman run --rm 6bbe0a62fec4702a9928a4dd8663d45ee463c6bfc7e626ec0d6852a9a129cbb2 du -b /vaultwarden
54528456	/vaultwarden
$ podman run --rm f311f128551f28e9eceb4f4c6ea496fc63cf700b48d491e53fd72a5fca02bf1f du -h /vaultwarden
46.9M	/vaultwarden
$ podman run --rm f311f128551f28e9eceb4f4c6ea496fc63cf700b48d491e53fd72a5fca02bf1f du -b /vaultwarden
49207744	/vaultwarden
$ math 54528456 / 49207744
1.108128
```

ProbstDJakob:
The commit https://github.com/dani-garcia/vaultwarden/pull/7412/commits/674fff782504c66c822e7c694078f3b82e40f5d7 is completely optional. I just thought it may be reasonable to adhere to the OCI standard. Though your `ghcr.io/blackdex/rust-musl` does also neither adhere to the OCI nor the POSIX spec (no judgement intended) and thus podman still warns about the usage of `SHELL`, but at least this container image can be build with OCI compliant builders (yours is to complex to "simply" patch it). But as I said it is completely optional and I can drop it again (or move it to a separate PR).

BlackDex:
The final images are OCI compliant and `HEALTHCHECK` will be ignored by tools adhering too it, so that is fine. Healthcheck for docker users is a nice to have, and since we build using Docker Buildx, that probably is just fine to keep in my opinion.

The other stuff might probably be a good change.

ProbstDJakob:
If I got you right, removing the `SHELL` instruction and non POSIX compliant commands is fine, then that's done by https://github.com/dani-garcia/vaultwarden/pull/7412/commits/97ac457910cb979c9abb4fe13bec2eb071f79f48.

BlackDex:
I did some more testing my self. But, `SHELL` works just fine via Podman. It only doesn't persist into the final image. Which I think isn't an issue. You can build just fine using `podman`, and the same goes for my rust-musl images btw.

ProbstDJakob:
The `SHELL` instruction itself is not a problem and will simply be ignored by podman like the `HEALTHCHECK` instruction, but as it is ignored, the shell still is `/bin/sh` and the most it "guarantees" is that it is POSIX compliant (more or less). That is the command `[[` and `source` may or may not work depending on what `/bin/sh` actually refers to and if the shell (like bash) has a special POSIX compliant mode which deactivates some features in such a case. In my case it broke without `--format=docker` and if I remember correctly it was caused by `source` not being supported by `/bin/sh`.

ProbstDJakob:
Yep, it was caused by `source`:

```console
$ podman build -f Dockerfile.alpine .
WARN[0000] missing "VW_VERSION" build argument. Try adding "--build-arg VW_VERSION=<VALUE>" to the command line
[1/7] STEP 1/1: FROM --platform=linux/amd64 docker.io/vaultwarden/web-vault@sha256:f004f72a5d357b87483839500a517da3d1b4ea0a57b9731989d298cccea7d02a AS vault
Trying to pull docker.io/vaultwarden/web-vault@sha256:f004f72a5d357b87483839500a517da3d1b4ea0a57b9731989d298cccea7d02a...
Getting image source signatures
Copying blob 63d44cf76346 done   |
Copying blob 217dab312907 done   |
Copying config c91785f078 done   |
Writing manifest to image destination
--> c91785f07873
[2/7] STEP 1/1: FROM --platform=linux/amd64 ghcr.io/blackdex/rust-musl:x86_64-musl-stable-1.96.1 AS build_amd64
Trying to pull ghcr.io/blackdex/rust-musl:x86_64-musl-stable-1.96.1...
Getting image source signatures
Copying blob 213d412e9015 done   |
Copying blob e2b5641cbe38 done   |
Copying blob d4d8465efe9f done   |
Copying blob 686cde56dcb8 done   |
Copying blob 2c1ce468d9f3 done   |
Copying blob c1b08c27184b done   |
Copying blob 5d61f0351ede done   |
Copying blob c7e05d2bc8c0 done   |
Copying blob 545e15c9c9a9 done   |
Copying blob fe55365f8df1 done   |
Copying blob 4fd10acafe73 done   |
Copying blob 61b9eafd67e5 done   |
Copying blob 3d27edb2f0b6 done   |
Copying blob 8d65bbefca10 done   |
Copying blob 0c1ee469dc0c done   |
Copying blob 05bc64b7a7ef done   |
Copying blob 920b98a8fb05 done   |
Copying blob f44fea201fa4 done   |
Copying blob b3fe3ef1b131 done   |
Copying blob 2c1ce468d9f3 skipped: already exists
Copying config 977872a325 done   |
Writing manifest to image destination
--> 977872a325eb
[6/7] STEP 1/19: FROM --platform=linux/amd64 977872a325ebd30a319c66518651c2a51898f80a9a5ac32c38ed9bce34fa1ac8 AS build
[6/7] STEP 2/19: ARG TARGETARCH
WARN[0027] SHELL is not supported for OCI image format, [/bin/bash -eux -o pipefail -c] will be ignored. Must use `docker` format
--> 27ef646ec354
[6/7] STEP 3/19: ARG TARGETVARIANT
WARN[0027] SHELL is not supported for OCI image format, [/bin/bash -eux -o pipefail -c] will be ignored. Must use `docker` format
--> d49194cd56c0
[6/7] STEP 4/19: ARG TARGETPLATFORM
WARN[0027] SHELL is not supported for OCI image format, [/bin/bash -eux -o pipefail -c] will be ignored. Must use `docker` format
--> a6778b907c6a
[6/7] STEP 5/19: SHELL ["/bin/bash", "-o", "pipefail", "-c"]
WARN[0027] SHELL is not supported for OCI image format, [/bin/bash -o pipefail -c] will be ignored. Must use `docker` format
--> 6415285c4217
[6/7] STEP 6/19: ENV DEBIAN_FRONTEND=noninteractive     LANG=C.UTF-8     TZ=UTC     TERM=xterm-256color     CARGO_HOME="/root/.cargo"     USER="root"     PQ_LIB_DIR="/usr/local/musl/pq17/lib"
WARN[0027] SHELL is not supported for OCI image format, [/bin/bash -o pipefail -c] will be ignored. Must use `docker` format
--> fcb9550d2cbf
[6/7] STEP 7/19: RUN mkdir -pv "${CARGO_HOME}" &&     rustup set profile minimal
info: profile set to minimal
WARN[0028] SHELL is not supported for OCI image format, [/bin/bash -o pipefail -c] will be ignored. Must use `docker` format
--> 9a51ae04f06a
[6/7] STEP 8/19: RUN USER=root cargo new --bin /app
    Creating binary (application) `app` package
note: see more `Cargo.toml` keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
WARN[0028] SHELL is not supported for OCI image format, [/bin/bash -o pipefail -c] will be ignored. Must use `docker` format
--> e17f9ac2eb42
[6/7] STEP 9/19: WORKDIR /app
WARN[0029] SHELL is not supported for OCI image format, [/bin/bash -o pipefail -c] will be ignored. Must use `docker` format
--> 98838735bc11
[6/7] STEP 10/19: RUN echo "export CARGO_TARGET=${RUST_MUSL_CROSS_TARGET}" >> /env-cargo &&     cat /env-cargo
export CARGO_TARGET=x86_64-unknown-linux-musl
WARN[0029] SHELL is not supported for OCI image format, [/bin/bash -o pipefail -c] will be ignored. Must use `docker` format
--> 31716f045218
[6/7] STEP 11/19: RUN source /env-cargo &&     rustup target add "${CARGO_TARGET}"
/bin/sh: 1: source: not found
Error: building at STEP "RUN source /env-cargo &&     rustup target add "${CARGO_TARGET}"": while running runtime: exit status 127
```

BlackDex:
> The `SHELL` instruction itself is not a problem and will simply be ignored by podman like the `HEALTHCHECK` instruction, but as it is ignored, the shell still is `/bin/sh`

That is not true during my tests at least.
This is what a podman build returns.
> WARN[0075] SHELL is not persisted in the OCI image format, [/bin/bash -eux -o pipefail -c] will be used for RUN subsequent instructions to build this image but will not be present in saved image which may affect any images that use this as a base. Must use `docker` format to persist in saved image 

It will use that to start the shell per `RUN`, maybe it's a newer feature though, that i have not checked.

Edit:
I'm using podman v6.0.1 btw

ProbstDJakob:
Okay, yeah 6 has some breaking changes (and some very good ones), but currently I still use the one Fedora delivers which is 5.8.4. Nonetheless I personally always prefer staying away from `bash` and write POSIX compliant scripts, but that is a personal preference. So just say and I drop the commit. But still, OCI does not support that feature. It may be added in the future or the other way around, builders will remove support. The POSIX solution will work either way.

DrissiReda:
If having the tag versions of the image container is a deal breaker we can still have something like `latest-s3` or `v1.36.0-s3`. I'm looking forward to using vaultwarden with 2 replicas and this would avoid me having to rebuild images at every release.

carstenblt:
> If having the tag versions of the image container is a deal breaker we can still have something like `latest-s3` or `v1.36.0-s3`. I'm looking forward to using vaultwarden with 2 replicas and this would avoid me having to rebuild images at every release.

+1!

Having a stateless deployment is so much more convenient!

ProbstDJakob:
Just to make clear, this PR has nothing to do with a stateless deployment. That’s a completely different problem.

PR #7421, opened 2026-07-14
Make 2FA remember duration configurable

tom27052006:
## Summary

This makes the lifetime of remembered two-factor authentication logins configurable instead of always using the hard-coded 30-day duration.

- Adds the editable `/admin` setting `Two-Factor remember duration`
- Adds the `TWO_FACTOR_REMEMBER_DAYS` environment variable
- Keeps 30 days as the default for backward compatibility
- Validates values between 1 and 90 days
- Uses the configured duration when issuing new 2FA remember tokens

Existing remember tokens keep the expiration time with which they were issued. The new duration applies to newly issued tokens. `DISABLE_2FA_REMEMBER` continues to disable the feature entirely.

## Testing

- `cargo clippy --features sqlite -- -D warnings`
- `git diff --check`


BlackDex:
I'm not sure if 365 days as a max is wise.
30 might be too short for some, which i can understand. But a whole year, almost feels like better to not have it enabled at all.
I want to prevent people just setting the max because it's annoying. Since it is there for a reason.

tom27052006:
> I'm not sure if 365 days as a max is wise. 30 might be too short for some, which i can understand. But a whole year, almost feels like better to not have it enabled at all. I want to prevent people just setting the max because it's annoying. Since it is there for a reason.

What would you think about a max of 90 days? It would still offer enough flexibility but wouldn´t be too much.

tom27052006:
Changed the Max to 90 days

PR #7422, opened 2026-07-14
Add default organization for SSO users

tom27052006:
## Summary

This adds an optional default organization for users when they complete their first SSO sign-in.

- Adds the editable `/admin` setting `SSO default organization UUID`
- Adds the `SSO_DEFAULT_ORGANIZATION_UUID` environment variable
- Validates that the configured value is a UUID and normalizes it to the canonical lowercase hyphenated form (uppercase, braced, and non-hyphenated input works too)
- Adds first-time SSO users to the configured organization with `Accepted` status
- Does not send an invitation email
- Leaves users pending until an organization administrator confirms them
- Preserves an existing membership instead of replacing its status or permissions

If the configured organization does not exist, SSO enrollment fails with a clear error instead of assigning the user incorrectly. Collections or groups must still be assigned through the normal organization administration flow.

## Testing

- `cargo test --features sqlite sso::tests`
- `cargo clippy --features sqlite -- -D warnings`
- `git diff --check`


src/sso.rs
```diff
@@ -354,6 +359,80 @@ pub async fn redeem(
...
+
+fn new_default_sso_membership(user_uuid: UserId, org_id: OrganizationId) -> Membership {
+    let mut membership = Membership::new(user_uuid, org_id, None);
+    membership.status = MembershipStatus::Accepted as i32;
+    membership.atype = MembershipType::User as i32;
+    membership
+}
+
+#[cfg(test)]
+mod default_organization_tests {
```
RaphaelRoumezin: the convention of the repository would have this module named `tests`, and be at the bottom of the file

src/sso.rs
```diff
@@ -354,6 +359,80 @@ pub async fn redeem(
...
+        let membership = new_default_sso_membership(
+            UserId::from("00000000-0000-0000-0000-000000000001"),
+            OrganizationId::from("00000000-0000-0000-0000-000000000002"),
+        );
+
+        assert!(!membership.access_all);
+        assert_eq!(membership.status, MembershipStatus::Accepted as i32);
+        assert_eq!(membership.atype, MembershipType::User as i32);
+        assert!(membership.invited_by_email.is_none());
+    }
```
RaphaelRoumezin: this feels extra, given the simplicity of the function it is testing

src/sso.rs
```diff
@@ -354,6 +359,80 @@ pub async fn redeem(
...
+    };
+    Ok(OrganizationId::from(parsed.to_string()))
+}
+
+fn new_default_sso_membership(user_uuid: UserId, org_id: OrganizationId) -> Membership {
+    let mut membership = Membership::new(user_uuid, org_id, None);
+    membership.status = MembershipStatus::Accepted as i32;
+    membership.atype = MembershipType::User as i32;
+    membership
+}
```
RaphaelRoumezin: This function is only called once, and to build a single structure. I think it could be flattened

.env.template
```diff
@@ -501,6 +501,10 @@
 ## Allow unknown email verification status. Allowing this with `SSO_SIGNUPS_MATCH_EMAIL=true` open potential account takeover.
 # SSO_ALLOW_UNKNOWN_EMAIL_VERIFICATION=false
 
+## Automatically add users on their first SSO sign-in as accepted members of this organization.
+## An administrator must confirm users and assign collections or groups. No invitation email is sent.
```
RaphaelRoumezin: This description should match the one in config.rs

src/config.rs
```diff
@@ -1086,6 +1088,12 @@ fn validate_config(cfg: &ConfigItems, on_update: bool) -> Result<(), Error> {
         validate_sso_master_password_policy(cfg.sso_master_password_policy.as_ref())?;
     }
 
+    if let Some(org_uuid) = &cfg.sso_default_organization_uuid
+        && uuid::Uuid::parse_str(org_uuid).is_err()
+    {
+        err!("`SSO_DEFAULT_ORGANIZATION_UUID` must be a valid UUID")
+    }
+
```
RaphaelRoumezin: Maybe this could use `normalize_organization_uuid`

RaphaelRoumezin (commented):
I'm not a maintainer, but I've made a few reviews for your PR from my personal standpoint.

src/sso.rs
tom27052006: Good point — I checked and all other test modules in the repo follow that pattern. Renamed the module to tests and moved it to the bottom of sso.rs in 9fc9b7c

src/sso.rs
tom27052006: Yes, it mostly asserted the defaults that Membership::new already sets, so it didn't test any real logic. I removed it — and since the helper only existed so the test could call it, this also enabled your suggestion below. The remaining tests cover the UUID normalization, which is actual logic.

src/sso.rs
tom27052006: Done — flattened into enroll_user_in_default_organization. I kept the explicit status/atype assignments even though they currently match the defaults of Membership::new, so the enrollment behavior doesn't silently change if those defaults ever do.

.env.template
tom27052006: .env.template and the config doc in config.rs now use the same wording, including the note that no invitation email is sent

src/config.rs
tom27052006: Good idea, the validation in config.rs now calls normalize_organization_uuid directly instead of duplicating the parse. That also guarantees validation and the organization lookup always accept exactly the same set of inputs, so they can't drift apart (which was the root cause of the earlier lookup bug this function fixed).

tom27052006:
> I'm not a maintainer, but I've made a few reviews for your PR from my personal standpoint.

Thanks for taking the time to review! I've pushed a commit addressing all your points — details in the replies to the individual comments below :)

Timshel:
Not a maintainer either but some comments.
I think long term group mapping from the provider would be better (I intend to open a PR with it but waiting for others PR to be merged), but in the meantime it probably makes sense :).

Another way to do that would be to return it in [get_org_domain_sso_verified](https://github.com/dani-garcia/vaultwarden/blob/660faee68e3406d33244b67eadc18524c47674c2/src/api/core/organizations.rs#L917).
This would mean the resulting state is not accepted, but the user would be notified of the enrollment when setting the password (important when the Organization has the admin reset policy) and the correct password requirements would be applied.

src/sso.rs
```diff
@@ -328,6 +331,8 @@ pub async fn redeem(
     sso_auth.delete(conn).await?;
 
     if sso_user.is_none() {
+        enroll_user_in_default_organization(user, conn).await?;
```
Timshel: I think it's better to move the enrollment in the [post_set_password](https://github.com/dani-garcia/vaultwarden/blob/660faee68e3406d33244b67eadc18524c47674c2/src/api/core/accounts.rs#L439).

It's the usual place where organization invitation is accepted.
Additionally, this would ensure that the user has set a password, without-it the `Accepted` status would cause issues (Admin confirmation would fail).

src/sso.rs
```diff
@@ -354,6 +359,38 @@ pub async fn redeem(
...
+    if Organization::find_by_uuid(&org_id, conn).await.is_none() {
+        err!("The organization configured in `SSO_DEFAULT_ORGANIZATION_UUID` does not exist")
+    }
+
+    let mut membership = Membership::new(user.uuid.clone(), org_id.clone(), None);
+    membership.status = MembershipStatus::Accepted as i32;
+    membership.atype = MembershipType::User as i32;
+    membership.save(conn).await?;
+
+    info!("Added SSO user {} to default organization {} pending confirmation", user.uuid, org_id);
```
Timshel: Could use the Organization `billing_email`  to send [send_invite_accepted](https://github.com/dani-garcia/vaultwarden/blob/660faee68e3406d33244b67eadc18524c47674c2/src/mail.rs#L481), might be a way to notify the admin that the user was added.

src/sso.rs
```diff
@@ -354,6 +359,38 @@ pub async fn redeem(
...
+
+    if Membership::find_by_user_and_org(&user.uuid, &org_id, conn).await.is_some() {
+        return Ok(());
+    }
+
+    if Organization::find_by_uuid(&org_id, conn).await.is_none() {
+        err!("The organization configured in `SSO_DEFAULT_ORGANIZATION_UUID` does not exist")
+    }
+
+    let mut membership = Membership::new(user.uuid.clone(), org_id.clone(), None);
```
Timshel, line 383: In [Organization.send_invite](https://github.com/dani-garcia/vaultwarden/blob/660faee68e3406d33244b67eadc18524c47674c2/src/api/core/organizations.rs#L1138) after creating the membership there is additional logic such as logging an event `OrganizationUserInvited`, adding the user to collections and groups.

src/sso.rs
tom27052006: I´ve changed it, the first SSO login now creates an Invited membership. It only transitions to Accepted after the user’s password has been saved, ensuring admin confirmation works correctly

src/sso.rs
tom27052006: Done, now after the user gets accepted, the existing notification is now sent to Organization.billing_email, SMTP failures are logged without failing the otherwise successful enrollment

src/sso.rs
tom27052006: An OrganizationUserInvited event is now logged with the user, device, and IP information. I wasn’t entirely sure what you intended regarding collections and groups, since, to my knowledge, they cannot be assigned directly through the current SSO flow, or did i miss something?

tom27052006:
> Not a maintainer either but some comments. I think long term group mapping from the provider would be better (I intend to open a PR with it but waiting for others PR to be merged), but in the meantime it probably makes sense :).
> 
> Another way to do that would be to return it in [get_org_domain_sso_verified](https://github.com/dani-garcia/vaultwarden/blob/660faee68e3406d33244b67eadc18524c47674c2/src/api/core/organizations.rs#L917). This would mean the resulting state is not accepted, but the user would be notified of the enrollment when setting the password (important when the Organization has the admin reset policy) and the correct password requirements would be applied.

Thanks for the review, Ive watched you comments and gave a feedback / result. To the get_org_domain_sso_verified point:
Now, when a default organization is configured, its actual UUID is returned instead of the dummy value, this ensures the Web Vault applies that organization’s password policies during enrollment

src/sso.rs
Timshel: A yes sorry about that, had a quick look and misread it :(.

PR #7433, opened 2026-07-17
Add support for Fill Assist

RaphaelRoumezin:
See #7428 and https://bitwarden.com/blog/fill-assist-improving-autofill-every-time-for-everyone/

Since this feature is only client-side and is opt-in on the client anyway, I feel like it can just be added as a static always-enabled feature flag. But it could also be added as a supported experimental flag.

### DIFF
--- a/src/api/core/mod.rs
+++ b/src/api/core/mod.rs
@@ -220,6 +220,7 @@ fn config() -> Json<Value> {
         &FeatureFlagFilter::ValidOnly,
     );
     feature_states.insert("pm-19148-innovation-archive".to_owned(), true);
+    feature_states.insert("fill-assist-targeting-rules".to_owned(), true);
 
     Json(json!({
         // Note: The clients use this version to handle backwards compatibility concerns
--- a/src/api/core/mod.rs
+++ b/src/api/core/mod.rs
@@ -245,6 +245,9 @@ fn config() -> Json<Value> {
           "notifications": format!("{domain}/notifications"),
           "sso": "",
           "cloudRegion": null,
+          // fillAssistRules as returned by the Bitwarden server (https://github.com/bitwarden/server @ 045113a),
+          // and hardcoded in the Bitwarden clients (https://github.com/bitwarden/clients @ 2581be6)
+          "fillAssistRules": "https://github.com/bitwarden/map-the-web/releases/latest/download",
         },
         // Bitwarden uses this for the self-hosted servers to indicate the default push technology
         "push": {


CompKingCanada:
Do I understand that this is hard coding an option that would appear in vaultwarden.env?

fill-assist-targeting-rules = true

Can admins add this enviornment variable if users aren't seeing the new AutoFill option in the Browser Add-ons?

RaphaelRoumezin:
I'm not sure what is that "vaultwarden.env" you're referring to, but this PR effectively hard codes and enables this feature.

Another option that could be done for this PR, as mentioned in the PR message, is to add the flag as a supported experimental flag, so that admins can add it to their config file (or env variables) to enable the experimental feature as an opt-in on their deployment.

dani-garcia:
The official server also adds the fill assist URL to the /config endpoint, we should probably add it too in case the clients remove the default value.

```json
{
  "version": "2026.6.2",
  "gitHash": "5e5b0881",
  "server": null,
  "environment": {
    "cloudRegion": "US",
    "vault": "https://vault.bitwarden.com",
    "api": "https://api.bitwarden.com",
    "identity": "https://identity.bitwarden.com",
    "notifications": "https://notifications.bitwarden.com",
    "sso": "https://sso.bitwarden.com",
    "fillAssistRules": "https://github.com/bitwarden/map-the-web/releases/latest/download"
  },
  ...
}
```

(From https://api.bitwarden.com/config)

RaphaelRoumezin:
> The official server also adds the fill assist URL to the /config endpoint, we should probably add it too in case the clients remove the default value.

Thanks for the insight @dani-garcia, change is implemented.

src/api/core/mod.rs
```diff
@@ -220,6 +220,7 @@ fn config() -> Json<Value> {
         &FeatureFlagFilter::ValidOnly,
     );
     feature_states.insert("pm-19148-innovation-archive".to_owned(), true);
+    feature_states.insert("fill-assist-targeting-rules".to_owned(), true);
```
BlackDex, line 223: I'm not sure if we should enable this by default?
Mainly because Bitwarden it self doesn't add it by default as `true`.

Adding it as an optional flag would be fine by me.
I think having the `fillAssistRules` in as default shouldn't be an issue though.
It would only become an issue if it will change.

BlackDex (changes requested):

src/api/core/mod.rs
timkgh: Users would benefit from having it work by default and it would avoid questions why autofill doesn't work. I would enable it by default with an option to disable it if it really causes issues for someone, but the default should be to try to make the experience _better_ for everyone.

RaphaelRoumezin: I think that would make a justifiable option to be in the config so much for users that want to enable it as users that want to disable it

But having it as default is also a good option, considering it can be disabled client side in settings.

BlackDex: These are experimental features, and they are toggled via flags for a reason. Adding this as a default with an option to disable it needs again a new config flag, as where all flags are already configurable with our current setup. 

And again, Bitwarden has it disabled by default to, and it would surely benefit their users too i works imagine.

tom27052006:
Small note for web-v2026.9.0: fill-assist-targeting-rules now also exposes the new Fill Assist policy (PolicyType.FillAssist = 22) in the web vault.

Since Vaultwarden doesn’t support policy 22 yet, opening or saving that policy would fail once this flag is enabled.

Might be worth adding support for policy 22 directly here.

PR #7434, opened 2026-07-17
Remove old compatibility code

Timshel:
Does not appear to be used any more (tested with `v2026.6.2`).

It starts to cause issues with `v2026.6.4` since it's reused for blobs cf https://github.com/bitwarden/clients/pull/20765/files

### DIFF
--- a/src/db/models/cipher.rs
+++ b/src/db/models/cipher.rs
@@ -306,16 +306,6 @@ impl Cipher {
             type_data_json = Value::Null;
         }
 
-        // Clone the type_data and add some default value.
-        let mut data_json = type_data_json.clone();
-
-        // NOTE: This was marked as *Backwards Compatibility Code*, but as of January 2021 this is still being used by upstream
-        // data_json should always contain the following keys with every atype
-        data_json["fields"] = json!(fields_json);
-        data_json["name"] = json!(self.name);
-        data_json["notes"] = json!(self.notes);
-        data_json["passwordHistory"] = Value::Array(password_history_json.clone());
-
         let collection_ids = if let Some(cipher_sync_data) = cipher_sync_data {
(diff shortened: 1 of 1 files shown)


BlackDex (approved):
Seems to work ok with v2026.6.1 clients too, like FireFox/Chrome extensions.

stefan0xC (approved):
Thanks. I can confirm that this is needed for the new web-vault `v2026.6.4`.

BlackDex:
Seems i broke the Alpine builds with my rust-musl changes.
The target seems to be using a variable i removed and wasn't really something to be used by others.

So, i need to fix that before this PR will build.

Nerevar-de:
Is there an ETA for the release with that fixincluded?
Unfortunatelly this is a severe problem after the auto-update of the browser extension to 2026.7.0

BlackDex:
For now you could use `:testing` tagged container images which has this fix already.

coolaj86:
For those of us not familiar, what does `:testing` mean? Is that a suffix for a Docker image name or something? Or a cargo or git build tag or something like that?

What's the minimum information a person needs to know to follow documentation on how to use that?

BlackDex:
It's the container image. Crates do not have those tags, and we only provide container images.

coolaj86:
Thanks.

I got it working with the `:testing` binaries: https://github.com/bitwarden/clients/issues/22013#issuecomment-5064079493

BlackDex:
Strangely enough, i do still see this on `vault.bitwarden.com`, but it also doesn't seem to cause any harm/issues not having it anymore.

Haran:
This has broken various 3rd party integrations and libraries. 

API response schema changes must be announced explicitly and in a verbose manner, not just like _"Remove old compatibility code by @Timshel"_

PR #7435, opened 2026-07-17
Support admin reset 2fa

Timshel:
When looking at https://bitwarden.com/help/recover-a-member-account/ it's mentionned that of the require 2fa policy is on the user will be revoked.

As such though it made sense to support `EMAIL_2FA_AUTO_FALLBACK`, this way a better account security is kept and the user is nor revoked if the policy is activated in any org.
It will be activated only if not already present (since if working it would make no sense to reset).

Need a client based on `2026.6.4`, tested with https://github.com/Timshel/oidc_web_vault/releases/tag/v2026.6.4-1

### DIFF
--- a/src/static/templates/email/admin_account_recovery.hbs
+++ b/src/static/templates/email/admin_account_recovery.hbs
@@ -1,0 +1,12 @@
+Admin account recovery from {{org_name}} organization
+<!---------------->
+{{#if reset_password}}
+The master password for {{user_name}} has been changed.
+{{/if}}
+{{#if reset_2fa}}
+Your two steps providers have been reset. {{#if fallback_2fa_email}} Email two factor has been activated as a fallback.{{/if}}
+{{/if}}
+
+If you did not initiate this request, please reach out to your administrator immediately.
+
+{{> email/email_footer_text }}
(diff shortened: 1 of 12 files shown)


BlackDex (changes requested):
Looks fine at a quick glance over it.
I have not yet tested it though, but there are some conflicts currently.

src/static/templates/email/admin_account_recovery.hbs
```diff
@@ -0,0 +1,12 @@
+Admin account recovery from {{org_name}} organization
+<!---------------->
+{{#if reset_password}}
+The master password for {{user_name}} has been changed.
+{{/if}}
+{{#if reset_2fa}}
+Your two steps providers have been reset. {{#if fallback_2fa_email}} Email two factor has been activated as a fallback.{{/if}}
```
BlackDex: I think the first part isn't fully correct.
This is probably better.
```suggestion
Your two-step verification providers have been reset.
```

BlackDex (changes requested):
Works great, only one remark regarding the mail though.
And you probably need to adjust that at multiple locations.

BlackDex (approved):

bmm-alc:
Hi
I can confirm now I can revoke 2fa for a user.
The user then can login directly.

PR #7444, opened 2026-07-19
Add optional TOTP two-factor authentication for the admin page

tom27052006:
Adds an optional TOTP second factor for the `/admin` page.

Setting `ADMIN_TOTP_SECRET` to a Base32 secret requires a 6-digit one-time code in addition to the admin token at login. It allows one 30-second step of time drift, rejects replayed codes, and keeps the existing admin rate limiting. TOTP verification is shared with the existing user 2FA implementation.

Admin TOTP can be managed from the normal config page. A new 20-byte secret and QR code are generated entirely in the browser using Web Crypto and a locally bundled QR generator. No external service is used. The secret is only saved through the normal config flow, and existing saved secrets are never sent back to the browser or shown again. Setting, replacing, and removing the secret all use the existing config flow, without a separate enrollment page or an additional TOTP check when disabling it.

Environment-provided secrets still work and can be overridden by the saved config. Removing the saved override falls back to the environment value.

Has no effect when `DISABLE_ADMIN_TOKEN` is enabled.


src/api/admin.rs
```diff
@@ -246,6 +255,153 @@ fn validate_token(token: &str) -> bool {
...
+        return true;
+    };
+    let Some(code) = code.map(str::trim) else {
+        return false;
+    };
+
+    // A matching recovery code logs in and disables the admin page 2FA, like the user 2FA recovery flow
+    if let Some(stored_hash) = CONFIG.admin_totp_recovery_code()
+        && !code.is_empty()
+        && crate::crypto::ct_eq(crate::crypto::sha256_hex(code.to_uppercase().as_bytes()), stored_hash.trim())
```
RaphaelRoumezin: This doesn't allow matching if the config-stored hash is lowercase.
Plus, if you keep the recovery code logic, consider using hash algorithms as strong as the ones used for the admin token, and maybe push users to use the same `./vaultwarden hash` flow to generate it.

src/api/admin.rs
```diff
@@ -246,6 +255,153 @@ fn validate_token(token: &str) -> bool {
...
+    {
+        warn!("Admin page 2FA recovery code used, disabling TOTP");
+        if let Err(e) = CONFIG.set_admin_totp(None, None).await {
+            error!("Unable to disable admin page TOTP after recovery code use: {e:?}");
+            return false;
+        }
+        return true;
+    }
```
RaphaelRoumezin: I dont think having a "recovery code" is really useful in this case, as it is meant for admins, who can just reset the TOTP in the config directly. It's just another static weakness and TOTP bypass, especially if admins store it in the same location as the admin password (which is likely to be done)

src/api/admin.rs
```diff
@@ -246,6 +255,153 @@ fn validate_token(token: &str) -> bool {
...
+            if time_step <= *last_used {
+                warn!("This admin TOTP code has already been used");
+                return false;
+            }
+            *last_used = time_step;
+            return true;
+        }
+    }
+    false
+}
```
RaphaelRoumezin, line 305: Perhaps, TOTP generation code could be commoned with the VW TOTP own implementation.

src/api/admin.rs
```diff
@@ -246,6 +255,153 @@ fn validate_token(token: &str) -> bool {
...
+        err!("Failed to generate QR code")
+    };
+    let qr_svg = qr.render::<svg::Color<'_>>().min_dimensions(240, 240).build();
+
+    Ok(Json(json!({
+        "secret": secret,
+        "uri": uri,
+        "qr_svg": qr_svg,
+    })))
+}
```
RaphaelRoumezin: This could be done client side

src/api/admin.rs
```diff
@@ -246,6 +255,153 @@ fn validate_token(token: &str) -> bool {
...
+    }
+    if source == "environment" {
+        err!("The TOTP secret is set via `ADMIN_TOTP_SECRET` and can only be removed from the environment")
+    }
+    let Some(secret) = CONFIG.admin_totp_secret() else {
+        err!("Admin page TOTP is not enabled")
+    };
+    if !check_totp_code(&secret, data.code.trim()) {
+        err!("Invalid TOTP code, please try again")
+    }
```
RaphaelRoumezin: The admin is already authenticated, he shouldn't need to re-input the TOTP.

src/static/templates/admin/totp.hbs
RaphaelRoumezin, line 1: In my opinion, the whole TOTP configuration flow could be done entirely client-side, on the config page. The ADMIN_TOTP_SECRET should be as editable as the others, and a modal could provide a helper for configuring it (with QR code and everything).
It would then just use the configuration edit flow to set/clear the TOTP secret.

This would also avoid the unnecessary complex code in `src/config.rs` to make the value non-editable from the admin page and have special states whether it was set from the environment.

src/config.rs
```diff
@@ -1588,6 +1615,33 @@ impl Config {
...
+    pub fn admin_totp_status(&self) -> (bool, &'static str) {
+        let inner = self.inner.read().unwrap();
+        let is_set = |v: &Option<String>| v.as_deref().is_some_and(|s| !s.trim().is_empty());
+        if is_set(&inner._usr.admin_totp_secret) {
+            (true, "enrolled")
+        } else if is_set(&inner._env.admin_totp_secret) {
+            (true, "environment")
+        } else {
+            (false, "none")
+        }
```
RaphaelRoumezin: This doesn't really account for admins that use the `config.json` as a main configuration file. Just because a value is set in the config file doesn't mean it originates from the admin panel.

RaphaelRoumezin (changes requested):
Overall, the idea and implementation is good, and I'd love to see that become a feature.
However I feel pretty major design flaws in the overall execution, mainly from the configuration and enrollment flow in the admin panel.

Do consider I'm not a VW maintainer, and this is my personal opinion.

src/api/admin.rs
tom27052006: This doesn't matter anymore — I removed the recovery code completely (see your next comment), so there's no stored hash left.

src/api/admin.rs
tom27052006: Good point, I agree. I removed the recovery code. If someone loses their authenticator, they can just delete ADMIN_TOTP_SECRET from the config or env and restart — so no extra secret is needed.

src/api/admin.rs
tom27052006: I made one shared verify_totp() function in two_factor::authenticator, and both the user 2FA and the admin check now use it, so the code isn't duplicated. Each side still keeps its own "last used" value. The user path is now constant-time too, and I added a small test.

src/api/admin.rs
tom27052006: I kept this on the server on purpose. The TOTP secret is the root secret, so I'd rather generate it in one place with the server's RNG than in the browser, where it lives in the page and depends on the page not being tampered with. It also has to reach the server to be saved anyway, so doing it in the browser doesn't really save anything — but it would mean adding a JS crypto/QR library to the admin panel, which is intentionally small, plain JS with no external/CDN code. The qrcode crate is build-time only, so there's no runtime cost. That said, this is a preference — if you or the maintainers would rather keep it fully client-side, I'm happy to change it.

src/api/admin.rs
tom27052006: I would keep this too. Most 2FA systems ask for the second factor to turn the second factor off, and for a good reason: 2FA is meant to protect the case where the admin token leaked but the attacker doesn't have the device. If disabling needed no code, then just stealing a session cookie would be enough to remove 2FA, which defeats the whole point. It also doesn't cause lockouts, because the real recovery path for a lost authenticator is removing the secret from the config/env. But of course this is a judgment call — if the maintainers feel an active admin session should be trusted fully, I can drop the check.

src/config.rs
tom27052006: I changed admin_totp_status() to check the environment first and to say "config" instead of "enrolled", and I added a note that a secret set by hand in config.json looks the same as one set from the panel. This also fixes a case where the secret is set in both env and config — it's now correctly shown as not removable, since the env value still applies

src/static/templates/admin/totp.hbs
tom27052006: I would keep this server-side. As a normal editable field, any admin session could read or delete the secret in one click, and there'd be no server check that a code works before enabling it (lockout risk). A browser-side check can also be skipped by posting the value directly, so it has to run on the server anyway. Part of the config.rs code also fixes an existing bug (_duo_akey being dropped on save), so I'd rather keep it. But it's the biggest design call here — happy to switch to the simpler approach if the maintainers prefer it.

tom27052006:
> Overall, the idea and implementation is good, and I'd love to see that become a feature. However I feel pretty major design flaws in the overall execution, mainly from the configuration and enrollment flow in the admin panel.
> 
> Do consider I'm not a VW maintainer, and this is my personal opinion.

Thanks for the review! I've made a few changes, removed the recovery code, shared the TOTP check with the existing user 2FA, and made the env-vs-config status more honest. I kept the enable/disable flow on the server on purpose and explained why in the threads, but I'm happy to change more if you or the maintainers see it differently.

src/api/admin.rs
RaphaelRoumezin: If the browser has been tampered with and the sysadmin logs in the admin panel anyway, I think other problems might appear before RNG manipulation.
Plus, client-side secret generation is already performed for the cryptographic keys in Vaultwarden and for the password generator in clients, so staying on this track seems more appropriate.

Also, the qrcode crate is not build-time only, since you use it in runtime.

src/api/admin.rs
RaphaelRoumezin: If someone steals a session cookie they don't have to remove 2FA, they can just attack the configuration.
This panel is mainly a convenience for the config/env, and shouldn't constrain the admin in what they can edit or not (other than technical limitations ofc). If that was the case, the admin token should be protected in the same manner.

src/static/templates/admin/totp.hbs
RaphaelRoumezin: Again, it's also a risk having the admin token be unprotected, but it's still done anyway. This panel doesn't need "restrict you for your own good" features.

src/config.rs
RaphaelRoumezin: This wasn't my main concern. Your code assumes that a totp secret cannot be disabled from the admin panel just because it was set through env variables, however this doesn't account for configuration that was set manually in `config.json`. It should at least offer a possibility to disable in any context, since `config.json` is supposed to be acounted over env variables.

RaphaelRoumezin:
I still believe most of the flows in this code should be done client-side, without even having server-side running code. This code is way too complex for the admin panel, which is supposed to be a helper web page for sysadmins.
At the very core, this feature should only be enabled through yet another configuration entry, and a client side helper interface should come second to complete the feature.

As I take time to read your code and write those reviews, I'd also appreciate if you actually read them instead of pasting them into your AI agent.

tom27052006:
> I still believe most of the flows in this code should be done client-side, without even having server-side running code. This code is way too complex for the admin panel, which is supposed to be a helper web page for sysadmins. At the very core, this feature should only be enabled through yet another configuration entry, and a client side helper interface should come second to complete the feature.
> 
> As I take time to read your code and write those reviews, I'd also appreciate if you actually read them instead of pasting them into your AI agent.

Thanks, you are right about two things. I was wrong about qrcode. It is used at runtime. I also got the config order wrong. config.json overrides the environment, so my check is wrong.
I looked at the code again. The admin page can show the TOTP secret. The normal config reset can also bypass the extra TOTP check. So my earlier security argument was not right
I will make the code simpler and follow the existing config system. I will keep the TOTP secret hidden. I will move the other config fix into a separate change.
I used AI to help with some replies, but I am responsible for what I write and for the code. I should have checked these points better. Thanks for taking the time to review it!

tom27052006:
I’ve simplified the setup and moved it into the normal config page. The secret and QR code are now generated entirely in the browser. Saving, replacing, and removing the secret uses the existing config flow, and saved secrets are not shown again. I also removed the separate TOTP page, the server-side QR generation, and the extra code check when disabling TOTP.
@RaphaelRoumezin  does this match what you had in mind?

PR #7446, opened 2026-07-20
fix: use nix access for USE_SENDMAIL executable check

### LINKED ISSUE #7445: USE_SENDMAIL rejects valid owner/group-only executable commands
### Prerequisites

Not available: Vaultwarden exits during startup configuration validation before /admin/diagnostics is reachable. Reproduced with the official vaultwarden/server:1.36.0 linux/arm64 image; source main 169aa5efcc8d94684ff3bc813a00e6bcc0cc537a has the same validation.

### Vaultwarden Build Version

1.36.0; also present on main 169aa5efcc8d94684ff3bc813a00e6bcc0cc537a

### Deployment method

Official Container Image

### Custom deployment method

Official vaultwarden/server:1.36.0 image, run in an isolated container with --network none and no published ports.

### Reverse Proxy

None (failure occurs before the HTTP listener starts)

Linux

### Operating System Version

Ubuntu 24.04 LTS host; linux/arm64 container

### Clients

CLI

### Client Version

Not client-specific; Vaultwarden exits during startup configuration validation

### Steps To Reproduce

1. Create a minimal sendmail-compatible executable with owner-only execute permission:

```sh
mkdir -p /tmp/vw-sendmail-repro
printf '#!/bin/sh\nexit 0\n' > /tmp/vw-sendmail-repro/sendmail
chmod 0700 /tmp/vw-sendmail-repro/sendmail
```

2. Run the official image:

```sh
docker run --rm --network none \
  -e USE_SENDMAIL=true \
  -e SENDMAIL_COMMAND=/repro/sendmail \
  -e SMTP_FROM=vaultwarden@example.test \
  -e DOMAIN=http://localhost \
  -v /tmp/vw-sendmail-repro:/repro:ro \
  vaultwarden/server:1.36.0
```

3. Observe that startup rejects the command.
4. Change only the mode to 0755 and repeat. The sendmail validation then passes (the isolated container proceeds to the next unrelated startup check).

### Expected Result

A sendmail command that is executable by the Vaultwarden process (for example, a 0700 file owned by that process user, or a 0750 file executable through its group) should pass configuration validation.

### Actual Result

Vaultwarden rejects 0700 and 0750 commands as “isn't executable”. Only modes with all three execute bits set, such as 0755, pass the current check.

### Logs

```text
[CAUSE] Error loading config:
sendmail command at '"/repro/sendmail"' isn't executable
```

### Screenshots or Videos

_No response_

### Additional Context

The Unix validation in `src/config.rs` currently uses:

```rust
if !metadata.permissions().mode() & 0o111 != 0 {
    err!(format!("sendmail command at {path:?} isn't executable"));
}
```

This condition is true whenever *any* of the three execute bits is absent, so it effectively requires mode `0111` in full. To test whether a file has at least one execute bit, the mask should be checked as `mode & 0o111 != 0` (or rejected when that value equals zero).

I reproduced the failure on the released official image and confirmed the same source condition at main commit `169aa5efcc8d94684ff3bc813a00e6bcc0cc537a`. A 0755 control passes the validation.

I searched open and closed issues/PRs and Discussions for `USE_SENDMAIL`, `SENDMAIL_COMMAND`, executable permissions, and 0700/0750 modes. #3147 is the original sendmail implementation, and discussion #4417 concerns NixOS queue permissions rather than this startup validation.

H-TTTTT:
## Problem

`USE_SENDMAIL` startup validation rejected valid sendmail binaries that the **current process** can execute, but whose mode bits are not world-executable (e.g. `0700` / `0750`).

The old check was based on mode bits and also had operator-precedence issues:

```rust
if !metadata.permissions().mode() & 0o111 != 0
```

## Fix

On Unix, use `nix::unistd::access(path, AccessFlags::X_OK)` so validation matches real execute permission for the running process (same approach suggested in https://github.com/dani-garcia/vaultwarden/issues/7445#issuecomment-5023808210).

- Add Unix-only dependency `nix = 0.31.3` with `fs` feature
- Drop mode-bit-only checks / related unit tests (per review)

## Test plan

- [x] Logic: `access(X_OK)` accepts owner-executable sendmail for the running user
- [x] `nix` pinned to **0.31.3** in `Cargo.toml` / `Cargo.lock`

Fixes #7445

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,7 +61,7 @@ unstable = []
 # Logging
 syslog = "7.0.0"
 # Filesystem access checks (e.g. sendmail executability for the current user)
-nix = { version = "0.30.1", default-features = false, features = ["fs"] }
+nix = { version = "0.31.3", default-features = false, features = ["fs"] }
 
 [dependencies]
 macros = { path = "./macros" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,6 +60,8 @@ unstable = []
 [target."cfg(unix)".dependencies]
 # Logging
 syslog = "7.0.0"
+# Filesystem access checks (e.g. sendmail executability for the current user)
+nix = { version = "0.30.1", default-features = false, features = ["fs"] }
 
 [dependencies]
 macros = { path = "./macros" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -949,6 +949,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
 
 [[package]]
+name = "cfg_aliases"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f079e83a288787bcd14a6aea84cee5c87a67c5a3e660c30f557a3d24761b3527"
+
+[[package]]
 name = "chacha20"
 version = "0.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3249,6 +3255,18 @@ name = "ndk-context"
 version = "0.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "27b02d87554356db9e9a873add8782d4ea6e3e58ea071a9adb9a2e8ddb884a8b"
+
+[[package]]
+name = "nix"
+version = "0.30.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "74523f3a35e05aba87a1d978330aef40f67b0304ac79c1c00b294c9830543db6"
+dependencies = [
+ "bitflags 2.13.0",
+ "cfg-if",
+ "cfg_aliases",
+ "libc",
+]
 
 [[package]]
 name = "nom"
@@ -5930,6 +5948,7 @@ dependencies = [
  "macros",
  "mimalloc",
  "moka",
+ "nix",
  "num-derive",
  "num-traits",
  "opendal",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3258,9 +3258,9 @@ checksum = "27b02d87554356db9e9a873add8782d4ea6e3e58ea071a9adb9a2e8ddb884a8b"
 
 [[package]]
 name = "nix"
-version = "0.30.1"
+version = "0.31.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "74523f3a35e05aba87a1d978330aef40f67b0304ac79c1c00b294c9830543db6"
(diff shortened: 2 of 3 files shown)


BlackDex (changes requested):
I would suggest to use the suggestion provided here https://github.com/dani-garcia/vaultwarden/issues/7445#issuecomment-5023808210

Also, from my point of view those tests are too much, no reason to test if modes work or not. Remove those and adjust the code as suggested.

H-TTTTT:
Addressed the review feedback:

- Switched the Unix check to `nix::unistd::access(path, AccessFlags::X_OK)` as suggested in https://github.com/dani-garcia/vaultwarden/issues/7445#issuecomment-5023808210
- Removed the mode-bit helper and unit tests
- Added `nix` (unix-only, `fs` feature, no default features)

`cargo check --features sqlite` passes on this branch.

H-TTTTT:
@BlackDex follow-up ready for re-review:

- Switched to `nix::unistd::access(..., X_OK)` as suggested in https://github.com/dani-garcia/vaultwarden/issues/7445#issuecomment-5023808210
- Removed the mode helper and unit tests
- Minimal `Cargo.lock` delta (`nix` + `cfg_aliases` only)

Happy to adjust further if anything else is needed.

RaphaelRoumezin:
You may want to update the PR's message to match the current implementation

Cargo.toml
```diff
@@ -60,6 +60,8 @@ unstable = []
 [target."cfg(unix)".dependencies]
 # Logging
 syslog = "7.0.0"
+# Filesystem access checks (e.g. sendmail executability for the current user)
+nix = { version = "0.30.1", default-features = false, features = ["fs"] }
```
BlackDex: The latest version of nix is 0.31.3, please adjust to match.

BlackDex (changes requested):

Cargo.toml
H-TTTTT: Updated: `nix` is now pinned to **0.31.3** in `Cargo.toml` and `Cargo.lock`.

H-TTTTT:
Updated the PR description to match the current implementation (`nix::unistd::access` + `nix` 0.31.3). Ready for re-review when you have a moment. Thanks!

H-TTTTT:
@BlackDex Gentle ping for re-review on the latest commit (`ca9602ef`).

All previous feedback should now be addressed:
1. Using `nix::unistd::access(..., X_OK)` (per your suggestion on #7445)
2. Mode-bit helper / unit tests removed
3. `nix` bumped to **0.31.3** as requested on the Cargo.toml review thread
4. PR description updated to match the implementation

CI is green and the PR is mergeable. Happy to adjust anything else if needed. Thanks!

H-TTTTT:
Closing this PR; will not continue work on it.

RaphaelRoumezin:
How come ? This was a good fix fixing a recognized issue.

PR #7452, opened 2026-07-21
Add support for collection management settings

RaphaelRoumezin:
[Collection Management settings](https://bitwarden.com/resources/resource-collections-management-settings/) is a brick of 4 flags in organization settings, allowing to fine-tune what roles and/or permissions are required for certain critical operations.

## Design choices

- The `allow_admin_access_to_all_collection_items` flag defaults to true, to match Vaultwarden current behavior.
- The settings are stored as an `i16` (SMALLINT in schema), in the form of a bit flagset.

## TODO

- [x] Implement settings form and persistance
- [ ] Add checks for "Allow owners and admins to manage all collections and items from the Admin Console"
- [x] Add checks for "Restrict collection creation to owners and admins"
- [x] Add checks for "Restrict collection deletion to owners and admins"
- [x] Add checks for "Restrict item deletion to members with the `Manage collection` permission"

### DIFF
--- a/migrations/mysql/2026-07-21-120000_add_collection_management/down.sql
+++ b/migrations/mysql/2026-07-21-120000_add_collection_management/down.sql
@@ -1,4 +1,1 @@
-ALTER TABLE organizations DROP COLUMN allow_admin_access_to_all_collection_items;
-ALTER TABLE organizations DROP COLUMN limit_collection_creation;
-ALTER TABLE organizations DROP COLUMN limit_collection_deletion;
-ALTER TABLE organizations DROP COLUMN limit_item_deletion;
+ALTER TABLE organizations DROP COLUMN collection_settings;
--- a/migrations/mysql/2026-07-21-120000_add_collection_management/down.sql
+++ b/migrations/mysql/2026-07-21-120000_add_collection_management/down.sql
@@ -1,0 +1,4 @@
+ALTER TABLE organizations DROP COLUMN allow_admin_access_to_all_collection_items;
+ALTER TABLE organizations DROP COLUMN limit_collection_creation;
+ALTER TABLE organizations DROP COLUMN limit_collection_deletion;
+ALTER TABLE organizations DROP COLUMN limit_item_deletion;
--- a/migrations/mysql/2026-07-21-120000_add_collection_management/up.sql
+++ b/migrations/mysql/2026-07-21-120000_add_collection_management/up.sql
@@ -1,0 +1,4 @@
+ALTER TABLE organizations ADD COLUMN allow_admin_access_to_all_collection_items BOOLEAN NOT NULL DEFAULT TRUE;
+ALTER TABLE organizations ADD COLUMN limit_collection_creation BOOLEAN NOT NULL DEFAULT FALSE;
+ALTER TABLE organizations ADD COLUMN limit_collection_deletion BOOLEAN NOT NULL DEFAULT FALSE;
+ALTER TABLE organizations ADD COLUMN limit_item_deletion BOOLEAN NOT NULL DEFAULT FALSE;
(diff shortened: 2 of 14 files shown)


RaphaelRoumezin:
In current implementation:
https://github.com/dani-garcia/vaultwarden/blob/169aa5efcc8d94684ff3bc813a00e6bcc0cc537a/src/db/models/organization.rs#L512-L513

This line indicates to the user's client that org policy "Restrict collection creation to owners and admins" is off **only** if the user is a Custom role with Manage collections.
I feel like this is useless, as this could just be statically off for all users, since having it on or off for any role other than Custom with Manage collections has no effect.

This doesn't affect this PR because this flag is now set by organization settings instead of being deduced, but if I don't understand something about this (or if this line is actually useful for something after all), this might have an effect on this PR's code.

RaphaelRoumezin:
I don't know what to think about this, but it is worth noting that in Bitwarden, the "Restrict item deletion to members with the Manage collection permission" option can be bypassed if:
- A user wants to delete a cipher which is in a collection where he has the permission "Edit items"
- The user has access to another collection with the "Manage collection" permission

The cipher can just be added to that second collection by the user, and he will gain the power to delete it permanently in both collections.

This feature kind of breaks the Bitwarden permission system, i think

RaphaelRoumezin:
The 3 last settings have been implemented.

The "Allow owners and admins to manage all collections and items from the Admin Console" setting would probably be easier to implement whenever proper permissions are done, such as:
- Remove the ambiguous `access_all` flag
- Have correct collection permissions management, and do not make owner/admins automatically have the "Manage Collection" permission in the UI.

I believe this is part of the work done in #7397.

Then, all arbitrary permissions admin/owners have on collections and items should be dependent of the setting flag being set.
The Bitwarden implementation also has those anti-lockouts, which should be implemented too:
- If the setting is disabled, all collections must have at least one member with the permission "Manage collection" in them. This affects collection creation and updates.
- If the setting is disabled, any collection without at least one member with the "Manage collection" permission will be excluded from the setting, allowing owner/admins full control over it anyway. Warnings will also appear in the UI, indicating to the admins that this must be changed. This can happen if the last managing members are deleted from the org or the server, or if the collection was already like that and the setting got turned off (since the "must have at least one managing member left" check only occur when the setting is already off).

The rest of this PR I feel is pretty clean, except for the choice to make a flag byte for the database, that is unprecedented and increase code complexity. Maybe this could just be done through 4 boolean fields.

Timshel:
Not a maintainer, but I wonder if it would not make sense to use a struct and serialize it in json in the DB reusing [impl_FromToSqlText](https://github.com/dani-garcia/vaultwarden/blob/660faee68e3406d33244b67eadc18524c47674c2/src/db/mod.rs#L358).
It would take more space in DB, not sure if it's an issue, but this would allow to easily support any non-boolean setting.
It could be included in the Org and not need its own db put/get.

RaphaelRoumezin:
> Not a maintainer, but I wonder if it would not make sense to use a struct and serialize it in json in the DB reusing [impl_FromToSqlText](https://github.com/dani-garcia/vaultwarden/blob/660faee68e3406d33244b67eadc18524c47674c2/src/db/mod.rs#L358). It would take more space in DB, not sure if it's an issue, but this would allow to easily support any non-boolean setting. It could be included in the Org and not need its own db put/get.

Not sure this solves anything. The whole point of having specific db fetcher is to lighten db requests, since the contexts where these settings need to be checked don't really call for the whole org object to be pulled.
Also, I don't see in what case non-boolean values would need to be stored here, since its only 4 checkboxes at the moment. I believe we can have other DB fields added or the schema altered if at some point in the future, non-boolean settings are added to Bitwarden, but as of now a JSON is overkill.

Timshel:
> or the schema altered if at some point in the future

That's kind of the point that doing this for the four database engine can be a pain ^^.
But I might be biased due to crap like this  https://github.com/dani-garcia/vaultwarden/pull/7265 :sweat_smile:

RaphaelRoumezin:
Another option would be to make a 1:1 model for org settings, such as not to bloat the current org model and be able to expand new fields (and convert the current flagset integer to 4 bools)

PR #7453, opened 2026-07-21
Fix compilation with newer `rust-musl` version

dfunkt:


docker/Dockerfile.j2
```diff
@@ -106,7 +106,7 @@ WORKDIR /app
 
 {% if base == "alpine" %}
 # Environment variables for Cargo on Alpine based builds
-RUN echo "export CARGO_TARGET=${RUST_MUSL_CROSS_TARGET}" >> /env-cargo && \
+RUN echo "export CARGO_TARGET=${TARGET}" >> /env-cargo && \
```
BlackDex: Maybe better to use `CARGO_BUILD_TARGET`?

That is also used by cross-rs if I'm not mistaking.

BlackDex (changes requested):

docker/Dockerfile.j2
BlackDex: Instead of `TARGET`. Because that is also defined in my images.

dfunkt: Changed as requested.

BlackDex (approved):

PR #7460, opened 2026-07-23
Update version to 2026.4.1 in mod.rs

ghost:
the hard-coded version here is lagging behind.


BlackDex:
This is not really an issue, but it will be updated in a different upcoming PR.
The version mentioned here is mainly used by clients to verify if some features can be enabled or not.
And currently this is not used by the clients anymore as far as i can tell.

Thanks though for creating a PR 😄

PR #7465, opened 2026-07-23
SSO: Set verified_at for existing users associated via SSO_SIGNUPS_MATCH_EMAIL

fredrikekre:
When an existing non-SSO account is associated with an SSO login via
`SSO_SIGNUPS_MATCH_EMAIL`, the user's `verified_at` is never set, even though
the provider vouches for the address on every SSO login (new SSO users get
`verified_at` set on signup under the same conditions). Such users keep seeing
the "verify your email" nudge unless they go through the manual email
verification flow.

Two commits:

1. Backfill `verified_at` when associating an existing account, if the provider
   email matches the locally stored one. This mirrors the existing behavior for
   new SSO users.

2. Edit: This commit was dropped following discussion in https://github.com/dani-garcia/vaultwarden/pull/7465#discussion_r3643625402 ~Only set `verified_at` when the provider asserts `email_verified == true`, in
   both the new-user path and the backfill, regardless of
   `SSO_ALLOW_UNKNOWN_EMAIL_VERIFICATION`. That flag still allows logging in
   without the claim, but the account is no longer marked verified on the
   strength of a claim that was never made; the user can still verify through
   the regular email flow. This is a behavioral change for deployments using
   `SSO_ALLOW_UNKNOWN_EMAIL_VERIFICATION`, so this commit can be dropped if that
   isn't wanted. The first commit stands on its own.~

Issue identified by me, patches written by 🤖 and reviewed by me.

### DIFF
--- a/src/api/identity.rs
+++ b/src/api/identity.rs
@@ -337,6 +337,13 @@ async fn sso_login(
                 }
 
                 user.save(conn).await?;
+            } else if user.verified_at.is_none()
+                && user.email == user_infos.email
+                && user_infos.email_verified.unwrap_or(CONFIG.sso_allow_unknown_email_verification())
+            {
+                // The email verification is handled by the provider, backfill for accounts created before SSO
+                user.verified_at = Some(now);
+                user.save(conn).await?;
             }
 
             if user.email != user_infos.email {


src/api/identity.rs
```diff
@@ -307,7 +307,9 @@ async fn sso_login(
             }
 
             let mut user = User::new(&user_infos.email, user_infos.user_name.clone());
-            user.verified_at = Some(now);
+            if user_infos.email_verified == Some(true) {
+                user.verified_at = Some(now);
+            }
```
RaphaelRoumezin: This check here is unnecessary, as it has already been performed above. Plus, this doesnt consider, and thus bypasses the `sso_allow_unknown_email_verification` configuration.

src/api/identity.rs
```diff
@@ -307,7 +307,9 @@ async fn sso_login(
             }
```
RaphaelRoumezin, line 308: check done here

src/api/identity.rs
fredrikekre: This point is reachable also with `user_infos.email_verified == None` if `SSO_ALLOW_UNKNOWN_EMAIL_VERIFICATION` is set. The intention of the second commit is to only set `user.verified_at` in the explicit `user_infos.email_verified == Some(true)` case. My interpretation of the flag was that it would just let users log in but not necessarily that the email was to be regarded as verified. Not setting `user.verified_at` would allow Vaultwarden's regular email verification flow to be used for verification.

RaphaelRoumezin: To me, `SSO_ALLOW_UNKNOWN_EMAIL_VERIFICATION` allows SSO to be used with an IdP that doesn't report user email verification, and blindly trust that the users are verified, and I believe most deployments use this config to bypass the whole email verification flow, when using `SSO_ONLY` with an IdP they trust.
Adding your fix would probably call for an extension of the configuration, such as `SSO_IGNORE_UNKNOWN_EMAIL_VERIFICATION`, or maybe a field `SSO_UNKNOWN_EMAIL_VERIFICATION` that could be set to `reject` (forbid login), `unverified` (still requires verification, your implementation) or `verified` (bypass, current implementation).

fredrikekre: Okay, I will just drop the second commit for now then.

src/api/identity.rs
```diff
@@ -336,6 +336,10 @@ async fn sso_login(
                     user.name = user_name.clone();
                 }
 
+                user.save(conn).await?;
+            } else if user.verified_at.is_none() && user.email == user_infos.email {
```
Timshel, line 340: The previous block rely on the fact that it's a new user (`private_key.is_none()`) and does not check again the verified status. 

But here it's not the case, so I would add something like:
`user_infos.email_verified.unwrap_or(CONFIG.sso_allow_unknown_email_verification())`

fredrikekre: Added.

Issue #5440, opened 2025-01-25
Autofill on page load field - Browser Extensions


Gerardv514:
As discussed in #5388 I am noticing when viewing vault items in the Chrome Browser extension, the field called "Autofill on Page Load" is completely blank. If I open the browser extension and click the + add sign, I see the extension automatically completes this field with a default of "Use Default Setting."

I believe these entries of null/blank happen, when you save a new vault item from the mobile apps, since I can reproduce that on demand. I am specifically using ios mobile app.

**I also just tried to reproduce this using Bitwarden cloud, and it doesn't happen there.**

~~My question is, for~~ all the items that are blank (or probably null), is ~~this~~ caused by adding vault items from ~~some other BW app such as~~ the mobile ios app, ~~or could it be the prompt in a browser after logging into a site for the first time and then clicking yes to save?~~

It may not be an issue now, but seeing how the mobile apps have become more strict with the ways things are saved, could this cause an issue down the line? I am using a sqlite3 database, according to the data folder, can anyone recommend a sqlite3 viewer for Windows and then where in the database can I find this field.

In the below example, you will see how my existing item is saved into the vault which is blank/null for auto fill on page load field. This item was saved either two ways, through ios mobile app or Chrome Extensions using the pop up of "Do you want to save this item," and it was created sometime back in 2023. Yet if I try and add it using + sign (add) in chrome extension if populates this field with default setting.

I am on latest 1.32.7, for the sake of it I did check my test container as well and same result there.

I have actually noticed this null/blank field ever since using Vaultwarden, and in 2021. I actually have an email where I wrote this up to Bitwarden support, but after learning more of the difference of Vaultwarden & Bitwarden apps at the time, I let that discussion fizzle out when they said they couldn't reproduce. In retrospect with more experience now, I should have started the discussion here.

Attempting to look at this in the DB. I created two different test login items, one was created using mobile ios app, and the second was created using the bitwarden chrome browser extension. All I filled in was item name. After saving, On the chrome extension I viewed both test items and I see one that has the Auto Fill on Page load filled out (showing the default option to use in options) and the 2nd one (created from ios app) auto fill on page load is completely blank.

{"autofillOnPageLoad":null,"password":null,"passwordRevisionDate":null,"totp":null,"uris":

{"password":null,"passwordRevisionDate":null,"totp":null,"uris":

Existing item saved into the vault (Chrome Extension)

[...]

PR #7466, opened 2026-07-24
fix: add autofillOnPageLoad and fido2Credentials fields for SDK 0.2.0 compatibility

### LINKED ISSUE #7462: Bitwarden Windows, Latest Version - App dont show any entry
1) Update / Install the latest Windows Bitwarden Version
2) Open Bitwarden
3) No entries are shown, everything is empty

alefsanderribeiro:
## Description

The Bitwarden browser extension v2026.7.0, released on July 23, 2026, ships with SDK internal `0.2.0` which introduced stricter deserialization rules for cipher data. After upgrading, users of self-hosted Vaultwarden can successfully authenticate and sync (HTTP 200), but the vault appears completely empty — no credentials are displayed.

The root cause was identified as two issues in `src/db/models/cipher.rs`:

### 1. Missing fields in login cipher data

The SDK's `CipherLoginModel` gained two new fields that the browser extension's WASM deserializer expects to exist in the JSON response: `autofillOnPageLoad` and `fido2Credentials`. Vaultwarden did not emit these keys, causing the WASM deserializer to panic.

### 2. Strict password history deserialization

The SDK's `CipherPasswordHistoryModel` uses **non-optional** `String` fields for `password` and `lastUsedDate`. Entries with null-like password values could trigger deserialization failures.

## Changes

- **autofillOnPageLoad**: Now emitted as `null` when not set by the user
- **fido2Credentials**: Now emitted as `[]` (empty array) when the user has no FIDO2 credentials
- **passwordHistory**: Non-string password values are replaced with empty string `""` instead of dropping the entry

## Compatibility

- ✅ **Browser extension v2026.7.0**: Fixed — vault now displays correctly
- ✅ **CLI v2026.7.0**: Works (CLI does not use WASM SDK)
- ✅ **Web Vault**: No regression
- ✅ **Mobile apps**: No regression

## Related

- Closes #7462
- Ref: https://github.com/bitwarden/clients/issues/22013
- SDK update PR: https://github.com/bitwarden/clients/pull/21233
- SDK CipherLoginModel: https://github.com/bitwarden/sdk-internal/blob/main/crates/bitwarden-api-api/src/models/cipher_login_model.rs

### DIFF
--- a/src/db/models/cipher.rs
+++ b/src/db/models/cipher.rs
@@ -233,6 +233,11 @@ impl Cipher {
                 d.into_iter()
                     .filter_map(|d| match d.data.get("password") {
                         Some(p) if p.is_string() => Some(d.data),
+                        Some(_) => {
+                            let mut data = d.data;
+                            data["password"] = json!("");
+                            Some(data)
+                        }
                         _ => None,
                     })
                     .map(|mut d| {
@@ -279,6 +284,14 @@ impl Cipher {
             // Check if `passwordRevisionDate` is a valid date, else convert it
             if let Some(pw_revision) = type_data_json["passwordRevisionDate"].as_str() {
                 type_data_json["passwordRevisionDate"] = json!(validate_and_format_date(pw_revision));
+            }
+
+            // SDK 0.2.0 compatibility: Bitwarden extension v2026.7.0 expects these fields
+            if type_data_json.get("autofillOnPageLoad").is_none() {
+                type_data_json["autofillOnPageLoad"] = Value::Null;
+            }
+            if type_data_json.get("fido2Credentials").is_none() {
+                type_data_json["fido2Credentials"] = Value::Array(Vec::new());
             }
         }
 


dani-garcia:
I don't believe this is the cause of the error, these two fields are marked as optional on the SDK: https://github.com/bitwarden/sdk-internal/blob/db3ed96daf9a17b9620cc97d5e1df3bd6ceb97cc/crates/bitwarden-vault/src/cipher/login.rs#L296-L298

Setting the password history password to an empty string is also more likely to cause issues than to fix them.

Please try the latest `testing` image to see if the issue is solved, until then I'll close this PR.

Gerardv514:
@dani-garcia just a fyi despite the auto fill on page load being optional, I had brought this up a very long time ago. 

When an entry is added to the vault via iOS app, when you view that entry in a browser extension the data is blank. This means that you have to go into each vault item and set this option if it was created by iOS app. 

On the other hand if you create an item in the browser extension auto fill on page load is automatically set to a default setting , to use the option in settings. 

*not saying this is causing an issue, but items being added from iOS app to a vault leads to many entries in the vault with this field blank. This leads to having to touch every one of these items separately from iOS app and fix the items to show default. 

https://github.com/dani-garcia/vaultwarden/issues/5440

BlackDex:
I think i mentioned it there too, but that sounds like a client side issue. Unless there is a default if not provided by the client during save/create

Gerardv514:
I believe it was mentioned that Bitwarden server returns more values than Vaultwarden does. Although I'm not an expert in the deeper workings of all this. 

https://github.com/dani-garcia/vaultwarden/issues/5440#issuecomment-2613927188

Gerardv514:
Also it's not reproducible via Bitwarden cloud.

Issue #7361, opened 2026-06-22
Client 2026.5.1 does not load entries


fzenger7:
### Prerequisites

- [x] I have searched the existing **Closed _AND_ Open** [Issues](https://github.com/dani-garcia/vaultwarden/issues?q=is%3Aissue%20) **_AND_** [Discussions](https://github.com/dani-garcia/vaultwarden/discussions?discussions_q=)
- [x] I have searched and read the [documentation](https://github.com/dani-garcia/vaultwarden/wiki/)

### Your environment (Generated via diagnostics page)

* Vaultwarden version: v1.36.0
* Web-vault version: v2026.4.1
* OS/Arch: linux/x86_64
* Running within a container: true (Base: Debian)
* Database type: SQLite
* Database version: 3.51.3
* Uses config.json: false
* Uses a reverse proxy: true
* IP Header check: true (X-Real-IP)
* Internet access: true
* Internet access via a proxy: false
* DNS Check: true
* Browser/Server Time Check: true
* Server/NTP Time Check: true
* Domain Configuration Check: true
* HTTPS Check: true
* Websocket Check: true
* HTTP Response Checks: true

### Config & Details (Generated via diagnostics page)

<details><summary>Show Config & Details</summary>

**Config:**

```json
{
  "_duo_akey": null,
  "_enable_duo": false,
  "_enable_email_2fa": true,
  "_enable_smtp": true,
  "_enable_yubico": true,
  "_icon_service_csp": "",
  "_icon_service_url": "",
  "_ip_header_enabled": true,
  "_max_note_size": 10000,
  "_smtp_img_src": "***:",
  "admin_ratelimit_max_burst": 3,
  "admin_ratelimit_seconds": 300,
  "admin_session_lifetime": 20,
  "admin_token": "***",
  "allowed_connect_src": "",
  "allowed_iframe_ancestors": "",
  "attachments_folder": "data/attachments",
  "auth_request_purge_schedule": "30 * * * * *",
  "authenticator_disable_time_drift": false,
  "data_folder": "data",
  "database_conn_init": "",
  "database_idle_timeout": 600,
  "database_max_conns": 10,
  "database_min_conns": 2,
  "database_timeout": 30,
  "database_url": "***************",
  "db_connection_retries": 15,
  "disable_2fa_remember": false,
  "disable_admin_token": false,
  "disable_icon_download": false,
  "dns_prefer_ipv6": false,
  "domain": "*****://***************",
[...]

PR #7467, opened 2026-07-24
Strip null values from optional Login cipher fields

0xjjjjjj:
## Description

Fixes vault-load crashes on Bitwarden 2026.5+ clients caused by explicit `null` values on optional Login cipher fields.

Refs: #7361, [bitwarden/clients#21135](https://github.com/bitwarden/clients/issues/21135). Same class of bug as #7308 (SSH-key) but for the Login (`atype=1`) type.

## Symptom

Angular UI hangs on the skeleton loading screen after successful unlock. Console:

```
Unhandled error in angular Error: Error: invalid type: JsValue(Object({...Login cipher shape...})), expected a string
    je bitwarden_wasm_internal_bg.js:6830
    ...
    getStore browser-local-storage.service.ts:78
```

Serde aborts on the first failing cipher, so a single bad item blanks the entire vault.

## Root cause

The 2026.5+ `sdk-internal` WASM deserializer routes JS `null` values into `EncString::deserialize` (a custom `Deserialize` impl expecting a string) rather than short-circuiting `Option<T>` to `None` the way `serde-json` does. Vaultwarden emits explicit `null` values for optional Login fields; upstream Bitwarden's server serializes them with `#[serde(skip_serializing_if = "Option::is_none")]` and always omits the key instead. Missing key becomes `None` and works. Present as null crashes the new decoder.

Wire schema for reference: [`CipherLoginModel`](https://github.com/bitwarden/sdk-internal/blob/main/crates/bitwarden-api-api/src/models/cipher_login_model.rs), every optional field is `#[serde(skip_serializing_if = "Option::is_none")]`.

## Fix

Post-process `type_data_json` for `atype == 1` (Login) ciphers to strip `null`-valued keys before emit, matching upstream's shape. Covers: `password`, `username`, `totp`, `autofillOnPageLoad`, `fido2Credentials`, `passwordRevisionDate`, `uris`.

The legacy `uri` (singular) field, unconditionally set to `Value::Null` at line 261 as backwards-compat for old mobile clients per the existing comment, is deliberately left untouched.

## Verification

Tested against a real vault of ~1000 Login ciphers, 703 of which (71%) hold at least one `null` on the affected keys:

| Binary | Result |
|---|---|
| Stock `1.36.0` | Vault fails to load, console shows the trace above |
| Patched (this PR) | Vault loads cleanly, Firefox extension v2026.5.x, no console errors |

Rollback / reapply cycle confirms the patch is the necessary and sufficient change.

## Scope

- Read-path only, no schema change, no migration, trivially revertible.
- Login (`atype=1`) only. Card (`atype=3`) and Identity (`atype=4`) may share the same root cause but are not covered here (no reports found, would prefer a separate PR once reproduced).

### DIFF
--- a/src/db/models/cipher.rs
+++ b/src/db/models/cipher.rs
@@ -280,6 +280,23 @@ impl Cipher {
             if let Some(pw_revision) = type_data_json["passwordRevisionDate"].as_str() {
                 type_data_json["passwordRevisionDate"] = json!(validate_and_format_date(pw_revision));
             }
+
+            // Strip `null`s from optional Login fields — newer WASM clients panic (see #7361).
+            if let Value::Object(ref mut map) = type_data_json {
+                for key in [
+                    "password",
+                    "username",
+                    "totp",
+                    "autofillOnPageLoad",
+                    "fido2Credentials",
+                    "passwordRevisionDate",
+                    "uris",
+                ] {
+                    if map.get(key).is_some_and(Value::is_null) {
+                        map.remove(key);
+                    }
+                }
+            }
         }
 
         // Fix secure note issues when data is invalid


0xjjjjjj:
Closing. #7434 (already merged to main) is the actual fix. My patched binary happened to work because it was built from main, not because of the null-strip in this PR. Verified by rebuilding from pure upstream/main without my change and confirming the vault loads. Apologies for the noise.

PR #7477, opened 2026-07-25
Add `pm-34171-card-scanner` feature flag

bdd:
Introduced in 2026.4.1 Mobile releases (Android and iOS), enables use of device camera to autofill a card type item.

N.B. This feature relies on Google ML Kit. F-Droid policy disallows Google ML Kit. Because of this, F-Droid builds do not ship with card scanner feature. (See https://github.com/bitwarden/android/pull/6890)

PR #7478, opened 2026-07-25
Add `pm-32009-new-item-types` feature flag

bdd:
Introduced in 2026.4.0 for all clients, enables new item types bank account, driver's license, and passport.

Tested with 1.37.0's web build 2026.6.4 and Android 2026.6.1(21713) from Google Play Store.

`EXPERIMENTAL_CLIENT_FEATURE_FLAGS=desktop-ui-migration-milestone-1,desktop-ui-migration-milestone-2,desktop-ui-migration-milestone-3,desktop-ui-migration-milestone-4,pm-34171-card-scanner,pm-32009-new-item-types`

Screenshot From 2026-07-24 17-17-15

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -393,6 +393,7 @@
 ## - "cxp-import-mobile": Enable the import via CXP on iOS (Clients >= 2025.9.2)
 ## - "cxp-export-mobile": Enable the export via CXP on iOS (Clients >= 2025.9.2)
 ## - "pm-30529-webauthn-related-origins":
+## - "pm-32009-new-item-types": Enable new item types: Bank Account, Driver's License, and Passport (Clients >= 2026.4.0)
 ## - "desktop-ui-migration-milestone-1": Special feature flag for desktop UI (Desktop >= 2026.2.0)
 ## - "desktop-ui-migration-milestone-2": Special feature flag for desktop UI (Desktop >= 2026.2.0)
 ## - "desktop-ui-migration-milestone-3": Special feature flag for desktop UI (Desktop >= 2026.2.0)
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -266,6 +266,9 @@ pub struct CipherData {
     Card = 3,
     Identity = 4,
     SshKey = 5
+    BankAccount = 6
+    DriversLicense = 7
+    Passport = 8
     */
     pub r#type: i32,
     pub name: String,
@@ -278,6 +281,9 @@ pub struct CipherData {
     card: Option<Value>,
     identity: Option<Value>,
     ssh_key: Option<Value>,
+    bank_account: Option<Value>,
+    drivers_license: Option<Value>,
+    passport: Option<Value>,
 
     favorite: Option<bool>,
     reprompt: Option<i32>,
@@ -510,6 +516,9 @@ pub async fn update_cipher_from_data(
         3 => data.card,
         4 => data.identity,
         5 => data.ssh_key,
+        6 => data.bank_account,
+        7 => data.drivers_license,
+        8 => data.passport,
         _ => err!("Invalid type"),
     };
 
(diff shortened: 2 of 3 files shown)


tessus:
@dani-garcia @BlackDex sorry for using this PR, but in this case it might be better than using a discussion, otherwise please tell me to do so.

I've noticed that the number of open PRs increased quite a lot lately. I guess mostly because of AI generated ones. I know from experience that a great number of PRs in a project can be daunting and thus I always try to either close the ones that are problematic or merge low hanging fruits like this one or #7477. I hope you are not totally annoyed by my comment, but I also know from experience that sometimes a little reminder can be helpful.

BlackDex:
> or merge low hanging fruits like this one

I understand the gist, but there is a time effort too. But for this specific item, that isn't really low hanging fruit in my opinion. This item needs to be tested and checked with multiple clients, validated it all looks correct etc.. etc...

And i do try to merge simple or needed PR's from time to time at least once a week of possible, sometimes two.

tessus:
> But for this specific item, that isn't really low hanging fruit in my opinion. This item needs to be tested and checked with multiple clients, validated it all looks correct etc.. etc...

Unfortunately I agree to disagree. For any other "normal" PR I would agree with you, but this PR adds a feature flag, which in fact is actually the best way for people to test it:

1) in the test image
2) behind a feature flag (which is not set by default)

P.S.: Btw, I've been running both flags for about 3 weeks now. I haven't found any issues yet.

BlackDex (approved):

BlackDex:
> Unfortunately I agree to disagree. For any other "normal" PR I would agree with you, but this PR adds a feature flag, which in fact is actually the best way for people to test it:

If we only would have users which use testing and report back systematically, then that would be cool.
But it isn't. Also, i needed to test what happens if i remove the feature flag afterwards, and have these items in there already and edit them. Stuff like that needs to be checked and validated.

But, besides that, i had some time this week to test it, and seems to work great.
Thanks @bdd

tessus:
> If we only would have users which use testing and report back systematically, then that would be cool.
But it isn't. 

To be fair, how can someone test new things with the test image, if they are not in the test image? It seems to me that the test image is not a test image, but a more current stable. 

The only way for people to currently test something is by retrieving PRs and compile vw. It is not a problem for me, but for others it might be. For more people to test things, the main branch should be updated more quickly. Or maybe there is a need for a `dev` branch and a real test image should be created from it. I don't know what the best soution is. I can only say that with the current process it is rather unlikely that a lot of people test something. Unless you have a few nerds (like me) who compile manually and test PRs.

BlackDex:
@tessus, I'm not going to just merge anything that looks right on sight. Some stuff i would like to test first before I'll approve and merge. That might be a bit too much safety for some items. Also, after i have tested it and approved and then merged people can start test and use it too.

tessus:
@BlackDex I understand what you are saying. I merely pointed out that this is the reason why people don't test and report back. It's because you do the testing and merge only if it is stable. there is no need to test the testing image, except maybe if it fixed a bug for users.

BlackDex:
Well, i disagree, but that is fine. In my opinion maintainers need to check and validate, and if i have some ideas which might fail or need attention i want to test it. This PR could be a potential item to breaks stuff with other clients, and if reverting would still break, that would also be very bad for testing users.

Also, there are other PR which have been merged and contained issues (even if checked before) and i do not fully check and test all PR's, even if you might think so.

tessus:
> In my opinion maintainers need to check and validate,

I agree with you and I am not saying you shouldn't do that. 

I believe we might be talking past each other. For me a test image is used to test stuff that can be reverted for the next test image. A test image is allowed to break stuff - in fact it is expected. A test image is usually cut from a dev branch. When the code has been validated and tested and does not need reverting, it is merged to the master branch, which can then be used for pre-prod images. If it breaks stuff, the image fulfilled its purpose and the code is fixed or reverted. Only with a "real" test image from a dev branch you get more non-dev people to test it. But maybe we just have different views on dev processes.
I am also not asking to change the dev process of this project.

The only thing I wanted to say was that I understand why people do not test like you want them to.

PR #7483, opened 2026-07-25
Fix sendmail executable permission check

### LINKED ISSUE #7445: USE_SENDMAIL rejects valid owner/group-only executable commands

p-boenisch:
Fixes #7445

The previous permission check inspected the Unix mode bits directly and
could reject commands that were executable by the current process.

Use `nix::unistd::access` with `AccessFlags::X_OK` to check whether the
sendmail command is executable by the running process.

Checks performed:

- `cargo fmt --check`
- `cargo check --features sqlite`
- `cargo test --features sqlite,mysql,postgresql`
- `cargo clippy --features sqlite,mysql,postgresql -- -D warnings`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -935,6 +935,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
 
 [[package]]
+name = "cfg_aliases"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f079e83a288787bcd14a6aea84cee5c87a67c5a3e660c30f557a3d24761b3527"
+
+[[package]]
 name = "chacha20"
 version = "0.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3227,6 +3233,18 @@ name = "ndk-context"
 version = "0.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "27b02d87554356db9e9a873add8782d4ea6e3e58ea071a9adb9a2e8ddb884a8b"
+
+[[package]]
+name = "nix"
+version = "0.31.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cf20d2fde8ff38632c426f1165ed7436270b44f199fc55284c38276f9db47c3d"
+dependencies = [
+ "bitflags 2.13.1",
+ "cfg-if",
+ "cfg_aliases",
+ "libc",
+]
 
 [[package]]
 name = "nom"
@@ -5867,6 +5885,7 @@ dependencies = [
  "macros",
  "mimalloc",
  "moka",
+ "nix",
  "num-derive",
  "num-traits",
  "opendal",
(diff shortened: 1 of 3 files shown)


stefan0xC:
As explained in the thread where I've suggested this fix https://github.com/dani-garcia/vaultwarden/issues/7445#issuecomment-5017775252 this is not actually sufficient. Can you update your PR to use the access check from the suggested `nix` crate?

p-boenisch:
Thanks for the correction. I've replaced the mode-bit check with `nix::unistd::access` using `AccessFlags::X_OK` and added `nix` as a Unix-only dependency with the `fs` feature. All tests and Clippy checks pass locally.

BlackDex (approved):

Issue #7493, opened 2026-07-28
Return non-200 HTTP status code when admin panel is disabled


spyesx:
### Prerequisites

- [x] I have searched the existing **Closed _AND_ Open** [Issues](https://github.com/dani-garcia/vaultwarden/issues?q=is%3Aissue%20) **_AND_** [Discussions](https://github.com/dani-garcia/vaultwarden/discussions?discussions_q=)
- [x] I have searched and read the [documentation](https://github.com/dani-garcia/vaultwarden/wiki/)

### Vaultwarden Support String

When the admin panel is disabled because the `ADMIN_TOKEN` is not configured, navigating to the `/admin` path returns a plain text message:

> The admin panel is disabled, please configure the 'ADMIN_TOKEN' variable to enable it

While this is useful for a human, this response currently returns an HTTP status code of 200 OK. Thus, monitoring tools are misled by the actual status.

The [`admin_disabled`](https://github.com/dani-garcia/vaultwarden/blob/main/src/api/admin.rs#L100) route in `src/api/admin.rs` should be updated to return a `503 NOT IMPLEMENTED`. A `403` could work as well but this is not an issue from the client if the admin is disabled.

### Vaultwarden Build Version

v1.37.0

### Deployment method

Official Container Image

### Reverse Proxy

traefik

### Host/Server Operating System

Linux

### Clients

Web Vault

### Steps To Reproduce

```
➜  ~ curl -I https://OBFUSCATED/admin
HTTP/2 200
[...]
```

### Expected Result

```
➜  ~ curl -I https://OBFUSCATED/admin
HTTP/2 503
[...]
```

### Actual Result

```
➜  ~ curl -I https://OBFUSCATED/admin
HTTP/2 200
[...]
```

### Logs

```text

```

PR #7494, opened 2026-07-28
refactor(admin): return 503 status code when admin panel is disabled

spyesx:
This PR addresses #7493 

Currently, when the Vaultwarden instance does not have the `ADMIN_TOKEN` configured, navigating to the `/admin` path  returns a plain text message with an HTTP 200.

This PR changes the `HTTP 200` by an `HTTP 503 service unavailable` to allow health checks and other monitoring systems to get the actual status of the Admin panel.

### DIFF
--- a/src/api/admin.rs
+++ b/src/api/admin.rs
@@ -98,8 +98,11 @@ static CAN_BACKUP: LazyLock<bool> = LazyLock::new(|| ACTIVE_DB_TYPE.get().is_som
 static CAN_BACKUP: LazyLock<bool> = LazyLock::new(|| false);
 
 #[get("/")]
-fn admin_disabled() -> &'static str {
-    "The admin panel is disabled, please configure the 'ADMIN_TOKEN' variable to enable it"
+fn admin_disabled() -> (Status, &'static str) {
+    (
+        Status::ServiceUnavailable,
+        "The admin panel is disabled, please configure the 'ADMIN_TOKEN' variable to enable it"
+    )
 }
 
 const COOKIE_NAME: &str = "VW_ADMIN";


BlackDex:
503 is not the right response for this.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/503

If anything it should either be a 404 or just let it stay as-is.

spyesx:
Hello @BlackDex :)

Well, I guess it depends on where we stand from.

`HTTP 404` means Not Found. The endpoint is found here, the functionality exists. It is only disabled.

`HTTP 503` means the endpoint exists, but it is currently out of service or unconfigured.

The message is "The admin panel is disabled" not "The admin panel is not found". It wouldn't make sense to return a 404 with some text in it.

4xx errors are Client response error. Here, the client can't change anything to fix the error. No better path, no auth... The problem comes from a misconfiguration of the server, thus a Server error response in 5xx.

stefan0xC:
> The problem comes from a misconfiguration of the server, thus a Server error response in 5xx.

It's not a misconfiguration if the endpoint is intentionally disabled. As explained in the link above "503 should be used for temporary conditions", which a disabled `/admin` endpoint is not.

spyesx:
@stefan0xC yes :) I understood it the same way as well, temporary.

The token is missing from the configuration. By security, the admin is disabled. This is to not allow access without token. Is it common for an instance owner to deliberately disable the admin? Or is this situation temporary until the config is fixed?

I'm ok to change to whatever everyone agrees. I'm just trying to make the message clear.

I would also agree with a `200 OK`, because the endpoint is found, ok. But then `/admin` should serve a `401 Unauthorized` when the feature is available then? Or the opposite ahah

The only thing is we can't keep 200 when the feature is available and when it's not. It cannot be monitored. And 404 doesn't fit here. The API replies in every cases.

Which one do you guys prefer?

stefan0xC:
Why would a 404 not fit?

spyesx:
Because 404 means Not Found. The problem isn't that the endpoint does not exist. It exists :) It even replies the server is missing something to provide the feature. The problem is the config and its `ADMIN_TOKEN`.

RouL:
I also don't feel too happy with a 404. Doesn't feel right, also doesn't feel (completely) wrong. But for sure it would be better than 200. Another option might be 403, depending on how you're interpreting the setting. If you see it from a "I don't allow /admin to anybody" point of view, it could fit.

But 200 is just plain wrong, I think we should be able to agree on that one. :)

spyesx:
Should we need a poll ? :)

1. `200` when no `ADMIN_TOKEN` is set + `401` when `ADMIN_TOKEN` is set
2. `200` when no `ADMIN_TOKEN` is set + `403` when `ADMIN_TOKEN` is set
3. `503` when no `ADMIN_TOKEN` is set + `200` when `ADMIN_TOKEN` is set

Trying to move on with the PR so my monitoring can do its job properly.

stefan0xC:
> Trying to move on with the PR so my monitoring can do its job properly.

You know that you can override this path using a reverse proxy? 
```nginx
location /admin {
    return 404;
}
```

BlackDex:
Well, `503` is definitely not the right status, it is used for (temporary) outage, and not a disable environment.
This could cause weird issues also if for example Vaultwarden isn't reachable at all by a reverse proxy, then it would also generate a `503`, then what is it, is Vaultwarden down or is the admin panel disabled.

>[503 Service Unavailable](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/503)
>
>    The server is not ready to handle the request. Common causes are a server that is down for maintenance or that is overloaded. Note that together with this response, a user-friendly page explaining the problem should be sent. This response should be used for temporary conditions and the [Retry-After](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Retry-After) HTTP header should, if possible, contain the estimated time before the recovery of the service. The webmaster must also take care about the caching-related headers that are sent along with this response, as these temporary condition responses should usually not be cached.

From my point of view this should be the correct way as that is what the response codes are intended to be for.

|ADMIN_TOKEN|DISABLE_ADMIN_TOKEN|Authenticated|Status|
|:-------------------:|:----------------------------------:|:----------------:|:-------:|
| ✔️ | ❌ | ❌ | [401](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/401) |
| ✔️ | ❌ | ✔️ | [200](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/200) |
| ❌ | ❌ | - | [403](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/403) or [404](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/404) |
| ❌ | ✔️ | - | [200](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/200) |

`403` means forbidden, and that is what the user intended with disabling it.
I would actually prefer `404` to at least try and mitigate simple scripts trying to search for admin panels or what ever to just get a `Not Found` response instead.

Our main reason for using `200` is because we just provide a simple message that it is disabled.
Also, most monitoring tools are also able to read the body response and if it matches or does not matches a subset of that content, like search for `The admin panel is disabled`, and if that doesn't exists, alert.

RouL:
I totally agree with @BlackDex here, even though I prefer 403, but that is a matter of preference, not of correctnes. And I could totally live with a 404 as well.

The only thing I feel I have to comment on. Not as critique, but more as a short addition.
> Our main reason for using 200 is because we just provide a simple message that it is disabled.
Also, most monitoring tools are also able to read the body response and if it matches or does not matches a subset of that content, like search for The admin panel is disabled, and if that doesn't exists, alert.

While it is true, that a monitoring system can (usually) check the body, it is also true, that you can send a message as well in the body of a 403 / 404.

BlackDex:
> While it is true, that a monitoring system can (usually) check the body, it is also true, that you can send a message as well in the body of a 403 / 404.

I know, that is fine too, but it's just not the default of the framework.
But then still, you could match text.

forkrails:
Whole point of disabling admin panel is for security. Anything that is not 404 tells scanners, attackers, etc that something lives here. 404 should be default. Users can just use their proxy to return a different status if needed.

spyesx:
> Whole point of disabling admin panel is for security. Anything that is not 404 tells scanners, attackers, etc that something lives here. 404 should be default. Users can just use their proxy to return a different status if needed.

Today we have 200 only in both cases so a reverse proxy can't help much. Additionally, `/admin` is documented everywhere so an attacker knows what's here and he can read the body as well, which says "config error". Pushing the ball on this logic would tell an attacker he should focus on this instance because it is misconfigured then. The HTTP code is debatable but not the justification.

BlackDex:
It doesn't matter what you return with or without any status code or body. If an attacker knowns it's Vaultwarden and really wants to attack a Vaultwarden instance, then they probably also known the status codes which are returned depending on which config.

A 404 in my opinion is the best option, since 1, you have disabled the interface, and it should just return a 404 since it's not available. And it will keep simple script-kiddies and other tools a bit away if they just check the 404 status and continue as if nothing is available.

RouL:
> Whole point of disabling admin panel is for security. Anything that is not 404 tells scanners, attackers, etc that something lives here. 404 should be default. Users can just use their proxy to return a different status if needed.

Actually, if the admin panel is disabled, it doesn't really matter if an attacker detects, that in theory, with the right config, something could live there. Also, as mentioned by @spyesx, if the attacker detects vaultwarden, they know the location of the admin panel anyways.

Sending 404 gives no real security gain. What it does is making a few "simple script kiddies" be less interested, as mentioned by @BlackDex.

PR #7500, opened 2026-07-28
Add configurable user verification policy for WebAuthn 2FA

### LINKED ISSUE #7437: WebAuthn times out with PIN-configured CTAP 2.1 authenticator when user verification is discouraged
## Steps to reproduce

1. Configure a FIDO2 PIN on a Swissbit iShield Key 2 Pro MIFARE.
2. Open the Web Vault and navigate to Settings > Security > Two-step login > Passkey.
3. Attempt to register the security key.
4. No PIN prompt appears and the WebAuthn operation eventually times out.
5. Resetting the FIDO2 applet and registering the key before configuring a PIN works, but this is not a complete workaround because authentication is also initiated with `userVerification: discouraged`.

The same Vaultwarden installation successfully registers YubiKeys that already have a FIDO2 PIN. A PIN prompt appears for those keys.

The iShield also registers and authenticates successfully on webauthn.io with its PIN configured.

## Expected result

Vaultwarden should interoperate with the authenticator: either the non-UV operation should complete successfully, or user verification should be requested so the browser presents a PIN prompt.

## Actual result

With a FIDO2 PIN configured, no PIN dialog appears and the operation times out. Registration works only while no PIN is configured.

## Authenticator information

```text
Product:              Swissbit iShield Key 2 Pro MIFARE
Device version:       1.1.0
FIDO version:         v1.4.0-0-gd69b47b
AAGUID:               7787a482-13e8-4784-8a06-c7ed49a7aaf4
Supported protocols:  U2F, CTAP 2.0, CTAP 2.1
clientPin:            yes
alwaysUv:             no
makeCredUvNotRqd:     yes
```

The authenticator advertises `makeCredUvNotRqd: yes`. Therefore, with Vaultwarden requesting `userVerification: discouraged`, Chrome does not present a PIN dialog. The resulting non-UV `makeCredential` operation does not complete.

## Confirmed diagnostic workaround

I built Vaultwarden 1.36.0 with only the WebAuthn user-verification policy changed from `discouraged` to `preferred` for both registration and authentication:

- registration challenge: `Discouraged_DO_NOT_USE` > `Preferred`
- stored registration state: `"discouraged"` > `"preferred"`
- authentication challenge: `Discouraged_DO_NOT_USE` > `Preferred`
- stored authentication state: `"discouraged"` > `"preferred"`

With this test build:

1. Chrome presents the FIDO2 PIN dialog.
2. After entering the PIN and touching the key, registration completes.
3. The key appears in the WebAuthn security-key list.
4. Authentication in a new Incognito window also succeeds with PIN and touch.

No other source or configuration changes were made.

This indicates that Vaultwarden's explicit `userVerification: discouraged` policy triggers the incompatible path. The underlying issue may be in the authenticator firmware, Windows WebAuthn, or their interaction, but allowing `preferred` provides a working interoperability option.

Would the maintainers consider making the WebAuthn 2FA user-verification policy configurable, while keeping `discouraged` as the default for backward compatibility? I would be willing to prepare a PR once the preferred approach and configuration scope are agreed upon.

## Environment

- Vaultwarden: 1.36.0
- Successful diagnostic build: 1.36.0-uv-preferred-test
- Web Vault: 2026.4.1
- Server: Debian 13, Linux x86_64
- Database: SQLite 3.51.3
- Deployment: built from source by the Proxmox VE Community Script
- Reverse proxy: none; Vaultwarden serves HTTPS directly
- Client: Web Vault
- Browser: Google Chrome 150.0.7871.124

* Vaultwarden version: v1.36.0-uv-preferred-test
* Web-vault version: v2026.4.1
* OS/Arch: linux/x86_64
* Running within a container: false (Base: Not applicable)
* Database type: SQLite
* Database version: 3.51.3
* Uses config.json: true
* Uses a reverse proxy: false
* Internet access: true
* DNS Check: true
* Server/NTP Time Check: true
* HTTPS Check: true
* Websocket Check: true
* HTTP Response Checks: true

**Config:**

```json
{
  "_duo_akey": "***",
  "_enable_duo": false,
  "_enable_email_2fa": true,
  "_enable_smtp": true,
  "_enable_yubico": true,
  "_icon_service_csp": "",
  "_icon_service_url": "",
  "_ip_header_enabled": true,
  "_max_note_size": 10000,
  "_smtp_img_src": "***:",
  "admin_session_lifetime": 20,
  "admin_token": "***",
  "allowed_connect_src": "",
  "attachments_folder": "/opt/vaultwarden/data/attachments",
  "data_folder": "/opt/vaultwarden/data",
  "database_conn_init": "",
  "database_idle_timeout": 600,
  "database_max_conns": 10,
  "database_min_conns": 2,
  "database_timeout": 30,
  "database_url": "********************************",
  "db_connection_retries": 15,
  "disable_admin_token": false,
  "dns_prefer_ipv6": false,
  "domain": "*****://*******************",
  "domain_origin": "*****://*******************",
  "domain_path": "",
  "domain_set": true,
  "duo_host": null,
  "duo_ikey": null,
  "duo_skey": null,
  "duo_use_iframe": false,
  "email_attempts_limit": 3,
  "email_change_allowed": true,
  "email_expiration_time": 600,
  "email_token_size": 6,
  "enable_db_wal": true,
  "enable_websocket": true,
  "events_days_retain": null,
  "extended_logging": true,
  "helo_name": null,
  "hibp_api_key": null,
  "icon_blacklist_regex": null,
  "icon_cache_folder": "/opt/vaultwarden/data/icon_cache",
  "icon_cache_negttl": 259200,
  "icon_cache_ttl": 2592000,
  "icon_download_timeout": 10,
  "icon_redirect_code": 302,
  "icon_service": "internal",
  "invitations_allowed": true,
  "log_file": null,
  "log_level": "info",
  "org_attachment_limit": null,
  "org_creation_users": "",
  "org_events_enabled": false,
  "org_groups_enabled": false,
  "purge_incomplete_sso_auth": "0 20 0 * * *",
  "push_enabled": false,
  "reload_templates": false,
  "rsa_key_filename": "/opt/vaultwarden/data/rsa_key",
  "sendmail_command": null,
  "sends_allowed": false,
  "sends_folder": "/opt/vaultwarden/data/sends",
  "show_password_hint": true,
  "signups_allowed": false,
  "signups_verify": true,
  "smtp_auth_mechanism": "\"Login\"",
  "smtp_debug": false,
  "smtp_embed_images": true,
  "smtp_explicit_tls": null,
  "smtp_from_name": "*********",
  "smtp_host": "********************",
  "smtp_password": "***",
  "smtp_port": 587,
  "smtp_security": "starttls",
  "smtp_ssl": null,
  "smtp_timeout": 15,
  "smtp_username": "***********************",
  "sso_allow_unknown_email_verification": false,
  "sso_audience_trusted": null,
  "sso_authority": "",
  "sso_callback_path": "*****://************************************************",
  "sso_client_id": "",
  "sso_client_secret": "***",
  "sso_debug_tokens": false,
  "sso_enabled": false,
  "sso_master_password_policy": null,
  "sso_only": false,
  "sso_pkce": true,
  "sso_scopes": "email profile",
  "templates_folder": "/opt/vaultwarden/data/templates",
  "tmp_folder": "/opt/vaultwarden/data/tmp",
  "use_sendmail": false,
  "use_syslog": false,
  "user_send_limit": null,
  "web_vault_enabled": true,
  "web_vault_folder": "/opt/vaultwarden/web-vault",
  "yubico_client_id": "51567",
  "yubico_secret_key": "***",
  "yubico_server": null
}
```

</details>

all-solutions:
## Description

This adds an optional server setting to use `preferred` user verification for WebAuthn 2FA registration and authentication.

The setting is available through the admin panel as:

> Set WebAuthn 2FA user verification to preferred

It can also be configured using:

```env
WEBAUTHN_2FA_USER_VERIFICATION=true
```

The default remains `false`, preserving the existing `discouraged` behavior.

When enabled, `preferred` is applied consistently to both the browser challenge and the server-side WebAuthn state during registration and authentication.

This improves compatibility with authenticators such as the Swissbit iShield Key 2 Pro when a PIN is configured and other WebAuthn credentials are already registered.

Closes #7437

## Testing

Tested with:

- Swissbit iShield Key 2 Pro with a configured FIDO2 PIN
- YubiKeys with configured FIDO2 PINs
- WebAuthn credential registration and authentication
- Both `discouraged` and `preferred`
- Configuration persistence across restarts
- Native Debian 13 installation
- Docker deployment

## Admin panel

image


BlackDex:
I'm against this, as Bitwarden does not provide the same option in anyway.
This might cause issues with their clients in ways we can not predict.

I would like to keep as close as possible to Bitwarden's way of working for these kind of items.

all-solutions:
Thanks for the feedback. I understand the reasoning behind staying as close as possible to Bitwarden's behavior, especially when client compatibility is involved.

For me, this is mainly an interoperability issue rather than a request to change Vaultwarden's default behavior. The iShield Key 2 Pro simply does not work reliably with `userVerification: discouraged`, while `preferred` fixes both registration and authentication.

The iShield is still relatively new, but it is also one of the more advanced CTAP 2.1 authenticators currently available. Other vendors are also moving to newer generations of security keys, so I would not be surprised if similar cases show up with additional devices over time.

That is also why I intentionally kept `discouraged` as the default in the PR. Existing installations would behave exactly as before, and only administrators who actually need `preferred` would enable it.

For my own setup, this is not a blocker anymore. I have a small patch script that reapplies the change after Vaultwarden updates, and that has already worked without issues during an update. So I am fine either way.

My main reason for submitting the PR was that I expect more users to run into this eventually, and an optional setting would give them a clean workaround without changing the behavior for anyone else.

The code change is quite small, and in my testing the regular Bitwarden clients continue to work normally with `preferred` enabled.

So even if you do not want to merge it at this point, I would appreciate keeping it in mind in case this starts appearing with more authenticators.

all-solutions:
A small update from my side, since there are now a few more data points around this.

In addition to the original iShield case, another user has reported a similar interoperability problem with a Google Titan Security Key when `userVerification` is set to `discouraged`.

I also tested the original iShield/YubiKey scenario directly against Bitwarden Cloud and was able to reproduce the same behavior there:

- iShield registers successfully when no other WebAuthn credential exists
- with an existing YubiKey credential, the iShield registration times out
- removing the YubiKey makes the iShield registration work again

I have reported this upstream to Bitwarden as well:

https://github.com/bitwarden/server/issues/8193

Bitwarden has already added the issue to their internal tracking system.

So at this point, this seems less likely to be a Vaultwarden-specific or single-authenticator edge case, and more like a broader interoperability issue around `userVerification: discouraged` with certain authenticators.

I still think the approach in this PR is a reasonable compromise: the default remains `discouraged`, so existing behavior does not change, while administrators who run into these compatibility problems can opt into `preferred`.

Of course, if Bitwarden changes its own behavior upstream, Vaultwarden may eventually be able to follow that instead. I mainly wanted to add the new information here because the situation has changed a bit since the PR was opened.

BlackDex:
The problem here would be that it would need to probably keep already existing keys as discouraged, else those users can't login anymore, or need to use a different 2FA method, remove all the keys and add them again if that setting is changed. Just flipping the switch will probably cause issues for all existing keys i think.

all-solutions:
That's a fair concern.

I agree that changing the policy globally could be problematic if credentials created under `discouraged` behave differently when authentication is later requested with `preferred`.

I can test this explicitly before making any further assumptions. I have both YubiKeys and the iShield Key 2 Pro available here, so I can reproduce the relevant combinations locally.

For example, I can verify:

- register a YubiKey with `discouraged`, then authenticate with `preferred`
- register an iShield with `discouraged`, then authenticate with `preferred`
- register credentials with `preferred`, then switch back to `discouraged`
- test mixed existing credentials after changing the setting
- verify whether already registered credentials remain usable when the global policy changes

That should tell us whether the current global setting is safe for existing credentials.

If existing credentials really do depend on the policy that was used when they were created, then I agree that simply flipping a global switch would not be the right implementation.

In that case, I could also look into adding a fallback instead of forcing existing credentials onto the new policy. For example, existing credentials could continue to use the behavior they were registered with, while `preferred` is used for new registrations or only when required for compatibility.

I would prefer to test the behavior first and then adjust the implementation based on the results rather than making assumptions here.

If a fallback is needed, I am happy to update the PR accordingly.

all-solutions:
I tested this explicitly now, using a clean test Vaultwarden instance and resetting the LXC back to the same starting point between the individual test cases.

I tested both YubiKeys and the iShield Key 2 Pro.

The cases I checked were:

- YubiKey registered with `discouraged`, then authenticated with `preferred`
- iShield registered with `discouraged`, then authenticated with `preferred`
- credentials registered with `preferred`, then authenticated after switching back to `discouraged`
- mixed existing credentials after changing the global setting
- already registered credentials after switching the policy in both directions
- YubiKey registered with `discouraged`, then switching to `preferred`, adding an iShield afterward, and using both credentials successfully
- the reverse mixed-device combinations as well

All of these tests succeeded.

In particular, I also tested a mixed migration scenario:

1. Register a YubiKey while the policy is `discouraged`.
2. Change the server setting to `preferred`.
3. Register an iShield Key 2 Pro.
4. Authenticate with the previously registered YubiKey.
5. Authenticate with the newly registered iShield.

Both credentials continued to work normally.

I also tested the corresponding combinations with the devices in the opposite order, and did not encounter any case where switching the policy caused an existing credential to stop working.

So based on these tests, existing credentials do not appear to depend on the policy that was active when they were registered.

That means the current global setting seems to be safe with the authenticators I can test here, and a per-credential fallback does not appear to be necessary at this point.

Of course, this does not prove that no authenticator could ever behave differently, but with both YubiKeys and the iShield tested individually, in mixed configurations, and across policy changes in both directions, I currently do not see evidence that changing the policy breaks existing credentials.

If you have a specific additional scenario in mind that you would like me to test, I am happy to add it.

BlackDex:
Thanks @all-solutions, that was helpful.
Ill see if i can test Windows Hello, and Apple Biometric too.
Maybe also Android. Just to be sure that also doesn't fail.

BlackDex:
I did some quick testing my self using the following options.
- YubiKey 5C (via FireFox and Desktop Flatpak on Linux)
- Bitwarden Extension (via FireFox on Linux)
- Android Biometrics (Chrome and App - via Android Virtual Device)
- Windows Hello (via FireFox on Windows 11)
- SoftU2F (Rust-U2F) (via FireFox and Desktop Flatpak on Linux)

I switched to `preferred` after registering all options, and i was still able to login.
Then i removed some of the keys and re-registered them, the only difference now is that they mention `preferred` in the json.

Than, after switching the option back to `false` these all kept working.
So, now I'm wondering, if setting this as a default and not even a configurable option would be the best way to go.

all-solutions:
In principle, there's no reason not to use this as the default.

My production Vaultwarden runs exactly this way, since I definitely need `preferred` due to the iShields. I had pushed for it to be a configurable option to ensure 100% backward compatibility.

all-solutions:
Given the test results so far, would you prefer me to rework this PR so that `preferred` becomes the default behavior and the configuration option is removed entirely?

My original reason for making it configurable was simply to keep the current behavior untouched and minimize any risk for existing installations.

Since both of us have now tested quite a few combinations and none of the existing credentials stopped working after switching between `discouraged` and `preferred`, that extra configuration may not actually be necessary.

If that is the direction you prefer, I can simplify the PR accordingly and change WebAuthn 2FA registration and authentication to use `preferred` by default.

That would also make the implementation smaller and avoid adding another server setting.

Just let me know if you want me to rework the PR in that direction.

all-solutions:
Update: 
I got a call today from MTRIX (SwissBit Sales). You asked about iShield templates? Seriously? This isn't just an iShield issue. Sorry, that's too bad.
What's the point of that? The solution is already there. 
Or is this just a case of “we're double-checking”?

BlackDex:
> Update: I got a call today from MTRIX (SwissBit Sales). You asked about iShield templates? Seriously? This isn't just an iShield issue. Sorry, that's too bad. What's the point of that? The solution is already there. Or is this just a case of “we're double-checking”?

I'm not sure what is too bad here? That I want to check, verify and test? And maybe other parts of Vaultwarden other than the 2FA part? Not really seeing the issue here.

Regarding the PR, as it looks like, it seems that preferred doesn't break stuff, though not yet tested apple devices.

srg-hammerbacher:
We would prefer that `discouraged` stays the default, because it accelerates the authentication and the "third factor" PIN is not needed for many people. `preferred` can be configured if needed.

We would be happy to see the merge soon, as we also cannot use swissbit iShield keys as intended.

PR #7502, opened 2026-07-28
Update supported Bitwarden client feature flags for 2026.7

balcsida:
## Summary

- Update immutable upstream feature-flag source references.
- Add client-compatible Boolean flags with tagged-source consumers.
- Remove flags no longer consumed by current tagged clients.
- Document every available flag in `.env.template`.

## Upstream release baselines

| Component | Tag | Full commit SHA |
| --- | --- | --- |
| Vaultwarden base | 1.37.0 | 46ae59eaf444f0ae0a799070cf2bd6c415284a51 |
| Bitwarden Server | v2026.7.1 | 97ad380f0f82b560d81c1e2e684cef9e85b3379e |
| Browser/Web/Desktop | browser/web/desktop-v2026.7.0 | adf0337e4a0f788b895933792fc04fa162669eff |
| Android Password Manager | v2026.7.0-bwpm | 36c892e1887b5051270ead6021afb16420663566 |
| iOS Password Manager | v2026.7.0-bwpm | e9514dc0f85092ef3cf2a2ccbe07dc5a4f661b0c |

## Added

- `undetermined-cipher-scenario-logic`, `macos-native-credential-sync`, `fill-assist-targeting-rules`, and `pm-39071-default-password-manager-prompt`: client-local autofill behavior.
- `windows-desktop-autotype`, `windows-desktop-autotype-ga`, `ssh-ecdsa`, and `desktop-ui-settings-dialog`: current Desktop UI/native behavior; ECDSA requires `ssh-agent-v2`.
- `innovation-sprint-shared-unlock-part-1` and `innovation-sprint-shared-unlock-part-2`: client-local Desktop/Browser/Web shared unlock; both parts must be enabled together.
- `pm-37785-vault-batch-bar`, `pm-37785-desktop-vault-batch-bar`, `pm-31039-item-action-in-extension`, `pm-29968-fill-after-save`, `pm-32380-btn-text-add-create`, and `pm-28091-add-copy-and-quick-launch-actions`: tagged Vault UI behavior; Desktop batch-bar support requires both batch-bar flags.
- `pm-32016-remove-at-risk-callout`: suppresses the at-risk password callout when explicitly enabled.
- `pm-32783-electron-storage-cache` and `content-script-ipc-channel-framework`: internal/experimental Desktop and Browser client rollouts.
- `pm-32413-multi-client-password-management`: authentication-sensitive Browser/Desktop client rollout.
- `pm-4516-devices-add-last-activity-date`: displays last activity only when supplied by the server; Vaultwarden does not currently supply this field.
- `pm-34224-mobile-attachment-updates`: current mobile UI behavior using existing APIs.
- `pm-34410-attachment-upload-progress` and `phishing-detection`: existing client upload UI and Browser phishing warning behavior.

## Retained

- `pm-5594-safari-account-switching`, `ssh-agent-v2`, and `pm-30529-webauthn-related-origins` remain consumed by tagged clients.

## Removed

- Removed legacy SSH, Desktop biometrics, alias, mTLS, CXP, and desktop UI-milestone flags. Exact-string searches of Browser/Web/Desktop, Android, and iOS 2026.7 sources found no current consumer.

## Deliberately excluded

- Numeric/string, QA/debug, billing/cloud, encryption/KDF, SDK migration, and unsupported server-behavior flags.
- `pm-32009-new-item-types`: excluded because open #7478 provides the required Vaultwarden cipher model/API support.
- `pm-34171-card-scanner`: excluded because open #7477 already adds it.

## Parallel work

- #7477 adds `pm-34171-card-scanner` and is intentionally not duplicated here.
- #7478 remains open and is intentionally not duplicated here.
- `phishing-detection` has self-hosted test discussion: https://github.com/dani-garcia/vaultwarden/discussions/7036

## Testing

- `cargo fmt --all -- --check`
- `cargo test --profile ci --features sqlite` (32 passed), including the focused feature-flag regressions
- `cargo test --profile ci --features mysql`
- `cargo test --profile ci --features sqlite,mysql,postgresql,enable_mimalloc,s3`
- `cargo clippy --profile ci --features sqlite,mysql,postgresql,enable_mimalloc,s3`
- Documentation parity check (27 flags)
- `git diff --check`
- MySQL-enabled commands use Homebrew `mysql-client@8.4` through per-command `MYSQLCLIENT_LIB_DIR` and `MYSQLCLIENT_VERSION` overrides; the default MySQL 9.7 client remains unchanged.

Static source verification only; no runtime tagged-client test was run.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -382,14 +382,19 @@
 ##
 ## The following flags are available:
 ## - "pm-5594-safari-account-switching": Enable account switching in the Safari browser extension. (Verified in Browser 2026.7.0)
+## - "pm-32413-multi-client-password-management": Enable this authentication-sensitive client rollout. (Verified in Browser/Desktop 2026.7.0)
+## - "pm-4516-devices-add-last-activity-date": Show device last activity when supplied by the server. Vaultwarden does not currently supply this field. (Verified in Browser/Desktop 2026.7.0)
 ## - "undetermined-cipher-scenario-logic": Enable the browser extension's updated autofill matching logic. (Verified in Browser 2026.7.0)
 ## - "macos-native-credential-sync": Enable native macOS credential-provider sync in Desktop. (Verified in Desktop 2026.7.0)
 ## - "fill-assist-targeting-rules": Enable Fill Assist targeting rules for autofill. (Verified in Browser/Android/iOS 2026.7.0)
 ## - "pm-39071-default-password-manager-prompt": Prompt new browser-extension installations to make Bitwarden the default password manager. (Verified in Browser 2026.7.0)
+## - "pm-29968-fill-after-save": Fill matching logins after saving them. (Verified in Browser 2026.7.0)
+## - "content-script-ipc-channel-framework": Enable this internal/experimental client rollout. (Verified in Browser 2026.7.0)
 ## - "windows-desktop-autotype": Enable Windows Desktop auto-type for Premium accounts. (Verified in Desktop 2026.7.0)
 ## - "windows-desktop-autotype-ga": Show Windows Desktop auto-type controls; also enable "windows-desktop-autotype". (Verified in Desktop 2026.7.0)
 ## - "ssh-agent-v2": Enable the current SSH agent implementation in Desktop. (Verified in Desktop 2026.7.0)
 ## - "ssh-ecdsa": Allow importing supported ECDSA SSH keys in Desktop; also enable "ssh-agent-v2". (Verified in Desktop 2026.7.0)
+## - "pm-32783-electron-storage-cache": Enable this internal/experimental client rollout. (Verified in Desktop 2026.7.0)
 ## - "desktop-ui-settings-dialog": Use the Desktop settings dialog. (Verified in Desktop 2026.7.0)
 ## - "innovation-sprint-shared-unlock-part-1": Enable the shared-unlock Desktop/Browser leader. Must be enabled with part 2.
 ## - "innovation-sprint-shared-unlock-part-2": Enable the shared-unlock Browser/Web follower and controls. Must be enabled with part 1.
@@ -397,6 +402,12 @@
 ## - "pm-30529-webauthn-related-origins": Allow passkey operations across domains a site declares as related. (Verified in Browser 2026.7.0)
 ## - "pm-34410-attachment-upload-progress": Show attachment upload progress in supported clients. (Verified in Browser/Desktop 2026.7.0)
 ## - "phishing-detection": Warn in the browser extension when a visited site matches its phishing data. (Verified in Browser 2026.7.0)
+## - "pm-37785-vault-batch-bar": Enable the Vault batch bar. Both batch-bar flags are required for Desktop. (Verified in Browser/Desktop 2026.7.0)
+## - "pm-37785-desktop-vault-batch-bar": Enable the Desktop Vault batch bar. Both batch-bar flags are required for Desktop. (Verified in Desktop 2026.7.0)
+## - "pm-31039-item-action-in-extension": Show item actions in the browser extension. (Verified in Browser 2026.7.0)
+## - "pm-32380-btn-text-add-create": Use Add/Create button text. (Verified in Browser/Desktop 2026.7.0)
+## - "pm-28091-add-copy-and-quick-launch-actions": Add Copy and Quick Launch actions. (Verified in Browser/Desktop 2026.7.0)
+## - "pm-32016-remove-at-risk-callout": Suppress the at-risk password callout. (Verified in Browser/Desktop 2026.7.0)
 # EXPERIMENTAL_CLIENT_FEATURE_FLAGS=
 
 ## Require new device emails. When a user logs in an email is required to be sent.
--- a/.env.template
+++ b/.env.template
@@ -381,22 +381,21 @@
 ## Note that clients cache the /api/config endpoint for about 1 hour and it could take some time before they are enabled or disabled!
 ##
 ## The following flags are available:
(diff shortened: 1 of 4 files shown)


src/config.rs
```diff
@@ -1409,33 +1409,33 @@ pub enum PathType {
...
+    "ssh-ecdsa",
+    // Desktop Team
+    "desktop-ui-settings-dialog",
     // Mobile Team
-    "anon-addy-self-host-alias",
-    "simple-login-self-host-alias",
-    "mutual-tls",
-    "cxp-import-mobile",
-    "cxp-export-mobile",
+    "pm-34171-card-scanner",
```
bdd: A prior PR for this exists at https://github.com/dani-garcia/vaultwarden/pull/7477

balcsida: Ohh, let me remove this as soon as I'm in front of my computer

hemagx:
Any reason why `innovation-sprint-shared-unlock-part-1` and `innovation-sprint-shared-unlock-part-2` weren't included?

balcsida:
@hemagx: I forgot those 😅

src/util.rs
```diff
@@ -804,6 +804,64 @@ pub fn parse_experimental_client_feature_flags(
         .collect()
 }
 
+#[cfg(test)]
```
bdd, line 807: I'm not quite following the value proposition of having a test like this.
In the end we will embedding naturally ephemeral strings that have no corresponding implementation / support code in Vaultwarden. `SUPPORTED_FEATURE_FLAGS` is essentially a Vw specific allow list of string nominals that's, I'm guessing, is there to ensure one cannot turn on an upstream feature flag willy-nilly that requires server side support that Vw clearly lacks yet.

What was your original intention with this test addition? Or was it your co-author, Codex's decision to add it?

.env.template
```diff
@@ -381,22 +381,33 @@
...
-## - "simple-login-self-host-alias": Enable configuring self-hosted Simple Login alias generator. (Android >= 2025.3.0, iOS >= 2025.4.0)
-## - "mutual-tls": Enable the use of mutual TLS on Android (Clients >= 2025.2.0)
-## - "cxp-import-mobile": Enable the import via CXP on iOS (Clients >= 2025.9.2)
-## - "cxp-export-mobile": Enable the export via CXP on iOS (Clients >= 2025.9.2)
-## - "pm-30529-webauthn-related-origins":
-## - "desktop-ui-migration-milestone-1": Special feature flag for desktop UI (Desktop >= 2026.2.0)
-## - "desktop-ui-migration-milestone-2": Special feature flag for desktop UI (Desktop >= 2026.2.0)
-## - "desktop-ui-migration-milestone-3": Special feature flag for desktop UI (Desktop >= 2026.2.0)
-## - "desktop-ui-migration-milestone-4": Special feature flag for desktop UI (Desktop >= 2026.2.0)
+## - "pm-5594-safari-account-switching": Enable account switching in the Safari browser extension. (Verified in Browser 2026.7.0)
```
bdd, line 384: These "Verified in 2026.7.0" changes actually remove information. What is in parens is an annotation of when and on what platform(s) the feature flag was first introduced.

It'd be a lot better if you kept the same convention. This will also keep the patch self contained, introducing the feature flags and only adding their documentation, rather than touching the brief explanation and genesis of already defined ones.

Also, just curious, did you really verify all of these manually with your human hands and eyes?

bdd (changes requested):

src/util.rs
BlackDex: I agree, these tests are useless.

AI/LLM's tend to add these just because it seems good to add them.
But most of the time they test useless outcome, and test item's which could be volatile like these items.

I would say, remove them.

BlackDex (changes requested):
I have not checked all these items, but this looks like just using LLM to add all missing items to Vaultwarden without really checking if those are useful or not.

I also wonder if some of these items do not need server side changes to either request or response models. Enabling a feature might cause a lot of issues.

Some of these features might not even get into a final result, or get changes along the way which might need changes on the server side.

Having all these enabled by someone who thinks those are nice, might break their whole database and loos access.

I'm not against adding very useful features which are most likely going to be added, or just provide some UI tweak and does not cause API changes. But other than that, some features should not be implemented.

Also Bitwarden is going fast with a lot of some development, and trying to keep-up with all these thought-balloons is also not very useful.

balcsida:
> I have not checked all these items, but this looks like just using LLM to add all missing items to Vaultwarden without really checking if those are useful or not.

While I didn't test all of them one-by-one and asked the LLM to do it so it covers all of them (so admins at least have the option to send these flags to the clients connected to the server), I'm currently running a build from this branch on production so I can enable the experimental features, like browser extension integration outside of DuckDuckGo Browser and SSH agent in Bitwarden Desktop.

> Also Bitwarden is going fast with a lot of some development, and trying to keep-up with all these thought-balloons is also not very useful.

That is far point! The other option I see is to have the ability to send any flags to the clients, without any validation. What do you think?

.env.template
balcsida: Fair point, let me make it clear so it doesn't loose the info

> Also, just curious, did you really verify all of these manually with your human hands and eyes?

Not all of them, I didn't test the Safari and Windows related changes, but the macOS desktop, Chrome(ium)/Edge and SSH related changes I use actively.

BlackDex:
> That is far point! The other option I see is to have the ability to send any flags to the clients, without any validation. What do you think?

That is what we had before, and it would also go against my comment about ot might breaking the data or needing server side changes.

balcsida:
Fair enough. I'm going to close this PR then and whenever I have time to test and prove any and every specific feature flag that actually works, I'm going to reopen a separate PR for that.
Thanks for your patience and sorry for wasting your time.

PR #7507, opened 2026-07-29
Support for custom 2fa providers(e.g., 2fa via Telegram)

vd2org:
### Overview

This PR implements the external 2FA protocol. It allows admins to integrate any custom provider through a simple HTTP protocol.

### Motivation

None of my users use YubiKey or want to figure out WebAuthn/Duo/etc. However, all of them already use Telegram. So it makes sense to provide an additional layer of security through a familiar tool instead of trying to force them to use something they do not want.

Why not SSO? For a small installation like mine, spinning up the entire SSO infrastructure is overkill. Besides, it would not make a user's life easier.

### Authentication flow description

1. 
   * User: tries to log in to the system and receives an error (`2FA needed`).
   * Vaultwarden: sends a request to the external 2FA server with the user's credentials.
   * 2faServer: reaches the user through an external channel (e.g., Telegram).
2. 
   * User: on the external channel (Telegram), chooses to allow the login.
   * 2faServer: replies to the user with a login code.
3. 
   * User: tries again using a 2FA code.
   * Vaultwarden: validates the code by sending a validation request to 2faServer and allows or declines the login.

The request/response models can be found here: https://github.com/vd2org/vaultwarden-contrib/blob/contrib/external_2fa/src/api/core/two_factor/ext2fa.rs

### Clients’ behavior

Bitwarden clients do not support custom 2FA providers, so we are reusing the YubiKey login flow. There is a new type, `TwoFactorType::External2fa`, with the value `900`. If this two-factor authentication type is enabled for a user, it is changed to `TwoFactorType::YubiKey` before the reply is sent to the client. As a result, the client treats it as YubiKey and prompts for a code.

### Limitations

Currently, only the core support has been implemented, so enabling ext2fa for a user requires a manual database change.

I plan to implement an API for registration support and add support for it in the WebVault.

### Reference implementation

Here is an example implementation of an external 2FA server for Telegram, along with detailed instructions and a complete set of `compose` files to run it: https://github.com/vd2org/tg2fa

Here are Docker images with the patched Vaultwarden: https://github.com/vd2org/vaultwarden/pkgs/container/vaultwarden

### Showcase

I've prepared a showcase to demonstrate how it works. Login requests will be sent to a Telegram channel.

Installation: https://showcase.idsaas.win
Channel for 2FA requests: https://t.me/+spuCqGWOp6o4MWI0

Users:

* `spock@example.com` / `ThIsIsThEsTrOnGpAsSwOrD`
* `kirk@example.com` / `ThIsIsThEsTrOnGpAsSwOrD`
* `mccoy@example.com` / `ThIsIsThEsTrOnGpAsSwOrD`
* `uhura@example.com` / `ThIsIsThEsTrOnGpAsSwOrD`
* `scotty@example.com` / `ThIsIsThEsTrOnGpAsSwOrD`

### DIFF
--- a/src/api/core/two_factor/ext2fa.rs
+++ b/src/api/core/two_factor/ext2fa.rs
@@ -1,0 +1,75 @@
+use std::net::IpAddr;
+
+use reqwest::Method;
+
+use crate::{CONFIG, http_client::make_http_request};
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct SecondFactorData {
+    pub id: i64,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct MetaData {
+    pub username: String,
+    pub device_name: String,
+    pub ip_addr: IpAddr,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct AuthRequest {
+    pub id: String,
+    pub meta_data: MetaData,
+    pub second_factor_data: SecondFactorData,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct AuthValidate {
+    pub id: String,
+    pub code: String,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct AuthResponse {
+    pub ok: bool,
+    pub description: Option<String>,
+}
+
+fn build_url(path: &str) -> String {
+    let base = CONFIG.ext2fa_url();
+    if base.ends_with('/') {
+        format!("{}{}", base, path)
+    } else {
+        format!("{}/{}", base, path)
+    }
+}
+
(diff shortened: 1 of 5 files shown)


BlackDex:
I'm against this we want to keep as close as possible to Bitwarden, and only support the systems they have support for in all their clients. This might break future clients or other weird stuff.

So, from my point of view, this will not get merged.
If other users have an other opinion or see a good use-case for this without breaking anything please let us know.

Ill keep it open for a while, but will probably close this in a week or so.

vd2org:
> This might break future clients or other weird stuff.

It will not break clients. From the client's point of view, the server requested authentication using a Yubikey.

vd2org:
P.S. fixed showcase.

vd2org:
Okay! Thank you for the feedback! ☺️

PR #7534, opened 2026-07-31
Add trusted device encryption for SSO logins

tom27052006:
Continues #7034, which stopped after storing the device keys.

On top of that:

- the remaining device endpoints from bitwarden/server's `DevicesController`: `update-trust`, `untrust`, `retrieve-keys` and `lost-trust`
- rotating the account keys drops the copies of the old user key that sit on the devices, so a stale one cannot be handed out
- approval of a new device by an administrator of the organization, for the case where no other device of the user is around to ask

`SSO_TRUSTED_DEVICE_ENCRYPTION` turns it on, off by default, requires `SSO_ENABLED`.

### Admin approval

Per organization this needs **Account Recovery Administration with auto-enrollment** and a working SMTP configuration, since the enrollment endpoint refuses to run on a mail-less instance. Bitwarden switches those three policies on automatically when trusted device encryption is selected in the organization's SSO configuration; here SSO is configured for the whole server, so they have to be set by hand.

The page for answering those requests lives in `bitwarden_license/bit-web/` in bitwarden/clients and is therefore in no web vault build. The server side of it is under `src/` and AGPL, which is why it could be reimplemented here at all.

No page ships here. An earlier revision carried a stand-in under `/admin`; it asked an organization administrator for their vault master password inside the panel of the server operator, which are two different roles, and it brought its own partial login: no Argon2, no two-step login, and no fingerprint of the asking device to compare against. It is gone.

So a member can raise a request from a stock client, but nothing answers it out of the box. That is survivable rather than a missing safety net: a member who lost every trusted device is recovered through account recovery instead, which works today under the same conditions, confirmed member and enrolled, and costs them a new master password. Approval is the gentler version of that.

The page does exist, written against what the web vault already has, and is linked in a comment below. It is a web vault change, so nothing in this PR depends on it.

### Two adjacent fixes it depends on

- `POST /accounts/keys` did not notice that an account already has a key pair and would replace it, which makes every existing cipher undecryptable. The trusted device flow is what makes this reachable: a client that misjudges an existing account as new posts a fresh key pair here.
- `POST /accounts/set-password` keyed "account already initialized" off the key pair, so an account created without a master password could never gain one, and the offboarding path was dead.

Tested against Keycloak with a black-box suite covering the login signalling, trusting and untrusting, key rotation, offboarding, the admin approval flow and one run with the real client-side crypto end to end.


tom27052006:
Following up on the admin approval part: those requests need a page to be answered from, and none ships in a web vault build — upstream keeps theirs in `bitwarden_license/bit-web/`, which is not AGPL and cannot be reused, so it had to be written from scratch. Here is one, in case you want it.

- the page — https://github.com/vaultwarden/vw_web_builds/compare/v2026.7.0...tom27052006:vw_web_builds:feature/device-approvals-v2026.7.0
- the one server commit it needs — https://github.com/dani-garcia/vaultwarden/compare/aa6046cc...tom27052006:vaultwarden:feature/trusted-device-web-vault

I think the web vault is the better place for this, and not only because upstream puts it there: an administrator who opens it is already signed in and unlocked, so the keys the approval needs are in memory and answering one takes no further authentication. The `/admin` stand-in had no vault session to build on, so it had to carry a login of its own, and that is where its shortcomings came from.

Writing it turned out cheap anyway: the unwrap is the one account recovery already does in `apps/web`, and the navigation entry and the string are in every build already. It shows the request fingerprint, as upstream does. Built against `v2026.6.4`; the files it touches are identical in `v2026.7.0`.

The server commit reports `ssoEnabled` and `ssoMemberDecryptionType` per organization. Nothing in a stock client acts on them, so they only matter to a vault that has the page — drop it if you would rather not.

Both are outside this PR on purpose.

Amsal1:
Looking forward for this to get released soon, would reduce pain for many users to use long master password everytime.

anthonyalayo:
Also looking forward to this.

tom27052006:
Maybe it would make sense to do something similar to the initial SSO rollout and let this sit in `:testing` or even a special release for a while before it goes into a stable release.

Since this touches key handling and quite a few client flows, getting some real-world testing with different clients and SSO setups could be useful.

Also, the diff looks bigger than the actual implementation really is: out of the ~2.7k added lines, around 680 are tests and ~450 are comments/docs

tom27052006:
Following up on the testing idea above: I threw together a full test build so you don’t have to build all the pieces yourself.

`ghcr.io/tom27052006/vaultwarden:tde-testing`

For Docker Compose, just replace the image with:

`image: ghcr.io/tom27052006/vaultwarden:tde-testing`

Combined source is here:
`https://github.com/tom27052006/vaultwarden/tree/test/trusted-device-full-build`

It includes the current PR plus the Device Approvals web vault changes. Everything should work fine, but it’s still just a test build, so use it at your own risk 😄

If you run into anything weird, feel free to drop it here!

Timshel:
Hey,
Sorry for opening a competing PR but, as I mentioned, though it made sense to merge the base feature first.
And I would prefer if the migrations stay with the same name :sweat_smile:

tom27052006:
> Hey, Sorry for opening a competing PR but, as I mentioned, though it made sense to merge the base feature first. And I would prefer if the migrations stay with the same name 😅

Yeah no worries 😄 I think we’re all happy if we can get TDE in at least in its basic form first.

I just hope the remaining parts don’t get lost afterwards or end up sitting around forever once I rebase them on top of #7769 😅

PR #7535, opened 2026-08-01
Implement V2 encryption, registration & rotation

dani-garcia:
Adds support for Bitwarden's v2 crypto, plus added support for setting up v2 encryption during user registration, and key rotation. 

This is currently gated behind two feature flags by default (which Bitwarden currently has disabled) but it will be enabled by default some time in the future and we'll need to make sure we support it before then to avoid application issues. 

To test this, you'll need to enable these two feature flags:
`pm-27278-v2-password-registration`: v2 registration
`enable-account-encryption-v2-jit-password-registration`: SSO JIT set-password


dani-garcia:
replaced by #7764, #7765, and #7766, which split this into multiple parts, as it was getting too big.

PR #7547, opened 2026-08-03
Add "Create new collections" custom role permission

bigbgh:
# Add "Create new collections" custom role permission

Bitwarden's custom role has a [`Create new collections`](https://bitwarden.com/help/user-types-access-control/#custom-roles) permission, which allows a member to create collections without giving them access to every collection in the organization.

That role could not be configured. The three collection options were collapsed into the single `access_all` flag and all three had to be checked to set it, so checking only `Create new collections` left the member a manager without `access_all`, and `post_organization_collections` then refused the request. The option could not be checked on its own either, since the CSS overrides hid the whole block nested below `Manage all collections`.

This stores the permission in a new `create_new_collections` column on `users_organizations` so it can be granted by itself.

Things which are not obvious from the diff:

- `access_all` is now determined by `Edit any collection` and `Delete any collection` alone. Those two are the only options the `Manage all collections` parent checkbox still drives, so they represent it, and `Create new collections` no longer takes part in that check.
- The creator is assigned to the collection they just created, because a member without `access_all` would otherwise lose access to it immediately. Skipped when they already assigned themselves, so the access set in the request is kept.
- `Edit any collection` and `Delete any collection` stay hidden, as they are still only supported together through `access_all`. Supporting them separately needs their own columns and enforcement.
- Existing members are unaffected. Anyone who had all three options checked already has `access_all = 1`, which implies the new permission.

## Testing

Verified with web-vault v2026.6.4 that a custom member with only `Create new collections` is stored as `access_all = 0, create_new_collections = 1`, can create collections and gets `manage` on each one, has no access to collections they were not assigned to, and still cannot invite or edit members. Members with `Manage all collections`, and owners and admins, are unchanged.

`cargo build` and `cargo clippy -- -D warnings` are clean for `sqlite`, `mysql` and `postgresql`.

## The indeterminate parent checkbox

Checking `Create new collections` on its own leaves `Manage all collections` indeterminate, because two of its children are hidden. The saved result is correct, as the parent checkbox is not sent to the server.

I left that visible rather than styling it away, because the indeterminate state is reachable two ways which mean different things: checking only `Create new collections` (`access_all = 0`), and unchecking `Create new collections` while the parent is checked, which leaves the other two set and so still means `access_all = 1`. Rendering it as unchecked would hide a real grant of full access in the second case.

### DIFF
--- a/migrations/mysql/2026-07-24-000000_add_create_new_collections_permission/down.sql
+++ b/migrations/mysql/2026-07-24-000000_add_create_new_collections_permission/down.sql
@@ -1,0 +1,1 @@
+ALTER TABLE users_organizations DROP COLUMN create_new_collections;
--- a/migrations/mysql/2026-07-24-000000_add_create_new_collections_permission/up.sql
+++ b/migrations/mysql/2026-07-24-000000_add_create_new_collections_permission/up.sql
@@ -1,0 +1,1 @@
+ALTER TABLE users_organizations ADD COLUMN create_new_collections BOOLEAN NOT NULL DEFAULT FALSE;
--- a/migrations/postgresql/2026-07-24-000000_add_create_new_collections_permission/down.sql
+++ b/migrations/postgresql/2026-07-24-000000_add_create_new_collections_permission/down.sql
@@ -1,0 +1,1 @@
+ALTER TABLE users_organizations DROP COLUMN create_new_collections;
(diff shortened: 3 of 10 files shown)


PureOxygen-Dev:
Is this not already covered by this pull request?

https://github.com/dani-garcia/vaultwarden/pull/7397

bigbgh:
> Is this not already covered by this pull request?
> 
> #7397

You're right. Thanks for pointing that out before I started working on implementing the remaining features which he has already done in his!

PR #7551, opened 2026-08-04
chore(deps): bump opendal to 0.58.1

Xuanwo:
## Summary

Bump the direct Apache OpenDAL dependency from `0.57.0` to `0.58.1` so Vaultwarden stays on the current stable release for local filesystem and optional S3 storage.

## Changes

- `Cargo.toml` / `Cargo.lock`: `opendal` `0.57.0` → `0.58.1` (including `opendal-core`, `opendal-service-fs`, and `opendal-service-s3`)
- `src/storage.rs`: adapt to OpenDAL 0.58 construction API
  - `Operator::new(builder)` now returns a finished `Operator`, so remove the obsolete `.finish()` calls for both FS and S3 builders
  - When detecting URI options that should skip the AWS SDK credential chain, also accept `skip_signature` (new preferred name) while keeping `allow_anonymous` for compatibility

No `object_store_opendal` dependency is present in this project.

## Validation

- `cargo check --features "sqlite,s3"`
- `cargo test --features "sqlite,s3" storage` (4 tests passed)

## Notes

Runtime S3 behavior against real AWS/SSO credentials was not exercised in this change; only compile-time coverage and storage path unit tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -715,6 +715,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
+name = "base64"
+version = "0.23.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b25655df2c3cdd83c5e5b293b88acd880332b2ddadd7c30ac43144fdc0033da9"
+
+[[package]]
 name = "base64-simd"
 version = "0.8.0"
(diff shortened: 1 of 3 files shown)


BlackDex:
I Already have a change locally which updates a bit more then only this dep/crate including some other fixes.
I have to finish it and check some other stuff. Therefor I'm going to close this one.

Xuanwo:
> I Already have a change locally which updates a bit more then only this dep/crate including some other fixes. I have to finish it and check some other stuff. Therefor I'm going to close this one.

Got it, thank you! Btw, thank you for using opendal, let me know if there are anything you want from opendal side.

Issue #6415, opened 2025-10-30
Incoherent user "enabled" status across admin interfaces


dawagner:
### Prerequisites

- [x] I have searched the existing **Closed _AND_ Open** [Issues](https://github.com/dani-garcia/vaultwarden/issues?q=is%3Aissue%20) **_AND_** [Discussions](https://github.com/dani-garcia/vaultwarden/discussions?discussions_q=)
- [x] I have searched and read the [documentation](https://github.com/dani-garcia/vaultwarden/wiki/)

### Your environment (Generated via diagnostics page)

* Vaultwarden version: v1.34.3
* Web-vault version: v2025.7.0
* OS/Arch: linux/x86_64
* Running within a container: true (Base: Debian)
* Database type: PostgreSQL
* Database version: PostgreSQL 15.14 (Debian 15.14-0+deb12u1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 12.2.0-14+deb12u1) 12.2.0, 64-bit
* Uses config.json: false
* Uses a reverse proxy: true
* IP Header check: true (X-Forwarded-For)
* Internet access: true
* Internet access via a proxy: false
* DNS Check: true
* Browser/Server Time Check: true
* Server/NTP Time Check: true
* Domain Configuration Check: false
* HTTPS Check: true
* Websocket Check: true
* HTTP Response Checks: true

### Config & Details (Generated via diagnostics page)

<details><summary>Show Config & Details</summary>

**Config:**

```json
{
  "_duo_akey": null,
  "_enable_duo": true,
  "_enable_email_2fa": true,
  "_enable_smtp": true,
  "_enable_yubico": true,
  "_icon_service_csp": "",
  "_icon_service_url": "",
  "_ip_header_enabled": true,
  "_max_note_size": 10000,
  "_smtp_img_src": "***:",
  "admin_ratelimit_max_burst": 3,
  "admin_ratelimit_seconds": 300,
  "admin_session_lifetime": 20,
  "admin_token": "***",
  "allowed_connect_src": "",
  "allowed_iframe_ancestors": "",
  "attachments_folder": "data/attachments",
  "auth_request_purge_schedule": "30 * * * * *",
  "authenticator_disable_time_drift": false,
  "data_folder": "data",
  "database_conn_init": "",
  "database_max_conns": 10,
  "database_timeout": 30,
  "database_url": "**********://*******************************************************",
  "db_connection_retries": 60,
  "disable_2fa_remember": false,
[...]

PR #7554, opened 2026-08-04
Fix revoked org members retaining access to org ciphers

abhiShandy:
The cipher access restriction queries (direct collection access, group collection access, and full access via groups) never checked the organization membership status. Since users_collections and groups_users rows are kept when a membership is revoked, revoked (and not yet confirmed) members kept read, write, delete, attachment and collection-move access to org ciphers via the direct-by-UUID endpoints.

Require a confirmed membership in the cipher's organization in all three queries, matching the behavior of the sync queries.

This might be related to https://github.com/dani-garcia/vaultwarden/issues/6415

PS: this bug was found by Kimi K3

### DIFF
--- a/src/db/models/cipher.rs
+++ b/src/db/models/cipher.rs
@@ -665,16 +665,27 @@ impl Cipher {
     }
 
     async fn get_user_collections_access_flags(&self, user_uuid: &UserId, conn: &DbConn) -> Vec<(bool, bool, bool)> {
+        let user_uuid = user_uuid.to_string();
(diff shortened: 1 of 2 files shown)


DasCanard:
this needs to be pushed asap!
@stefan0xC @BlackDex 

https://github.com/advisories/GHSA-hm7r-36jx-j25q
https://nvd.nist.gov/vuln/detail/cve-2026-95814

dani-garcia (approved):

PR #7558, opened 2026-08-05
Misc fixes and updates

### LINKED ISSUE #7545: Web vault: saving collection changes hangs with "invalid type: unit value, expected u8"
### Prerequisites

-

### Vaultwarden Build Version

1.37.1

### Deployment method

Official Container Image

### Custom deployment method

_No response_

### Reverse Proxy

-

Linux

### Operating System Version

linux / x86_64

### Clients

Web Vault

### Client Version

Web-vault version: 2026.6.4

### Summary

Hi, thanks as always for maintaining this project.

I ran into a small but annoying issue in the web vault after updating to 1.37.1.
When I edit an existing organization item and change only its collection assignment,
the Save button spins forever and the dialog never closes. The change itself is saved
correctly — if I close the dialog and reopen the item, the new collections are there —
so it seems to be purely a client-side failure after the request succeeds.

The browser console shows:

    Unhandled error in angular Error: Error: invalid type: unit value, expected u8

The browser extension doesn't show this at all; it only happens in the web vault.

### Steps to reproduce

1. Open the web vault and edit an existing item that belongs to an organization.
2. Change only the collection assignment, leave everything else untouched, and save.
3. The Save button keeps spinning and the error above is logged in the console.
4. Close the dialog and reopen the item — the collection change has been applied.

### Expected Result

-

### Actual Result

-

### Logs

```text

```

### Screenshots or Videos

_No response_

### What I think is going on

I did some digging and this is my best guess, though I'd appreciate a second opinion
since I may well have missed something.

When only the collections change, `DefaultCipherFormService.saveCipher()` splits the
work into two requests: it first calls `updateWithServer()` (which explains why the
item is persisted), and then saves the collection changes separately. If `config.admin`
is set, or the item previously had no collections, it goes through
`saveCollectionsWithServerAdmin()`.

That path ends up in `libs/common/src/services/api.service.ts`:

    putCipherCollectionsAdmin(id: string, request: CipherCollectionsRequest): Promise<any> {
      return this.send("PUT", "/ciphers/" + id + "/collections-admin", request, true, true);
    }

The trailing `true` is `hasResponse`, so the client expects the updated cipher JSON back
and builds `new CipherData(response)` / `new Cipher(data)` out of it.

On the server side, in `src/api/core/ciphers.rs`, `post_collections_admin()` and
`put_collections_admin()` both return `EmptyResult`, i.e. no body — whereas
`post_collections_update()` (`/collections`) and `post_collections2_update()`
(`/collections_v2`) do return the cipher JSON.

If that's correct, the client ends up with a Cipher whose `type` and `reprompt` are
undefined, and passing that to `cipherService.decrypt()` makes sdk-internal fail to
deserialize `CipherType` (a `#[repr(u8)]` enum), which matches the
`invalid type: unit value, expected u8` message exactly. The rejected promise never
resolves the dialog, which would explain the endless spinner.

### Possible fix

If the above holds, having `/ciphers/<cipher_id>/collections-admin` (POST and PUT)
return the updated cipher JSON — the same way `post_collections_update()` already does —
should be enough.

BlackDex:
### General

- Update GHA and pre-commit
- Fix showing events for a specific user
- Update `web-vault` to `v2026.7.0`
- Updated all the crates
- Updated `opendal` and adjusted code where needed
- Updated `yubico_ng` and adjusted code where needed
  This version now supports using an own `HttpClient` and it pulls in no `reqwest` dependency anymore. Now it will use our own client which uses custom hickory DNS and other features.
- Fix updating collections for a cipher
  Fixes #7545
  Fixes #7546

### CSS Caching

Currently we set a cache ttl of 24 hours, and users need to do a force refresh if there is anything changed to the CSS file.
In the past we have had several issue reported which were related to a still cached CSS file.

This commit will change the caching and also cache the generated CSS file in memory.
Instead of letting the browser cache it for 24 hours we generate an ETag, this is just a hash of the contents.
This ETag is returned by the browser during a request, and we can match this, and if so, just return a `304` `Not Modified`.
If the ETag is not known, we return the new content.

This should make simple refreshes by clients get updated settings or a new version of Vaultwarden which has other CSS entries get updated instantly.
If a user does a hard refresh, we will not receive the ETag and the content will be served.

The same goes if someone has the `reload_templates` feature enabled, since then we should not cache anyway.
If someone adjust settings via the `/admin` interface, the cache will be invalidated and a new CSS will be generated.

### Extra admin diagnostics checks

Added a check if the templates are overridden and return which specific folder, `admin`, `email` or `scss`.
This way we could more quickly point users to possible outdated templates which they are using.

Also updated the Support String to use some emojis so we should be able to quicker see if there is something wrong.
Just checking `true` or `false` could be difficult sometimes, and sometimes what we had as `false` wasn't bad either.

Also adjusted the eslint comments so it will work with the latest version of eslint.

### DIFF
--- a/.github/workflows/hadolint.yml
+++ b/.github/workflows/hadolint.yml
@@ -41,12 +41,12 @@ jobs:
       # Uses the Docker-based action (hadolint pre-bundled in ghcr.io/hadolint/hadolint:v2.14.0-debian)
       # so no binary is downloaded at runtime. Pinned by commit SHA for supply-chain safety.
       - name: Run hadolint on Dockerfile.debian
-        uses: hadolint/hadolint-action@2332a7b74a6de0dda2e2221d575162eba76ba5e5 # v3.3.0
+        uses: hadolint/hadolint-action@2a66e89f53d0771bb131a7fa31f3136336094aa6 # v3.4.0
         with:
           dockerfile: docker/Dockerfile.debian
 
       - name: Run hadolint on Dockerfile.alpine
-        uses: hadolint/hadolint-action@2332a7b74a6de0dda2e2221d575162eba76ba5e5 # v3.3.0
+        uses: hadolint/hadolint-action@2a66e89f53d0771bb131a7fa31f3136336094aa6 # v3.4.0
         with:
           dockerfile: docker/Dockerfile.alpine
       # End Test Dockerfiles with hadolint
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -106,7 +106,7 @@ jobs:
 
       # Login to Docker Hub
       - name: Login to Docker Hub
-        uses: docker/login-action@abd2ef45e78c5afb21d64d4ca52ee8550d9572c7 # v4.5.1
+        uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f # v4.6.0
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
@@ -121,7 +121,7 @@ jobs:
 
       # Login to GitHub Container Registry
       - name: Login to GitHub Container Registry
-        uses: docker/login-action@abd2ef45e78c5afb21d64d4ca52ee8550d9572c7 # v4.5.1
+        uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f # v4.6.0
         with:
           registry: ghcr.io
           username: ${{ github.repository_owner }}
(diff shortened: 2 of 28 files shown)


dani-garcia (approved):

calvin-li-developer:
@BlackDex  @dani-garcia, I think today's GitHub problems cause this PR not to trigger a release action. Might need to do this one manually for this PR.

BlackDex:
Looks like it indeed. I do not think we have a manual trigger for this workflow.

BlackDex:
@calvin-li-developer there was a PR which i approved already which was not yet merged, so i did that right now, and there will be a new release shortly.

PR #7563, opened 2026-08-06
[web-v2026.7.1] 2FA userVerificationToken for Authenticator, Email and Duo

Timshel:
Latest client change the way 2FA providers are activated/deactivated.
The password is now sent only when requiring the provider current data, further change are validated using a `userVerificationToken`.

This adds this logic for the Authenticator, Email and Duo provider (Duo is untested, others were tested with this [build](https://github.com/Timshel/oidc_web_vault/releases/tag/v2026.7.1-1)).

I won't be around for ~10 days, feel free to modify this PR or create a new one; otherwise I'll continue it when I'm back.

### DIFF
--- a/src/api/core/two_factor/authenticator.rs
+++ b/src/api/core/two_factor/authenticator.rs
@@ -3,7 +3,7 @@ use rocket::{Route, serde::json::Json};
 
 use crate::{
     api::{EmptyResult, JsonResult, PasswordOrOtpData, core::log_user_event, core::two_factor::generate_recover_code},
-    auth::{ClientIp, Headers},
+    auth::{ClientIp, Headers, two_factor},
     crypto,
     db::{
         DbConn,
@@ -20,7 +20,6 @@ pub fn routes() -> Vec<Route> {
 
 #[post("/two-factor/get-authenticator", data = "<data>")]
 async fn generate_authenticator(data: Json<PasswordOrOtpData>, headers: Headers, conn: DbConn) -> JsonResult {
-    let data: PasswordOrOtpData = data.into_inner();
     let user = headers.user;
 
     data.validate(&user, false, &conn).await?;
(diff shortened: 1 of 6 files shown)


Timshel:
Made the change for Webauthn and Yubikey too.
Was able to test Authenticator, Email, Duo and Webauthn providers but can't test Yubikey.

BlackDex:
@Timshel did a quick test with v2026.8.0 and Yubico OTP doesn't work. This is what i get (With some debugging added), when i try to active a key when there are non registered yet. There is no HTTP call done yet, it stops at `validate_yubikey`

```
[2026-08-25 15:20:22.568][request][INFO] PUT /api/two-factor/yubikey
[src/auth/two_factor.rs:217:9] keys = [
    "vvrcexxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
]
[src/auth/two_factor.rs:217:23] claims.keys = []
[2026-08-25 15:20:22.574][vaultwarden::auth::two_factor][ERROR] Invalid verification token: Invalid keys
```

BlackDex:
Logging in with an already saved token does still work though, but adding new ones is just out of the question.

Timshel:
A yes I see the error.

Timshel:
Should be better, need to use the keys in db when checking the token.

Timshel:
Added two minor commits to use the Enum instead of `i32` as parameter (since it concerns code modified by the PR did not make a separate one).

tom27052006:
I played a little bit around with the PR a bit using web-v2026.8.1 and found two small WebAuthn issues

Adding a second WebAuthn key failed with Invalid verification token: Invalid state. Looks like activate_webauthn() always validates with enabled = false, even if WebAuthn is already enabled. I changed that locally to use the current enabled state and adding a second key worked after that

Then I tried removing just one of the two keys. That also fails, since the web vault calls DELETE /api/two-factor/webauthn for a single key, while the PR currently only seems to have /api/two-factor/webauthn/all

Timshel:
Hey,
Thanks for testing, latest code should fix both issues :)

PR #7566, opened 2026-08-07
Add configurable OIDC claim for user display name

kittygaming99:
Resolves discussion #7359 

### Background
Currently, when creating a user via SSO, Vaultwarden hardcodes the `preferred_username` claim to set the user's display name. This causes issues with Identity Providers (IdPs) like Google that don't issue a `preferred_username` claim or allow custom claim mapping, often resulting in the user's email address being used as their display name instead. The OIDC specification also suggests that the `name` claim is a better fit for this.

### Proposed changes
This PR introduces a new configuration option `(sso_name_claim)` in the admin panel, allowing administrators to customise which OIDC claim is mapped to the user's display name.

- Added `sso_name_claim` to the SSO configuration block (`src/config.rs`), which automatically generates the relevant input field in the admin settings UI.
- Updated the OIDC login flow (`src/sso.rs`) to extract the configured claim dynamically.
- The setting defaults to `name`. If the configured claim is missing from the IdP's response, the logic safely falls back to preferred_username to ensure backwards compatibility with existing setups. Consequently, if the configuration option is left blank, it will fall back to `preferred_username`. This ensures backwards compatibility with existing setups and retains the historical display name behaviour.

### Testing

sso_name_claim_configuration_option_1
- Configuration option is added to the admin panel
sso_name_claim_configuration_option_2
- Custom OIDC claim, `nickname`, is configured as the display name
sso_name_claim_configuration_option_3
- Mock OAuth2 Server Sign-in is used to simulate an IdP with that claim
sso_name_claim_configuration_option_4
- Login is successful, and the display name is correctly configured
sso_name_claim_configuration_option_5
- Configuration is changed to an empty value. This is expected to fall back to the `preferred_username` claim
sso_name_claim_configuration_option_6
- Mock OAuth2 Server Sign-in is used to simulate an IdP that sends the `name` claim. The expected behaviour is that the `preferred_username` will be used because the `sso_name_claim` option was left blank
sso_name_claim_configuration_option_7
- The display name is set to `preferred_username` as intended
- Configuring `name` as the display name claim but not sending a value also fell back to `preferred_username` as intended

### Possible improvements
- Adding a filter to ensure the mapped display name claim does not contain an empty string or only whitespace could be a good idea. For example:
```
let extract_claim = |claims: &serde_json::Value, claim_key: &str| -> Option<String> {
    claims.get(claim_key)
        .and_then(|v| v.as_str())
        .map(str::trim)
        .filter(|s| !s.is_empty())
        .map(String::from)
};
```


sistason:
I stumbled on the same problem, but when wanting to change the identifier not being `sub` (for a expand&contract migration of SSO-Ids).

Maybe it would be better to add a complete SSO_CLAIM_MAPPING config option for all claims and not do one for `preferred_username` (which I'd like to change as well to `name`), one for `sub` and later one for `email` and maybe `verified`?

kittygaming99:
That could be a good shout. I don't see why we shouldn't allow the user to configure other claims. We'd have to consider error handling for more critical claims like sub. I'll have a look at how missing claims are managed at the moment.

kittygaming99:
@sistason I think that complete custom claim mapping lies a bit out of the scope for this PR. While I agree that would be a really good feature I think it's beyond my capabilities as I am still learning Rust.

I would be happy to take a stab at it if we were to get guidance from a Maintainer on how to implement it. It would require a big overhaul of SSO and some deep thought on error handling. I'm also not sure how to structure the Admin Panel UI. Should we have a JSON object (`sso_claim_mapping`) that can be edited by distinct configuration fields? Or should each claim have its own config option?

For now I am just looking for a fix for the name claim as having this feature before onboarding the bulk of our users will mean display names don't have to be manually fixed.

Should we add error handling for empty name claim values or name claims containing white spaces?

Timshel:
Not at maintainer, but at the time I used `preferred_username` because it seemed to match and did not have the issue of a locale.

If the main goal is to read the `name` claim then it's probably better to use:

- https://docs.rs/openidconnect/latest/openidconnect/struct.IdTokenClaims.html#method.name
- https://docs.rs/openidconnect/latest/openidconnect/struct.IdTokenClaims.html#method.locale

Especially since the spec mention that:

> End-User's full name in displayable form including all name parts, possibly including titles and suffixes, ordered according to the End-User's locale and preferences.

Not sure how the ordered part is handled but reading `name` as a `String` might not be best.

Outside of it :

- Since `sso_name_claim` has a default value it means that it will default to the `name` claim for everyone.
  To not change the current logic for everyone I think an empty default would make more sense.
- I don't think remapping in general is a good idea, especially for `sub` and `email` since those are fields defined in the spec.

PR #7584, opened 2026-08-12
Reject SSH key ciphers with null or empty required fields

### LINKED ISSUE #7514: SSH key cipher with `null` members in `sshKey` is accepted and silently discarded (Bitwarden cloud rejects the same request)
### Prerequisites

n/a - I'm not actually a vaultwarden user, discovered this only while debugging a test script submitted to me.

### Vaultwarden Build Version

1.37.0

### Deployment method

Official Container Image

### Custom deployment method

Discovered this issue while testing https://github.com/cachix/secretspec/pull/166.

## Summary

Creating an SSH key cipher (`type: 5`) whose `sshKey` object has `null` in any of
`privateKey`, `publicKey` or `keyFingerprint` succeeds against Vaultwarden — the
API returns a cipher id — but the key material does not survive the round trip.
Reading the item back shows the `sshKey` object gone.

Bitwarden's own server rejects the identical request with a validation error, so
a client that is correct against cloud gets a success response and silent data
loss against Vaultwarden. That asymmetry is the actual problem: the write looks
like it worked.

## Additional context

- **Non-empty strings are all that is required.** We tested three payload shapes
  against both servers: (A) nulls in two members, (B) arbitrary non-empty
  placeholder strings in all three, (C) realistic-looking OpenSSH key material in
  all three. Cloud refuses A and accepts both B and C. So this is a presence
  check, not key-material validation — the fix on the server side is a
  null/type check, not a parser.

- **The first-party clients do not hit this.** The web vault and browser
  extension always populate all three members, so this only reaches API and CLI
  clients that build the cipher JSON themselves. That is probably why it has
  gone unnoticed.

- **We only tested `type: 5`.** Whether other cipher types silently drop their
  type-specific sub-object when a member is `null` is untested and worth a look
  while the relevant code is open — if the cause is generic handling of the
  type-specific data blob rather than something SSH-specific, the same shape of
  bug likely exists elsewhere.

- **Where we would start looking** (a guess, offered only as a starting point —
  we did not read the Vaultwarden source): the cipher create/update path in
  `src/api/core/ciphers.rs`, where the incoming type-specific `data` is stored.
  If members are copied through without a per-type required-field check, a `null`
  would land in the stored blob and later collapse the object. Upstream Bitwarden
  appears to validate at the model-binding layer, which matches the shape of the
  error message it returns.

- **Reported by** the SecretSpec Bitwarden provider work
  (https://github.com/cachix/secretspec, PR #166), which uses a disposable
  Vaultwarden container as its CI backend precisely because it needs no real
  vault credentials. Worked around it client-side by never emitting a `null`:
  unaddressed members are initialised to a placeholder string. That workaround is
  fine for us and we are not blocked — filing this because the silent-success
  behaviour will bite other API clients the same way, and it cost us a debugging
  session to find.

### Reverse Proxy

n/a

Linux

### Operating System Version

the host OS for the docker image was mac/podman, but not relevant for this case

### Clients

CLI

### Client Version

CLI 2025.11.0

### Steps To Reproduce

Against a Vaultwarden instance:

```bash
export BW_SESSION=$(bw unlock --raw)

ID=$(jq -nc '{
  type: 5,
  name: "vw-sshkey-null-probe",
  notes: "probe, safe to delete",
  sshKey: {
    privateKey: "probe-private-value",
    publicKey: null,
    keyFingerprint: null
  },
  fields: []
}' | bw encode | bw create item --nointeraction | jq -r '.id')

echo "created: $ID"

bw sync --nointeraction
bw get item "$ID" --nointeraction | jq '.sshKey'

bw delete item "$ID" --nointeraction
```

Then run exactly the same script against `https://vault.bitwarden.com`.

Against a Vaultwarden instance:

```bash
export BW_SESSION=$(bw unlock --raw)

ID=$(jq -nc '{
  type: 5,
  name: "vw-sshkey-null-probe",
  notes: "probe, safe to delete",
  sshKey: {
    privateKey: "probe-private-value",
    publicKey: null,
    keyFingerprint: null
  },
  fields: []
}' | bw encode | bw create item --nointeraction | jq -r '.id')

echo "created: $ID"

bw sync --nointeraction
bw get item "$ID" --nointeraction | jq '.sshKey'

bw delete item "$ID" --nointeraction
```

Then run exactly the same script against `https://vault.bitwarden.com`.

## Observed

**Vaultwarden** — `bw create item` exits 0 and prints a cipher with an `id`. After
`bw sync`, the item is present and correctly named, but the key is not there:
`.sshKey` reads back as `null`. Driving the same write through an application
layer, the user-visible symptom is a write that reports success followed by a
read that returns `[error: cannot decrypt]`.

> Note for whoever picks this up: our notes record both `sshKey: null` and
> `[error: cannot decrypt]` from this configuration. We did not pin down which
> read path produces which — plausibly the raw `bw get item` shows the null while
> the CLI renders a partially-populated, undecryptable object as the error string.
> Worth re-measuring precisely; it does not change the headline, which is that the
> write is accepted and the key is unrecoverable.

**Bitwarden cloud** — the create is refused outright, no cipher is created, and
the server's reason is:

```
invalid type: unit value, expected a valid string
```

### Expected Result

Reject the request, as cloud does. A `400` naming the offending member would be
ideal; anything other than `200`-plus-discard is an improvement. The failure mode
to avoid is the current one, where the only way a client can discover the write
did not take is to read the item back and compare.

### Actual Result

See reprooduction steps - the short version is the vaultwarden server fails to create the item but returns a 200 response code with an empty response.

### Logs

```text

```

### Screenshots or Videos

_No response_

### Additional Context

_No response_

xhon-pelushi:
## Summary
- Validate SSH key (`type: 5`) payloads before save.
- Require non-empty string values for `privateKey`, `publicKey`, and `keyFingerprint`.
- Reject null/empty/missing fields with an error instead of accepting the write and dropping `sshKey` on read-back.

Fixes #7514

## Test plan
- [x] `cargo test --profile ci --features sqlite ssh_key_validation_tests`
- [ ] POST/PUT a type-5 cipher with any required sshKey field set to `null` returns an error.
- [ ] POST/PUT with non-empty placeholder strings for all three fields succeeds and round-trips.
- [ ] Existing non-SSH cipher types are unchanged.

### DIFF
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -377,17 +377,6 @@ async fn post_ciphers(data: Json<CipherData>, headers: Headers, conn: DbConn, nt
 /// Enforces the personal ownership policy on user-owned ciphers, if applicable.
 /// A non-owner/admin user belonging to an org with the personal ownership policy
 /// enabled isn't allowed to create new user-owned ciphers or modify existing ones
-/// Ensure SSH key type-data has the required non-empty string fields.
-fn validate_ssh_key_data(type_data: &Value) -> EmptyResult {
-    for field in ["privateKey", "publicKey", "keyFingerprint"] {
-        match type_data.get(field).and_then(Value::as_str) {
-            Some(value) if !value.is_empty() => {}
-            _ => err!(format!("SSH key field '{field}' must be a non-empty string")),
-        }
-    }
-    Ok(())
-}
-
 /// (that were created before the policy was applicable to the user). The user is
 /// allowed to delete or share such ciphers to an org, however.
 ///
@@ -398,6 +387,20 @@ async fn enforce_personal_ownership_policy(data: Option<&CipherData>, headers: &
         let policy_type = OrgPolicyType::PersonalOwnership;
         if OrgPolicy::is_applicable_to_user(user_id, policy_type, None, conn).await {
             err!("Due to an Enterprise Policy, you are restricted from saving items to your personal vault.")
+        }
+    }
+    Ok(())
+}
+
+/// Ensure SSH key type-data carries the mandatory non-empty string fields.
+///
+/// `Cipher::to_json` already discards the type-data of stored SSH ciphers whose
+/// fields are missing or empty; rejecting them here keeps such ciphers from
+/// being written in the first place.
+fn validate_ssh_key_data(type_data: &Value) -> EmptyResult {
+    for field in ["privateKey", "publicKey", "keyFingerprint"] {
+        if type_data[field].as_str().is_none_or(str::is_empty) {
+            err!(format!("SSH key field '{field}' must be a non-empty string"))
         }
     }
     Ok(())
@@ -2246,41 +2249,18 @@ mod ssh_key_validation_tests {
     use serde_json::json;
 
     #[test]
-    fn accepts_non_empty_required_fields() {
-        let data = json!({
-            "privateKey": "priv",
-            "publicKey": "pub",
-            "keyFingerprint": "fp"
-        });
-        assert!(validate_ssh_key_data(&data).is_ok());
-    }
+    fn validate_ssh_key_data_required_fields() {
+        let cases = [
+            ("all fields present", json!({"privateKey": "priv", "publicKey": "pub", "keyFingerprint": "fp"}), true),
+            ("null private key", json!({"privateKey": null, "publicKey": "pub", "keyFingerprint": "fp"}), false),
+            ("empty public key", json!({"privateKey": "priv", "publicKey": "", "keyFingerprint": "fp"}), false),
+            ("missing fingerprint", json!({"privateKey": "priv", "publicKey": "pub"}), false),
+            ("non-string fingerprint", json!({"privateKey": "priv", "publicKey": "pub", "keyFingerprint": 42}), false),
+            ("no fields at all", json!({}), false),
+        ];
 
-    #[test]
-    fn rejects_null_private_key() {
-        let data = json!({
-            "privateKey": null,
-            "publicKey": "pub",
-            "keyFingerprint": "fp"
-        });
-        assert!(validate_ssh_key_data(&data).is_err());
-    }
-
-    #[test]
-    fn rejects_empty_public_key() {
-        let data = json!({
-            "privateKey": "priv",
-            "publicKey": "",
-            "keyFingerprint": "fp"
-        });
-        assert!(validate_ssh_key_data(&data).is_err());
-    }
-
-    #[test]
-    fn rejects_missing_fingerprint() {
-        let data = json!({
-            "privateKey": "priv",
-            "publicKey": "pub"
-        });
-        assert!(validate_ssh_key_data(&data).is_err());
+        for (case, type_data, expected_ok) in cases {
+            assert_eq!(validate_ssh_key_data(&type_data).is_ok(), expected_ok, "case: {case}");
+        }
     }
 }
--- a/src/api/core/ciphers.rs
+++ b/src/api/core/ciphers.rs
@@ -377,6 +377,17 @@ async fn post_ciphers(data: Json<CipherData>, headers: Headers, conn: DbConn, nt
 /// Enforces the personal ownership policy on user-owned ciphers, if applicable.
 /// A non-owner/admin user belonging to an org with the personal ownership policy
 /// enabled isn't allowed to create new user-owned ciphers or modify existing ones
+/// Ensure SSH key type-data has the required non-empty string fields.
+fn validate_ssh_key_data(type_data: &Value) -> EmptyResult {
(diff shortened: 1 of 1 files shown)


src/api/core/ciphers.rs
```diff
@@ -377,6 +377,17 @@ async fn post_ciphers(data: Json<CipherData>, headers: Headers, conn: DbConn, nt
...
+fn validate_ssh_key_data(type_data: &Value) -> EmptyResult {
+    for field in ["privateKey", "publicKey", "keyFingerprint"] {
+        match type_data.get(field).and_then(Value::as_str) {
+            Some(value) if !value.is_empty() => {}
+            _ => err!(format!("SSH key field '{field}' must be a non-empty string")),
+        }
+    }
+    Ok(())
+}
+
```
BlackDex: This function is located at a wrong place, it cuts into the comment of an other function.

I also think it could be a bit more optimized maybe?

```suggestion
fn validate_ssh_key_data(type_data: &Value) -> EmptyResult {
    for field in ["privateKey", "publicKey", "keyFingerprint"] {
        type_data
            .get(field)
            .and_then(Value::as_str)
            .filter(|v| !v.is_empty())
            .ok_or_else(|| err!(format!("SSH Key field '{field}' is required!")))?;
    }
}
```

BlackDex (changes requested):
I think the validation function can be a bit optimized.

The tests i find a bit too much t.b.h.
It might also be an option to just create one test function and have a vec of cases to check and then just run those?

Something like this maybe?

```rust
    let cases = [
        // missing field
        (json!({"publicKey":"pub","keyFingerprint":"fp"}), false),
        // null field
        (json!({"privateKey":null,"publicKey":"pub","keyFingerprint":"fp"}), false),
    ]
```

And then loop over those cases and use the `false` or `true` as an assert outcome?
That would make it easier to add other tests if needed in the future and thinking about different function names. The failed test is shown in detail if i'm correct, so we should see what would failed (but i might be wrong here).

xhon-pelushi:
Thanks — all three addressed in `11f4175`.

**Placement.** You were right, and it was worse than misplaced: the function had landed between the third and fourth lines of the `enforce_personal_ownership_policy` doc comment, so `/// (that were created before the policy was applicable to the user)` was orphaned below it. Moved below that function, comment is one block again.

**The function.** I went with the idiom already in the tree rather than my own — `Cipher::to_json` checks these same three fields at `src/db/models/cipher.rs`:

```rust
if self.atype == 5
    && (type_data_json["keyFingerprint"].as_str().is_none_or(str::is_empty)
        || type_data_json["privateKey"].as_str().is_none_or(str::is_empty)
        || type_data_json["publicKey"].as_str().is_none_or(str::is_empty))
```

so this now reads:

```rust
fn validate_ssh_key_data(type_data: &Value) -> EmptyResult {
    for field in ["privateKey", "publicKey", "keyFingerprint"] {
        if type_data[field].as_str().is_none_or(str::is_empty) {
            err!(format!("SSH key field '{field}' must be a non-empty string"))
        }
    }
    Ok(())
}
```

I added a doc comment noting how the two relate — `to_json` discards the type-data of ciphers already stored with bad fields, this stops them being written in the first place.

One note on your suggested snippet: `err!` expands to `return Err(...)` (`src/error.rs`), so inside `.ok_or_else(|| err!(...))` it would return from the closure rather than the function and wouldn't type-check. Same shape otherwise.

**Tests.** Collapsed to one table as you suggested, and added two cases the originals missed — a non-string field, and type-data with no fields at all:

```rust
let cases = [
    ("all fields present", json!({...}), true),
    ("null private key",   json!({...}), false),
    ...
];
for (case, type_data, expected_ok) in cases {
    assert_eq!(validate_ssh_key_data(&type_data).is_ok(), expected_ok, "case: {case}");
}
```

To your question about failure output — with a bare `assert!` you'd only get the line number, so I gave each case a label that `assert_eq!` prints. A failure now reads `case: empty public key`.

Verified locally on the pinned 1.97.1 toolchain: `cargo test --features sqlite` 30 passed / 0 failed, `cargo fmt --check` clean.

Unrelated, but you may want to know: `cargo clippy --features sqlite --all-targets -- -D warnings` fails on `src/api/admin.rs:908-910` with "used `assert!` with an equality comparison" —

```rust
assert!(web_vault_compare("2025.12.0", "2025.12.1") == -1);
```

Nothing to do with this PR (it only touches `ciphers.rs`), but it does mean that clippy job should be red on main too.

src/api/core/ciphers.rs
xhon-pelushi: Done in `11f4175` — moved below `enforce_personal_ownership_policy` (it had landed between the third and fourth lines of that function's doc comment, orphaning the last line), and rewritten to the `as_str().is_none_or(str::is_empty)` form already used for these same three fields in `Cipher::to_json`. Details and the note about `err!` expanding to `return Err(...)` are in the PR comment.

PR #7639, opened 2026-08-24
Route service clients through shared HTTP setup

### LINKED ISSUE #7630: 1.37.2: S3 backend broken at runtime — "default HTTP transport is not installed" (opendal 0.58 + default-features = false)
### Prerequisites

Note: the S3 bucket name and the organisation name were replaced with placeholders.
Nothing else was modified. Support string is from 1.37.1 (we rolled back) — 1.37.2
never stays up long enough to serve /admin/diagnostics.

* Vaultwarden version: v1.37.1
* Web-vault version: v2026.6.4
* OS/Arch: linux/x86_64
* Running within a container: true (Base: Alpine)
* Database type: PostgreSQL
* Database version: PostgreSQL 16.13 on x86_64-pc-linux-gnu, compiled by x86_64-pc-linux-gnu-gcc (GCC) 12.4.0, 64-bit
* Uses config.json: true
* Uses a reverse proxy: true
* IP Header check: false (X-Forwarded-For)
* Internet access: true
* DNS Check: true
* Server/NTP Time Check: true
* HTTPS Check: true
* Websocket Check: true
* HTTP Response Checks: true

**Environment settings which are overridden:** DOMAIN, SIGNUPS_ALLOWED, ORG_CREATION_USERS, INVITATIONS_ALLOWED, PASSWORD_HINTS_ALLOWED, ADMIN_TOKEN, INCREASE_NOTE_SIZE_LIMIT, SSO_ENABLED, SSO_ONLY, SSO_SIGNUPS_MATCH_EMAIL, SSO_CLIENT_ID, SSO_CLIENT_SECRET, SSO_AUTHORITY, SSO_AUTHORIZE_EXTRA_PARAMS, SSO_MASTER_PASSWORD_POLICY, SMTP_HOST, SMTP_SECURITY, SMTP_PORT, SMTP_FROM, SMTP_USERNAME, SMTP_PASSWORD

**Config:**

```json
{
  "data_folder": "s3://<bucket>",
  "database_url": "**********://******************************************************************************************************************************************************************************",
  "icon_cache_folder": "s3://<bucket>/icon_cache",
  "attachments_folder": "s3://<bucket>/attachments",
  "sends_folder": "s3://<bucket>/sends",
  "tmp_folder": "/data/tmp",
  "templates_folder": "/data/templates",
  "rsa_key_filename": "s3://<bucket>/rsa_key",
  "enable_websocket": true,
  "push_enabled": false,
  "purge_incomplete_sso_auth": "0 20 0 * * *",
  "domain": "*****://*******************",
  "domain_set": true,
  "domain_origin": "*****://*******************",
  "domain_path": "",
  "web_vault_enabled": true,
  "sends_allowed": true,
  "hibp_api_key": null,
  "org_attachment_limit": null,
  "user_send_limit": null,
  "trash_auto_delete_days": 90,
  "signups_allowed": false,
  "signups_verify": false,
  "org_events_enabled": true,
  "org_creation_users": "****",
  "invitations_allowed": true,
  "email_change_allowed": true,
  "password_hints_allowed": false,
  "show_password_hint": false,
  "admin_token": "***",
  "invitation_org_name": "<org name>",
  "events_days_retain": 90,
  "ip_header": "X-Real-IP",
  "_ip_header_enabled": true,
  "ip_header_trusted_proxies": "local",
  "icon_service": "internal",
  "_icon_service_url": "",
  "_icon_service_csp": "",
  "icon_redirect_code": 302,
  "icon_cache_ttl": 2592000,
  "icon_cache_negttl": 259200,
  "icon_download_timeout": 10,
  "icon_blacklist_regex": null,
  "reload_templates": false,
  "extended_logging": true,
  "use_syslog": false,
  "log_file": null,
  "log_level": "debug",
  "enable_db_wal": true,
  "db_connection_retries": 15,
  "database_timeout": 30,
  "database_idle_timeout": 600,
  "database_max_conns": 10,
  "database_min_conns": 2,
  "database_conn_init": "",
  "disable_admin_token": false,
  "allowed_connect_src": "",
  "unauthenticated_ratelimit_seconds": 60,
  "unauthenticated_ratelimit_max_burst": 50,
  "admin_session_lifetime": 20,
  "org_groups_enabled": true,
  "increase_note_size_limit": true,
  "_max_note_size": 100000,
  "enforce_single_org_with_reset_pw_policy": true,
  "dns_prefer_ipv6": false,
  "sso_enabled": true,
  "sso_only": true,
  "sso_allow_unknown_email_verification": false,
  "sso_client_id": "************************************************************************",
  "sso_client_secret": "***",
  "sso_authority": "*****://*******************",
  "sso_scopes": "email profile",
  "sso_authorize_extra_params": "access_type=offline&prompt=consent",
  "sso_pkce": true,
  "sso_audience_trusted": null,
  "sso_callback_path": "*****://************************************************",
  "sso_master_password_policy": "{\"enforceOnLogin\":true,\"minComplexity\":4,\"minLength\":12,\"requireLower\":true,\"requireNumbers\":true,\"requireSpecial\":true,\"requireUpper\":true}",
  "sso_debug_tokens": false,
  "_enable_yubico": true,
  "yubico_client_id": null,
  "yubico_secret_key": null,
  "yubico_server": null,
  "_enable_duo": false,
  "duo_use_iframe": false,
  "duo_ikey": null,
  "duo_skey": null,
  "duo_host": null,
  "_duo_akey": null,
  "_enable_smtp": true,
  "use_sendmail": false,
  "sendmail_command": null,
  "smtp_host": "**************",
  "smtp_ssl": null,
  "smtp_explicit_tls": null,
  "smtp_security": "starttls",
  "smtp_port": 587,
  "smtp_from": "**********************",
  "smtp_from_name": "***********",
  "smtp_username": "********************",
  "smtp_password": "***",
  "smtp_auth_mechanism": null,
  "smtp_timeout": 15,
  "helo_name": null,
  "smtp_embed_images": true,
  "_smtp_img_src": "***:",
  "smtp_debug": false,
  "_enable_email_2fa": true,
  "email_token_size": 6,
  "email_expiration_time": 600,
  "email_attempts_limit": 3,
  "email_2fa_auto_fallback": false
}
```

</details>

### Vaultwarden Build Version

1.37.2 (built from tag 1.37.2, commit 46d71107f5094460dd5ecbe1dbac6e6c71e5189a)

### Deployment method

Build from source

### Custom deployment method

Built from the upstream release tag with ./docker/bake.sh alpine, single-platform
linux/amd64, build arg DB=postgresql,s3 (--provenance=false --sbom=false).
Alpine/musl target. Run via Docker Compose. PostgreSQL backend, and DATA_FOLDER
pointed at an S3 bucket (s3://<bucket>) for config + attachments.

The identical build recipe on tag 1.37.1 works correctly.

### Reverse Proxy

AWS Application Load Balancer (managed, no version). Not relevant to this issue — the server process exits before it binds/serves.

Linux

### Operating System Version

Ubuntu 22.04.5 LTS (x86_64), Docker 29.1.1

### Clients

Web Vault

### Client Version

N/A — the server fails at startup, before any client can connect.

### Steps To Reproduce

1. Check out tag 1.37.2.
2. Build with the s3 feature, e.g. `cargo build --features postgresql,s3`
   (we build via `./docker/bake.sh alpine` with build arg `DB=postgresql,s3`).
   The build succeeds with no warnings or errors.
3. Configure `DATA_FOLDER=s3://<bucket>` so config.json and attachments live in S3.
4. Start the server.

The same steps on tag 1.37.1 work correctly.

### Expected Result

The server reads its configuration from the S3 data folder and starts normally,
as it does on 1.37.1 built the same way — logging
"Using saved config from `s3://<bucket>/config.json`" followed by
"Rocket has launched".

### Actual Result

The server fails during startup, before Rocket launches, and the container
crash-loops. It cannot access the S3 data folder at all, reporting that no
default HTTP transport is installed.

Root cause appears to be the opendal 0.57.0 -> 0.58.1 bump in 1.37.2 combined
with `default-features = false` on the opendal dependency:

* Cargo.toml declares (identical in 1.37.1 and 1.37.2 apart from the version):

      opendal = { version = "0.58.1", default-features = false, features = ["services-fs"] }

  and the `s3` feature adds only `opendal/services-s3`. The `s3` feature block is
  byte-identical between 1.37.1 and 1.37.2.

* opendal 0.58 made the HTTP transport pluggable ("refactor: introduce HTTP
  transporter API", apache/opendal#7759; "RFC: Add HTTP transporter design",
  apache/opendal#7749) behind the `http-transport-reqwest` feature, which is a
  DEFAULT feature — and therefore stripped by `default-features = false`.

* This is visible in the resolved lock files: `reqwest` is a dependency of
  `opendal-core` 0.57.0 but NOT of `opendal-core` 0.58.1:

      # Cargo.lock @ 1.37.1
      name = "opendal-core"
      version = "0.57.0"
      dependencies = [ ..., "reqsign-core", "reqwest", "serde", ... ]

      # Cargo.lock @ 1.37.2
      name = "opendal-core"
      version = "0.58.1"
      dependencies = [ ..., "reqsign-core",            "serde", ... ]

  In 0.57.0 the transport came in unconditionally via opendal-core, so
  `default-features = false` was harmless. In 0.58.1 it does not, so no transport
  is compiled in and every HTTP-based service (i.e. S3) fails on first use.

Suggested fix — re-enable a transport alongside the S3 service:

      s3 = [
          "opendal/services-s3",
          "opendal/http-transport-reqwest",   # required since opendal 0.58
          "dep:aws-config",
          "dep:aws-credential-types",
          "dep:aws-smithy-runtime-api",
          "dep:http",
          "dep:reqsign-aws-v4",
          "dep:reqsign-core",
      ]

### Logs

```text
[WARNING] Secure Note size limit is increased to 100_000!
[WARNING] This could cause issues with clients. Also exports will not work on Bitwarden servers!.
[reqsign_aws_v4::sign_request][DEBUG] calculated scope: 20260824/us-east-2/s3/aws4_request
[reqsign_aws_v4::sign_request][DEBUG] calculated string to sign: AWS4-HMAC-SHA256
[vaultwarden][ERROR] Could not access S3 data folder 's3://<bucket>': ConfigInvalid (permanent) at list => default HTTP transport is not installed

Context:
   advice: call HttpTransporter::install_default before using HTTP services
   called: HttpTransporter::default
   service: s3
   path: /
   listed: 0

(then the startup banner repeats as the container restarts, looping)
```

### Screenshots or Videos

_No response_

### Additional Context

The "Clients" dropdown is set to Web Vault only because the field is required —
this is a server-side startup failure with no client involved.

On CI coverage: build.yml does run cargo test and cargo clippy with
sqlite,mysql,postgresql,enable_mimalloc,s3, so the feature compiles and unit
tests pass. The failure only surfaces on a live S3 request, and no published
image ships the s3 feature (Dockerfile.alpine defaults to
DB=sqlite,mysql,postgresql,enable_mimalloc), which is likely why it went
unnoticed.

txase:
Use Vaultwarden’s shared HTTP client configuration for OpenDAL storage
and AWS SDK traffic.

- Inject a client built by `get_reqwest_client_builder(false)` into
  OpenDAL’s reqwest transport. This avoids OpenDAL creating a standalone
  client that bypasses Vaultwarden’s DNS, redirect, proxy, TLS, timeout,
  and request configuration.
- Make `enforce_block = false` apply consistently to redirects as well
  as DNS resolution. Administrator-configured storage, metadata, and
  service endpoints may legitimately use private addresses.
- Make DNS resolver construction safe while `CONFIG` is initializing.
  `CONFIG` could be backed by a remote OpenDAL file fetched over HTTP.
  The resolver now queries both address families without consulting
  `CONFIG`, then orders each lookup using the loaded IPv6 preference.
  During bootstrap it falls back to the env var `DNS_PREFER_IPV6`, then
  IPv4-first.
- Construct the AWS SDK connector through the same shared HTTP builder.
  This was an additional latent issue in the current implementation.

### Validation

- Runtime smoke testing of remote configuration loading, login, sync,
  cipher operations, and attachment upload/download
- No errors, throttles, or HTTP 5xx responses during testing

Fixes #7630

### DIFF
--- a/src/http_client.rs
+++ b/src/http_client.rs
@@ -356,10 +356,29 @@ pub(crate) mod aws {
     };
     use reqwest::Client;
 
+    use super::get_reqwest_client_builder;
+
     // Adapter that wraps reqwest to be compatible with the AWS SDK
     #[derive(Debug)]
     pub(crate) struct AwsReqwestConnector {
-        pub(crate) client: Client,
+        client: Client,
+    }
+
+    impl AwsReqwestConnector {
+        pub(crate) fn new() -> Self {
+            let client = get_reqwest_client_builder(false).build().expect("Failed to build AWS HTTP client");
+            Self {
+                client,
+            }
+        }
+    }
+
+    fn connector_error(error: reqwest::Error) -> ConnectorError {
+        if error.is_timeout() {
+            ConnectorError::timeout(Box::new(error))
+        } else {
+            ConnectorError::io(Box::new(error))
+        }
     }
 
     impl HttpConnector for AwsReqwestConnector {
@@ -379,10 +398,10 @@ pub(crate) mod aws {
                     req_builder = req_builder.body(body_bytes.to_vec());
                 }
 
-                let response = req_builder.send().await.map_err(|e| ConnectorError::io(Box::new(e)))?;
+                let response = req_builder.send().await.map_err(connector_error)?;
 
                 let status = response.status().into();
-                let bytes = response.bytes().await.map_err(|e| ConnectorError::io(Box::new(e)))?;
+                let bytes = response.bytes().await.map_err(connector_error)?;
 
                 Ok(HttpResponse::new(status, bytes.into()))
             };
--- a/src/http_client.rs
+++ b/src/http_client.rs
@@ -36,7 +36,7 @@ pub fn get_reqwest_client_builder(enforce_block: bool) -> ClientBuilder {
     let mut headers = header::HeaderMap::new();
     headers.insert(header::USER_AGENT, header::HeaderValue::from_static("Vaultwarden"));
 
-    let redirect_policy = reqwest::redirect::Policy::custom(|attempt| {
+    let redirect_policy = reqwest::redirect::Policy::custom(move |attempt| {
         if attempt.previous().len() >= 5 {
             return attempt.error("Too many redirects");
         }
@@ -45,7 +45,7 @@ pub fn get_reqwest_client_builder(enforce_block: bool) -> ClientBuilder {
             return attempt.error("Invalid host");
         };
 
-        if let Err(e) = should_block_host(&host) {
+        if enforce_block && let Err(e) = should_block_host(&host) {
             return attempt.error(e);
         }
 
--- a/src/http_client.rs
+++ b/src/http_client.rs
@@ -14,7 +14,10 @@ use reqwest::{
 };
 use url::Host;
 
-use crate::{CONFIG, util::is_global};
+use crate::{
+    CONFIG,
+    util::{get_env_bool, is_global},
+};
 
 pub fn make_http_request(method: reqwest::Method, url: &str) -> Result<reqwest::RequestBuilder, crate::Error> {
     static INSTANCE: LazyLock<Client> =
@@ -57,6 +60,14 @@ pub fn get_reqwest_client_builder(enforce_block: bool) -> ClientBuilder {
         .redirect(redirect_policy)
         .dns_resolver(CustomDns::instance(enforce_block))
         .timeout(Duration::from_secs(10))
+}
+
+fn dns_prefer_ipv6() -> bool {
+    // CONFIG may require DNS to initialize, so avoid forcing it during bootstrap.
+    match LazyLock::get(&CONFIG) {
+        Some(config) => config.dns_prefer_ipv6(),
+        None => get_env_bool("DNS_PREFER_IPV6").unwrap_or(false),
+    }
 }
 
 fn should_block_ip(ip: IpAddr) -> bool {
@@ -258,12 +269,8 @@ impl CustomDnsResolver {
     fn new() -> Arc<Self> {
         TokioResolver::builder(TokioRuntimeProvider::default())
             .and_then(|mut builder| {
-                // Hickory's default since v0.26 is `Ipv6AndIpv4`, which sorts IPv6 first
-                // This might cause issues on IPv4 only systems or containers
-                // Unless someone enabled DNS_PREFER_IPV6, use Ipv4AndIpv6, which returns IPv4 first which was our previous default
-                if !CONFIG.dns_prefer_ipv6() {
-                    builder.options_mut().ip_strategy = hickory_resolver::config::LookupIpStrategy::Ipv4AndIpv6;
-                }
+                // Query both families; the preferred order is applied per lookup below.
+                builder.options_mut().ip_strategy = hickory_resolver::config::LookupIpStrategy::Ipv4AndIpv6;
                 builder.build()
             })
             .inspect_err(|e| warn!("Error creating Hickory resolver, falling back to default: {e:?}"))
@@ -286,6 +293,14 @@ impl CustomDnsResolver {
         }
 
         Ok(results)
+    }
+}
+
+fn sort_addresses(addresses: &mut [SocketAddr], prefer_ipv6: bool) {
+    if prefer_ipv6 {
+        addresses.sort_by_key(SocketAddr::is_ipv4);
+    } else {
+        addresses.sort_by_key(SocketAddr::is_ipv6);
     }
 }
 
@@ -320,7 +335,9 @@ impl Resolve for CustomDns {
         let this = Arc::clone(&self.resolver);
         Box::pin(async move {
             let name = name.as_str();
-            let results = this.resolve_domain(name, enforce_block).await?;
+            let mut results = this.resolve_domain(name, enforce_block).await?;
+            // Recheck after bootstrap so long-lived clients adopt the loaded config.
+            sort_addresses(&mut results, dns_prefer_ipv6());
             if results.is_empty() {
                 warn!("Unable to resolve {name} to any valid IP address");
             }
@@ -391,7 +408,7 @@ pub(crate) mod aws {
 mod tests {
     use super::*;
     use crate::util::is_global_hardcoded;
-    use std::net::Ipv4Addr;
+    use std::net::{Ipv4Addr, Ipv6Addr};
     use url::Host;
 
     // ===
@@ -402,6 +419,26 @@ mod tests {
             Host::Ipv6(v6) => Some(IpAddr::V6(v6)),
             Host::Domain(_) => None,
         }
+    }
+
+    #[test]
+    fn dns_setup_does_not_initialize_config() {
+        assert!(LazyLock::get(&CONFIG).is_none());
+        drop(CustomDns::instance(false));
+        assert!(LazyLock::get(&CONFIG).is_none());
+    }
+
+    #[test]
+    fn dns_preference_orders_addresses() {
+        let ipv4 = SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), 0);
+        let ipv6 = SocketAddr::new(IpAddr::V6(Ipv6Addr::LOCALHOST), 0);
+        let mut addresses = [ipv6, ipv4];
+
+        sort_addresses(&mut addresses, false);
+        assert_eq!(addresses, [ipv4, ipv6]);
+
+        sort_addresses(&mut addresses, true);
+        assert_eq!(addresses, [ipv6, ipv4]);
     }
 
     #[test]
--- a/src/http_client.rs
+++ b/src/http_client.rs
@@ -297,6 +297,9 @@ impl CustomDnsResolver {
 }
 
 fn sort_addresses(addresses: &mut [SocketAddr], prefer_ipv6: bool) {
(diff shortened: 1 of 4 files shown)


src/http_client.rs
```diff
@@ -289,6 +296,14 @@ impl CustomDnsResolver {
     }
 }
 
+fn sort_addresses(addresses: &mut [SocketAddr], prefer_ipv6: bool) {
+    if prefer_ipv6 {
+        addresses.sort_by_key(SocketAddr::is_ipv4);
+    } else {
+        addresses.sort_by_key(SocketAddr::is_ipv6);
+    }
```
tessus, line 307: something seems off here. Or maybe I misunderstood, but if you prefer ipv6 shouldn't they be sorted by the ipv6 addresses?

txase: The current ordering is correct because boolean keys sort false before true. When IPv6 is preferred, IPv6 addresses return false for is_ipv4() and therefore sort first; the inverse applies when IPv4 is preferred. The unit test below covers both cases. That said, this is harder to read than I'd like. If I add a comment above this condition block like the following would it address your concern?

```rust
    // `sort_by_key()` puts `false` values before `true` values
    if prefer_ipv6 {
        addresses.sort_by_key(SocketAddr::is_ipv4);
    } else {
        addresses.sort_by_key(SocketAddr::is_ipv6);
    }
```

Though that's sort of pedantic about core Rust behavior. I can't figure out anything more clear, though. At least the unit test proves the right behavior?

tessus: All good. It just means that it was my fault and my misunderstanding.

stefan0xC: Since false is returned before true you could get rid of the if and do the sorting like that: 
```rust
addresses.sort_by_key(|addr| addr.is_ipv4() == prefer_ipv6);
```
i.e. if it's an ipv6 address it will be false and thus depending on whether you prefer it or not it will be sorted first, no?

txase: I thought about ideas like that, but I think the root issue is more that `sort_by_key()` orders `false` before `true`, rather than what the conditional is itself. I think most people might read your suggestion like @tessus read the PR's version, still: "Why are we comparing `is_ipv4()` to `prefer_ipv6`? Why shouldn't this be `addr.is_ipv4() == prefer_ipv4`?"

stefan0xC: Ah, okay. I feel like it wouldn't be as obvious as with the verbose if else branch but what about

```rust
addresses.sort_by_key(|addr| !(addr.is_ipv6() == prefer_ipv6)); // we negate here because false (0) sorts before true (1)
```

BlackDex: Well, it does look weird, if anything it should at least have good comments explaining why/how.

BlackDex (changes requested):
I think the IP sort order should at least have some comments explaining it, as it now kinda looks as an error.

Besides that, this still does bring in reqwest, while for now probably not an issue, it might in the future. The only way to fully not bring in reqwest via any other crate in this case is creating the HttpTransport your self instead of relying on the crate.

But again, it might be fine for now. We might need to revisit this if we bump into a next incompatibility with reqwest maybe?

txase:
I'm happy to add a comment to make the conditional more clear.

On bringing in reqwest, vaultwarden already depends on it. Are you concerned about bringing in the new opendal-http-transport-reqwest crate specifically? Or that this creates a separate dependency chain for reqwest? Solving either would require reimplementing significant amounts of code from the new package, which felt worse to me when I looked into it. It was starting to become hundreds of lines of tricky buffer handling and other forms of difficult-to-prove-correct code.

But I'd be happy to give it a go, even if it's just to compare the two approaches? Let me know what you'd like to see here, @BlackDex.

BlackDex:
Yea, a comment would be best, not that someone (or some LLM) thinks otherwise and tries to adjust it, and we maybe think, ah, yes that sounds logical.

Regarding reqwest. The HttpTransport didn't looked that hard, but i might have missed some items though.
But then again, only copying the same code is probably useless and error prone.
I'm only concerned for a future breaking release of reqwest, and if that would either bring in multiple versions of reqwest or keep us at an older version for a longer time, as we had for a while with `openidconnect`/`oauth2` crate.

But again, currently that should not be an issue, and if needed, we can either update that crate by sending a PR or still copy/paste that code into Vaultwarden it self if really needed.

BlackDex (approved):

txase:
Thanks @BlackDex! I meant to get to the comment, but the US holiday weekend got in the way. Thanks for doing it for me and merging it in.

BlackDex:
Yea no prob! I know some users were waiting for it, and better to have it in testing yet right now.
And, also updated opendal, since that had some changes too, but none which needed code adjustments as far as i could tell.

PR #7651, opened 2026-08-25
fix(deps): bump h2 to 0.4.16 to patch RUSTSEC-2026-0258

Svector-anu:
Automated dependency bump to address a disclosed advisory.

- **Advisory:** [RUSTSEC-2026-0258](https://rustsec.org/advisories/RUSTSEC-2026-0258) (unbounded empty DATA frames → DoS)
- **Severity:** medium
- **Package:** `h2` `0.4.15` → `0.4.16`
- **Reachability:** pulled via `hyper` ← `reqwest` (icon fetch / outbound HTTP)

### Verification
- Reproduced locally: yes (lockfile resolution)
- Command: `cargo update -p h2 --precise 0.4.16`
- Before: `Cargo.lock` pinned `h2 0.4.15`
- After: `Cargo.lock` pins `h2 0.4.16`
- Environment: osv-scanner 2.5.1 flagged the advisory on current `main`

Detected by [osv-scanner](https://google.github.io/osv-scanner/). No code changes outside the lockfile.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1817,7 +1817,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


tessus:
why did you downgrade other packages?

BlackDex:
Also, we normally update multiple packages at the same time.
And, there is also a Rust version update, so we will probably create a PR which holds all package updates and not just a single item.

Also, if we would have liked these kind of updates, we would have enabled dependabot hehe, which we do not for the reason i mentioned above.

And, i also do find it strange that those packages are downgraded.
Since this isn't going to be a PR I'm going to merge, I will close this.

PR #7656, opened 2026-08-26
log_event take enum parameter not i32

Timshel:
Simpler to cast in the function rather than at all the calling site.


BlackDex:
Seems reasonable

BlackDex (approved):

PR #7671, opened 2026-08-31
Fix group import to update groups on data change

Jogius:
The `/public/organization/import` endpoint on the public API, which is used for example by Bitwarden's [directory-connector](https://github.com/bitwarden/directory-connector), correctly updates a user's information on change, but ignores any changes made to groups. I believe this is not the expected behavior, as it prevents the tool from correctly keeping the data in sync with the directory.

The Bitwarden server implementation of the same endpoint ([see here](https://github.com/bitwarden/server/blob/9ebe16587175b1c0e9208f84397bb75d0d595510/src/Core/AdminConsole/Services/Implementations/OrganizationService.cs#L1382-L1395)) checks whether the group name has changed and updates it.

```cs
foreach (var group in updateGroups)
{
    var updatedGroup = groupsDict[group.ExternalId].Group;
    if (group.Name != updatedGroup.Name)
    {
      group.RevisionDate = DateTime.UtcNow;
      group.Name = updatedGroup.Name;

      await _groupRepository.ReplaceAsync(group);
    }

    await UpdateUsersAsync(group, groupsDict[group.ExternalId].ExternalUserIds,
      existingExternalUsersIdDict,
      existingGroupUsers.ContainsKey(group.Id) ? existingGroupUsers[group.Id] : null);
}
```

I applied the same logic as found in Bitwarden's implementation to update group names too.

Hope this helps, and I'm happy to iterate / update this if you have any feedback!


Jogius:
Hello again, can I get brief feedback on whether this is something you could see merging? Depending on your answer, I can decide whether the [directory-connector](https://github.com/bitwarden/directory-connector) is a viable option for my org, or whether we need to build something ourself for directory syncing! Appreciate your work, thanks :)

PR #7676, opened 2026-09-01
Misc Updates

BlackDex:
- Update Rust to v1.98.0
- Update all the crates and adjusted code where needed
- Updated JavaScript libraries Removed jquery as this isn't needed anymore, adjusted code where needed
- Updated all GitHub Actions
- Fixed nightly clippy lint warnings


src/util.rs
```diff
@@ -537,10 +537,7 @@ pub fn is_valid_email(email: &str) -> bool {
     let Ok(email_url) = url::Url::parse(&format!("https://{}", email.domain())) else {
         return false;
     };
-    if email_url.path().ne("/") || email_url.domain().is_none() || email_url.query().is_some() {
-        return false;
-    }
-    true
+    !(email_url.path().ne("/") || email_url.domain().is_none() || email_url.query().is_some())
```
stefan0xC: Isn't this still too complicated? Wouldn't it be easier to assert the expected conditions are met?
```suggestion
    email_url.path() == "/" && email_url.domain().is_some() && email_url.query().is_none()
```

stefan0xC (changes requested):

src/util.rs
BlackDex: Too complicated? It's the exact same as it was, but then the clippy lint fixed.

The `&&` statement stops at the first wrong item, the `||` needs to check them all. The order might be changed maybe. Not sure if it would matter that much though.

But positive checks do look better though, but not sure if that makes this check more clear in general.

stefan0xC: I don't think the order matters and I am pretty sure that [`||` is also a lazy boolean operator](https://doc.rust-lang.org/reference/expressions/operator-expr.html#lazy-boolean-operators) that does not check everything if one condition is true. The happy path (a valid email address) would have to evaluate each term in both cases since it only returns true if either everything is true (with positive checks) or if every negative check is false.

I think it's more clear, that a valid email address does need to have a domain and neither path nor query. So for better readability we could sort them like that so the checks are done in a more reasonable order?
```rust
email_url.domain().is_some() && email_url.path() == "/" && email_url.query().is_none()
```

BlackDex: Adjusted, thanks.

dani-garcia (approved):

PR #7677, opened 2026-09-01
Add `pm-32413-multi-client-password-management` feature flag

tom27052006:
Adds the `pm-32413-multi-client-password-management` flag to the supported client feature flags allowlist. Bitwarden gates changing the master password directly in the browser extension and desktop client behind this flag (clients >= 2026.4.0); without it self-hosted users are sent to the web vault instead.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -390,6 +390,7 @@
 ##
 ## The following flags are available:
 ## - "pm-5594-safari-account-switching": Enable account switching in Safari. (Safari >= 2026.2.0)
+## - "pm-32413-multi-client-password-management": Enable changing the master password directly in the client. (Desktop/Extension >= 2026.4.0)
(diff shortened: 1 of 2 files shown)


Timshel:
Not a maintainer but I'll mention that restricting to only the web vault has it's use since it's the only client where the version is controlled.

BlackDex:
> Not a maintainer but I'll mention that restricting to only the web vault has it's use since it's the only client where the version is controlled.

And yet, we missed that too haha. But that is a fair point. But, ill reckon Bitwarden will make it default sometime. But then it would be at least more tested and mainstream by them too.

tom27052006:
Yeah, I get that. Still, maybe it could make sense to keep it as an optional opt-in until Bitwarden enables it by default anyway, since it would already be quite useful in some environments like mine

BlackDex (approved):

PR #7678, opened 2026-09-02
perf(db): add indexes on foreign key and lookup columns

BryanFRD:
The schema has no secondary indexes. Every index today comes from a `PRIMARY KEY` or a `UNIQUE` constraint, so any lookup on a foreign key column, or on the trailing column of a composite key, falls back to a full table scan.

This adds 18 indexes on SQLite and PostgreSQL, and 17 on MySQL. MySQL skips `archives.cipher_uuid`, the one table declaring a real `FOREIGN KEY` clause, since InnoDB already maintains an index for it. The other tables use inline `REFERENCES`, which MySQL parses and ignores, so no index exists there either.

Measured on PostgreSQL 17 with 200 users, 70k ciphers, 20k attachments and 300k events:

| Query | Before | After |
|---|---|---|
| `Cipher::find_owned_by_user` | 4.50 ms | 0.09 ms |
| `Cipher::find_by_org` | 2.66 ms | 0.12 ms |
| Attachment sync prefetch | 4.91 ms | 1.41 ms |
| `Event::find_by_organization_uuid` | 18.35 ms | 0.19 ms |
| `event_cleanup_job` daily delete | 52.12 ms | 4.28 ms |

Write cost is about 10% on a bulk insert of 5000 ciphers, 30.1 ms against 31.8 ms median.

Note that the main sync query in `Cipher::find_by_user` still sequential-scans `ciphers`. Its `WHERE` clause is a chain of `OR`s across several joined tables, so the planner cannot drive it from an index. The gains here are on the surrounding lookups, not on that scan.

All migrations were applied in order against fresh SQLite, PostgreSQL 17 and MySQL 8 databases, and the down migrations were verified to roll back cleanly.

On large PostgreSQL instances the `CREATE INDEX` on `ciphers` and `event` takes a table lock. `CREATE INDEX CONCURRENTLY` cannot run inside a diesel migration transaction, so this is worth a note in the release notes.


Timshel:
Not a maintainer but recently had some [issue](https://github.com/dani-garcia/vaultwarden/pull/7265) where MariaDB changed the default naming convention for foreign keys.
Wonder if it could affect some names here too ?

BryanFRD:
Good instinct, it does affect this PR, though not through the names.

The indexes here are named explicitly, so the constraint naming change does not reach them. I did reproduce it while checking: MariaDB 11.8 names the constraint `b_ibfk_1`, 12.3 names it `1`.

What it did surface is that MySQL ignores inline `REFERENCES` while MariaDB honours them, and has since at least 11.x. On MariaDB those columns already carry an auto-named backing index, so the comment in `up.sql` was wrong about them.

The up migration is fine either way: InnoDB drops the auto-named index once the named one can serve the constraint, so MySQL 8, MariaDB 11.8 and 12.3 all end with exactly one index per column, no duplicates.

The down migration was not fine. On MariaDB it failed with `ERROR 1553: Cannot drop index 'idx_ciphers_user_uuid': needed in a foreign key constraint`, since these indexes had become the only ones backing those constraints. Fixed by setting `FOREIGN_KEY_CHECKS = 0` in `down.sql`, matching what `run_migrations` already does for the up path.

Up and down now both verified on all three engines. Thanks for the pointer.

PR #7694, opened 2026-09-05
feat: allow overriding sso endpoint

nico151999:
The goal is to introduce an environment variable (`SSO_INTERNAL_ENDPOINT`) that can override the base URL of the discovery endpoint (by default derived from the SSO authority) and, if the returned discovered URLs have the same same base URL as the SSO authority, override their base URL with the `SSO_INTERNAL_ENDPOINT` one as well.

However, if set, the variables `SSO_TOKEN_ENDPOINT`, `SSO_USERINFO_ENDPOINT` and `SSO_JWKS_URI`, which contain full URIs (not only base URLs) take precedence. These 3 variables mentioned here are for discussion. Do we want to introduce them or stick to  `SSO_INTERNAL_ENDPOINT` only?

I tested the overrides via variables in my K8s cluster with a cluster-internal SSO provider (Hydra) and it works well.

### DIFF
--- a/src/sso_client.rs
+++ b/src/sso_client.rs
@@ -3,9 +3,9 @@ use std::{borrow::Cow, collections::HashSet, future::Future, pin::Pin, sync::Laz
 use openidconnect::{
     AccessToken, AsyncHttpClient, AuthDisplay, AuthPrompt, AuthType, AuthenticationFlow, AuthorizationCode,
     AuthorizationRequest, ClientId, ClientSecret, CsrfToken, EmptyAdditionalClaims, EmptyExtraTokenFields,
-    EndpointNotSet, EndpointSet, HttpClientError, HttpRequest, HttpResponse, IdTokenClaims, IdTokenFields, Nonce,
-    OAuth2TokenResponse, PkceCodeChallenge, PkceCodeVerifier, RefreshToken, ResponseType, Scope, StandardErrorResponse,
-    StandardTokenResponse,
+    EndpointNotSet, EndpointSet, HttpClientError, HttpRequest, HttpResponse, IdTokenClaims, IdTokenFields,
+    JsonWebKeySetUrl, Nonce, OAuth2TokenResponse, PkceCodeChallenge, PkceCodeVerifier, RefreshToken, ResponseType,
+    Scope, StandardErrorResponse, StandardTokenResponse, TokenUrl, UserInfoUrl,
     core::{
         CoreAuthDisplay, CoreAuthPrompt, CoreClient, CoreClientAuthMethod, CoreErrorResponseType, CoreGenderClaim,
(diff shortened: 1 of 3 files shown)


Timshel:
I know it was mentioned, but is there an identified need to have separate url for endpoints ?

src/sso_client.rs
```diff
@@ -77,12 +77,45 @@ impl OidcHttpClient {
...
+pub(crate) fn rewrite_oidc_url(public: &str, internal: &str, original: &str) -> Option<String> {
+    if public.trim().is_empty() || internal.trim().is_empty() {
+        return None;
+    }
+    let public_trim = public.trim_end_matches('/');
+    let internal_trim = internal.trim_end_matches('/');
+    if public_trim == internal_trim {
+        return None;
+    }
+    let suffix = original.strip_prefix(public_trim)?;
```
Timshel: I believe it will cause issue if the url has been set with for ex `sso_userinfo_endpoint` ?

nico151999:
> I know it was mentioned, but is there an identified need to have separate url for endpoints ?

Well, yes, it was mentioned, but I also find it reasonable because the openid discovery endpoint returns URLs that might be routed differently internally. In my specific use case I only need to define an internal override for scheme (http/https), hostname and port, so basically the base URL to support internal routing that doesn't needlessly leave the cluster before it enters it again. So for my specific use case we only need one variable. The other variables I added for people with different use cases where they have different routing rules, e.g. a different internal reverse proxy.

On the other hand we should expect people to be able to at least be able to reconfigure the internal load balancer for such use cases. When it comes to TLS termination and DNS rules I think we should not expect people to change that since it would feel more like a workaround in many setups (e.g. I do not want to reconfigure coredns in a K8s cluster for the requirements of a single service). Also, a sidecar for re-routing traffic would be overkill. So we should at least offer the `sso_internal_endpoint`. The others are debatable and we can remove them again if desired. What do you think @Timshel and @BlackDex?

src/sso_client.rs
nico151999: I refactored the implementation and also tested the specific overrides now. Works well in my K8s setup. The goal is to introduce an environment variable (`SSO_INTERNAL_ENDPOINT`) that can override the base URL of the discovery endpoint (by default derived from the SSO authority) and, if the returned discovered URLs have the same same base URL as the SSO authority, override their base URL with the `SSO_INTERNAL_ENDPOINT` one as well.

However, if set, the variables `SSO_TOKEN_ENDPOINT`, `SSO_USERINFO_ENDPOINT` and `SSO_JWKS_URI`, which contain full URIs (not only base URLs) take precedence. These 3 variables mentioned here are for discussion. Do we want to introduce them or stick to  `SSO_INTERNAL_ENDPOINT` only?

nico151999:
@BlackDex just wanted to ping you in case this got lost. I'd like to hear your opinion on this. Thanks!

PR #7714, opened 2026-09-08
Fix iOS registration token response

tom27052006:
When SIGNUPS_VERIFY is disabled, /identity/accounts/register/send-verification-email returns the token as a JSON string ("eyJ..."). The iOS client reads the raw response body as the token, so the quotes become part of it and /identity/accounts/register/finish fails with Error decoding JWT: Error(Base64(InvalidByte(0, 34))).

The iOS client sends Accept: */*. This change now returns JSON only when the client explicitly requests application/json; otherwise, the token is returned as text/plain:

* Accept: application/json → "eyJ..."
* all other cases → raw eyJ...

Returning a plain String for every request would not be enough: api.service.ts only parses the body when the content type is application/json, so clients expecting JSON could silently get undefined. Clients that require JSON explicitly request it, while all other requests receive the raw token as plain text. The 204 No Content response when email verification is required is unchanged.

Tested on iOS, signup works now. Web Vault also works as expected. I don’t have an Android device, so I could only test the response behavior with curl.

Discussion: https://github.com/dani-garcia/vaultwarden/discussions/7712

### DIFF
--- a/src/api/identity.rs
+++ b/src/api/identity.rs
@@ -1086,10 +1086,9 @@ enum RegisterVerificationResponse {
     PlainToken(String),
 }
 
-// The iOS client sends `Accept: */*` and reads the raw body as its token, so it would end up with
-// the quotes of a JSON string. Every other client keeps the JSON string.
+// Return JSON only when the client explicitly requests it, otherwise return plain text.
 fn accepts_json(accept: Option<&Accept>) -> bool {
-    accept.is_none_or(|accept| accept.preferred().media_type() != &MediaType::Any)
+    accept.is_some_and(|accept| accept.preferred().media_type() == &MediaType::JSON)
 }
 
 #[post("/accounts/register/send-verification-email", data = "<data>")]
--- a/src/api/identity.rs
+++ b/src/api/identity.rs
@@ -3,7 +3,7 @@ use num_traits::FromPrimitive;
 use rocket::{
     Route,
     form::{Form, FromForm},
-    http::{Cookie, CookieJar, SameSite},
+    http::{Accept, Cookie, CookieJar, MediaType, SameSite},
     response::Redirect,
     serde::json::Json,
 };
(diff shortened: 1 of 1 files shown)


BlackDex:
What is the exact header the iOS client sends to the server?
Because, this current change now changes the default response to be text if no header is sent, which might be a breaking change. If anything, it would probably be better to check for text header to keep our current way of working.

But, if you are able to provide the headers sent by the iOS client (Since i do not have an iOS device) then we can beter determine what to do.

tom27052006:
> What is the exact header the iOS client sends to the server? Because, this current change now changes the default response to be text if no header is sent, which might be a breaking change. If anything, it would probably be better to check for text header to keep our current way of working.
> 
> But, if you are able to provide the headers sent by the iOS client (Since i do not have an iOS device) then we can beter determine what to do.

I just checked the actual request from the iOS app, it sends

```http
Accept: */*
```

So you're right about the no-header case, I'll adjust it so `*/*` returns the raw token, while no `Accept` header keeps the current JSON response.

BlackDex:
Ok, the Android Client does send type it wants.

Android: `Content-Type: text/plain; charset=utf-8`
Web-Vault (FireFox): `Content-Type: application/json; charset=utf-8`
Desktop: `Content-Type: application/json; charset=utf-8`

I have not tested the Browser Extension, but that probably does the same as web and desktop i would guess.

tom27052006:
> Ok, the Android Client does send type it wants.
> 
> Android: `Content-Type: text/plain; charset=utf-8` Web-Vault (FireFox): `Content-Type: application/json; charset=utf-8` Desktop: `Content-Type: application/json; charset=utf-8`
> 
> I have not tested the Browser Extension, but that probably does the same as web and desktop i would guess.

Also tested the browser extension: it also sends `Accept: application/json` like the Web Vault or the Desktop App

src/api/identity.rs
```diff
@@ -1083,11 +1083,19 @@ enum RegisterVerificationResponse {
...
     NoContent(()),
     Token(Json<String>),
+    PlainToken(String),
+}
+
+// The iOS client sends `Accept: */*` and reads the raw body as its token, so it would end up with
+// the quotes of a JSON string. Every other client keeps the JSON string.
+fn accepts_json(accept: Option<&Accept>) -> bool {
+    accept.is_none_or(|accept| accept.preferred().media_type() != &MediaType::Any)
 }
```
BlackDex: This of course is not correct anymore.

Also, as seen in all the results, all clients which expect and except the JSON format send a header which states that they support it. The Android client does except the same JSON format though, while it states it just want `plain/text`.

So i think the first iteration would have been better in this case.
If the client says it wants JSON, return JSON, otherwise return text.

We just needed to be sure what the clients are sending before just assuming.

BlackDex (changes requested):

src/api/identity.rs
tom27052006: Ah, I misunderstood that before. It now only returns JSON when the client explicitly asks for it, otherwise it returns plain text.

BlackDex (approved):

PR #7735, opened 2026-09-13
Normalize URL comparison in admin diagnostics checkDomain

7heMech:
Should resolve 
Screenshot_20260913-221915


BlackDex:
That duplicate `/` shouldn't be there it might also cause issues with the web-vault of that is there

7heMech:
> That duplicate `/` shouldn't be there it might also cause issues with the web-vault of that is there 

oh, well I mistyped but realized script thought it was issue, but server parses ok.

PR #7743, opened 2026-09-16
Add optional Docker updates to the admin diagnostics page

Skyline-23:
The Diagnostics `Update` badge currently indicates that a newer version exists but cannot start an update. This change makes the existing badge actionable when an administrator explicitly configures `UPDATER_SOCKET`: it shows `Updating…` during the update and reloads the page when the new container is healthy.

## Implementation

- Run Docker operations in a separate host process so they survive Vaultwarden container replacement. Vaultwarden receives an updater Unix socket, not the Docker socket.
- Restrict updates to one host-configured Compose service and image repository. Require admin authentication, reject overlapping requests and unapplied Compose changes, and pin the downloaded image ID before deployment.
- Stop the selected service for a complete data-directory backup, recreate only that service, and verify its Docker health check. Persist deployment state and block further updates when host recovery is needed.
- Mark compatible images so an update cannot silently remove the admin update endpoint. Include host service configuration, setup instructions, and recovery guidance.

Updates remain disabled by default. Without updater configuration, the existing informational badge is unchanged.

## Supported scope

The initial implementation supports a Linux Docker host with Python 3.10+, Docker Compose v2+, one service, and SQLite with all persistent vault data in a dedicated directory bind-mounted at `/data`. External databases/storage, named data volumes, custom environment files, additional writable mounts, and symlinks in the data directory are rejected. The first compatible image and host updater must be installed by the operator.

Failures after the service stop boundary require operator inspection. Database rollback is deliberately manual because a new version may have migrated the database or accepted writes. The image override and backup recovery procedure are documented in `docker/updater/README.md`.

## Validation

- `cargo test --locked --features sqlite -q`: 31 tests passed.
- `cargo clippy --locked --features sqlite -- -D warnings`: passed.
- `cargo fmt --all -- --check`: passed.
- `python3 -m unittest discover -s docker/updater -v`: 10 flow tests passed, including failure, concurrency, storage validation, configuration drift, and interrupted-update recovery.
- `node --check src/static/scripts/admin_updates.js`: passed.
- Regenerated both Dockerfiles from `Dockerfile.j2` and verified byte-for-byte equality.
- Exercised the existing yellow badge in the Codex in-app browser against a local Vaultwarden build. Verified `Updating…`, status reconnection, automatic reload, and the unchanged-image result.
- Ran actual Docker replacements using fixture images and a local registry in an isolated Colima instance. Verified backup contents, pinned-image deployment, and that a neighboring service was not recreated.

The Docker integration run used fixture images on macOS/Colima; it did not validate a production Vaultwarden migration or the Linux systemd unit. No production deployment or public image publication was performed.


qianlongzt:
Why not use GitOps with Doco-CD?

BlackDex:
I'm not really a fan of this. If you want something like this i would suggest to just use watchtower so it will update automatically.

From my point of view it's out of scope for this project.

PR #7746, opened 2026-09-16
[Web 2026.9.0] Support new Emergency Access password payload

tom27052006:
Bitwarden Web 2026.9 changed the Emergency Access takeover password request to use `authenticationData` and `unlockData`.

This updates Vaultwarden to support the new payload while keeping compatibility with older versions.

Upstream:
https://github.com/bitwarden/clients/pull/20643


src/api/core/emergency_access.rs
Timshel, line 689: Question for the maintainers, but I wonder if instead of this logic, the enforce should be used : 
```
two_factor::enforce_2fa_policy(&user, &headers.user.uuid, headers.device.atype, &headers.ip.ip, &conn).await?;
```
I believe when granting the emergency access it's mentioned that the user will gain access to the organizations.

Thinking on it probably would make senseto  revoke access instead,

Timshel:
Wrote a Playwright test which could be included: https://github.com/Timshel/OIDCWarden/commit/31cb915078b2b8222f454521dcb2d9844e6f270b
It relies on adding `data-testid="emergency"` in the template `send_emergency_access_invite.html.hbs`.

tom27052006:
> Wrote a Playwright test which could be included: [Timshel/OIDCWarden@31cb915](https://github.com/Timshel/OIDCWarden/commit/31cb915078b2b8222f454521dcb2d9844e6f270b) It relies on adding `data-testid="emergency"` in the template `send_emergency_access_invite.html.hbs`.

Thanks! I’ve included the Playwright test in the PR and it passes successfully on my side

PR #7749, opened 2026-09-17
SSO config to skip 2FA on login based on ACR and AMR

kowalski7cc:
Updated #6833 with support for ACR and AMR values. AMR claims are used with OIDC providers such Authentik. Remains a flag to fully skip 2FA if SSO is having issue with passing 2FA flag. In my case for some reason even if I have 2FA on all my social accounts, when logging in with let's say Google/Microsoft, VW still asks me 2FA. No issues with user+psw+otp or passkey login.
Code is made with AI assistance, but asked to a friend to do some basic code review.
Tested on k3s with Authentik 2026.8.


src/api/identity.rs
```diff
@@ -353,7 +353,20 @@ async fn sso_login(
         Some((mut user, sso_user)) => {
             let mut device = get_device(&data, conn, &user).await?;
 
-            let twofactor_token = twofactor_auth(&mut user, &data, &mut device, ip, client_version, conn).await?;
+            let skip_2fa = match CONFIG.sso_2fa_skip().as_str() {
```
Timshel, line 356: Logic could probably be moved in `OIDCAuthenticatedUser` to keep most of SSO logic out if identity.

src/db/models/sso_auth.rs
```diff
@@ -35,6 +35,8 @@ pub struct OIDCAuthenticatedUser {
     pub email: String,
     pub email_verified: Option<bool>,
     pub user_name: Option<String>,
+    pub acr: Option<String>,
+    pub amr: Option<Vec<String>>,
```
Timshel: The class is persisted to the database, should add the columns.

src/sso_client.rs
```diff
@@ -202,6 +202,21 @@ impl Client {
             .add_scopes(scopes)
             .add_extra_params(CONFIG.sso_authorize_extra_params_vec());
 
+        if CONFIG.sso_2fa_skip() == "auto" {
+            auth_req = auth_req.add_extra_param(
```
Timshel, line 206: Seems simpler to use `auth_req.add_auth_context_value(AuthenticationContextClass::new(acr))`

Timshel:
Not a maintainer but played a bit with `acr` and Keycloak (still has issues since I'm unable to the correct acr value for a TOTP login :().

From what I could find `acr` values are not restricted to integer and can be mapped to LoA (level of Authentication or level of Assurance) where I have seen mention of level 3/4.
As such I think it should be a configuration, Ex: `SSO_ACR_MIN: "3,2,1"`.

